PR #6, opened 2019-11-04
Use sdfv instead of xdot during transformations

am-ivanov:


tbennun:
Thanks! Please take a look at Commit 711036e979c9b3e416e4d7bf345b2f9aa9122d23, I modified SDFV to do something like your sdfv1.py. I think we can keep your changes to the rest of the files and the new SDFV instead of sdfv1.py. Thoughts?

tbennun:
Thanks for updating. I still think we should create a temporary file (in the current folder) and delete it when sdfv exits.

am-ivanov:
It is good to have html for debug purposes. Why it is so important to delete them? How it is possible to see all sdfg's at the same time after all transformation passes? We can add a command-line option like "--preserve-html=true/false", but why it is needed at all?

PR #7, opened 2019-11-05
Serialize patch

definelicht:
- [x] As a start, don't catch all exceptions, and don't fail silently if a field is missing.
- [x] Tolerate string inputs in `set_properties_from_json`
- [x] There's a monstrous list of string-to-type mappings in `dace.properties.known_types`. This seems brittle and hideous. We can replace this by calling an optionally implemented method.
- [x] Naming scheme is off (should be `to_json`, not `toJSON`
- [x] Move `set_properties_from_json` out of `Property` class
- [x] Right now every implementation of `toJSON`/`fromJSON` needs to call `json.dumps` and `json.loads`. Reduce this to receive the JSON objece.


tbennun:
Jenkins, do your thing.

tbennun:
Jenkins, do your thing. Again.

tbennun:
Jenkins, do your thing.

tbennun:
Jenkins, do your thing.

tbennun:
Jenkins, do your thing.

definelicht:
@tbennun The test now fails because `dace.types.struct` is constructed with mix of types and typeclass fields. Unfortunately this does not seem to happen only in the initializer, so we have to track down where this happens. I assume we want it to only contain typeclasses?

definelicht:
Turns out the problem is that `json_to_typeclass` in `dtypes` is doing two different things at the same time (converting from string to type, and from json to object), and was failing at one of them. I've fixed the branch that was failing, but eventually we want to split this into two different functions (or refactor entirely).

PR #17, opened 2019-11-08
Fix Nesting Transformation

TizianoDeMatteis:
Fix for nesting transformation:
- previously only source and sink node in the SDFG were considered and moved to the nested one. Now all the input/output node are considered
- change of transient property if an array has been moved to a Nested SDFG

### DIFF
--- a/dace/transformation/interstate/sdfg_nesting.py
+++ b/dace/transformation/interstate/sdfg_nesting.py
@@ -50,58 +50,35 @@ class NestSDFG(pattern_matching.Transformation):
         transients = {}
 
         for state in nested_sdfg.nodes():
-
-            for node in nxutil.find_source_nodes(state):
+            #  Input and output nodes are added as input and output nodes of the nested SDFG
+            for node in state.nodes():
                 if (isinstance(node, nodes.AccessNode)
                         and not node.desc(nested_sdfg).transient):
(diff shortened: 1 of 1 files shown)


dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -117,24 +94,25 @@ def apply(self, sdfg):
         for oldarrname, newarrname in transients.items():
             nested_sdfg.arrays.pop(oldarrname)
             nested_sdfg.arrays[newarrname].transient = False
-            outer_sdfg.arrays[oldarrname].transient = False
+            outer_sdfg.arrays[oldarrname].transient = True
```
tbennun: What's this doing? Do we even need this line, e.g. in the case that the outer array is not transient?

TizianoDeMatteis: Probably it is not needed. Putting it to `False` was wrong by the way, as additional kernel parameters were created

tbennun: ok. Can you try removing this line and see if the tests still pass?

TizianoDeMatteis: I confirm that that was unecessary

PR #18, opened 2019-11-09
Minor Xilinx code generation fixes

tbennun:
To retain support for math functions (sqrt etc.) inside tasklets


TizianoDeMatteis:
Looks ok to me!

PR #24, opened 2019-11-12
Call dace.serialize.set_properties_from_json from Diode

definelicht:
For some reason we had two implementations of `set_properties_from_json`: one in `dace.serialize`, and one in `diode.diode_rest`. Get rid of the latter, and fix the call that was breaking transformations in Diode.


definelicht:
@tbennun Still running tests, but this seems to fix it

PR #31, opened 2019-11-13
Make local mode default in DIODE and other cleanups

tbennun:


diode/remote_execution.py
```diff
@@ -40,48 +57,90 @@ def config_get(self, *key_hierarchy):
...
+            runpy.run_path(driver_file, run_name='__main__')
+        # Catching all exceptions, including SystemExit
+        except (Exception, SystemExit) as ex:
+            # Corner case: If exited with error code 0, it is a success
+            if isinstance(ex, SystemExit) and ex.code == 0:
+                pass
+            else:
+                traceback.print_exc()
+                self.show_output(traceback.format_exc())
+                sys.exit(1)
```
definelicht: Doesn't this work with just raising the exception?

diode/remote_execution.py
```diff
@@ -162,79 +202,19 @@ def run(self, dace_state, fail_on_nonzero=False):
...
-            except:
-                print(
-                    "WARNING: results.log could not be transmitted (probably not created)"
-                )
+            except RuntimeError:
+                pass
 
             self.remote_delete_file(remote_dace_file)
             self.remote_delete_dir(remote_dace_dir)
+        except:  # Literally anything can happen here
```
definelicht: Explain why we cannot specify the exception to catch here

PR #44, opened 2019-11-28
Fix propagated connectors

alexnick83:
The purpose of this branch is to fix a number of issues manifesting in large dace programs with multi-nested calls to other dace programs. A (non exhaustive) list of the issues:
- ~~Subsets of connectors-arrays do not propagate properly to the upper levels.~~ Suggested fix implemented in b23756bf.
- Multi-nested calls to dace programs cause read/write connector issues. More specifically, all parameters to dace program calls are recognized as inputs, which leads to written connector-arrays to be declared as inputs, causing issue during compilation (trying to write ArrayViewIn). Under investigation.
- There seems to be a code-generation issue with reduce and arrays of size 1. The output is generated as scalar instead of pointer.


dace/codegen/targets/cpu.py
```diff
@@ -1935,19 +1935,33 @@ def _generate_Reduce(self, sdfg, dfg, state_id, node, function_stream,
...
-        for d in range(input_dims):
+        input_subset = input_memlet.subset
+        output_subset = output_memlet.subset
+        for d in range(input_num_dims):
             if d in axes:
                 axis_vars.append("__i%d" % d)
             else:
                 axis_vars.append("__o%d" % octr)
+                ri = input_subset[d]
+                rngi = ri[1] - ri[0] + 1
```
tbennun: use subset.size()

dace/codegen/targets/cpu.py
```diff
@@ -2002,11 +2016,13 @@ def _generate_Reduce(self, sdfg, dfg, state_id, node, function_stream,
         outvar = ("__tmpout" if use_tmpout else cpp_array_expr(
             sdfg,
             output_memlet,
-            offset=["__o%d" % i for i in range(output_dims)],
+            offset=["__o%d" % output_axis_vars[i]
```
tbennun:
```suggestion
            offset=[("__o%d" % output_axis_vars[i])
```

dace/subsets.py
```diff
@@ -493,7 +493,7 @@ def squeeze(self):
         if not squeezed_ranges:
             squeezed_ranges = [(0, 0, 1)]
             squeezed_tsizes = [1]
-            non_ones = [len(shape) - 1]
+            # non_ones = [len(shape) - 1]
```
tbennun:
```suggestion
```

dace/codegen/targets/cpu.py
alexnick83: Implemented in  47f8c7046c1d6955fa21e2ada2ad8bc19cd975e1

Issue #49, opened 2019-12-03
Floor division inside index, makes the compilation fail.


nikvas88:
**Describe the bug**
When floor division (//) instead of division (/) is used inside the index, compilation fails.

**To Reproduce**
Try to run this program:
```python
@dace.program(dace.float64[N], dace.float64[N])
def floor_div(Input, Output): 
    @dace.map(_[0:N])
    def div(i):
        inp << Input[i//N]
        out >> Output[i]
        out = inp
```

tbennun:
Issue is fixed in PR #55

PR #55, opened 2019-12-05
Output floor division properly (fixes Issue #49)

tbennun:


tbennun:
Exactly, this fixes a bug

PR #57, opened 2019-12-08
Fix MapFusion transformation

tbennun:


dace/sdfg.py
```diff
@@ -1828,9 +1832,8 @@ def apply_transformations(self,
         while applied:
             applied = False
             # Find and apply immediately
-            for match in opt.get_pattern_matches(
+            for sdfg, match in opt.get_pattern_matches(
                     strict=strict, patterns=patterns, states=states):
-                sdfg = self.sdfg_list[match.sdfg_id]
```
alexnick83: I want to note that this is OK only when transforming the SDFG in-place, like we do through the command-line. The changes are completely unsuitable for other use-cases, e.g. DIODE.

tbennun: I tested DIODE and it works.  Also, this method (`sdfg.apply_transformations`) is always in-place.

alexnick83: Did you test that you can go back, i.e. "undo"?

tbennun: yes, I did multiple undos and re-transformations too

tbennun: as a side-note, transformations don't really work on nested SDFGs in master. (just checked)

tbennun:
@alexnick83 The SDFG list removal is indeed unrelated to this PR directly. I'll refactor it out. However, when I started applying transformations on loaded SDFGs it started crashing in unexpected locations. This was found to be because `sdfg_list` has different SDFG objects than the actual nested SDFGs. This could either result from bad deserialization or transformations not updating `sdfg_list` properly. Not sure how to address this, but I don't think we really need `sdfg_list`, there can be better location identifiers.

dace/sdfg.py
alexnick83: Let's open another branch/PR to fix this then.

PR #63, opened 2019-12-13
Fix strict transformations after loading and nice rendering of small graphs

tbennun:
Probably stems from improper saving/loading of the `sdfg_list` property


tbennun:
@alexnick83 now it's fixed the right way

PR #65, opened 2019-12-15
Serialize constant types

tbennun:
- [x] Serialize
- [x] Deserialize

### DIFF
--- a/dace/codegen/targets/framecode.py
+++ b/dace/codegen/targets/framecode.py
@@ -7,7 +7,7 @@ from dace.codegen.prettycode import CodeIOStream
 from dace.codegen.targets.target import TargetCodeGenerator, TargetDispatcher
 from dace.sdfg import SDFG, SDFGState, ScopeSubgraphView
 from dace.graph import nodes
-from dace import dtypes, config
+from dace import dtypes, data, config
 
 from dace.frontend.python import wrappers
 from dace.codegen import cppunparse
@@ -39,10 +39,9 @@ class DaCeCodeGenerator(object):
 
     def generate_constants(self, sdfg: SDFG, callsite_stream: CodeIOStream):
         # Write constants
-        for cstname, cstval in sdfg.constants.items():
-            if isinstance(cstval, np.ndarray):
-                dtype = dtypes.typeclass(cstval.dtype.type)
-                const_str = "constexpr " + dtype.ctype + \
+        for cstname, (csttype, cstval) in sdfg.constants_prop.items():
+            if isinstance(csttype, data.Array):
+                const_str = "constexpr " + csttype.dtype.ctype + \
                     " " + cstname + "[" + str(cstval.size) + "] = {"
                 it = np.nditer(cstval, order='C')
                 for i in range(cstval.size - 1):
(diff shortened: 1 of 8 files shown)


dace/codegen/targets/framecode.py
```diff
@@ -52,7 +51,8 @@ def generate_constants(self, sdfg: SDFG, callsite_stream: CodeIOStream):
                 callsite_stream.write(const_str, sdfg)
             else:
                 callsite_stream.write(
-                    "constexpr auto %s = %s;\n" % (cstname, str(cstval)), sdfg)
+                    "constexpr %s %s = %s;\n" % (csttype.dtype.ctype, cstname,
```
definelicht: This is great, now we can remove `generate_constants` from the FPGA codegen and just call this instead :-)

dace/sdfg.py
```diff
@@ -220,13 +220,10 @@ def __init__(self,
...
             raise InvalidSDFGError('Invalid SDFG name "%s"' % name, self, None)
 
-        #if not isinstance(arg_types, collections.OrderedDict):
-        #    raise TypeError
-
-        #self._arg_types = arg_types  # OrderedDict(str, typeclass)
-        #self._constants = constants  # type: Dict[str, Any]
         self.arg_types = arg_types or collections.OrderedDict()
-        self.constants_prop = constants or {}
+        self.constants_prop = {}
```
definelicht: Why is it that we need a separate `constants` and `constants_prop`? Seems pretty ugly

dace/sdfg.py
```diff
@@ -530,23 +528,49 @@ def constants(self):
...
+                return value
+            elif isinstance(dtype, dt.Scalar):
+                return dtype.dtype(value)
+            raise TypeError('Unsupported data type %s' % dtype)
+
+        result.update({k: cast(*v) for k, v in self.constants_prop.items()})
         return result
 
-    def add_constants(self, new_constants: Dict[str, Any]):
+    def add_constant(self, name: str, value: Any):
```
definelicht: I guess we cannot enforce that users don't directly update the constants dictionary, but we hope that they do this by convention?

dace/sdfg.py
tbennun: Having only one structure would require many more changes in the code. Feel free to amend, no particular reason why it is this way.

dace/sdfg.py
tbennun: We can hide away `add_constants` and make it an internal function, or just not exist at all. Your choice

dace/codegen/targets/framecode.py
tbennun: :)

dace/sdfg.py
definelicht: Ok, not worth it now

dace/sdfg.py
definelicht: Why don't we let `add_constants` call `add_constant`?

tbennun: Sure. I can quickly write this up and let you know

tbennun: Fixed now

dace/sdfg.py
```diff
@@ -220,13 +220,10 @@ def __init__(self,
...
-        #if not isinstance(arg_types, collections.OrderedDict):
-        #    raise TypeError
-
-        #self._arg_types = arg_types  # OrderedDict(str, typeclass)
-        #self._constants = constants  # type: Dict[str, Any]
         self.arg_types = arg_types or collections.OrderedDict()
-        self.constants_prop = constants or {}
+        self.constants_prop = {}
+        if constants is not None:
+            self.constants_prop.update(constants)
```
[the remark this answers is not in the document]
tbennun: now it does

PR #67, opened 2019-12-17
DIODE: Use multiple processes while executing

tbennun:


definelicht:
I ran a segfaulting test, and I get this in the terminal:

```
ERRORS: [DuplicateDLLError('Library libgemm.so is already loaded somewhere else, either unload it or use a different name for the SDFG/program.')]
An error occurred
```

But no error in DIODE

PR #68, opened 2019-12-17
Do not save generated code files unless they changed

tbennun:
Speeds up build time

### DIFF
--- a/dace/codegen/compiler.py
+++ b/dace/codegen/compiler.py
@@ -298,10 +298,10 @@ def unique_flags(flags):
     return set(re.findall(pattern, flags))
 
 
-def file_changed(filename: str, file_contents: str):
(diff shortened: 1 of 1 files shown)


dace/codegen/compiler.py
```diff
@@ -297,6 +298,29 @@ def unique_flags(flags):
     return set(re.findall(pattern, flags))
 
 
+def file_changed(filename: str, file_contents: str):
+    # If file did not exist before, it has changed
```
definelicht: Small nitpick: maybe it's more accurate to call this function something like `identical_file_exists`, to make it clear that it covers both whether the file exists at all, and that the content should be the same? This of course requires negating the return value.

tbennun: Done

PR #103, opened 2020-01-23
Make pipeline entries/exists serializable/deserializable

definelicht:
The pipeline entry and exit were missing the functionality to be serializable and deserializable. I've implemented it such that it uses the methods from MapEntry/MapExit by generalizing those methods.


definelicht:
@TizianoDeMatteis Could you review this?

PR #107, opened 2020-01-29
WIP: Augassign WCR fix

alexnick83:
Augmented assignments should go back to be implemented with WCR, instead of using the output as input as well. The generation of a map for array arguments should remain. A strict transformation that detects when it is safe to remove WCR must be created.


alexnick83:
This PR is obsolete. A new PR will be opened that will fix the issue in a different manner.

PR #108, opened 2020-01-30
Python frontend: Automatic callback generation

tbennun:
- [x] Callback in Python (outside tasklets)
- [x] Callback that returns arrays
- [x] Get return type from lhs of expression
- [x] Callback with kwargs / default values
- [x] Callback that returns a tuple of array, scalar
- [x] Callback to module function
- [x] Callback to class
- [x] ~~Callback in tasklet~~
- [x] Print warning upon use of callbacks
- [x] ~~Make a replacement for `print` to convert to std::cout?~~
- [x] ~~Callbacks in generators? `for i in myfunction(N)`~~
- [x] Check callbacks in nested SDFGs and how it affects pystate
- [x] Callbacks with GPU arrays (using cupy)


tests/callback_automatic_test.py
```diff
@@ -0,0 +1,80 @@
...
+    return alpha * A @ B
+
+
+def scale(C, beta):
+    C *= beta
+
+
+@dace.program
+def autocallback(A: dace.float32[N, N], B: dace.float32[N, N], C: dace.float32[N, N], beta: dace.float32):
+    tmp = almost_gemm(A, 0.5, B)
```
tbennun, line 18:
```suggestion
    tmp: dace.float32[N, N] = almost_gemm(A, 0.5, B)
    # OR
    # tmp[:] = almost_gemm(A, 0.5, B)
```

dace/frontend/python/newast.py
```diff
@@ -3973,7 +4041,238 @@ def _parse_sdfg_call(self, funcname: str,
...
+
+                argtypes.append(atype)
+
+        # Return type inference
+        return_type = None
+
+        # If the parent is a top level expression, the return value is unused
+        if isinstance(node.parent, ast.Expr) and node.parent.toplevel:
+            return_type = dtypes.typeclass(None)
+        # TODO: Assign, augassign
```
tbennun, line 4122: is this still a todo?

dace/frontend/python/newast.py
```diff
@@ -3141,7 +3183,13 @@ def visit_AnnAssign(self, node: ast.AnnAssign):
             dtype = None
             warnings.warn('typeclass {} is not supported'.format(type_name))
         if node.value is None and dtype is not None:  # Annotating type without assignment
-            self.annotated_types[rname(node.target)] = dtype
+            try:
+                array_type = eval(astutils.unparse(node.annotation), globals(),
```
tbennun, line 3187: should this maybe be `self.globals`?

dace/frontend/python/newast.py
```diff
@@ -3669,16 +3718,30 @@ def _parse_sdfg_call(self, funcname: str,
...
+                closure_arrays_only = closure_arrays
             for aname, arr in closure_arrays.items():
-                desc = data.create_datadescriptor(arr)
-                outer_name = self.sdfg.add_datadesc(aname,
-                                                    desc,
-                                                    find_new_name=True)
-                if not desc.transient:
-                    self.nested_closure_arrays[outer_name] = (arr, desc)
-                    # Add closure arrays as function arguments
+                if aname in sdfg.symbols:
```
tbennun, line 3726: @alexnick83 I'm just wondering why this doesn't cover the "arrays only" case?

tbennun (commented):
@alexnick83

dace/frontend/python/newast.py
alexnick83: Hmm, could be I guess. In any case, this is supposed to work for array types, i.e. `dace.whatever[shape]`, so both should work (both have dace).

dace/frontend/python/newast.py
alexnick83: I am not sure what you mean. The main issue is that closure_arrays that have to do with callbacks must not add descriptors (line 3735 and below) because they are symbols. If you are suggesting that the existence of `aname` in sdfg.symbols is enough to make the distinction, then I don't know if this could always be the case (perhaps some kind of name clashing?).

dace/frontend/python/preprocessing.py
```diff
@@ -403,7 +403,20 @@ def has_replacement(callobj: Callable,
     """
     from dace.frontend.common import op_repository as oprepo
 
+    # Nothing from the `dace` namespace needs preprocessing
+    mod = None
```
alexnick83, line 407: This apparently doesn't work for `math`, at least inside a tasklet :(

tbennun: Fixed

dace/frontend/python/newast.py
tbennun: Fixed.

dace/frontend/python/newast.py
tbennun: Fixed (waiting for response from #891)

dace/frontend/python/newast.py
```diff
@@ -3176,20 +3176,13 @@ def visit_Assign(self, node: ast.Assign):
     def visit_AnnAssign(self, node: ast.AnnAssign):
         type_name = rname(node.annotation)
         try:
-            dtype = eval(type_name)
-            if not isinstance(dtype, dtypes.typeclass):
-                raise NotImplementedError
+            dtype = eval(astutils.unparse(node.annotation), self.globals,
```
alexnick83, line 3179: We lost here the `dtypes.typeclass` check. The previous code in those lines would extract the `dace.whatever` name, even from an array definition, and check if this is a supported typeclass. Now, it may be possible to annotate an assignment with an unsupported 3rd party type without raising an exception.

tbennun: changing `eval` to something that uses the globals (like N for example) can now correctly resolve type annotations. This caused new issues for array types that would previously silently fail in any case.

How do you propose we solve this?

dace/frontend/python/newast.py
```diff
@@ -3178,6 +3178,12 @@ def visit_AnnAssign(self, node: ast.AnnAssign):
         try:
             dtype = eval(astutils.unparse(node.annotation), self.globals,
                          self.defined)
+            if isinstance(dtype, data.Data):
+                simple_type = dtype.dtype
+            else:
+                simple_type = dtype
+            if not isinstance(simple_type, dtypes.typeclass):
+                raise
```
tbennun, line 3186: @alexnick83 raise what? should this be `TypeError('Type annotations of non-dace types not supported')` ?

dace/frontend/python/newast.py
alexnick83: What I mean is that we had two checks, one of typeclass and one evaluating for array types. We lost the typeclass one. I restored it (in a more modern version) and added a test. I am not sure which should be our policy. It will now throw the warning and ignore the annotation, but maybe we want to throw NotImplementedError.

dace/frontend/python/newast.py
alexnick83: it is raising so that it is caught by the `except` below.

alexnick83: We could make it NotImplementedError or TypeError but I am not about the policy. It appears that the idea now is to just warn and ignore the annotation.

alexnick83: The added test reflects that.

alexnick83 (approved):

PR #109, opened 2020-01-31
Remove static symbol storage

tbennun:
Removes the static symbol store in `dace.symbolic`. This is the first step in allowing nested SDFGs to have their own symbol systems.

- [x] Remove methods
- [x] Adapt symbolic analysis in dace core
- [x] ~Ensure all SDFG symbols are stored in `sdfg.symbols`~
- [x] Fix SDFG specialization
- [x] Add symbol name->value shape inference in dace.program invocations
- [x] Adapt samples and tests
- [x] Add sample with two different symbol values in a nested SDFG
- [x] Fix workaround in FPGA code generation


dace/sdfg.py
```diff
@@ -3776,14 +3740,15 @@ def validate(self, sdfg, state_id) -> None:
...
                 if not isinstance(src_node, nd.AccessNode):
-                    raise InvalidSDFGEdgeError(
-                        "Memlet creates an "
-                        "invalid path (source node %s should "
-                        "be a data node)" % str(src_node),
-                        sdfg,
-                        state_id,
-                        eid,
-                    )
+                    pass
```
alexnick83: Why is this not an error any longer?

tests/memlet_lifetime_validation_test.py
```diff
@@ -17,33 +17,33 @@
...
-    # Left for debugging purposes
-    sdfg1.draw_to_file()
-
-    try:
-        sdfg1.validate()
-        print("SDFG 1 passed validation, test FAILED")
-        exit(1)
-    except InvalidSDFGError:
-        print("Test 1 passed, exception successfully caught")
+    # sdfg1 = SDFG('shouldntwork1')
```
alexnick83: Is this test obsolete or is it temporarily broken?

dace/sdfg.py
tbennun: Because we can connect a tasklet directly through a map. This removes the need for this valuation check and the test below

tests/memlet_lifetime_validation_test.py
tbennun: See above. We don't need the test any longer as this is a valid SDFG.

PR #110, opened 2020-02-03
Nested Symbols

tbennun:
Allow nested SDFGs to be written once and have different symbol values based on their invocation.

- [x] Add symbol mapping to `NestedSDFG` nodes
- [x] Adapt nested SDFG calling convention
    - [x] Frontend
    - [x] Code generation
- [x] Adapt `undefined_symbols` to look at `symbol_mapping` and ignore nested SDFGs
- [x] Adapt the `InlineSDFG` transformation to take nested symbols into account

### DIFF
--- a/dace/properties.py
+++ b/dace/properties.py
@@ -537,9 +537,11 @@ class ListProperty(Property):
         # Otherwise, convert to strings
         return list(map(str, l))
 
-    @staticmethod
-    def from_string(s):
-        return list(s)
+    def from_string(self, s):
+        if s.startswith('[') and s.endswith(']'):
+            return [self.element_type(d.strip()) for d in s[1:-1].split(',')]
+        else:
+            return list(s)
 
     def from_json(self, data, sdfg=None):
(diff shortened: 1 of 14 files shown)


dace/sdfg.py
```diff
@@ -310,9 +310,6 @@ def from_json(cls, json_obj, context_info={'sdfg': None}):
 
         return ret
 
-        # Counter to make it easy to create temp transients
-        self._temp_transients = 0
```
tbennun: This code still exists in `__init__`. It was duplicated as a result of a merge conflict and is unreachable.

dace/properties.py
```diff
@@ -662,31 +748,6 @@ def from_string(s):
         return di
 
 
-class ParamsProperty(Property):
```
alexnick83: Why is this removed? Is it deprecated by the new DictProperty?

tbennun: Yes, `ListProperty` to be precise.

PR #111, opened 2020-02-04
FPGA codegen fixes

definelicht:


definelicht:
@tbennun I put the three fixes in three different commits

PR #112, opened 2020-02-05
Use __array_interface__ and __cuda_array_interface__ in calls to SDFG

tbennun:


tbennun:
@definelicht done, ptal.

Issue #10, opened 2019-11-06
Running multiple SDFGs with the same name using CUDA fails with DuplicateDLLError


definelicht:
### Problem

When compiling and running multiple SDFGs with the same name from the same Python executable, the SDFG is usually deleted in between (I assume by garbage collection?), which unloads the dynamic libraries.

However, when CUDA is involved, it seems that the SDFGs are not deleted even when they are no longer references by Python, which results in the loaded library to stick around. In practice, this can result in DuplicateDLLError when trying to load a new SDFG using the same name.

### Expected behavior

SDFGs should be cleaned up when they are no longer referenced, and it should be possible to run multiple SDFGs with the same name from the same Python executable without explicitly deleting them.

### To reproduce

Run test `tests/library/blas_dot.py` on the `library_nodes` branch, but change the initialization of `dace.SDFG` to always have the same name, then run the test including the cuBLAS runs.

### Workaround

Name each SDFG differently, *or* explicitly call `del my_sdfg` between executions.

tbennun:
Fixed in #114

PR #114, opened 2020-02-05
Reload DLLs on failure (fixes Issue #10)

tbennun:


tbennun:
@definelicht please check this w.r.t. known issue in library nodes
@am-ivanov please check this w.r.t. your reported issue in Jupyter

Thanks!

tbennun:
I can confirm that the Jupyter problem does not reproduce on Windows any longer.

PR #116, opened 2020-02-08
Extensible subclass and enumeration registries

tbennun:
Used for generalizing extensible transformations, code generators, storage/schedule/instrumentation types etc.

- [x] Make decorators for registries
- [x] Use extensible enumerations
- [x] Use registries in:
    - [x] Transformations
    - [x] Code generators
    - [x] Instrumentation providers
    - [x] Memlet propagation patterns


dace/runtime/include/dace/types.h
```diff
@@ -37,6 +37,7 @@
     #define DACE_HDFI inline
     #define DACE_HFI inline
     #define DACE_DFI inline
+    typedef short half;
```
definelicht: `short` != `float16`, what's the idea here?

definelicht: This also seems to be what's failing the tests.

tbennun: CPU/C++ code does not (yet) natively support half-precision floating point. Since CUDA and NumPy do, we need a way to communicate that. In CUDA, `float16` is defined as `short` in certain points of host code, so this fits. Now the tests will hopefully pass.

PR #121, opened 2020-02-12
WIP: BLAS Implementations

definelicht:
We need to flesh out our repertoire of BLAS implementations, and can gradually merge these into master:

- [ ] cuBLAS GEMM (@alexnick83)
- [ ] cuBLAS Transpose (@alexnick83)
- [ ] "Pure" GEMV
- [ ] MKL GEMV
- [ ] cuBLAS GEMV


definelicht:
@alexnick83 When you get back, could you finish the two implementations above you have already started? Then we can assign the others on demand later.

tbennun:
so it was set to be merged, even though it wasn't.

PR #124, opened 2020-02-14
Map Fission Transformation

tbennun:
Example:

Before:
before

After:
after


dace/transformation/helpers.py
```diff
@@ -120,6 +120,8 @@ def nest_state_subgraph(sdfg: SDFG,
     input_names = []
     output_names = []
     for edge in inputs:
+        if edge.data.data is None:  # Skip edges with an empty memlet
+            continue
```
definelicht: What are these cases exactly?

tbennun: When you have a tasklet or access node that is not coming from the outside, but still is replicated in the map, it needs to be dominated by the map-entry and post-dominated by the map-exit. In these cases we add an edge with an empty memlet.

PR #134, opened 2020-02-19
Extensible scope/tasklet preamble/postamble methods

tbennun:
Some codegen cleanup, refactoring, and addition of extra methods that can be extended to support behavior. These will later be used in further refactoring to move some of the CPU codegen into.

Extending CPUCodeGen with `generate_scope_{preamble,postamble}` allows the code to be customized as follows:
```
<OUTER PREAMBLE>
for(...) {
  for(...) {
    <INNER PREAMBLE>
    scope generated code...
    <INNER POSTAMBLE>
  }
}
<OUTER POSTAMBLE>
```

and `generate_tasklet_{preamble,postamble}`:
```
<BEFORE MEMLETS PREAMBLE>
memlet definitions and copy in
<AFTER MEMLETS PREAMBLE>
tasklet code...
<BEFORE MEMLETS POSTAMBLE>
output memlet copies
<AFTER MEMLETS POSTAMBLE>
```


dace/codegen/targets/target.py
```diff
@@ -49,7 +49,7 @@ def generate_state(self, sdfg, state, function_stream, callsite_stream):
                                     to the current location (call-site)
                                     in the code.
         """
-        raise NotImplementedError('Abstract class')
+        pass
```
alexnick83: The except/pass monster shows up again  :laughing:

tbennun: This changed to pass since not all code generators implement or must implement this method :)

tbennun:
Yes, there will be a cpp code generator later which CPU will extend (i.e., with openmp pragmas). Same will apply for GPU/FPGA. Will be much cleaner than now

PR #139, opened 2020-02-22
Fix various issues in MapFission

tbennun:
Adds tests and fixes various issues in MapFission with multi-state nested SDFGs and complex subgraphs.

### DIFF
--- a/dace/sdfg.py
+++ b/dace/sdfg.py
@@ -4433,8 +4433,15 @@ def replace(subgraph: Union[SDFGState, ScopeSubgraphView, SubgraphView],
             if isinstance(propclass, properties.RangeProperty):
                 setattr(node, pname, replsym(propval))
             if isinstance(propclass, properties.CodeProperty):
-                for stmt in propval['code_or_block']:
-                    ASTFindReplace({name: new_name}).visit(stmt)
+                if isinstance(propval['code_or_block'], str):
+                    # TODO: C++ AST parsing for replacement?
+                    if name != str(new_name):
+                        warnings.warn(
+                            'Replacement of %s with %s was not made '
+                            'for string tasklet code' % (name, new_name))
+                else:
+                    for stmt in propval['code_or_block']:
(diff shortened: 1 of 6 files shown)


dace/sdfg.py
```diff
@@ -4433,8 +4433,15 @@ def replsym(symlist):
             if isinstance(propclass, properties.RangeProperty):
                 setattr(node, pname, replsym(propval))
             if isinstance(propclass, properties.CodeProperty):
-                for stmt in propval['code_or_block']:
-                    ASTFindReplace({name: new_name}).visit(stmt)
+                if isinstance(propval['code_or_block'], str):
+                    # TODO: C++ AST parsing for replacement?
+                    if name != str(new_name):
+                        warnings.warn(
```
definelicht: Are we sure we don't want to fail here? Won't this break in most cases?

dace/transformation/helpers.py
```diff
@@ -230,3 +230,45 @@ def nest_state_subgraph(sdfg: SDFG,
...
+                     or internal_memlet.subset[0] == 0)):
+            to_unsqueeze = ones[1:]
+        else:
+            to_unsqueeze = ones
+
+        result.subset.unsqueeze(to_unsqueeze)
+    elif len(internal_memlet.subset) > len(external_memlet.subset):
+        raise ValueError(
+            'Unexpected extra dimensions in internal memlet '
+            'while inlining SDFG.\nExternal memlet: %s\n'
```
definelicht: "while inlining SDFG"?

dace/sdfg.py
tbennun: It will only break in cases where you have symbols (e.g. `a[i] = i`) in the tasklet itself. In most cases it doesn't break.

definelicht: Ok, sure

dace/transformation/helpers.py
tbennun: Fixed, thanks

PR #143, opened 2020-02-25
Transformation to reuse transients

targetsm:


tbennun:
@targetsm it's unnecessary to remove and re-add edges if you change a field inside the memlet. It's only necessary to do so if you redirect the edge (hence `.src` and `.dst` are properties)

tests/transients.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python
```
tbennun: @targetsm if you want the test to run every time the CI runs, it has to be called `*_test.py`. I'd recommend putting it in `tests/transformations/transient_reuse_test.py`

tests/transients.py
```diff
@@ -0,0 +1,53 @@
...
+from __future__ import print_function
+
+import argparse
+import dace
+import numpy as np
+from dace.transformation.interstate.transient_reuse import TransientReuse
+
+M = dace.symbol('M')
+K = dace.symbol('K')
+N = dace.symbol('N')
```
tbennun:
```suggestion
```

tests/transients.py
```diff
@@ -0,0 +1,53 @@
...
+    C[:] = A @ B @ A @ B @ A
+
+
+if __name__ == "__main__":
+    print("==== Program start ====")
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("M", type=int, nargs="?", default=5)
+    parser.add_argument("K", type=int, nargs="?", default=5)
+    parser.add_argument("N", type=int, nargs="?", default=5)
```
tbennun:
```suggestion
```

tests/transients.py
```diff
@@ -0,0 +1,53 @@
...
+    args = vars(parser.parse_args())
+
+    M.set(args["M"])
+    K.set(args["K"])
+    N.set(args["N"])
+
+    print('Matrix multiplication %dx%dx%d' % (M.get(), K.get(), N.get()))
```
tbennun:
```suggestion
```

tests/transients.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Initialize arrays: Randomize A and B, zero C
+    A = np.random.rand(M.get(), K.get()).astype(np.float64)
+    B = np.random.rand(K.get(), N.get()).astype(np.float64)
```
tbennun: sizes should all be M x M

tests/transformations/transient_reuse_test.py
```diff
@@ -0,0 +1,56 @@
...
+    N.value = args["N"]
+
+    # Initialize arrays: Randomize A and B, zero C
+    A = np.random.rand(M.value, M.value).astype(np.float64)
+    B = np.random.rand(M.value, M.value).astype(np.float64)
+    D = np.random.rand(M.value, N.value).astype(np.float64)
+    C = np.zeros([M.value, N.value], dtype=np.float64)
+    C_regression = np.zeros_like(C)
+
+    sdfg = operation.to_sdfg(args)
```
targetsm, line 47: Can I somehow get M and N here into the operation? The memory savings output would look nicer...

dace/transformation/interstate/transient_reuse.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def _atomic_counter_generator():
+    ctr = 0
+    while True:
+        ctr += 1
+        yield ctr
+
+
+_atomic_count = _atomic_counter_generator()
```
tbennun, line 16: @targetsm Please don't use atomic counters or other design patterns that run on global code. When you add an array or data descriptor you have the `find_new_name` flag that you can set (so "transient_reuse" becomes "transient_reuse_1", 2, ...)

tests/transformations/transient_reuse_test.py
tbennun: What does this mean? also what is `M/N.value`?

targetsm: I thought there may be a way to pass the values of M and N to operation.to_sdfg() so that when I sum the transient array sizes I actually get a value instead of combinations of M and N. Another option is to just set the sizes instead of having M and N as variables.
I should use M/N.value no?

dace/transformation/interstate/transient_reuse.py
```diff
@@ -0,0 +1,177 @@
...
+                successors[n] = set(G.successors(n))
+                ancestors[n] = set(nx.ancestors(G, n))
+            mappings = {}
+            for n in transients:
+                mappings[n] = set()
+
+            # Find valid mappings. A mapping (n, m) is only valid if the successors of n
+            # are a subset of the ancestors of m and n is also an ancestor of m.
+            # Further the arrays have to be equivalent.
+            # Assumption: Arrays are never used in two AccessNodes
```
tbennun, line 100: you can actually check that

tests/transformations/transient_reuse_test.py
targetsm: Should now be ok I think

dace/transformation/interstate/transient_reuse.py
```diff
@@ -0,0 +1,177 @@
...
+                transients.add(a)
+
+        for state in sdfg.nodes():
+            # Copy the whole graph
+            G = nx.DiGraph()
+            for n in state.nodes():
+                G.add_node(n)
+            for n in state.nodes():
+                for e in state.all_edges(n):
+                    G.add_edge(e.src, e.dst)
```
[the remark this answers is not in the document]
targetsm: So I changed it to MultiDiGraph, but using copy.deepcopy(state._nx) doesn't work somehow. The tasklets strangely are changed somehow, we can look at it in two weeks.

dace/transformation/interstate/transient_reuse.py
```diff
@@ -0,0 +1,175 @@
...
+
+            # Collapse all mappings and their scopes into one node
+            scope_dict = state.scope_dict(node_to_children=True)
+            for n in scope_dict:
+                if n is not None:
+                    G.add_edges_from([(n, x) for (y, x) in G.out_edges(state.exit_nodes(n)[0])])
```
tbennun, line 63:
```suggestion
                    G.add_edges_from([(n, x) for (y, x) in G.out_edges(state.exit_node(n))])
```

dace/transformation/interstate/transient_reuse.py
```diff
@@ -0,0 +1,175 @@
...
+            for n in state.nodes():
+                G.add_node(n)
+            for n in state.nodes():
+                for e in state.all_edges(n):
+                    G.add_edge(e.src, e.dst)
+
+            # Collapse all mappings and their scopes into one node
+            scope_dict = state.scope_dict(node_to_children=True)
+            for n in scope_dict:
+                if n is not None:
```
tbennun, line 62: Are you sure this works with nested scopes (map in a map)? I would simply loop over `scope_dict[None]` looking for `EntryNode`s to be safe.

dace/transformation/interstate/transient_reuse.py
```diff
@@ -0,0 +1,175 @@
...
+            for n in transients:
+                mappings[n] = set()
+
+            # Find valid mappings. A mapping (n, m) is only valid if the successors of n
+            # are a subset of the ancestors of m and n is also an ancestor of m.
+            # Further the arrays have to be equivalent.
+            for n in G.nodes():
+                for m in G.nodes():
+                    if n is not m and n.data == m.data:
+                        raise Exception("Transients used in multiple nodes")
```
tbennun, line 100: I would not raise an exception, just not consider this array

dace/transformation/interstate/transient_reuse.py
targetsm: Running it works on e.g. lenet it finds all nested scopes... but if you don't trust the method I will change it.

tbennun: I don't trust it because the order of collapsing might leave uncollapsed top-level maps.

targetsm: Ok! Changed it yesterday ^^

PR #147, opened 2020-02-25
Fixes for fusion transformations and frontend

tbennun:
- [x] MapFusion with multiple inputs/outputs
- [x] StateFusion with Write-After-Write in one sub-component
- [x] Relax StateFusion to apply in trivial cases
- [x] MapFusion with arrays doesn't modify memlets
- [x] StripMining/MapTiling detect tile size of 1 or T


tbennun:
@TizianoDeMatteis please review the FPGA part of the diff. Thanks!

dace/transformation/dataflow/local_storage.py
```diff
@@ -54,7 +54,9 @@ def apply(self, sdfg):
 
         array = self.array
         if array is None or len(array) == 0:
-            array = graph.edges_between(node_a, node_b)[0].data.data
+            array = next(e.data.data
+                         for e in graph.edges_between(node_a, node_b)
+                         if e.data.data is not None and e.data.wcr is None)
```
alexnick83: Why are outputs with wcr excluded (assuming they have not been explicitly picked through the transformation parameters)?

alexnick83: Nvm, we have accumulate transient now

PR #149, opened 2020-02-26
WIP: Fix covered nested accesses

alexnick83:
In DaCe we still don't have the required tools to detect overlapped accesses in a generalized manner. Because of this, the python frontend always generates new accesses for data, even if there is already another access to the same data, unless the access ranges match exactly. This causes issues when an access is re-used in the same scope. For example:
```python
@dace.program
def nested_superrange(A: dace.float32[2, 3, 4], B: dace.float32[3, 4]):
    tmp = dace.define_local([3, 4, 2, 1, 1], dace.float32)
    for i, j in dace.map[0:3, 0:4]:
        tmp[i, j] = A[:, i, j]
        B[i, j] = tmp[i, j, 0] + tmp[i, j, 1]
```

There is an access to tmp (tmp[i, j]) that is subsequently re-used (tmp[i, j, 0] and tmp[i, j, 1]). The python frontend in master will generate new accesses for tmp[i, j, 0] and tmp[i, j, 1] instead of re-using the access for tmp[i, j]. Due to this, the result of the program execution will be wrong.

It is possible for DaCe to support the above case, because the re-used accesses are **covered** by the first access. We already have a suitable method to detect this, dace.subsets.Subset.covers. The purpose of this branch is to amend the frontend and any other core-files necessary to support re-use of covered nested accesses properly.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -977,7 +977,7 @@ def _parse_memlet_subset(array: data.Data,
 
 # Parses a memlet statement
 def ParseMemlet(visitor, defined_arrays_and_symbols: Dict[str, Any],
-                node: MemletType):
+                node: MemletType, subset: subsets.Subset=None):
     das = defined_arrays_and_symbols
     arrname = rname(node)
     if arrname not in das:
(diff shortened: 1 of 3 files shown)


tbennun:
@alexnick83 is this still WIP?

alexnick83:
> @alexnick83 is this still WIP?

Yes, the fix is quite hackish and is not even applied to all valid cases. I put it on low priority for now, since there are other branches that need more attention.

PR #154, opened 2020-02-28
cppunparse: add infer_type to all functions, support attributes

tbennun:


dace/codegen/cppunparse.py
```diff
@@ -513,9 +513,9 @@ def _TryExcept(self, t):
         for ex in t.handlers:
             self.dispatch(ex)
         if t.orelse:
-            raise SyntaxError('Invalid C++')
+            raise NotImplementedError('Invalid C++')
 
-    def _TryFinally(self, t):
+    def _TryFinally(self, t, infer_type=False):
```
TizianoDeMatteis: See comment for `assert`

dace/codegen/cppunparse.py
```diff
@@ -530,7 +530,7 @@ def _TryFinally(self, t):
         self.dispatch(t.finalbody)
         self.leave()
 
-    def _ExceptHandler(self, t):
+    def _ExceptHandler(self, t, infer_type=False):
```
TizianoDeMatteis: See comment for `assert`

dace/codegen/cppunparse.py
```diff
@@ -644,7 +644,7 @@ def _FunctionDef(self, t, infer_type=False):
     def _AsyncFunctionDef(self, t, infer_type=False):
         self._generic_FunctionDef(t, infer_type, is_async=True)
 
-    def _generic_For(self, t, is_async=False):
+    def _generic_For(self, t, is_async=False, infer_type=False):
```
TizianoDeMatteis: See comment for `assert`.

Actually, I recognized that this function has another problem, that is it generates `auto` for loop variables, not good for OpenCL

dace/codegen/cppunparse.py
```diff
@@ -1058,8 +1058,8 @@ def _BoolOp(self, t, infer_type=False):
         self.write(")", infer_type)
         return dace.dtypes.typeclass(np.bool) if infer_type else None
 
-    def _Attribute(self, t):
-        self.dispatch(t.value)
+    def _Attribute(self, t, infer_type=False):
```
TizianoDeMatteis: See comment for `assert`

dace/codegen/cppunparse.py
tbennun: yes, I noticed that too. this is not great for several reasons on other platforms too.

PR #155, opened 2020-02-28
Fix local-storage transformations for uneven partitions

tbennun:
- [x]  LocalStorage/MapTiling uneven copy does not produce SymExpr

### DIFF
--- a/dace/codegen/targets/cpp.py
+++ b/dace/codegen/targets/cpp.py
@@ -98,8 +98,6 @@ def memlet_copy_to_absolute_strides(dispatcher,
     # where the pointer type is changed to a vector type, and addresses
     # thus shouldn't take vectorization into account.
     copy_shape = memlet.subset.size_exact()
-    # copy_shape = memlet.subset.size()
-    # copy_shape = [symbolic.overapproximate(s) for s in copy_shape]
     src_nodedesc = src_node.desc(sdfg)
     dst_nodedesc = dst_node.desc(sdfg)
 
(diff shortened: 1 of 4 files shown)


dace/subsets.py
```diff
@@ -176,6 +176,30 @@ def size(self, for_codegen=False):
...
+            int_ceil = sp.Function('int_ceil')
+            return [
+                ts * int_ceil(
+                    ((iMax.expr if isinstance(iMax, symbolic.SymExpr) else
+                      iMax) + 1 - (iMin.expr if isinstance(
+                          iMin, symbolic.SymExpr) else iMin)),
+                    (step.expr
+                     if isinstance(step, symbolic.SymExpr) else step))
+                for (iMin, iMax, step), ts in zip(self.ranges, self.tile_sizes)
+            ]
```
tbennun: @alexnick83 If no one is calling this mode, do we need it?

dace/codegen/targets/cpp.py
```diff
@@ -97,8 +97,9 @@ def memlet_copy_to_absolute_strides(dispatcher,
     # Ignore vectorization flag is a hack to accommmodate FPGA behavior,
     # where the pointer type is changed to a vector type, and addresses
     # thus shouldn't take vectorization into account.
-    copy_shape = memlet.subset.size()
-    copy_shape = [symbolic.overapproximate(s) for s in copy_shape]
+    copy_shape = memlet.subset.size_exact()
+    # copy_shape = memlet.subset.size()
+    # copy_shape = [symbolic.overapproximate(s) for s in copy_shape]
```
tbennun:
```suggestion
```

dace/subsets.py
tbennun: Also, this is why we have `symcpp` in the codegen, it makes the conversion there.

alexnick83: I actually don't know why this mode exists. I just copied it as it was in the size method. Is it used there anywhere?

tbennun: Not sure. We can remove it then

alexnick83: I removed the path. We can just re-add it if we find a use-case for it.

PR #159, opened 2020-03-02
Store node instead of index for starting state

definelicht:
There was a bug where `sdfg._start_state` was stored as an index, but read as a node. I've changed it to be stored as a node, since this is more robust to the SDFG changing.

### DIFF
--- a/dace/sdfg.py
+++ b/dace/sdfg.py
@@ -649,7 +649,7 @@ class SDFG(OrderedDiGraph):
             raise TypeError("Expected SDFGState, got " + str(type(node)))
         super(SDFG, self).add_node(node)
         if is_start_state == True:
-            self.start_state = len(self.nodes())
+            self.start_state = len(self.nodes()) - 1
 
     def add_edge(self, u, v, edge):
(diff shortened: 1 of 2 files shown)


dace/sdfg.py
```diff
@@ -413,17 +413,21 @@ def start_state(self):
             return source_nodes[0]
         if self._start_state is None:
             raise ValueError('Ambiguous or undefined starting state for SDFG')
+        if self._start_state not in source_nodes:
+            raise ValueError("Start state {} not found in source nodes".format(
+                self._start_state))
+        return self._start_state
```
tbennun: What if the graph is one state in a loop and an exit? The start state will not be a source node..

definelicht: The only way around that as far as I can tell is to allow _any_ state to be a start state. I doubt our codegen supports that?

tbennun: I think this is what _start_state is for - to allow a non-source-node to be a starting state

definelicht: Not necessarily, it's possible to have multiple source nodes, which is the situation where you'd need user input to decide where to start (not that this ever happens in practice, as far as I know).

PR #165, opened 2020-03-09
Fix cppunparse: Assign, consider the case of Attribute

TizianoDeMatteis:


TizianoDeMatteis:
> Why not get the type by recursively visiting the target with infer_types=True?

@tbennun  Nice idea! I've implemented it

PR #168, opened 2020-03-10
Double argument in Xilinx kernels

definelicht:
When creating Xilinx OpenCL kernels, we were concatenating scalar arguments with symbol arguments. It seems symbol arguments sometimes include scalar arguments despite the `include_scalar_arguments` flag being set to `False` (and/or the other way around).

Work around this by deduplicating arguments in the Xilinx codegen.

### DIFF
--- a/dace/codegen/targets/xilinx.py
+++ b/dace/codegen/targets/xilinx.py
@@ -1,3 +1,5 @@
+import collections
+import itertools
 import os
 import re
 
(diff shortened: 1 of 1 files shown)


definelicht:
> I thought we fixed this

Me too... There's currently overlap in scalar parameters and symbol parameters, so this is the real issue. But I don't want to look into this right now (again... again).

PR #174, opened 2020-03-17
Add Gemm Library node and tests

orausch:
TODO:
- [x] update test to conform to dace testing code style (and maybe make it shorter)
- [x] fix bug when `complex=True` and `beta not in [1.0, 0.0]`
- [x] implement broadcasting for C
- [x] test missing C

### DIFF
--- a/dace/libraries/blas/environments/intel_mkl.py
+++ b/dace/libraries/blas/environments/intel_mkl.py
@@ -1,13 +1,19 @@
+import os
 import dace.library
 
-
+def _find_mkl_include():
+    if 'MKLROOT' in os.environ:
+        return [os.path.join(os.environ['MKLROOT'], 'include')]
+    else:
+        return []
+    
 @dace.library.environment
 class IntelMKL:
 
     cmake_minimum_version = None
     cmake_packages = ["BLAS"]
     cmake_variables = {"BLA_VENDOR": "Intel10_64lp"}
-    cmake_includes = []  # For some reason, FindBLAS does not find includes
+    cmake_includes = _find_mkl_include()
     cmake_libraries = ["mkl_rt"]
     cmake_compile_flags = []
     cmake_link_flags = []
--- a/dace/libraries/blas/nodes/gemm.py
+++ b/dace/libraries/blas/nodes/gemm.py
@@ -1,0 +1,281 @@
+from copy import deepcopy as dc
+import numpy as np
+from dace.config import Config
+import dace.library
+import dace.properties
+import dace.graph.nodes
+from dace.transformation.pattern_matching import ExpandTransformation
+from dace.properties import Property
+from .. import environments
+
+
+def _is_complex(dtype):
+    if hasattr(dtype, "is_complex"):
+        return dtype.is_complex()
+    else:
+        return dtype in [np.complex64, np.complex128]
+
+
+def _cast_to_dtype_str(value, dtype: dace.dtypes.typeclass) -> str:
+    if _is_complex(dtype) and _is_complex(type(value)):
+        raise ValueError("Cannot use complex beta with non-complex array")
+
+    if _is_complex(dtype):
+        cast_value = complex(value)
+
+        return "dace.{type}({real}, {imag})".format(
+            type=dace.DTYPE_TO_TYPECLASS[dtype].to_string(),
+            real=cast_value.real,
+            imag=cast_value.imag,
+        )
+    else:
+        return "dace.{}({})".format(dace.DTYPE_TO_TYPECLASS[dtype].to_string(),
+                                    value)
+
+
+def _get_gemm_inputs(node, state, sdfg):
+    """Returns the matrix multiplication input edges, arrays, and shape."""
+    res_a = None
+    res_b = None
+    res_c = None
+    for edge in state.in_edges(node):
+        if edge.dst_conn in ["_a", "_b", "_c"]:
+            size = edge.data.subset.size()
+            outer_array = sdfg.data(
+                dace.sdfg.find_input_arraynode(state, edge).data)
+            res = edge, outer_array, size
+            if edge.dst_conn == "_a":
+                res_a = res
+            elif edge.dst_conn == "_b":
+                res_b = res
+            else:
+                res_c = res
+    if res_a is None or res_b is None:
+        raise ValueError("Matrix multiplication input connectors \"_a\" and "
+                         "\"_b\" not found.")
+    return res_a, res_b, res_c
+
+
+@dace.library.expansion
+class ExpandGemmPure(ExpandTransformation):
+
+    environments = []
+
+    @staticmethod
+    def make_sdfg(node, parent_state, parent_sdfg):
+        sdfg = dace.SDFG(node.label + "_sdfg")
+        state = sdfg.add_state(node.label + "_state")
+
+        ((edge_a, outer_array_a, shape_a), (edge_b, outer_array_b, shape_b),
+         c_inputs) = _get_gemm_inputs(node, parent_state, parent_sdfg)
+
+        dtype_a = outer_array_a.dtype.type
+        dtype_b = outer_array_b.dtype.type
+        dtype_y = dace.DTYPE_TO_TYPECLASS[np.result_type(dtype_a,
+                                                         dtype_b).type]
+        if c_inputs is not None:
+            edge_c, outer_array_c, shape_c = c_inputs
+            dtype_c = outer_array_c.dtype.type
+
+        if node.transA:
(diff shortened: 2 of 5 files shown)


dace/libraries/blas/nodes/gemm.py
```diff
@@ -0,0 +1,239 @@
...
+                                   wcr_identity=0)
+            },
+            external_edges=True,
+            output_nodes=output_nodes)
+
+        if node.beta != 0:
+            add_program = "__y = ({} * __c) + __tmp".format(node.beta)
+
+            # addition map
+            state.add_mapped_tasklet(
```
orausch: We could avoid this map by adding c[i0, i1]/k in the multiplication tasklet, but I haven't tested if that is ok with numerically

tests/gemm_library_node.py
```diff
@@ -0,0 +1,143 @@
+import argparse
```
tbennun: You can use `tests/library/gemm_test.py`. If it ends with `_test.py`, it will run automatically

tests/gemm_library_node.py
```diff
@@ -0,0 +1,143 @@
...
+
+    sdfg(A=A, B=B, C=C, Y=Y)
+
+    diff = np.linalg.norm(Y_regression - Y) / (M * N)
+    print("Difference:", diff)
+    assert diff <= 1e-5
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
tbennun: No need for arguments in a test, you can shorten here.

dace/libraries/blas/environments/intel_mkl.py
```diff
@@ -7,12 +8,11 @@ class IntelMKL:
     cmake_minimum_version = None
     cmake_packages = ["BLAS"]
     cmake_variables = {"BLA_VENDOR": "Intel10_64lp"}
-    cmake_includes = []  # For some reason, FindBLAS does not find includes
+    cmake_includes = [os.path.join(os.environ['MKLROOT'], 'include')]
```
tbennun: I would add another function called `_find_mkl_include()`, which returns `[path.join(MKLROOT, include)]` if MKLROOT is in `os.environ`, or `[]` otherwise.

dace/libraries/blas/environments/intel_mkl.py
```diff
@@ -7,12 +8,11 @@ class IntelMKL:
...
     cmake_libraries = ["mkl_rt"]
     cmake_compile_flags = []
     cmake_link_flags = []
     cmake_files = []
-
```
tbennun: Please return deleted line

dace/libraries/blas/nodes/gemm.py
orausch: @tbennun do you prefer the extra map or adding in the multiplication map? Do you know how BLAS implementations usually do it?

tests/library/gemm_test.py
```diff
@@ -130,12 +117,9 @@ def params_generator(grid):
     for implementation in Gemm.implementations:
         print("Testing {}...".format(implementation))
         try:
-            if args["grid"]:
-                for params in params_generator(param_grid):
-                    print("testing:", params)
-                    run_test(implementation, params)
-            else:
-                run_test(implementation, args)
+            for params in params_generator(param_grid):
```
orausch: This takes a while to run currently (<5 min), hopefully this doesn't spam your CI. Depending on your opinion, we can scale this back a bit (although it did help me catch the weird bug with complex).

tbennun: @orausch That's too many options (good for an `*_exhaustivetest.py`). Please choose random {alpha,beta}, one size (23x24x25 or 10x11x12 are good because they're rectangular), and maybe run all the options in transA/transB. This will end up with four reasonable tests.

dace/libraries/blas/nodes/gemm.py
tbennun: extra map is more numerically stable.

dace/libraries/blas/nodes/gemm.py
```diff
@@ -0,0 +1,244 @@
...
+        self.alpha = alpha
+        self.beta = beta
+
+    def validate(self, sdfg, state):
+        in_edges = state.in_edges(self)
+        if len(in_edges) != 3:
+            raise ValueError("Expected exactly three inputs to gemm")
+        for _, _, _, dst_conn, memlet in state.in_edges(self):
+            if dst_conn == "_a":
+                a_size = memlet.subset.size()
```
orausch: Here (and in the validation), the MatMul node squeezes the shapes. I removed this so that broadcasting works.

dace/libraries/blas/nodes/gemm.py
```diff
@@ -0,0 +1,244 @@
...
+                                   wcr_str="lambda x, y: x + y",
+                                   wcr_identity=0)
+            },
+            external_edges=True,
+            output_nodes=output_nodes)
+
+        if node.beta != 0:
+            add_program = "__y = ({} * __c) + __tmp".format(node.beta)
+
+            # manually broadcasting C to [M, N]
```
orausch: It's probably worth adding a `_broadcasted_map` function that does something similar...

dace/libraries/blas/nodes/gemm.py
```diff
@@ -0,0 +1,246 @@
...
+                         location=location,
+                         inputs={"_a", "_b", "_c"},
+                         outputs={"_y"})
+        self.dtype = dtype
+        self.transA = transA
+        self.transB = transB
+        self.alpha = alpha
+        self.beta = beta
+
+    def validate(self, sdfg, state):
```
orausch: Should we check the dtypes here? Such a check doesn't happen in matmul.

tbennun:
@orausch please merge your branch with the latest version of the einsum branch. It's now ready for review.

tbennun:
Merged, please retarget PR against master.

dace/libraries/blas/nodes/gemm.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            mul_out, mul_out_array = tmp, array_tmp = sdfg.add_temp_transient(
+                shape_y, dtype_y, storage=storage)
+
+            access_tmp = state.add_read(tmp)
+            output_nodes = {mul_out: access_tmp}
+
+        # multiplication map
+        state.add_mapped_tasklet(
+            "_MatMult_",
```
tbennun:
```suggestion
            "_GEMM_",
```

dace/libraries/blas/nodes/gemm.py
```diff
@@ -0,0 +1,276 @@
...
+        node.validate(sdfg, state)
+        if node.dtype is None:
+            raise ValueError("Data type must be set to expand " + str(node) +
+                             ".")
+        return ExpandGemmPure.make_sdfg(node, state, sdfg)
+
+
+@dace.library.node
+class Gemm(dace.graph.nodes.LibraryNode):
+    """Executes alpha * (A @ B) + beta * C. Broadcasting C is NOT supported.
```
tbennun: Is it really not supported? Some cases are, maybe which ones should be given here.

dace/libraries/blas/nodes/gemm.py
tbennun: We want to support different input and output types in MatMul, maybe here as well. It's important for some optimizations.

dace/libraries/blas/nodes/gemm.py
tbennun: MatMul squeezes the shapes to do another kind of broadcasting in certain cases. Here it is ok because GEMM must explicitly work on two 2D arrays.

tests/library/gemm_test.py
```diff
@@ -0,0 +1,165 @@
...
+    else:
+        sdfg(A=A, B=B, Y=Y)
+
+    diff = np.linalg.norm(Y_regression - Y) / (M * N)
+    print("Difference:", diff)
+    assert diff <= 1e-5
+
+
+if __name__ == "__main__":
+    param_grid_trans = dict(
```
tbennun: please move this to a `def test_*()` so that pytest can work with this file.

dace/libraries/blas/nodes/gemm.py
orausch: Yup, that was an old docstring before I decided to add broadcasting to the library node

orausch:
Quick question on style: do you prefer to squash branches before merging? (I know some people are very adverse to force pushes)

tbennun:
> Quick question on style: do you prefer to squash branches before merging? (I know some people are very adverse to force pushes)

Squashing is better in your case, yes.

Issue #180, opened 2020-03-24
Cannot run two Intel FPGA programs in same executable


definelicht:
When compiling multiple SDFGs that use the Intel FPGA runtime in the same DaCe program, the second segfaults.

This seems to be an issue with hlslib's GlobalContext:

```
==29276== Invalid read of size 4
==29276==    at 0x3B81A90D: _M_find_before_node (hashtable.h:1548)
==29276==    by 0x3B81A90D: _M_find_node (hashtable.h:642)
==29276==    by 0x3B81A90D: find (hashtable.h:1425)
==29276==    by 0x3B81A90D: find (unordered_map.h:920)
==29276==    by 0x3B81A90D: GlobalContext (OpenCL.h:376)
==29276==    by 0x3B81A90D: __dace_init_intel_fpga (async_test_True.cpp:9)
==29276==    by 0x70D4DAD: ffi_call_unix64 (in /usr/lib/x86_64-linux-gnu/libffi.so.6.0.4)
==29276==    by 0x70D471E: ffi_call (in /usr/lib/x86_64-linux-gnu/libffi.so.6.0.4)
==29276==    by 0x6EC15C3: _ctypes_callproc (in /usr/lib/python3.6/lib-dynload/_ctypes.cpython-36m-x86_64-linux-gnu.so)
==29276==    by 0x6EC1D53: ??? (in /usr/lib/python3.6/lib-dynload/_ctypes.cpython-36m-x86_64-linux-gnu.so)
==29276==    by 0x5A067D: PyObject_Call (in /usr/bin/python3.6)
==29276==    by 0x50D965: _PyEval_EvalFrameDefault (in /usr/bin/python3.6)
==29276==    by 0x508244: ??? (in /usr/bin/python3.6)
==29276==    by 0x509641: _PyFunction_FastCallDict (in /usr/bin/python3.6)
==29276==    by 0x595310: ??? (in /usr/bin/python3.6)
==29276==    by 0x5A067D: PyObject_Call (in /usr/bin/python3.6)
==29276==    by 0x50D965: _PyEval_EvalFrameDefault (in /usr/bin/python3.6)
==29276==  Address 0x8 is not stack'd, malloc'd or (recently) free'd
==29276== Process terminating with default action of signal 11 (SIGSEGV)
==29276==  Access not within mapped region at address 0x8
==29276==    at 0x3B81A90D: _M_find_before_node (hashtable.h:1548)
[...]

PR #181, opened 2020-03-24
Fix FPGA context destruction

definelicht:
This PR addresses #180, which resulted in DaCe programs crashing when running multiple FPGA SDFGs, due to the singleton object for the FPGA context not being properly cleaned up, causing segfaults in the second program.

Following the pattern of CUDA, create the context as an extern object.


dace/config_schema.yml
```diff
@@ -342,7 +342,16 @@ required:
                         type: str
                         title: Kernel flags 
                         description: High-level synthesis C++ flags 
-                        default: "-fp-relaxed -cl-no-signed-zeros -cl-fast-relaxed-math -cl-single-precision-constant"
+                        default: "-fp-relaxed -cl-no-signed-zeros -cl-fast-relaxed-math -cl-single-precision-constant -fno-burst-interleaving=default"
```
TizianoDeMatteis: The flag is `-no-interleaving=default`

definelicht: Fixed

PR #182, opened 2020-03-24
Automatically build kernels when executing FPGA programs

definelicht:
To achieve this, we change `add_custom_target` to `add_custom_command`, so that kernels are not rebuilt if no files have changed. 

This also means we don't need a three-stage process in the Intel FPGA test anymore.


definelicht:
@tbennun I've now fixed two issues that are not related to the changes on this branch, and also occur on master. It seems master is broken, despite Pauli saying otherwise...

PR #186, opened 2020-03-25
Even more support for half-precision and CUDA events

tbennun:
- [x] CUDA events can benchmark tasklets


tbennun:
Deprecated. Resolved in other PRs.

Issue #59, opened 2019-12-09
Change how FPGA compiler is specified in the DaCe config


definelicht:
Right now, we are specifying the compiler with an executable name (e.g., "xocc"). Instead, we should pass the root folder of the installation (e.g., `/opt/Xilinx/Vitis/2019.2`). Furthermore, we should default to not setting this in the config, and instead letting the CMake script find the local installation.

definelicht:
@tbennun You want to update CUDA also?

tbennun:
Yes, it makes more sense this way, MPI too.

tbennun:
Fixed

PR #189, opened 2020-03-29
Let CMake detect toolkits/compilers by default

tbennun:
Addresses issue #59 and fixes cross-platform (i.e. Windows) builds without changing environment variables.

Still missing FPGA part @definelicht.


dace/config_schema.yml
```diff
@@ -132,7 +112,7 @@ required:
 
             allow_shadowing:
                 type: str
-                default: false 
+                default: true
```
definelicht: Is this related, or did it just sneak in here? :-)

definelicht:
Nice job Tal, should've done this a long time ago. I'll do the FPGA part.

dace/config_schema.yml
tbennun: Snuck in, can revert but I think it should be in master until the new codegen appears

PR #198, opened 2020-04-03
FPGA transform WCR fix

definelicht:


definelicht:
@manuelburger Could you see if this fixes your problem?

manuelburger:
> @manuelburger Could you see if this fixes your problem?

If I run the example provided with the issue I opened, I get a different Traceback now, but still ends in a `KeyError` :/

tbennun:
@manuelburger could you please share the full traceback?

manuelburger:
> @manuelburger could you please share the full traceback?

```
WARNING: FPGATransformSDFG::can_be_applied triggered a KeyError exception: '_x'
Traceback (most recent call last):
  File "dotNestedSDFG_example.py", line 132, in <module>
    compiledSDFG = test_sdfg.compile(optimizer=False)
  File "/home/burgerm/dace/dace/sdfg.py", line 1733, in compile
    sdfg.save(os.path.join('_dotgraphs', 'program.sdfg'))
  File "/home/burgerm/dace/dace/sdfg.py", line 1237, in save
    fp.write(dace.serialize.dumps(self.to_json()))
  File "/home/burgerm/dace/dace/sdfg.py", line 263, in to_json
    for k, v in sorted(self.undefined_symbols(True).items())
  File "/home/burgerm/dace/dace/sdfg.py", line 871, in undefined_symbols
    return undefined_symbols(self, self, include_scalar_data)
  File "/home/burgerm/dace/dace/sdfg.py", line 4335, in undefined_symbols
    obj.data_symbols(True) if isinstance(obj, SDFG) else obj.data_symbols(
  File "/home/burgerm/dace/dace/sdfg.py", line 830, in data_symbols
    symbols.update(state.data_symbols())
  File "/home/burgerm/dace/dace/sdfg.py", line 2590, in data_symbols
    return data_symbols(self)
  File "/home/burgerm/dace/dace/sdfg.py", line 4317, in data_symbols
    ddesc = d.desc(sdfg)
  File "/home/burgerm/dace/dace/graph/nodes.py", line 230, in desc
    return sdfg.arrays[self.data]
KeyError: '__z1_out'
```

definelicht:
@manuelburger Try again :-)

manuelburger:
> @manuelburger Try again :-)

Seems to work now! :)

dace/sdfg.py
```diff
@@ -2317,6 +2317,51 @@ def traverse(node):
...
+        """
+    if node.access == dace.dtypes.AccessType.ReadWrite:
+        raise NotImplementedError("Access node must be read or write only")
+    curr_node = node
+    curr_sdfg = sdfg
+    trace = [(node, state, sdfg)]
+    while curr_sdfg.parent is not None:
+        curr_state = curr_sdfg.parent
+        # Find the nested SDFG containing ourself in the parent state
+        for nested_sdfg in curr_state.nodes():
```
tbennun, line 2338: missing an `else` clause (what if the node was not found?)

dace/sdfg.py
```diff
@@ -2317,6 +2317,51 @@ def traverse(node):
...
+    """ Given an AccessNode in a nested SDFG, trace the accessed memory
+        back to the outermost scope in which it is defined.
+
+        :param node: An access node.
+        :param state: State in which the access node is located.
+        :param sdfg: SDFG in which the access node is located.
+        :return: A list of scopes (node, state, sdfg) in which the given
+                 data is accessed, from outermost scope to innermost scope.
+        """
+    if node.access == dace.dtypes.AccessType.ReadWrite:
```
tbennun, line 2330: Do you really want to rely on access types like this? there might be cases that an output node is accessed as readwrite in the middle of the nested SDFG (first written, then read, then written to again).

definelicht: I don't need this and I don't have a test case for it, so I won't spend time on it now :-) If we ever need it we can add this feature

dace/sdfg.py
definelicht: Fixed with d082a61

PR #202, opened 2020-04-09
SDFV: Draggable graph elements

phschaad:
- Nodes and edges can both be dragged
- Moving/panning of the view was moved from primary mouse button to secondary mouse button (for now?)
- Movement of nodes is restricted to remain within their parent element, e.g. state or nested SDFG

### DIFF
--- a/diode/webclient/renderer.js
+++ b/diode/webclient/renderer.js
@@ -205,12 +205,15 @@ class CanvasManager {
         this.user_transform = this.user_transform.translate(x / this.user_transform.a, y / this.user_transform.d);
     }
 
-    translate_element(el, x, y, g) {
+    translate_element(el, x, y) {
         this.stopAnimation();
+
+        const dx = x / this.user_transform.a;
+        const dy = y / this.user_transform.d;
 
         // Calculate the box to bind the element to. This is given by
         // the parent element, i.e. the element where out to-be-moved
-        // element is contained within.
+        // element is contained within
         const parent_layout = el.sdfg.nodes[0].attributes.layout;
         const min_x = (parent_layout.x - (parent_layout.width / 2)) +
             (el.width / 2);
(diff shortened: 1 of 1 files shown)


diode/webclient/renderer.js
```diff
@@ -205,6 +205,161 @@ class CanvasManager {
...
+
+            // Make sure we don't move our element outside its parent's
+            // bounding box
+            const target_x = el.x + dx;
+            const target_y = el.y + dy;
+            if (target_x <= min_x || target_x >= max_x) dx = 0;
+            if (target_y <= min_y || target_y >= max_y) dy = 0;
+        }
+
+        if (el.data.type && (el.data.type === 'Memlet' || el.data.type === 'InterstateEdge')) {
```
tbennun, line 276: you can check if it's a `typeof Edge` as far as I remember

diode/webclient/renderer.js
```diff
@@ -205,6 +205,161 @@ class CanvasManager {
...
+            if (el.points[2]) {
+                // Only allow dragging, if the memlet is 'making a curve'
+                el.points[1].x += dx;
+                el.points[1].y += dy;
+            }
+            // Don't do any of the other stuff, it doesn't apply here
```
tbennun, line 282:
```suggestion
            // The rest of the method doesn't apply here
```

diode/webclient/renderer.js
```diff
@@ -205,6 +205,161 @@ class CanvasManager {
...
+                node.out_connectors.forEach(c => {
+                    c.x += dx;
+                    c.y += dy;
+                });
+        }
+
+        // Allow recursive translation of nested SDFGs
+        function translate_recursive(ng) {
+            ng.nodes().forEach(state_id => {
+                const state = ng.node(state_id);
```
tbennun, line 307: why not use `let`?

diode/webclient/renderer.js
```diff
@@ -1183,7 +1412,7 @@ class SDFGRenderer {
                 return false;
             } else {
                 this.drag_start = null;
-                if (event.buttons & 1)
+                if (event.buttons & 1 || event.buttons & 2)
```
tbennun, line 1415: are you sure it should be 2 rather than 4 here?

diode/webclient/renderer.js
phschaad: I assumed `prefer-const`, is let generally preferred?

tbennun: you can use const if it is correct. I normally don't.

diode/webclient/renderer.js
phschaad: assuming you meant `instanceof`, yes that works, thank you!

phschaad:
Thank you!

PR #204, opened 2020-04-10
Enable non-strided maps for the Vectorization transformation.

tbennun:


tbennun:
@definelicht @TizianoDeMatteis please review this and see if it generates the right code for FPGAs. Thanks!

manuelburger:
I have tested this and if I now refactor my vectorized loops from `0:n:w` to `0:n/w` and access them with `i * w`, the accesses to memory are indexed as follows:
```
auto __inCon = dace::ArrayViewIn<float, 0, 4, 1> (fpga_y1 + i);
```
So no more `/w` in the array access. Synthesis now properly detects the memory bursts.

tbennun:
Great, thanks @manuelburger !

PR #208, opened 2020-04-12
Vector length conversion for within tasklets for Intel FPGA

definelicht:
We need this for StencilFlow. Implements the minimum required for it to work. It's pretty hacky and a bit of a mess, feel free to suggest how it can be improved in a reasonable time frame.

### DIFF
--- a/tests/intel_fpga_test.sh
+++ b/tests/intel_fpga_test.sh
@@ -44,7 +44,7 @@ run_sample() {
     echo -e ${3} | $PYTHON_BINARY ${1}.py ${@:4}
 
     if [ $? -ne 0 ]; then
-        bail "$1 (${RED}Wrong emulation result${NC})"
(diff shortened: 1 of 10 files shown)


tests/intel_fpga_test.sh
```diff
@@ -44,7 +44,7 @@ run_sample() {
     echo -e ${3} | $PYTHON_BINARY ${1}.py ${@:4}
 
     if [ $? -ne 0 ]; then
-        bail "$1 (${RED}Wrong emulation result${NC})"
```
tbennun, line 47: Why remove?

definelicht: This is printed no matter why the executable fails, 90% of the time it's _not_ due to wrong emulation results ;-) We could only reintroduce something like this if we had some convention for return codes

PR #210, opened 2020-04-16
Enable/disable automatically building bitstreams

definelicht:
Boolean flag in the config can enable/disable automatically building missing bitstreams for both Xilinx and Intel.


dace/codegen/targets/intel_fpga.py
```diff
@@ -67,12 +67,15 @@ def cmake_options():
         target_board = Config.get("compiler", "intel_fpga", "board")
         enable_debugging = ("ON" if Config.get_bool(
             "compiler", "intel_fpga", "enable_debugging") else "OFF")
+        autobuild = ("ON" if Config.get_bool(
```
TizianoDeMatteis, line 70: So are you disabling autobuild by default?

definelicht: Yes. Do you think this makes sense? It will still autobuild emulation kernels, just not builds that take many hours.

PR #213, opened 2020-04-17
Improve type inference

definelicht:
Use a more intelligent type inference system (would still be better to rely on something external for this?).

### DIFF
--- a/dace/codegen/cppunparse.py
+++ b/dace/codegen/cppunparse.py
@@ -811,43 +811,16 @@ class CPPUnparser:
         # double-complex (128-bit) otherwise
         dtype = self.dtype or 'dace::complex128'
 
-        if six.PY3:
-            if repr_n.endswith("j"):
-                self.write("%s(0, %s)" %
-                           (dtype, repr_n.replace("inf", INFSTR)[:-1]))
-            else:
-                self.write(repr_n.replace("inf", INFSTR), infer_type)
-                # If the number has a type, use it
-                if isinstance(t.n, np.uint):
-                    return dace.dtypes.typeclass(
-                        np.uint32) if infer_type else None
-                elif isinstance(t.n, np.int):
-                    return dace.dtypes.typeclass(
-                        np.int32) if infer_type else None
-                elif isinstance(t.n, np.float):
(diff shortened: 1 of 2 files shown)


tbennun:
have you looked at `typed-ast`? might give us what we need?

dace/codegen/cppunparse.py
```diff
@@ -811,43 +811,16 @@ def _Num(self, t, infer_type=False):
...
+            self.write(repr_n.replace("inf", INFSTR), infer_type)
+            # If the number has a type, use it
+            if infer_type:
+                return dace.dtypes.typeclass(type(t.n))
             else:
-                # Substitute overflowing decimal literal for AST infinities.
-                self.write(repr_n.replace("inf", INFSTR))
-
-            if repr_n.startswith("-"):
-                self.write(")")
```
tbennun, line 850: what happened to this?

dace/dtypes.py
```diff
@@ -260,159 +260,43 @@ def __repr__(self):
...
+    # Both are integers
+    if numpy.issubdtype(lhs_, numpy.integer) and numpy.issubdtype(
+            rhs_, numpy.integer):
+        # If one byte width is larger, use it
+        if lhs_.itemsize > rhs_.itemsize:
+            return lhs
+        elif lhs_.itemsize < rhs_.itemsize:
+            return rhs
+        # Sizes are the same
+        if numpy.issubdtype(lhs_, numpy.unsignedinteger):
```
tbennun, line 284: what about the other side of this?

dace/codegen/cppunparse.py
definelicht: It's above, just diff makes it look like it was deleted and re-added

dace/dtypes.py
definelicht: What do you mean?

tbennun: if `rhs_` is unsigned but `lhs_` isn't.

definelicht: Then we have to return `rhs_`, which is what we do

definelicht: It's in the comments :-)

PR #216, opened 2020-04-20
FPGA codegen, fix error message for different veclen to same memory

manuelburger:
The variable `nodedesc` was not initialized in this scope and threw an error.


definelicht:
Whoops, error inside an error. Thanks!

PR #218, opened 2020-04-21
Numpy Frontend Updates

orausch:
- Refractor existing replacements into new separate file
- Add `NestedCall` class to create nested replacements
- Add broadcasting for binary ops
- Add numpy replacements for `max`, `min`, `sum`, `argmax`, `argmin`
- Add `dace.elementwise`
- Add `diode.sdfv.view` and `SDFG.view()`


orausch:
Unfortunately the jenkins build on `pauli` requires a login, so I can't see why it's failing.

tbennun:
@orausch The build actually passed, but there is a sporadic error on the server that fails the build.

diode/sdfv.py
```diff
@@ -11,6 +12,49 @@
 import jinja2
 
 
+def view(sdfg, filename=None):
```
orausch, line 15: This is just a change from my autodiff branch: this allows you to view sdfgs from the terminal;
```python
from diode.sdfv import view
view(sdfg) # this will write the html to a tempfile and open it in your browser
```

dace/frontend/python/nested_call.py
```diff
@@ -0,0 +1,30 @@
+import dace
+from dace.sdfg import SDFG, SDFGState
+
+class NestedCall():
```
orausch, line 4: When writing the pytorch function replacements, I plan to reuse the existing replacements. To do this, I would propose this `NestedCall` object which allows you to call other replacements while building a replacement

orausch: It also works with replacements that return stuff, i.e. `result = nest(_func)(....)` should also work.

tbennun: @orausch please add an example use to the documentation, no one will read these comments once the code is merged :)

dace/frontend/python/newast.py
```diff
@@ -1,37 +1,69 @@
...
 import dace
 from dace import data, dtypes, subsets, symbolic, sdfg as sd
 from dace.config import Config
-from dace.frontend.common import op_impl
 from dace.frontend.common import op_repository as oprepo
 from dace.frontend.python import astutils
 from dace.frontend.python.astutils import ExtNodeVisitor, ExtNodeTransformer
 from dace.frontend.python.astutils import rname
+from dace.frontend.python import nested_call
+from dace.frontend.python.memlet_parser import DaceSyntaxError, parse_memlet, pyexpr_to_symbolic, ParseMemlet, inner_eval_ast, MemletExpr
```
tbennun, line 16: needs `yapf` and some parentheses around the imports (so that they split lines)

dace/frontend/python/memlet_parser.py
```diff
@@ -0,0 +1,236 @@
...
+from dace.frontend.python import astutils
+from dace.frontend.python.astutils import rname
+from dace.memlet import Memlet
+from dace.symbolic import pystr_to_symbolic
+
+MemletType = Union[ast.Call, ast.Attribute, ast.Subscript, ast.Name]
+MemletExpr = namedtuple('MemletExpr',
+                        ['name', 'accesses', 'wcr', 'wcr_identity', 'subset'])
+
+class DaceSyntaxError(Exception):
```
tbennun, line 17: not sure that it belongs here. Is there a reason why it's not in newast or a common.py?

dace/frontend/python/newast.py
```diff
@@ -3589,6 +2759,10 @@ def visit_Call(self, node: ast.Call):
 
         result = func(self.sdfg, self.last_state, *args, **keywords)
 
+        if isinstance(result, tuple) and type(result[0]) is nested_call.NestedCall:
```
tbennun, line 2762: why not `isinstance(result[0], nested_call.NestedCall)`?

dace/frontend/python/replacements.py
```diff
@@ -0,0 +1,789 @@
...
+    nest(_reduce)(  # comment for yapf
+        "lambda x, y: _val_and_idx(val={}(x.val, y.val), idx=(x.idx if x.val {} y.val else y.idx))"
+        .format(func, '>' if func == 'max' else '<'),
+        structs,
+        out_array=reduced_structs,
+        axis=axis)
+
+    # map to int64
+    out, outarr = sdfg.add_temp_transient(sdfg.arrays[reduced_structs].shape,
+                                          dace.int64)
```
tbennun, line 458: why int64? I would use a result_type flag, that is int32 by default.

dace/frontend/python/replacements.py
```diff
@@ -0,0 +1,789 @@
...
+
+def _is_scalar(sdfg: SDFG, arrname: str):
+    """ Checks whether array is pseudo-scalar (shape=(1,)). """
+    shape = sdfg.arrays[arrname].shape
+    if len(shape) == 1 and shape[0] == 1:
+        return True
+    return False
+
+
+def _inverse_dict_lookup(dict: Dict[str, Any], value: Any):
```
tbennun, line 630: so this function appears twice now. Yet another reason for common.py with that function too

diode/sdfv.py
tbennun: Nice

dace/frontend/python/newast.py
orausch: That would break things like reduce, which return `[]`, because `result[0]` would give an index error.

PR #226, opened 2020-04-28
Convert Reduce into a library node

tbennun:


.travis.yml
```diff
@@ -1,4 +1,4 @@
-dist: xenial
+dist: bionic
```
definelicht, line 1: What does this have to do with converting reduce into a library node? ;-)

tbennun: OpenMP implementation doesn't work unless GCC is 6/7 or newer

PR #227, opened 2020-04-29
Property additions/removals

tbennun:
- [x] Remove `map.flatten`
- [x] `CodeProperty` returns `CodeBlock`, rather than a string
- [x] Add allocation lifetime and alignment

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -48,23 +48,25 @@ class Data(object):
                        choices=dace.dtypes.StorageType,
                        default=dace.dtypes.StorageType.Default,
                        from_string=lambda x: dtypes.StorageType[x])
+    lifetime = Property(dtype=dace.dtypes.AllocationLifetime,
+                        desc='Data allocation span',
+                        choices=dace.dtypes.AllocationLifetime,
+                        default=dace.dtypes.AllocationLifetime.Scope,
+                        from_string=lambda x: dtypes.AllocationLifetime[x])
     location = DictProperty(
         key_type=str,
         value_type=None,
         desc='Full storage location identifier (e.g., rank, GPU ID)')
-    toplevel = Property(dtype=bool,
-                        desc="Allocate array outside of state",
-                        default=False)
     debuginfo = DebugInfoProperty(allow_none=True)
 
-    def __init__(self, dtype, shape, transient, storage, location, toplevel,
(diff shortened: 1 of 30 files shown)


dace/data.py
```diff
@@ -253,6 +259,11 @@ class Array(Data):
                          desc='This pointer may alias with other pointers in '
                          'the same function')
 
+    alignment = Property(dtype=int,
```
definelicht, line 262: This doesn't actually change anything for I/O arrays, right? In those cases, we should sanity check than I/O arrays are aligned to what is expected, otherwise we could end up with horrible sporadic runtime errors.

dace/dtypes.py
```diff
@@ -68,6 +68,17 @@ class ReductionType(aenum.AutoNumberEnum):
...
 
 
+@extensible_enum
+class AllocationLifetime(aenum.AutoNumberEnum):
+    """ Options for allocation span (when to allocate/deallocate) of data. """
+
+    Scope = ()  # Allocated/Deallocated on innermost scope start/end
+    State = ()  # Allocated throughout the containing state
+    SDFG = ()  # Allocated throughout the innermost SDFG (possibly nested)
+    TopLevel_SDFG = ()  # Allocated throughout the entire program (outer SDFG)
```
definelicht, line 78: Maybe "Global"? Would be more intuitive for newcomers

dace/data.py
tbennun: Correct. If referring to non-transient memory, alignment (same as the strides) has to match the external memory attributes. We cannot really verify this until SDFG call-site, and we try to make that as low overhead as possible.

dace/dtypes.py
tbennun: Sounds good, renamed 👍

dace/data.py
definelicht: It's a very cheap check though: `ptr % alignment == 0`, and it only emitted if alignment is actually set. Seems worth it?

tbennun: Sure, added to TODO

definelicht: For this PR or TODO(later)? Kappa

tbennun: As part of the new code generator Project. So later, but not that much later (once we actually use these properties to generate code).

PR #241, opened 2020-04-30
Innermost loop check

definelicht:
Make the check whether a loop is innermost more sophisticated by taking nested SDFGs and unrolled maps into account.

@manuelburger

### DIFF
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -880,6 +880,22 @@ class FPGACodeGen(TargetCodeGenerator):
     def _generate_PipelineEntry(self, *args, **kwargs):
         self._generate_MapEntry(*args, **kwargs)
 
+    def _is_innermost(self, scope, scope_dict):
+        to_search = list(scope)
+        while len(to_search) > 0:
+            x = to_search.pop()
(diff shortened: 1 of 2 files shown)


definelicht:
@manuelburger This fixes the extra pipelines in the SDFG you sent

dace/codegen/targets/fpga.py
```diff
@@ -880,6 +880,22 @@ def copy_memory(self, sdfg, dfg, state_id, src_node, dst_node, edge,
     def _generate_PipelineEntry(self, *args, **kwargs):
         self._generate_MapEntry(*args, **kwargs)
 
+    def _is_innermost(self, scope, scope_dict):
+        to_search = list(scope)
+        while len(to_search) > 0:
+            x = to_search.pop()
+            if (isinstance(x, dace.graph.nodes.MapEntry)
+                    or isinstance(x, PipelineEntry)):
```
tbennun, line 888: Check for `EntryNode` or at least isinstance with a tuple as a second argument.

definelicht: I want to be explicit here, it should only be an entry node that I know for sure that we would pipeline. But I'll add the tuple

definelicht: Fixed

PR #247, opened 2020-05-01
Make CI tests shorter

tbennun:


.travis.yml
```diff
@@ -26,6 +26,7 @@ script:
   - export PYTHONPATH=`pwd`:$PYTHONPATH
   - export COVERAGE_RCFILE=`pwd`/.coveragerc
   - export PYTHON_BINARY="coverage run --source=`pwd`/dace --parallel-mode"
+  - export DACE_optimizer_interface=" "
```
definelicht, line 29: Did you check if this now requires the optimizer interface to be exported in tests that need it? `intel_fpga_test.sh` comes to mind.

tbennun: Yep, this is happening in the travis tests (CPU only). The Jenkins tests don't have this setting set.

definelicht: Cool

PR #253, opened 2020-05-06
Don't pipeline degenerate loops

definelicht:
This PR fixes two things:
- If a loop is fully degenerate, don't insert pipeline pragmas, since no loops are actually emitted.
- When checking if a loop is innermost, don't take degenerate loops into account.


definelicht:
@manuelburger This should fix your issue with GEMV

Issue #237, opened 2020-04-29
Infer type from symbolic expression


tbennun:
It would be great if we reused our existing type inference infrastructure to get a type from symbolic expressions (which already have typed symbols).

We have a preliminary implementation in symbolic.py (`symtype`), but it doesn't infer anything at the moment.

definelicht:
I need type inference for an expression AST, _including vector length_.

PR #257, opened 2020-05-08
Type inference

TizianoDeMatteis:
- Type inference functionality has been moved outside of `CPPUnparser`.
- the `type_inference` module (`dace/codegen/tools`) provides a method `infer` that takes in input some code and a dictionary of already defined symbols (with type) and returns a dictionary of inferred symbols with types
- a new config property is defined: "default_data_types" specifies if default data types must follow Python (`int`s and `float`s represented with 64 bits) or C semantic (`int`s and `float`s represented with 32 bits)

Solves #237


dace/codegen/tools/type_inference.py
```diff
@@ -0,0 +1,580 @@
...
+import ast
+from dace import dtypes
+from dace.codegen import cppunparse
+from dace.symbolic import SymExpr
+from dace.symbolic import symstr
+from sympy import Basic
+import sys
+
+
+def infer(code, symbols):
```
definelicht, line 19: Suggestion: split into two inferfaces:
```python
infer_types(code, symbols)  # Returns a dictionary
infer_expr_type(expr_code, symbols)  # Returns a single type
```
Quite often you are just interested in the final type of an expression, so it would be convenient to have the bottom one to save boilerplate

dace/codegen/tools/type_inference.py
```diff
@@ -0,0 +1,580 @@
...
+    """
+    Perform type inference on the given code
+    :param code: a string, AST, or symbolic expression
+    :param symbols: already known symbols with their types. This is a dictionary "symbol name" -> dytpes.typeclass
+    :return: a dictionary "symbol name" -> dytpes.typeclass of inferred symbols
```
definelicht, line 24:
```suggestion
    :return: a dictionary "symbol name" -> dtypes.typeclass of inferred symbols
```

dace/codegen/tools/type_inference.py
```diff
@@ -0,0 +1,580 @@
+"""
+    Type inference: traverses code and returns types for all undefined symbols according to C semantics
+    infer() has a lenient implementation: if something it not inferred (for example an unsupported construct) it will not
+        return anything and it will not produce errors
```
definelicht, line 4: Are we sure we don't want to fail if something cannot be inferred? For OpenCL we need to fail in this case. What will the dictionary contain in this case, `None`?

dace/codegen/tools/type_inference.py
```diff
@@ -0,0 +1,580 @@
...
+    return dtypes.typeclass(np.result_type(t.value).name)
+
+
+def _Num(t, symbols, inferred_symbols):
+    # get the minimum between the minimum type needed to represent this number and the corresponding default data types
+    # e.g., if num=1, then it will be represented by using the default integer type (int32 if C data types are used)
+    return dtypes.result_type_of(
+        dtypes.typeclass(type(t.n)),
+        dtypes.typeclass(np.min_scalar_type(t.n).name))
+    np.promote_types(np.min_scalar_type(t.n), dtypes.typeclass(type(t.n)).type)
```
definelicht, line 271: Is something crazy happening with indentation here, or is this last statement never reached?

dace/config_schema.yml
```diff
@@ -111,7 +111,17 @@ required:
                 description: >
                    Allowing shadowing of variables in the code (reduces
                    exceptions to warnings when shadowing is encountered).
-                   
+
+            default_data_types:
+                type : str
+                default: Python default
```
definelicht, line 117: Maybe reduce to just "Python" and "C"? It's awkward to write this from the command line:
`DACE_default_data_types=Python\ default python3 samples/tizicoin.py`

dace/codegen/tools/type_inference.py
TizianoDeMatteis: Expression or assignment? E.g., `target = a+b+c ` and you want the type of target

dace/codegen/tools/type_inference.py
TizianoDeMatteis: Probably it will contain `None`.

In `cppunparse` we fail if we try to unparse a statement that is not expressible in C++.
Don't  know if it is ok to fail also here, because then we will  tailor type inference on C++

dace/codegen/tools/type_inference.py
definelicht: Both!

dace/codegen/tools/type_inference.py
definelicht: Do you mean C++ or C? If it's not expressible in C, we want to fail I think?

tbennun: I think it's fine to return `None`s for un-inferrable variables. Other interface users may replace that with e.g., default types or return a notification to the user.

dace/config_schema.yml
tbennun: Agreed

dace/dtypes.py
```diff
@@ -247,10 +247,22 @@ def __init__(self, wrapped_type):
                 wrapped_type = getattr(numpy, wrapped_type)
             except AttributeError:
                 raise ValueError("Unknown type: {}".format(wrapped_type))
+
+        config_data_types = Config.get('compiler', 'default_data_types')
         if wrapped_type is int:
-            wrapped_type = numpy.int64
+            if config_data_types == 'Python default':
```
tbennun, line 253: I would check `config_data_types.lower() == 'python'` to be sure with weird configuration inputs

tbennun: applies to the rest too

tests/type_inference_test.py
```diff
@@ -0,0 +1,193 @@
...
+import ast
+
+
+class TestTypeInference(unittest.TestCase):
+    def testSimpleAssignement(self):
+        # simple assignment tests
+
+        #bool
+        code_str = "value=True"
+        inf_symbols = type_inference.infer(code_str, {})
```
tbennun, line 15: perhaps infer should have a default argument for symbols (`=None`, **not** `={}`) just to make things cleaner.

tests/type_inference_test.py
```diff
@@ -0,0 +1,193 @@
...
+        self.assertEqual(inf_symbols["value"], dtypes.typeclass(np.int8))
+
+        # aug assignment
+        code_str = "value += 1.1"
+        inf_symbols = type_inference.infer(code_str, {})
+        self.assertEqual(inf_symbols["value"], dtypes.typeclass(float))
+
+        # annotated assignments
+        # in this case conversion is stricter (int-> int32)
+        code_str = "value : int  = 1"
```
tbennun, line 41: how do we know `int` is int32?

tests/type_inference_test.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def testDefaultDataTypes(self):
+        # check that configuration about defult data types is enforced
+        config_data_types = Config.get('compiler', 'default_data_types')
+
+        code_str = """value1 = 10
+value2=3.14
+value3=5000000000"""
+        inf_symbols = type_inference.infer(code_str, {})
+        if config_data_types == "Python default":
```
tbennun, line 179: remember to change the strings here if you change the configuration value

tests/type_inference_test.py
```diff
@@ -0,0 +1,193 @@
...
+            self.assertEqual(inf_symbols["value1"], dtypes.typeclass(np.int64))
+            self.assertEqual(inf_symbols["value2"],
+                             dtypes.typeclass(np.float64))
+        elif config_data_types == "C default":
+            self.assertEqual(inf_symbols["value1"], dtypes.typeclass(np.int32))
+            self.assertEqual(inf_symbols["value2"],
+                             dtypes.typeclass(np.float32))
+
```
tbennun, line 187: add an `else: raise NameError('Unsupported config data type')` or something similar

dace/codegen/tools/type_inference.py
tbennun: yep, and I would make symbols an optional argument (`symbols=None` and then inside do `symbols = symbols or {}` to avoid weird bugs with Python default arguments)

tests/type_inference_test.py
TizianoDeMatteis: Because for annotate assignments it uses (defined in `cppunparse._py2c_typeconversion`)
```
_py2c_typeconversion = {
    "uint": dace.dtypes.typeclass(np.uint32),
    "int": dace.dtypes.typeclass(np.int32),
    "float": dace.dtypes.typeclass(np.float32),
    "float64": dace.dtypes.typeclass(np.float64)
}

```

dace/codegen/cppunparse.py
```diff
@@ -683,37 +656,37 @@ def _generic_For(self, t, is_async=False):
         if t.orelse:
```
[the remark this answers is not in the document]
TizianoDeMatteis: do you mean the loop variable? like `for i in range(10)` understand the type of `i`?

tests/type_inference_test.py
tbennun: should it have a different definition than python? when I write `x: int = 4` I expect the value to be a Python int.

PR #273, opened 2020-05-19
Free symbol cleanup and subgraph API

tbennun:
- [x] Clean up all versions of `undefined_symbols` (Convert everything to `free_symbols`)
    - [x] `scalar_parameters`
    - [x] `interstate_symbols`/`data_symbols`/`scope_symbols`/`all_symbols`
    - [x] `defined_vars`
    - [x] `symbols_defined_at`
    - [x] `local_transients`
- [x] Each element should have a "free_symbols" function.
    - [x] Node 
    - [x] Edge
    - [x] Memlet
    - [x] State, Scope / State Subgraph
    - [x] SDFG
- [x] API: Function arguments/signature for SDFG, state subgraph
    - [x] `SDFG.arglist`
    - [x] `StateGraphView.arglist`
    - [x] ~~`local_transients` should be replaced with `top_level_transients(scope)` - `scope.arglist`~~


dace/sdfg/replace.py
```diff
@@ -0,0 +1,92 @@
...
+        :param name: Name to find.
+        :param new_name: Name to replace.
+    """
+    symrepl = {
+        symbolic.symbol(name):
+        symbolic.symbol(new_name) if isinstance(new_name, str) else new_name
+    }
+
+    # Replace in node properties
+    for node in subgraph.nodes():
```
definelicht, line 41: How does this work for nested SDFGs?

dace/sdfg/scope.py
```diff
@@ -0,0 +1,235 @@
...
+    curnode = other_node
+    nodescope = sdict[node]
+    while curnode is not None:
+        curnode = sdict[curnode]
+        if curnode == nodescope:
+            return True
+    return False
+
+
+def is_devicelevel(sdfg: 'dace.sdfg.SDFG', state: 'dace.sdfg.SDFGState',
```
definelicht, line 156: If this is for GPU only, make this clear from the function name

dace/subsets.py
```diff
@@ -62,6 +63,11 @@ def coord_at(self, i):
         """
         raise NotImplementedError
 
+    @property
+    def free_symbols(self) -> Set[str]:
+        """ Returns a set of undefined symbols in this subset. """
+        raise NotImplementedError
```
definelicht, line 69: Minor: you can make this a class method and print the class which hasn't implemented it

dace/sdfg/scope.py
tbennun: I want it to be any device-level (the API existed before and now moved to this file)

dace/sdfg/replace.py
tbennun: No need to replace into nested SDFGs, as the `symbol_mapping` ensures separate symbol systems and the data containers are different.

dace/sdfg/scope.py
tbennun: Fixed

dace/subsets.py
tbennun: While I cannot make this a class method (as it's a property), the latest commit prints out the class name that does not implement this method.

PR #277, opened 2020-05-29
Consider strides in BLAS nodes and global symbols as free symbols

tbennun:


definelicht:
Maybe we should add an explicit test for this?

manuelburger:
Also some bugs with regards to types. Here a float is recasted to int.
[bugTest.zip](https://github.com/spcl/dace/files/4700807/bugTest.zip)

tbennun:
@manuelburger This bug is unrelated to the BLAS nodes in this PR, but will be addressed (with your sample as a unit test) in the next PR (the "In Progress" cards in https://github.com/spcl/dace/projects/2)

Thanks!

tbennun:
@manuelburger updating you that this issue was fixed. Thanks!

PR #291, opened 2020-06-22
Memlet API revamp

tbennun:
- [x] Memlet API revamp
    * EmptyMemlet is removed in favor of Memlet() and `memlet.is_empty()`
    * Use legacy `Memlet.simple` API in all cases (Memlet API may be subject to further changes)
    * `num_accesses` is now two properties: `volume` and `dynamic` (for bounded but dynamic volume)
    * `wcr_conflict` is now `wcr_nonatomic` for clarity
    * New ease-of-use API for memlet construction from string (`Memlet('A[i:i+5]')`)
    * Memlet constructor accepts both strings and native types (AST, Subset) for supported fields
    * New `src_subset` and `dst_subset` properties
    * Remove unused `DYNAMIC` and `NA_DYNAMIC`

### DIFF
--- a/dace/memlet.py
+++ b/dace/memlet.py
@@ -143,8 +143,14 @@ class Memlet(object):
         attrs = dace.serialize.all_properties_to_json(self)
 
         # Fill in new values
-        attrs['src_subset'] = self.src_subset
-        attrs['dst_subset'] = self.dst_subset
+        if self.src_subset is not None:
+            attrs['src_subset'] = self.src_subset.to_json()
+        else:
+            attrs['src_subset'] = None
+        if self.dst_subset is not None:
+            attrs['dst_subset'] = self.dst_subset.to_json()
+        else:
+            attrs['dst_subset'] = None
 
         return {"type": "Memlet", "attributes": attrs}
 
--- a/dace/memlet.py
+++ b/dace/memlet.py
@@ -155,7 +155,7 @@ class Memlet(object):
 
         # Fill in legacy (DEPRECATED) values for backwards compatibility
         attrs['num_accesses'] = \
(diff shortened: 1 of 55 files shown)


dace/memlet.py
```diff
@@ -22,90 +24,196 @@ class Memlet(object):
...
+                 wcr_nonatomic: bool = False,
+                 allow_oob: bool = False):
+        """ 
+        Constructs a Memlet.
+        :param expr: A string expression of the this memlet, given as an ease
+                     of use API. Must follow one of the following forms:
+                     1. ``ARRAY``,
+                     2. ``ARRAY[SUBSET]``,
+                     3. ``ARRAY[SUBSET] -> OTHER_SUBSET``.
+        :param data: (DEPRECATED) Data descriptor name attached to this memlet.
```
definelicht, line 79: :pagesofohdeer: Want to add a deprecated warning?

dace/memlet.py
```diff
@@ -140,75 +248,178 @@ def simple(data,
...
-        return self.subset.bounding_box_size()
+        if self.src_subset:
+            return self.src_subset.bounding_box_size()
+        elif self.dst_subset:
+            return self.dst_subset.bounding_box_size()
+        return []
+
+    # New fields
+    @property
+    def src_subset(self):
```
definelicht, line 413: Ohhh I see what you're doing there :P sneaking in the new definition

dace/memlet.py
tbennun: Only that parameter is deprecated. I don't know how to add a deprecated parameter.

dace/memlet.py
tbennun: yep :) slowly but surely

PR #294, opened 2020-06-24
Add `GPU_Persistent` map

JanKleine:
Adds support for a persistent GPU map, that can be used to fuse kernels. It allocates a fixed number of thread blocks (a configurable fraction of #SMs on the device multiplied with a fixed multiplier to control occupancy).

The `GPU_Device` maps inside a `GPU_Persistent` map are implemented as loops, in case the device map has a larger range than the number of threads allocated.


dace/dtypes.py
```diff
@@ -42,6 +42,7 @@ class ScheduleType(aenum.AutoNumberEnum):
     GPU_Device = ()  # Kernel
     GPU_ThreadBlock = ()  # Thread-block code
     GPU_ThreadBlock_Dynamic = ()  # Allows rescheduling work within a block
+    GPU_Persistent = ()
```
tbennun, line 45: Could you add it to line 954 in this file as well?

dace/codegen/targets/cuda.py
```diff
@@ -955,6 +983,65 @@ def generate_devicelevel_state(self, sdfg, state, function_stream,
...
+        # the DFGs inside of a state are independent, so they don't need
+        # synchronization. DFGs in a GPU_Persistent scope are per se executed
+        # by a single thread only. (Device) Maps however can be distributed
+        # across multiple threads
+        elif self._toplevel_schedule == dtypes.ScheduleType.GPU_Persistent:
+
+            # reset streams in GPU persistent maps if the lifetime is scope,
+            # otherwise streams do not behave as expected becasue they are
+            # allocated on host side
+            # TODO discuss how exactly streams should behave (between states)
```
tbennun, line 996: Still need to discuss?

dace/codegen/targets/cuda.py
```diff
@@ -1002,12 +1094,43 @@ def generate_scope(self, sdfg, dfg_scope, state_id, function_stream,
...
+        is_persistent = (dfg_scope.source_nodes()[0].map.schedule
+                         == dtypes.ScheduleType.GPU_Persistent)
 
         # Get parameters of subgraph
         kernel_args = dfg_scope.arglist()
+
+        # handle dynamic map inputs
+        # TODO move this loop into dfg_scope.arglist()
+        for e in dace.sdfg.dynamic_map_inputs(state, scope_entry):
+            kernel_args[str(e.src)] = e.src.desc(sdfg)
```
tbennun, line 1108: Do we still need these lines?

dace/codegen/targets/cuda.py
```diff
@@ -424,8 +424,8 @@ def allocate_stream(self, sdfg, dfg, state_id, node, function_stream,
                     .format(**fmtargs), sdfg, state_id, node)
                 self._globalcode.write(
                     """
-DACE_EXPORTED void __dace_alloc_{location}(uint32_t size, dace::GPUStream<{type}, {is_pow2}>& result);
-void __dace_alloc_{location}(uint32_t size, dace::GPUStream<{type}, {is_pow2}>& result) {{
+DACE_EXPORTED void __dace_alloc_{location}(uint32_t {size}, dace::GPUStream<{type}, {is_pow2}>& result);
```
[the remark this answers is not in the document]
JanKleine: Line 429 used `{size}` , however lines 427/428 used `size`. Thus the first argument to the `__dace_alloc_<...>()` function was called `size`, but line 428 used the actual name of the variable (for example `N`, or `nnz`), which was never actually declared, resulting in a compiler error.

Another option would be to use `size` everywhere in the function, but since a value for `{size}` was passed in by `.format()` I assumed this was the intended behaviour.

dace/codegen/targets/cuda.py
JanKleine: I don't think so, it was from when we discussed whether to reset the streams each state or not.

dace/codegen/targets/cuda.py
JanKleine: If there was a fix for dynamic range maps on GPU in main repo then no, otherwise this is the fix. Though I can remove the TODO.

dace/codegen/targets/cuda.py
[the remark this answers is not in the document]
tbennun: Thanks!

dace/codegen/targets/cuda.py
tbennun: Yep

dace/codegen/targets/cuda.py
tbennun: I was going to ask "what happens if you print the arglist", if the dynamic range is there we can remove those loops from the code.

tbennun:
@JanKleine you know you could have applied all the updates in one batch, right? Now our CI is overloaded... I'll cancel the older builds.

JanKleine:
Oh, I'm Sorry about that, I only realised that later, won't happen again

PR #305, opened 2020-07-09
Allow "@dace" as a function decorator

tbennun:
This enables writing data-centric functions/programs in the most concise way possible:

```python
import dace
@dace
def myprog(A: dace.float32[20]):
    return A + A
```

### DIFF
--- a/dace/__init__.py
+++ b/dace/__init__.py
@@ -1,3 +1,4 @@
+import sys
 from .dtypes import *
 
 # Python frontend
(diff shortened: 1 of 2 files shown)


definelicht:
why

tbennun:
> no test :( also why

@definelicht Added test. The reason is to make it even easier to make a function a data-centric program, without `@dace.program` or `@dace.function`.

PR #308, opened 2020-07-14
Allowing reassignment of arrays of shape (1, )

alexnick83:
The purpose is to allow code like the following to work:
```python
@dace.program
def reassign(a: dace.float64[1]):
    s = 0.0
    s = 1.0
```
The solution is to interpret the second statement `s = 1.0` as `s[0] = 1.0` or `s[:] = 1.0`.
However, redefinition of arrays is still now allowed, i.e., this only works with arrays of shape (1, ).

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -2384,7 +2384,8 @@ class ProgramVisitor(ExtNodeVisitor):
                 true_array = defined_arrays[true_name]
 
             if (isinstance(target, ast.Name) and true_name and not op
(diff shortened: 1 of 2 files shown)


tbennun:
@alexnick83 Looks good, but one request: could you please revert the changes to the submodules? It is breaking the CI build (latest hlslib is not always supported).

alexnick83:
> @alexnick83 Looks good, but one request: could you please revert the changes to the submodules? It is breaking the CI build (latest hlslib is not always supported).

Fixed

PR #310, opened 2020-07-15
Support for matrix inversion in blas library

alexnick83:
The purpose is to add support for matrix inversion in the BLAS library. The following nodes are needed:
- [x] getf
- [x] getri
- [ ] getrs (optional, needed for cuSPARSE)
- [x] invert_matrix (uses getrf and getri/getrs)


alexnick83:
This was implemented in a newer PR.

PR #320, opened 2020-07-26
Allow for extensive optimization and history tracking of SDFGs via the VSCode interface

phschaad:
- [x] Crude history keeping capabilities
- <del>[ ] Allow for extensive history keeping with version tagging</del> This will be tackled at a later time in the very close future


phschaad:
@tbennun As discussed, extensive history and version tagging etc. is still on the tasklist, however I moved it out of this PR such that this one can be merged back without blocking us on the VSCode side of things in terms of releasing.

dace/properties.py
```diff
@@ -554,6 +554,39 @@ def from_json(self, data, sdfg=None):
...
+    def from_json(self, data, sdfg=None):
+        if data is None:
+            return data
+        if not isinstance(data, list):
+            raise TypeError(
+                'TransformationHistProperty expects a list input, got %s' % data
+            )
+        # This is imported here to avoid an import loop.
+        from dace.transformation.pattern_matching import Transformation
+        return [Transformation.from_json(elem) for elem in data]
```
tbennun, line 587: I think you should use `dace.serialize.from_json` because it can detect Transformations, expansions, and SubgraphTransformations

dace/transformation/interface/vscode/__main__.py
```diff
@@ -1,32 +1,68 @@
 import json
 from argparse import ArgumentParser
 
-def apply_transformation(sdfg, transformation):
+def reapply_history_until(sdfg_json, index):
+    """
+    Rewind a given sdfg back to a specific point in its history by reapplying
```
tbennun, line 6:
```suggestion
    Rewind a given SDFG back to a specific point in its history by reapplying
```

PR #328, opened 2020-08-03
Disable debugprint and optimization CLI by default

tbennun:


dace/config_schema.yml
```diff
@@ -29,14 +29,22 @@ required:
                     Automatically specialize every SDFG to the symbol values 
                     at call-time. Requires all symbols to be set.
 
-            interface:
+            autooptimize:
```
definelicht, line 32: Isn't this misleading? Auto-optimize sounds like DaCe will just make your program fast. Maybe just "optimizer" or "cli_optimizer" or...

dace/sdfg/state.py
```diff
@@ -1526,6 +1526,9 @@ def add_memlet_path(self,
                     if propagate:
                         cur_memlet = propagate_memlet(self, cur_memlet, snode,
                                                       True)
+        # Try to initialize memlets
+        for edge in edges:
+            edge.data.try_initialize(self.parent, self, edge)
```
definelicht, line 1531: Is this related or did it sneak in?

tbennun: It's related. The optimizer CLI, even when you don't optimize, actually serialized and deserialized, fixing the memlets. Now we just fix them during creation

dace/config_schema.yml
tbennun: You can still choose the optimizer in the next configuration to not be CLI (we use this in the transformation tester), so "cli_optimizer" is out. can we call it "optimize_on_call"?

definelicht: Maybe something with "transform" is also more appropriate than "optimize"?

tbennun: Done

definelicht: I still don't like that it's called "autooptimize" for the above reason :spooderman:

tbennun: But it's now called transform_on_call

tbennun:
@definelicht I renamed it

PR #331, opened 2020-08-05
Add simple RTL simulation (verilator) tasklets

andreaskuster:


samples/rtl/rtl_tasklet_scalar.py
```diff
@@ -44,8 +44,10 @@
         end else if (valid_i) begin // case: load a 
             b <= a;
             ready_o <= 1'b0;
-        end else // case: increment counter b
-            b <= b + 1; 
+        end if (b < 100) // case: increment counter b
```
tbennun, line 50: @andreaskuster Did you mean `end else if`?

andreaskuster: Yes, good catch! Thanks.
These typos should hopefully be catched by the unit tests soon :)

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,89 @@
...
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
+
+    # check result
+    assert b == 100
```
tbennun, line 89: @andreaskuster You need to have an `if __name__ == '__main__'` that calls all the tests in this file. This way we support both pytest and running the script directly.

andreaskuster: @tbennun Thanks. Yes, i checked out the other examples an thought that it would be either because of the way you are calling them or to run them locally. I actually included this in this file too (even in the first commit, see: https://github.com/andreaskuster/dace/blob/add_rtl_tasklets/tests/rtl/simulation_test.py)

dace/codegen/targets/rtl.py
```diff
@@ -154,14 +154,13 @@ def unparse_rtl_tasklet(sdfg, state_id, dfg, node, function_stream,
             idx_str = ""
             # catch symbolic (compile time variables)
             if symbolic.issymbolic(tasklet.in_connectors[inp].veclen, sdfg.constants):
-                pass
-                # TODO: raise RuntimeError("Please use sdfg.specialize to specialize the symbol in expression: {}".format(tasklet.in_connectors[inp].veclen))
+                raise RuntimeError("Please use sdfg.specialize to specialize the symbol in expression: {}".format(tasklet.in_connectors[inp].veclen))
             if symbolic.issymbolic(tasklet.in_connectors[inp].bytes, sdfg.constants):
-                pass
-                # TODO: raise RuntimeError("Please use sdfg.specialize to specialize the symbol in expression: {}".format(tasklet.in_connectors[inp].bytes))
+                raise RuntimeError("Please use sdfg.specialize to specialize the symbol in expression: {}".format(tasklet.in_connectors[inp].bytes))
```
tbennun, line 159: I would make a function that takes `expr.free_symbols` and subtracts from it the keys in `sdfg.constants`, in order to print out the symbols that need to be specialized.

andreaskuster: Good idea. I will do that.

andreaskuster:
All the bullet points of deliverable 1 (D1) are done, the code is refactored and follows the dace conventions (dispatchers, cmake integration, code generation,..) , and all tests are passing. I suggest we merge this pull request now.
I am aware of some open enhancements and code refactorization (see https://github.com/spcl/dace/projects/5), but i would do these afterwards separately, since ontherwise we will never come to an end  here.

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
...
+    get_filename_component(RTL_FILE_NAME "${RTL_FILE}" NAME_WE)
+
+    # add verilated .cpp files
+    set(VERILATOR_SRC "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.cpp")
+    set(VERILATOR_SRC "${VERILATOR_SRC}" "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}__Syms.cpp")
+    set(VERILATOR_SRC "${VERILATOR_SRC}" "${VERILATOR_ROOT}/include/verilated.cpp")
+    set(DACE_CPP_FILES "${DACE_CPP_FILES}" "${VERILATOR_SRC}")
+
+    # add verilated design
+    #add_library(VERILATED_LIB OBJECT)
```
definelicht, line 607:
```suggestion
```

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
...
+    add_library("${RTL_FILE_NAME}" OBJECT)
+
+    # include verilator
+    set(VERILATOR_INCLUDE "${VERILATOR_ROOT}/include" "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir")
+    include_directories(${VERILATOR_INCLUDE})
+
+    # verilate design
+    #verilate(VERILATED_LIB SOURCES ${RTL_FILE} VERILATOR_ARGS "${VERILATOR_FLAGS}")
+    verilate("${RTL_FILE_NAME}" SOURCES ${RTL_FILE} VERILATOR_ARGS "${VERILATOR_FLAGS}")
+    message(WARNING "ASDF: ${RTL_FILE}")
```
definelicht, line 617:
```suggestion
```

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
...
+    # add verilated design
+    #add_library(VERILATED_LIB OBJECT)
+    add_library("${RTL_FILE_NAME}" OBJECT)
+
+    # include verilator
+    set(VERILATOR_INCLUDE "${VERILATOR_ROOT}/include" "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir")
+    include_directories(${VERILATOR_INCLUDE})
+
+    # verilate design
+    #verilate(VERILATED_LIB SOURCES ${RTL_FILE} VERILATOR_ARGS "${VERILATOR_FLAGS}")
```
definelicht, line 615:
```suggestion
```

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
...
+  foreach(RTL_FILE ${DACE_RTL_FILES})
+
+    # extract design name
+    get_filename_component(RTL_FILE_NAME "${RTL_FILE}" NAME_WE)
+
+    # add verilated .cpp files
+    set(VERILATOR_SRC "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.cpp")
+    set(VERILATOR_SRC "${VERILATOR_SRC}" "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}__Syms.cpp")
+    set(VERILATOR_SRC "${VERILATOR_SRC}" "${VERILATOR_ROOT}/include/verilated.cpp")
+    set(DACE_CPP_FILES "${DACE_CPP_FILES}" "${VERILATOR_SRC}")
```
definelicht, line 604: Does this not cause the C++ files to be re-compiled as part of the main DaCe shared library object?

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
...
+    # include verilator
+    set(VERILATOR_INCLUDE "${VERILATOR_ROOT}/include" "${dace_program_BINARY_DIR}/CMakeFiles/${RTL_FILE_NAME}.dir/V${RTL_FILE_NAME}.dir")
+    include_directories(${VERILATOR_INCLUDE})
+
+    # verilate design
+    #verilate(VERILATED_LIB SOURCES ${RTL_FILE} VERILATOR_ARGS "${VERILATOR_FLAGS}")
+    verilate("${RTL_FILE_NAME}" SOURCES ${RTL_FILE} VERILATOR_ARGS "${VERILATOR_FLAGS}")
+    message(WARNING "ASDF: ${RTL_FILE}")
+    # add object library for linking
+    #set(DACE_LIBS ${DACE_LIBS} ${VERILATED_LIB})
```
definelicht, line 619:
```suggestion
```

dace/codegen/compiler.py
```diff
@@ -10,7 +10,12 @@
 import shutil
 import subprocess
 import re
-from typing import List
+from typing import Any, Dict, List
+import numpy as np
+import sympy as sp
+import warnings
+from os import listdir
+from os.path import isfile, join
```
definelicht, line 18: What are these changes for? Seems to be unused?

dace/codegen/compiler.py
```diff
@@ -35,10 +40,8 @@ def generate_program_folder(sdfg,
     """
 
     src_path = os.path.join(out_path, "src")
+    filelist = list()
 
-    os.makedirs(src_path, exist_ok=True)
```
definelicht, line 39: What happened here?

dace/codegen/targets/cpu.py
```diff
@@ -8,6 +8,8 @@
 from dace.codegen.prettycode import CodeIOStream
 from dace.codegen.targets import cpp
 from dace.codegen.targets.common import codeblock_to_cpp
+from dace.codegen.targets.cpp import *
```
definelicht, line 11: Avoid `import *`, instead import like `import dace.codegen.targets.cpp as cpp` or similar. This is to avoid polluting the namespace of this file with everything from `cpp`.

dace/codegen/targets/cpu.py
```diff
@@ -410,17 +412,17 @@ def copy_memory(
...
+            sdfg,
+            state_id,
+            src_node,
+            src_storage,
+            dst_node,
+            dst_storage,
+            dst_schedule,
+            edge,
+            dfg,
+            stream,
```
definelicht, line 425: Did you yapf this? I believe yapf makes 4 spaces, not 8.

dace/codegen/targets/cpu.py
```diff
@@ -492,13 +494,13 @@ def _emit_copy(
                 # Identify whether a stream is writing to an array
                 if isinstance(dst_nodedesc,
                               (data.Scalar, data.Array)) and isinstance(
-                                  src_nodedesc, data.Stream):
+                    src_nodedesc, data.Stream):
```
definelicht, line 497: Formatting

dace/codegen/targets/cpu.py
```diff
@@ -528,7 +530,7 @@ def _emit_copy(
                 # Array -> Stream - push bulk
                 if isinstance(src_nodedesc,
                               (data.Scalar, data.Array)) and isinstance(
-                                  dst_nodedesc, data.Stream):
+                    dst_nodedesc, data.Stream):
```
definelicht, line 533: Formatting (please yapf this)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+# Copyright 2019-2020 ETH Zurich and the DaCe authors. All rights reserved.
+
+import itertools
+
+from typing import List, Tuple, Dict
+
+import dace
```
definelicht, line 9: Please try to avoid importing `dace` if possible (this gives issues with cyclic dependencies): instead, import the specific submodules that you need (like you did on the following line)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+        # store reference to sdfg
+        self.sdfg: dace.SDFG = sdfg
+        # store reference to frame code generator
+        self.frame: DaCeCodeGenerator = frame_codegen
+        # get dispatcher to register callbacks for allocation/nodes/.. code generators
+        self.dispatcher: TargetDispatcher = frame_codegen.dispatcher
+        # register node dispatcher -> generate_node(), predicate: process tasklets only
+        self.dispatcher.register_node_dispatcher(self,
+                                                 lambda sdfg, node: isinstance(node,
+                                                                               nodes.Tasklet) and node.language == Language.RTL)
```
definelicht, line 42: You need to yapf this (with version 0.30 or higher)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+        # register all cpu type copies
+        cpu_storage_types = [StorageType.CPU_Pinned, StorageType.CPU_Heap, StorageType.CPU_ThreadLocal,
+                             StorageType.Register, StorageType.Default]
```
definelicht, line 45: This seems fishy, the RTL codegen should never touch CPU memory. I guess this is because we only have verilator in isolation right now, but can't we at least register ANY storage type when the predicate of RTL tasklet is fulfilled?

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+                                                              edge.dst_conn,
+                                                              dst_node.in_connectors[edge.dst_conn].ctype,
+                                                              edge.src.data)
+            else:  # scalar accessor
+                line: str = "{}* {} = &{}[0];".format(dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn,
+                                                      edge.src.data)
+        else:
+            raise RuntimeError("Not handling copy_memory case of type {} -> {}.".format(type(edge.src), type(edge.dst)))
+        # write accessor to file
+        callsite_stream.write(line)
```
definelicht, line 119: Why is this necessary instead of the regular CPU copy, it seems redundant? Can't we delegate this?

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+                                                              src_node.out_connectors[edge.src_conn].ctype,
+                                                              edge.dst.data)
+            else:  # scalar accessor
+                line: str = "{}* {} = &{}[0];".format(src_node.out_connectors[edge.src_conn].ctype, edge.src_conn,
+                                                      edge.dst.data)
+        else:
+            raise RuntimeError(
+                "Not handling define_out_memlet case of type {} -> {}.".format(type(edge.src), type(edge.dst)))
+        # write accessor to file
+        callsite_stream.write(line)
```
definelicht, line 149: Same as above: why can this not be delegated to the CPU codegen?

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+        else:
+            return "#(\n{}\n)".format(" " + "\n".join(
+                ["{} parameter {} = {}".format("," if i > 0 else "", key, constants[key]) for i, key in
+                 enumerate(constants)]))
+
+    def check_issymbolic(self, iterator: iter, sdfg):
+        for item in iterator:
+            # catch symbolic (compile time variables)
+            if symbolic.issymbolic(item, sdfg.constants):
+                raise RuntimeError("Please use sdfg.specialize to specialize the symbol in expression: {}".format(item))
```
definelicht, line 201: These functions doesn't use `self`, so they should not be in the class. Move them out as free functions

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+                                            linkable=True,
+                                            environments=None))
+
+        # generate verilator simulation cpp code components
+        inputs, outputs = self.generate_cpp_inputs_outputs(tasklet)
+        vector_init = self.generate_cpp_vector_init(tasklet)
+        num_elements = self.generate_cpp_num_elements()
+        internal_state_str, internal_state_var = self.generate_cpp_internal_state(tasklet)
+
+        # add header code do stream
```
definelicht, line 351:
```suggestion
        # add header code to stream
```

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+                                                                         num_elements=num_elements,
+                                                                         vector_init=vector_init,
+                                                                         internal_state_str=internal_state_str,
+                                                                         internal_state_var=internal_state_var),
+                              sdfg=sdfg,
+                              state_id=state_id,
+                              node_id=node)
+
+    @staticmethod
+    def header_template():
```
definelicht, line 367: You could just make this a string constant:
```VERILATOR_HEADER_TEMPLATE = """\```

File: dace/codegen/targets/rtl.py [hunk 24]
```diff
@@ -0,0 +1,519 @@
...
+#include "V{name}.h"
+"""
+
+    @staticmethod
+    def main_template():
+        """
+            Cpp Main Verilator Simulator Template
+        """
+        return """\
+std::cout << "SIM START" << std::endl;
```
definelicht, line 388: Be more specific about what is being simulated. If multiple tasklets are running simultaneously, this might get messy...

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+    // increment time
+    main_time++;
+
+    // check if valid_i and ready_o have been asserted at the rising clock edge -> input read handshake
+    if (model->ready_o == 1 && model->valid_i == 1){{
+        read_input_hs = true;
+    }} 
+    // feed new element
+    if(model->valid_i == 0 && in_ptr < num_elements){{
+        std::cout << "feed new element" << std::endl;
```
definelicht, line 433:
```suggestion
```

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+        {inputs}
+        model->valid_i = 1;
+    }}
+
+    // export element
+    if(model->valid_o == 1){{
+        std::cout << "export element" << std::endl;
```
definelicht, line 440:
```suggestion
```

dace/codegen/targets/rtl.py
definelicht: If we want to keep something like this, we need to add a verbosity flag. In many cases the user probably doesn't want every read and write to be printed to the terminal. Even then, it should be more descriptive (_what_ was fed)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+    if(DEBUG){{
+        VL_PRINTF("[t=%lu] clk_i=%u rst_i=%u valid_i=%u ready_i=%u valid_o=%u ready_o=%u \\n", main_time, model->clk_i, model->rst_i, model->valid_i, model->ready_i, model->valid_o, model->ready_o);
+        VL_PRINTF("{internal_state_str}", {internal_state_var});
+        std::cout << std::endl;
+    }}
+
+    // check if valid_i and ready_o have been asserted at the rising clock edge
+    if (read_input_hs){{
+        // remove valid_i flag
+        std::cout << "remove read_input_hs flag" << std::endl;
```
definelicht, line 463:
```suggestion
```

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+
+    // positive clock edge
+    model->clk_i = !model->clk_i;
+    model->eval();
+
+    // report internal state
+    if(DEBUG){{
+        VL_PRINTF("[t=%lu] clk_i=%u rst_i=%u valid_i=%u ready_i=%u valid_o=%u ready_o=%u \\n", main_time, model->clk_i, model->rst_i, model->valid_i, model->ready_i, model->valid_o, model->ready_o);
+        VL_PRINTF("{internal_state_str}", {internal_state_var});
+        std::cout << std::endl;
```
definelicht, line 457: Use `std::flush` instead, and put the newline on the previous `VL_PRINTF`. Otherwise you risk that they are split up

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+        // remove valid_i flag
+        std::cout << "remove read_input_hs flag" << std::endl;
+        model->valid_i = 0;
+        read_input_hs = false;
+    }}
+
+    // check if valid_o and ready_i have been asserted at the rising clock edge
+    if (write_output_hs){{
+        // remove valid_i flag
+        std::cout << "remove write_output_hs flag" << std::endl;
```
definelicht, line 471:
```suggestion
```

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+    // negative clock edge
+    model->clk_i = !model->clk_i;
+    model->eval();
+}}
+
+// report internal state
+if(DEBUG){{
+    VL_PRINTF("[t=%lu] clk_i=%u rst_i=%u valid_i=%u ready_i=%u valid_o=%u ready_o=%u \\n", main_time, model->clk_i, model->rst_i, model->valid_i, model->ready_i, model->valid_o, model->ready_o);
+    VL_PRINTF("{internal_state_str}", {internal_state_var});
+    std::cout << std::endl;
```
definelicht, line 486: Same as above, use `std::flush`

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,519 @@
...
+}}
+
+// final model cleanup
+model->final();
+
+// clean up resources
+delete model;
+model = NULL;
+
+std::cout << "SIM END" << std::endl;
```
definelicht, line 496: Same as above, be more descriptive (and perhaps make this dependent on verbosity flag)

dace/dtypes.py
```diff
@@ -96,6 +96,7 @@ class Language(aenum.AutoNumberEnum):
 
     Python = ()
     CPP = ()
+    RTL = ()
```
definelicht, line 99: I think this should be `SystemVerilog` to be symmetric with the others. We don't support _any_ RTL language here!

dace/runtime/include/dace/cuda/dynmap.cuh
```diff
@@ -11,6 +11,11 @@
 #include <hip/hip_cooperative_groups.h>
 #endif*/
 
+// HIP does not yet support features in cooperative groups used here.
+/*#elif defined(__HIPCC__)
+#include <hip/hip_cooperative_groups.h>
+#endif*/
+
```
definelicht, line 18: Something weird happened

samples/rtl/rtl_multi_tasklet.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

samples/rtl/rtl_multi_tasklet.py
```diff
@@ -0,0 +1,115 @@
...
+# add state
+state = sdfg.add_state()
+
+# add arrays
+sdfg.add_array('A', [1], dtype=dace.int32)
+sdfg.add_array('B', [1], dtype=dace.int32)
+sdfg.add_array('C', [1], dtype=dace.int32)
+
+# enable debugging output
+sdfg.add_constant("DEBUG", 1)
```
definelicht, line 25: Isn't this a DaCe configuration variable now?

samples/rtl/rtl_multi_tasklet.py
```diff
@@ -0,0 +1,115 @@
...
+    c = np.array([0]).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}, c={}".format(a, b, c))
+
+    # call program
+    sdfg(A=a, B=b, C=c)
+
+    # show result
+    print("a={}, b={}, c={}".format(a, b, c))
```
definelicht, line 115: You're missing a validation check that it actually produces the right result!

samples/rtl/rtl_tasklet_parameter.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

samples/rtl/rtl_tasklet_parameter.py
```diff
@@ -0,0 +1,97 @@
...
+
+# add state
+state = sdfg.add_state()
+
+# add arrays
+sdfg.add_array('A', [1], dtype=dace.int32)
+sdfg.add_array('B', [1], dtype=dace.int32)
+
+# enable debugging output
+sdfg.add_constant("DEBUG", 1)
```
definelicht, line 24: Same as above

samples/rtl/rtl_tasklet_parameter.py
```diff
@@ -0,0 +1,97 @@
...
+    b = np.array([0]).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 97: Missing validation

samples/rtl/rtl_tasklet_pipeline.py
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

samples/rtl/rtl_tasklet_pipeline.py
```diff
@@ -0,0 +1,151 @@
...
+    b = np.array([0] * num_elements).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 151: Missing validation

samples/rtl/rtl_tasklet_pipeline.py
```diff
@@ -0,0 +1,151 @@
...
+
+# define compile-time constant
+sdfg.specialize(dict(N=4))
+
+# add arrays
+sdfg.add_array('A', [N], dtype=dace.int32)
+sdfg.add_array('B', [N], dtype=dace.int32)
+
+# enable debugging output
+sdfg.add_constant("DEBUG", 1)
```
definelicht, line 30: I suggest to disable by default, and allow it to be set to true with a command line parameter

samples/rtl/rtl_tasklet_scalar.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

samples/rtl/rtl_tasklet_scalar.py
```diff
@@ -0,0 +1,94 @@
...
+
+# add state
+state = sdfg.add_state()
+
+# add arrays
+sdfg.add_array('A', [1], dtype=dace.int32)
+sdfg.add_array('B', [1], dtype=dace.int32)
+
+# enable debugging output
+sdfg.add_constant("DEBUG", 1)
```
definelicht, line 24: Suggest making command line parameter

samples/rtl/rtl_tasklet_scalar.py
```diff
@@ -0,0 +1,94 @@
...
+    b = np.array([0]).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 94: Missing validation

samples/rtl/rtl_tasklet_vector.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

samples/rtl/rtl_tasklet_vector.py
```diff
@@ -0,0 +1,101 @@
...
+
+# add state
+state = sdfg.add_state()
+
+# add arrays
+sdfg.add_array('A', [WIDTH], dtype=dace.int32)
+sdfg.add_array('B', [1], dtype=dace.int32)
+
+# enable debugging output
+sdfg.add_constant("DEBUG", 1)
```
definelicht, line 30: Suggest command line param

samples/rtl/rtl_tasklet_vector.py
```diff
@@ -0,0 +1,101 @@
...
+    b = np.array([0]).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 101: Missing validation

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
definelicht, line 2:
```suggestion
```

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
...
+
+    """
+        Execute    
+    """
+    # init data structures
+    a = np.random.randint(0, 100, dace.symbolic.evaluate(N, sdfg.constants)).astype(np.int32)
+    b = np.array([0]).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
```
definelicht, line 281: Don't print from tests

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
...
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 287: Don't print from tests

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
...
+    b = np.random.randint(0, 100, 1).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 189: Don't print from tests

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
...
+
+    """
+        Execute    
+    """
+    # init data structures
+    a = np.random.randint(0, 100, 1).astype(np.int32)
+    b = np.random.randint(0, 100, 1).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
```
definelicht, line 183: Don't print from tests

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
...
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 95: Don't print from tests

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,410 @@
...
+
+    """
+        Execute    
+    """
+    # init data structures
+    a = np.random.randint(0, 100, 1).astype(np.int32)
+    b = np.random.randint(0, 100, 1).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
```
definelicht, line 89: Don't print from tests

definelicht (changes requested):
High-level comments:
- Remember to yapf every file you have touched using version 0.30 or newer
- Don't add extra lines to the top of the files if they are not present in the rest of DaCe
- There's quite a lot of leftover debugging lines in the code that need to be purged

tests/rtl/simulation_test.py
andreaskuster: Thanks for the feedback. I didn't know about the yapf utility, it's pretty neat.

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
 
 endif() # DACE_ENABLE_XILINX
 
+# create verilator RTL simulation objects
+if(DACE_ENABLE_RTL)
+
+  # find verilator installation
+  find_package(verilator HINTS $ENV{VERILATOR_ROOT} ${VERILATOR_ROOT})
+  if (NOT verilator_FOUND)
+    message(FATAL_ERROR "Verilator was not found. Either install it, or set the VERILATOR_ROOT environment variable")
```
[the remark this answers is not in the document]
andreaskuster: Sure, I added this to the project page as a reminder for myself.

dace/codegen/CMakeLists.txt
```diff
@@ -512,6 +566,64 @@ if(DACE_ENABLE_XILINX)
...
+  endif()
+
+  # check minimal version requirements
+  set(VERILATOR_MIN_VERSION "4.028")
+  if("${verilator_VERSION}" VERSION_LESS VERILATOR_MIN_VERSION)
+    message(WARNING "Please upgrade verilator to version >=${VERILATOR_MIN_VERSION}")
```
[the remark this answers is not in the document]
andreaskuster: You are right, I agree that in our case it should be an error, since otherwise DaCe will get unstable. Verilator changed  too much from version 3.xxx to version 4.xxx.

dace/codegen/targets/cpu.py
andreaskuster: I didn't know about yapf before, but i just yapfed all files now..

dace/codegen/CMakeLists.txt
andreaskuster: Generally: I wouldn't call it re-compilation, but yes, they will be compiled into the DaCe library (this is intended)
For line 601 and 602: The verilator cmake command `verilate` basically creates these .cpp files (and does not compile them), which are the cpp equivalent of the verilator design. We need to have them in the binary file, such that we can call them from the tasklet cpp code location.
For line 603: I agree with you that this lines might be added multiple times (for multi rtl tasklets). I am going to move it ouside of the loop.

dace/codegen/CMakeLists.txt
```diff
@@ -252,7 +260,53 @@ if(DACE_ENABLE_HIP)
   # Get local AMD architectures
   if (NOT DEFINED LOCAL_HIP_ARCHITECTURES)
     # Compile and run a test program
-    execute_process(COMMAND ${HIP_HIPCC_EXECUTABLE} "${CMAKE_SOURCE_DIR}/tools/get_hip_arch.cpp" -o 
+    execute_process(COMMAND ${HIP_HIPCC_EXECUTABLE} "${CMAKE_SOURCE_DIR}/tools/get_hip_arch.cpp" -o
```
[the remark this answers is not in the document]
andreaskuster: You are right, something went wrong there. I copy&pasted the master version.

dace/codegen/compiler.py
andreaskuster: Good catch. They were used for the 'hack' before i solved the proper code file handling using CodeObjects.

dace/codegen/compiler.py
andreaskuster: That is weird, it is not present im my local version (which is up-to-date). It should have been removed.

dace/runtime/include/dace/cuda/dynmap.cuh
andreaskuster: I don't know what had happened here. I again just copy and paste the content from the master branch.

dace/codegen/targets/rtl.py
andreaskuster: Well, i didn't add the copy_dispatcher first, but that way the local variables that read/write from the function parameters didn't get created at all. Furthermore, i do have more flexibility by that, since i for example prefer to always use the pointers (instead of pointers only for non-scalar data like arrays).

dace/codegen/targets/rtl.py
andreaskuster: I just checked, the memlet to the tasklet edge for example are of type `StorageType.CPU_Heap` and `StorageType.Register`, which is why i do have to register. Sure, that is probably even cleaner to register all types of storages. I will do this.

dace/codegen/targets/rtl.py
andreaskuster: I totally agree with you. I intend to extend the `dace.conf` section of the rtl codegen by this (see project page, ToDo) and put all the debug output/internal state print statements into an if clause. We decided during the meeting that this is not part of deliverable 1 (but should definitely be done very soon afterwards). Since i want to move most of the simulator functionality to a proper class/functions instead of this flat code, it will be part of this refactorization.

dace/dtypes.py
andreaskuster: I agree, it is nicer for consistency of this enumeration, but this actually rises a bigger question on how general we want to keep or RTL CodeGen. Do we want to support other RTL languages in the future too through this CodeGen? (for example Verilog will also work for Verilator, and the Intel/Xiling tools).
I changed the language from RTL to SystemVerilog, but kept the target as rtl (i.e. target 'rtl', RTLCodeGen and the file is rtl.py)

samples/rtl/rtl_multi_tasklet.py
andreaskuster: It is in on the project todo list to add this as a `dace.conf` parameter. For the time being, i will add it as a command line parameter with default value=false.

dace/runtime/include/dace/cuda/dynmap.cuh
```diff
@@ -187,4 +187,4 @@ namespace dace {
 
 #endif // __CUDACC__
 
-#endif // __DACE_DYNMAP_CUH
+#endif // __DACE_DYNMAP_CUH
```
tbennun, line 190: please revert this (no newline at the end of the file)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+
+// verilator includes
+#include <verilated.h>
+
+// include model header, generated from verilating the sv design
+#include "V{name}.h"
+"""
+
+    CPP_MAIN_TEMPLATE = """\
+if(DEBUG){{
```
definelicht, line 398: This macro should have a safer name, like `DACE_VERILATOR_ENABLE_DEBUG`. Even better: turn it into a `constexpr bool kDebugVerilator = DACE_VERILATOR_ENABLE_DEBUG;`.

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+from dace.config import Config
+from dace.codegen.targets.target import TargetCodeGenerator
+from dace.codegen.prettycode import CodeIOStream
+from dace.codegen.targets.framecode import DaCeCodeGenerator
+from dace.codegen.dispatcher import TargetDispatcher
+from dace.sdfg.graph import MultiConnectorEdge
+from dace.sdfg.state import StateSubgraphView
+from dace.sdfg import find_input_arraynode, find_output_arraynode
+from dace.codegen.codeobject import CodeObject
+from dace.dtypes import pointer, vector
```
tbennun, line 17: Style: please import package names rather than classes (exceptions: `typing`, transformations, the `SDFG` and `SDFGState` classes)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+"""
+
+
+def check_issymbolic(iterator: iter, sdfg):
+    for item in iterator:
+        # catch symbolic (compile time variables)
+        if symbolic.issymbolic(item, sdfg.constants):
+            raise RuntimeError(
+                "Please use sdfg.specialize to specialize the symbol in expression: {}"
+                .format(item))
```
definelicht, line 544: I think this should be a `ValueError`, and make the error message a bit more descriptive, like "value of {} must be constant to generate hardware, please use sdfg.specialize..."

samples/rtl/rtl_tasklet_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+    b = np.array([0] * num_elements).astype(np.int32)
+
+    # show initial values
+    print("a={}, b={}".format(a, b))
+
+    # call program
+    sdfg(A=a, B=b)
+
+    # show result
+    print("a={}, b={}".format(a, b))
```
definelicht, line 153: Still a correctness check missing :-)

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+        # generate verilator simulation cpp code components
+        inputs, outputs = self.generate_cpp_inputs_outputs(tasklet)
+        vector_init = self.generate_cpp_vector_init(tasklet)
+        num_elements = self.generate_cpp_num_elements()
+        internal_state_str, internal_state_var = self.generate_cpp_internal_state(
+            tasklet)
+
+        # add header code to stream
+        sdfg.append_global_code(cpp_code=RTLCodeGen.CPP_HEADER_TEMPLATE.format(
+            name=unique_name))
```
tbennun, line 371: Please:
* Split to two headers (general and model headers)
* Check if you already printed out the general header and don't print it twice (you can add a field in `self` for that)

definelicht (changes requested):
Some minor details

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+#include <verilated.h>
+
+// include model header, generated from verilating the sv design
+#include "V{name}.h"
+"""
+
+    CPP_MAIN_TEMPLATE = """\
+if(DEBUG){{
+    std::cout << "SIM {name} START" << std::endl;
+}}
```
tbennun, line 400: Try to only output this code if debug is turned on.

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+            vector_init=vector_init,
+            internal_state_str=internal_state_str,
+            internal_state_var=internal_state_var),
+                              sdfg=sdfg,
+                              state_id=state_id,
+                              node_id=node)
+
+    CPP_HEADER_TEMPLATE = """\
+// generic includes
+#include <iostream>
```
tbennun, line 388: Only output this code if debug is turned on

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+
+    // check if valid_i and ready_o have been asserted at the rising clock edge -> input read handshake
+    if (model->ready_o == 1 && model->valid_i == 1){{
+        read_input_hs = true;
+    }} 
+    // feed new element
+    if(model->valid_i == 0 && in_ptr < num_elements){{
+        if(DEBUG){{
+            std::cout << "feed new element" << std::endl;
+        }}
```
tbennun, line 446: same

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,544 @@
...
+        {inputs}
+        model->valid_i = 1;
+    }}
+
+    // export element
+    if(model->valid_o == 1){{
+        if(DEBUG){{
+            std::cout << "export element" << std::endl;
+        }}
```
tbennun, line 455: same (and for the rest below)

dace/properties.py
```diff
@@ -1000,6 +1000,8 @@ def from_json(tmp, sdfg=None):
             lang = dace.dtypes.Language.Python
         elif lang.endswith("CPP"):
             lang = dace.dtypes.Language.CPP
+        elif lang.endswith("sv") or lang.endswith("systemverilog"):
+            lang = dace.dtypes.Language.SystemVerilog
```
tbennun, line 1004: this condition is inconsistent with the one below

samples/rtl/rtl_tasklet_pipeline.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2019-2020 ETH Zurich and the DaCe authors. All rights reserved.
+"""
+    Pipelined, axi-handshake compliant example that increments b from a up to 100.
```
tbennun, line 3:
```suggestion
    Pipelined, AXI-handshake compliant example that increments b from a up to 100.
```

samples/rtl/rtl_tasklet_pipeline.py
```diff
@@ -0,0 +1,153 @@
...
+    /*
+        Convention:
+           |----------------------------------------------------|
+        -->| clk_i (clock input)                                |
+        -->| rst_i (reset input, rst on high)                   |
+           |                                                    |
+        -->| {inputs}                             reg {outputs} |-->
+           |                                                    |
+        <--| ready_o (ready for data)       (data avail) valid_o|-->
+        -->| valid_i (new data avail)    (data consumed) yumi_i |<--
```
tbennun, line 41:
```suggestion
        -->| valid_i (new data avail)    (data consumed) ready_i |<--
```

tests/rtl/simulation_test.py
```diff
@@ -0,0 +1,377 @@
...
+    state.add_edge(A, None, tasklet0, 'a', dace.Memlet.simple('A', '0'))
+    state.add_edge(tasklet0, 'b', B_w, None, dace.Memlet.simple('B', '0'))
+    state.add_edge(B_r, None, tasklet1, 'b', dace.Memlet.simple('B', '0'))
+    state.add_edge(tasklet1, 'c', C, None, dace.Memlet.simple('C', '0'))
+
+    # validate sdfg
+    sdfg.validate()
+    """
+        Execute    
+    """
```
tbennun, line 356:
```suggestion
    # Execute
```

tbennun (changes requested):
Good work! Minor changes on my end. I think there are a bit too many samples...

samples/rtl/rtl_tasklet_pipeline.py
andreaskuster: Yeah I know. I omitted this, since we do not fully support pipelined designs yet, but i just added now a (temporry) assertion that checks the behaviour it should have at the moment.

definelicht: You can always add a comment to the code that explains this for future reference

dace/codegen/targets/rtl.py
andreaskuster: Done as soon as this pull request has been merged: https://github.com/spcl/dace/pull/463

dace/codegen/targets/rtl.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def __init__(self, frame_codegen: framecode.DaCeCodeGenerator,
+                 sdfg: sdfg.SDFG):
+        # store reference to sdfg
+        self.sdfg: sdfg.SDFG = sdfg
+        # store reference to frame code generator
+        self.frame: framecode.DaCeCodeGenerator = frame_codegen
+        # add debug constant if not existent
+        if "DACE_VERILATOR_ENABLE_DEBUG" not in sdfg.constants:
+            sdfg.add_constant("DACE_VERILATOR_ENABLE_DEBUG", False)
```
definelicht, line 30: Like Tal said, I think we should only emit the debugging code if the debugging flag is set, and have the debugging flag as an option in the DaCe config. But maybe we leave this for a future PR, since it's not critical, and it would be good to get this merged :-)

definelicht (approved):

dace/codegen/targets/rtl.py
andreaskuster: Yes, you are right, this is semi-properly solved. Since is it part of the dace.conf extension, i rather prefer to re-wire all the signals there at once.

tbennun (approved):
Good job with the fixes

PR #334, opened 2020-08-06
Generate nested SDFGs as function calls

tbennun:
This PR also uncovered issues in symbols only used within tasklets, as well as several other issues in tests.


PanicSheep:
This branch makes all my stencils compile. 👍

dace/codegen/prettycode.py
```diff
@@ -44,9 +44,21 @@ def write(self, contents, sdfg=None, state_id=None, node_id=None):
         for line in lines:
             opening_braces = line.count('{')
             closing_braces = line.count('}')
+
+            # Count closing braces before opening ones (e.g., for "} else {")
+            initial_closing_braces = 0
+            for char in line:
```
definelicht, line 50: I know this probably isn't performance critical, but... Looping over every character?

dace/codegen/targets/fpga.py
```diff
@@ -842,9 +842,14 @@ def generate_node(self, sdfg, dfg, state_id, node, function_stream,
             getattr(self, method_name)(sdfg, dfg, state_id, node,
                                        function_stream, callsite_stream)
         else:
+            old_codegen = self._cpu_codegen.calling_codegen
+            self._cpu_codegen.calling_codegen = self
+
             self._cpu_codegen.generate_node(sdfg, dfg, state_id, node,
```
definelicht, line 848: Could you do this with a callback instead?

definelicht: And get rid of `calling_codegen` entirely, if possible

dace/dtypes.py
```diff
@@ -316,6 +316,9 @@ def base_type(self):
     def veclen(self):
         return 1
 
+    def signature(self, name):
```
definelicht, line 319: Hmm, can a _type_ really have a signature? Isn't that a function term?

dace/symbolic.py
```diff
@@ -596,7 +598,7 @@ def pystr_to_symbolic(expr, symbol_map=None, simplify=None):
     locals.update(sympy.abc._clash)
 
     # Sympy processes "not" as direct evaluation rather than negation
-    if isinstance(expr, str) and 'not' in expr:
+    if isinstance(expr, str) and 'not ' in expr:
```
definelicht, line 601: `re.search(r"\\bnot\\b", expr) is not None`

definelicht (changes requested):

dace/codegen/prettycode.py
tbennun: ok, we can find the first `{` and count the `}` that come in the substring before it...

dace/codegen/targets/fpga.py
tbennun: because of nested SDFGs in nested SDFGs, which won't preserve the codegen unless I add a `codegen` argument everywhere

dace/dtypes.py
tbennun: I'd argue it's the signature part of that type in a function. I don't have a better name, suggestions?

dace/symbolic.py
tbennun: Fixed

dace/codegen/prettycode.py
tbennun: Done.

dace/dtypes.py
definelicht: I'm unsure about this. We currently have `ctype` and will base `ocltype` that tells you how to define the type in those languages, which is already dubious to put in `dtypes` rather than in, say, `cpp.py`.
If anything, maybe `as_arg` or `as_carg`?

dace/codegen/targets/fpga.py
definelicht: Fine, burninate in later PR

definelicht (changes requested):
See comment about signatures/args.

dace/dtypes.py
tbennun: @definelicht The reason why it needs a function of its own is because of C syntax unfortunately. If I want to define a callback, the callback type is not prepended before the argument name, but rather it's `return_type (*NAME)(arg1_t, arg2_t....)`. I'll change to as_arg

tbennun: Done

dace/codegen/codegen.py
```diff
@@ -47,16 +47,16 @@ def generate_dummy(sdfg) -> str:
     # first find all scalars and set them to 42
     for argname, arg in al.items():
         if isinstance(arg, data.Scalar):
-            allocations += "  " + str(
-                arg.signature(name=argname, with_types=True)) + " = 42;\n"
+            allocations += "  " + str(arg.as_arg(name=argname,
```
definelicht, line 50: lol damnit

dace/codegen/targets/cuda.py
```diff
@@ -56,14 +56,20 @@ def __init__(self, frame_codegen, sdfg):
         self._dispatcher = frame_codegen.dispatcher
         dispatcher = self._dispatcher
 
+        self.create_grid_barrier = False
+        self.extra_nsdfg_args = []
```
definelicht, line 60: I hate it

tbennun: I hate it too

dace/codegen/codegen.py
tbennun: I tried :)

PR #337, opened 2020-08-10
Wider support for FPGA vector width conversion

definelicht:
- [x] Reorganize common FPGA tests into `tests/fpga` subfolder
~~- [ ] Implement Xilinx in-tasklet vector length conversion~~
- [x] Consolidate `generate_Tasklet` of the codegens to use common infrastructure
- [x] Implement Intel incoming/outgoing memlet vector length conversion
- [x] Updated `generate_Tasklet` in `cpu.py` to have a callback codegen, and modularized the call to allow callbacks.
- [x] Introduced specific OpenCL types in `dace.dtypes`

Fixed some minor issues:
- [x] When assigning a memlet to an array interface, don't read out the value if the connector is a scalar.
- [x] Fixes which cases the memcopy size for array-to-device/device-to-array copies should be adjusted for the vector length.


dace/codegen/targets/intel_fpga.py
```diff
@@ -404,10 +412,23 @@ def generate_kernel_internal(self, sdfg, state, kernel_name, subgraphs,
...
+        # Generate data width converters
+        self.generate_converters(sdfg, kernel_header_stream)
+
+        res = (kernel_header_stream.getvalue() + kernel_body_stream.getvalue())
+
+        # TODO: To avoid re-implementing process_out_memlets and potentially
+        # other CPU codegen functionality, convert to OpenCL vector types here.
+        # It was either this or copy-pasting large amounts of code :-)
+        # We need a solution for this in the new code generator.
+        res = re.sub("dace::vec<([^,]+), ([2-9]+|[1-9][0-9]+)>", "\\1\\2", res)
```
tbennun, line 429: This won't work with `unsigned long`, for instance (which should be ulong4) or `long long` (longlong2)
the opencl type is planned to be returned in dtypes.

dace/codegen/targets/intel_fpga.py
```diff
@@ -1130,13 +1064,15 @@ def visit_Assign(self, node):
 
         if veclen_rhs > veclen_lhs:
             veclen = veclen_rhs
-            self.width_converters.add((True, dtype, veclen))
-            unpack_str = "unpack_{}{}".format(dtype.base_type.ctype, veclen)
+            ctype = dtype.base_type.base_type.ctype
```
tbennun, line 1067: same

dace/codegen/targets/fpga.py
```diff
@@ -1221,3 +1226,122 @@ def generate_host_function_boilerplate(self, sdfg, state, kernel_name,
         for arr_node in nested_global_transients:
             self._dispatcher.dispatch_allocate(sdfg, state, None, arr_node,
                                                None, host_code_stream)
+
+    def _generate_Tasklet(self, sdfg, dfg, state_id, node, function_stream,
+                          callsite_stream):
+
+        # TODO: this is mostly copy-pasta from the CPU-codegen. Refactor to
+        # inject hooks into the CPU codegen once the new codegen has been
+        # implemented.
```
tbennun, line 1235: at this point, shouldn't you be able to override generate_tasklet_preamble/postamble and replace the keyword remover class with `DaceOpenCLKeywordRemover`? I'm not sure there is anything to override in the former (looks like an exact copy of the CPU version?); and the latter is possible because `unparse_tasklet` is now a method rather than a static function.

dace/codegen/targets/xilinx.py
```diff
@@ -820,25 +753,5 @@ def _generate_Tasklet(self, sdfg, dfg, state_id, node, function_stream,
...
-                                   dst_node, edge, callsite_stream):
-        memlet = edge.data
-        if (self._dispatcher.defined_vars.get(
-                memlet.data)[0] == DefinedType.FPGA_ShiftRegister):
-            raise NotImplementedError("Shift register for Xilinx NYI")
-        else:
-            self._cpu_codegen.copy_memory(sdfg, dfg, state_id, src_node,
-                                          dst_node, edge, None, callsite_stream)
+    def unparse_tasklet(self, *args, **kwargs):
+        cpp.unparse_tasklet(*args, self, **kwargs)
```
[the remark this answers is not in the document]
definelicht: Changed the signature a bit to make it more clear and added a comment

test_all.sh
```diff
@@ -8,7 +8,7 @@ PYTHONPATH=$SCRIPTPATH
 DACE_debugprint="${DACE_debugprint:-0}"
 DACE_testing_serialization="${DACE_testing_serialization:-1}"
 DACE_testing_single_cache="${DACE_testing_single_cache:-1}"
-DACE_optimizer_interface="${DACE_optimizer_interface:-dace.transformation.optimizer.SDFGOptimizer}"
+DACE_optimizer_transform_on_call="${DACE_optimizer_transform_on_call:-dace.transformation.optimizer.SDFGOptimizer}"
```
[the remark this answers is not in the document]
definelicht: Ok, what does this syntax mean `${herp:-derp}`?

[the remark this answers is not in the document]
definelicht: Fixed, I think

dace/codegen/targets/intel_fpga.py
definelicht: Right, so do I implement this here? I can't access the dtype for the reason in the comment

dace/codegen/targets/intel_fpga.py
definelicht: Fixed with an explicit function

dace/codegen/targets/intel_fpga.py
```diff
@@ -327,6 +327,9 @@ def make_write(defined_type, dtype, var_name, write_expr, index, read_expr,
                         read_expr)
             else:
                 if is_unpack:
+                    ctype = dtype.base_type.base_type.ctype
```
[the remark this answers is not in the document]
definelicht: Fixed with an explicit function

dace/codegen/targets/intel_fpga.py
```diff
@@ -307,9 +305,11 @@ def make_write(defined_type, dtype, var_name, write_expr, index, read_expr,
...
```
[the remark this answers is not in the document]
definelicht: Fixed with an explicit function

dace/codegen/targets/intel_fpga.py
```diff
@@ -263,14 +261,14 @@ def make_read(defined_type, dtype, var_name, expr, index, is_pack,
             raise NotImplementedError(
                 "Unimplemented read type: {}".format(defined_type))
         if is_pack:
-            return "pack_{}{}(&({}))".format(dtype.base_type.ctype,
-                                             packing_factor, read_expr)
+            ctype = dtype.base_type.base_type.ctype
```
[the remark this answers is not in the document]
definelicht: It's not, it's `dace.pointer(dace.vector(dtype, veclen))`

dace/codegen/targets/fpga.py
definelicht: `FPGACodegen` does not inherit from `CPUCodegen`, so I cannot override anything (that's why we have this `self._cpu_codegen` object floating around).

I believe we need new codegen for this, so we can have callbacks or change the inheritance structure. Do you agree?

definelicht: Did a more "correct" solution by changing `cpu.py` to accept a callback codegen.

test_all.sh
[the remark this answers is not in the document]
tbennun: > Ok, what does this syntax mean `${herp:-derp}`?

Means "keep `$herp` as-is if defined, or set `$herp` to `derp` (i.e., some default value for the variable).

dace/codegen/targets/intel_fpga.py
tbennun: in dtypes.py we can have an `opencltype` (or `ctype`)?

dace/codegen/targets/cpu.py
```diff
@@ -1285,12 +1286,33 @@ def _generate_Tasklet(self, sdfg, dfg, state_id, node, function_stream,
                               node)
         callsite_stream.write('{', sdfg, state_id, node)
         callsite_stream.write(inner_stream.getvalue(), sdfg, state_id, node)
+        callsite_stream.write(after_memlets_stream.getvalue())
```
tbennun, line 1289: Thanks for the fix :)

dace/codegen/targets/cpu.py
```diff
@@ -1285,12 +1286,33 @@ def _generate_Tasklet(self, sdfg, dfg, state_id, node, function_stream,
...
         callsite_stream.write(outer_stream_end.getvalue(), sdfg, state_id, node)
 
         self._dispatcher.defined_vars.exit_scope(node)
 
-        callsite_stream.write(after_memlets_stream.getvalue())
+
+    def unparse_tasklet(self, sdfg, state_id, dfg, node, function_stream,
+                        inner_stream, locals, ldepth, toplevel_schedule):
+        # Call the generic CPP unparse_tasklet method
+        unparse_tasklet(sdfg, state_id, dfg, node, function_stream,
```
tbennun, line 1299: to be clearer, shall we just import `cpp` and call `cpp.unparse_tasklet` here?
Minor comment though

dace/codegen/targets/fpga.py
```diff
@@ -1221,3 +1236,35 @@ def generate_host_function_boilerplate(self, sdfg, state, kernel_name,
...
+                                   function_stream, before_memlets_stream,
+                                   after_memlets_stream):
+        # Inject dependency pragmas on memlets
+        for edge in dfg.out_edges(node):
+            datadesc = sdfg.arrays[edge.data.data]
+            if (isinstance(datadesc, dace.data.Array)
+                    and (datadesc.storage == dace.StorageType.FPGA_Local
+                         or datadesc.storage == dace.StorageType.FPGA_Registers)
+                    and not cpp.is_write_conflicted(dfg, edge)):
+                self.generate_no_dependence_post(edge.src_conn,
```
tbennun, line 1268: Nice and clean! 👍 
Shouldn't this be in xilinx.py though? (not sure OpenCL has a no_dependence pragma equivalent?)

test_all.sh
```diff
@@ -9,6 +9,7 @@ DACE_debugprint="${DACE_debugprint:-0}"
 DACE_testing_serialization="${DACE_testing_serialization:-1}"
 DACE_testing_single_cache="${DACE_testing_single_cache:-1}"
 DACE_optimizer_interface="${DACE_optimizer_interface:-dace.transformation.optimizer.SDFGOptimizer}"
+DACE_optimizer_transform_on_call="${DACE_optimizer_transform_on_call:-1}"
```
tbennun, line 12: Yep, fixed 👍

tests/intel_fpga/vec_sum.py
```diff
@@ -43,9 +40,33 @@ def sum(i: _[0:N]):
...
+            FPGATransformSDFG,
+            dace.transformation.dataflow.vectorization.Vectorization
+        ]
+        transformation_options = [{}, {
+            "propagate_parent": True,
+            "postamble": False
+        }]
+
+    for t, to in zip(transformations, transformation_options):
+        sdfg.apply_transformations(t, to)
```
tbennun, line 66: As far as I know, the API would work if you just fed in the two lists.

tests/fpga/veclen_conversion_connector.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
+import argparse
+import numpy as np
+from veclen_conversion import *
```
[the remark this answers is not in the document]
tbennun: @definelicht ?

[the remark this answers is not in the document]
definelicht: Apparently not! Do you want me to stick it there anyway?

dace/codegen/targets/intel_fpga.py
definelicht: Yes, I can take a look at implementing something like this, but I still can't implement this in a "correct" way (i.e. without the silly regex) before we have something like a global `make_read` and `make_write`.

dace/codegen/targets/cpu.py
definelicht: Yeah I agree, then I just have to track down every call to `cpp` in the code :wtfjoj: This is why `import *` sucks balls

dace/codegen/targets/fpga.py
definelicht: I chose to make this generic, so we have `no_dependence_pre` and `no_dependence_post` (symmetric for all the pragmas), so I wanna keep it this way for consistency :-) OpenCL has dependence pragmas, but before the code rather than after.

tests/intel_fpga/vec_sum.py
definelicht: Yeah I think you are right, it wasn't working for me initially, but for a different reason. Will change

dace/codegen/targets/cpu.py
definelicht: Went through every function in `cpp.py` and replaced calls in `cpu.py` :jojpride:

dace/codegen/targets/intel_fpga.py
definelicht: I've removed this for now, and left it for a future PR as part of new codegen.

dace/codegen/targets/cpu.py
```diff
@@ -1,12 +1,15 @@
 import itertools
 import warnings
 
-from dace import data, registry, memlet as mm
+import dace
+from dace import data, dtypes, registry, memlet, subsets, symbolic
```
tbennun, line 5: I don't think this is good to do at all. Within a python package you should import the modules as specifically as possible, so as to avoid import loops. I'm trying to avoid any instance of `import dace` from within dace

dace/codegen/targets/cpu.py
```diff
@@ -67,7 +70,7 @@ def __init__(self, frame_codegen, sdfg):
         # Register dispatchers
         dispatcher.register_node_dispatcher(self)
         dispatcher.register_map_dispatcher(
-            [dtypes.ScheduleType.CPU_Multicore, dtypes.ScheduleType.Sequential],
+            [dace.ScheduleType.CPU_Multicore, dace.ScheduleType.Sequential],
```
tbennun, line 73: see import comment above

dace/codegen/targets/cpu.py
tbennun, line 5: also, `from dace import memlet` might be problematic due to shadowing any variable called `memlet`

dace/codegen/targets/cpu.py
```diff
@@ -850,111 +855,9 @@ def process_out_memlets(self,
...
-            "Out" if is_output else "In",
-            dtype.ctype,
-            dims,
-            ", ".join(memlet_params),
-        )
-
     def memlet_definition(self,
                           sdfg: SDFG,
-                          memlet: mm.Memlet,
+                          memlet: memlet.Memlet,
```
tbennun, line 860: See above

tbennun (changes requested):
Almost LGTM

dace/codegen/targets/cpu.py
definelicht: You are commenting on `import dace`, not on `from dace import`, right?

definelicht: Removed `import dace`

tbennun (approved):
Looks good

PR #339, opened 2020-08-20
Improve error reporting for VSCode interactions

phschaad:
Moves Flask's debug and info prints to stdout instead of stderr, so stderr can be used for problem detection rather than verbose logging. Als sends potential exceptions/errors along with the response so VSCode can handle them accordingly.


dace/transformation/interface/vscode/__main__.py
```diff
@@ -1,6 +1,32 @@
...
+    # delays when booting in daemon mode.
+    from dace.sdfg import SDFG
+    try:
+        sdfg = SDFG.from_json(json)
+        error = None
+    except Exception as e:
+        error = {
+            'error': {
+                'message': 'Failed to parse the provided SDFG',
+                'details': get_exception_message(e),
```
tbennun, line 21: I would just use `str(e)` and include the type there too, i.e., `'%s: %s' % (type(e).__name__, e)`

dace/transformation/interface/vscode/__main__.py
```diff
@@ -1,6 +1,32 @@
...
+    else:
+        return str(exception)
+
+def load_sdfg_from_json(json):
+    # We lazy import SDFGs, not to break cyclic imports, but to avoid any large
+    # delays when booting in daemon mode.
+    from dace.sdfg import SDFG
+    try:
+        sdfg = SDFG.from_json(json)
+        error = None
```
tbennun, line 16: what if the SDFG has a validation error internally? Then the SDFG JSON (`json`) has a field called `error`....

tbennun (changes requested):

dace/transformation/interface/vscode/__main__.py
phschaad: Very good idea, the received message is a lot cleaner this way - the `get_exception_message` method has been changed accordingly.

dace/transformation/interface/vscode/__main__.py
phschaad: Also a very good point, I added a feature that will report internal SDFG validation errors back as well.

dace/transformation/interface/vscode/__main__.py
```diff
@@ -1,6 +1,42 @@
...
 
+def get_exception_message(exception):
+    return '%s: %s' % (type(exception).__name__, exception)
+
+def load_sdfg_from_json(json):
+    # We lazy import SDFGs, not to break cyclic imports, but to avoid any large
+    # delays when booting in daemon mode.
+    from dace.sdfg import SDFG
+
+    if ('error' in json):
```
tbennun, line 12:
```suggestion
    if 'error' in json:
```

dace/transformation/interface/vscode/__main__.py
```diff
@@ -9,11 +45,10 @@ def reapply_history_until(sdfg_json, index):
     :param sdfg_json:  The SDFG to rewind.
     :param index:      Index of the last history item to apply.
     """
-    # We lazy import DaCe, not to break cyclic imports, but to avoid any large
-    # delays when booting in daemon mode.
-    from dace.sdfg import SDFG
-
-    sdfg = SDFG.from_json(sdfg_json)
+    loaded = load_sdfg_from_json(sdfg_json)
+    if (loaded['error']):
```
tbennun, line 49:
```suggestion
    if loaded['error'] is not None:
```

dace/transformation/interface/vscode/__main__.py
```diff
@@ -37,18 +80,36 @@ def apply_transformation(sdfg_json, transformation):
     # We lazy import DaCe, not to break cyclic imports, but to avoid any large
     # delays when booting in daemon mode.
     from dace.transformation.pattern_matching import Transformation
-    from dace.sdfg import SDFG
 
-    sdfg = SDFG.from_json(sdfg_json)
-
-    revived_transformation = Transformation.from_json(transformation)
+    loaded = load_sdfg_from_json(sdfg_json)
+    if (loaded['error']):
```
tbennun, line 85:
```suggestion
    if loaded['error'] is not None:
```

dace/transformation/interface/vscode/__main__.py
```diff
@@ -59,9 +120,12 @@ def get_transformations(sdfg_json):
     # We lazy import DaCe, not to break cyclic imports, but to avoid any large
     # delays when booting in daemon mode.
     from dace.transformation.optimizer import SDFGOptimizer
-    from dace.sdfg import SDFG
 
-    sdfg = SDFG.from_json(sdfg_json)
+    loaded = load_sdfg_from_json(sdfg_json)
+    if (loaded['error']):
```
tbennun, line 125:
```suggestion
    if loaded['error'] is not None:
```

tbennun (approved):
Please fix remaining suggestions

PR #354, opened 2020-09-01
Interstate edge validation issue

alexnick83:
While validating interstate edges, there can be issues with assignments of Scalar values to symbols. These issues arise due to:
- The equality method `__eq__(self, other)` of the typeclass class, when `other` is a Scalar (or other Data object).
- The method `result_type_of(lhs, *rhs)`, when lhs/rhs are not typeclass objects or symbols.

To resolve the first issue, we add a `ctype` property to the Data class, that redirects to `Data.dtype.ctype`. For the second issue, we check if lhs/rhs are data objects. If this is true, we substitute them, for the rest of the method, with their dtype.


dace/dtypes.py
```diff
@@ -403,9 +403,10 @@ def result_type_of(lhs, *rhs):
 
     rhs = rhs[0]
 
-    # Extract the type if symbolic
-    lhs = lhs.dtype if type(lhs).__name__ == 'symbol' else lhs
```
alexnick83, line 407: Why do we check the name of the type, instead of something along the lines of `isinstance(lhs, (dace.symbol, sympy.symbol))`?

tbennun: Import loop

tbennun: Import loop

dace/data.py
```diff
@@ -126,6 +126,14 @@ def __repr__(self):
     def veclen(self):
         return self.dtype.veclen if hasattr(self.dtype, "veclen") else 1
 
+    # @property
+    # def type(self):
+    #     return self.dtype.type
```
tbennun, line 131:
```suggestion
```

tbennun (changes requested):
super minor change

tbennun (approved):

PR #364, opened 2020-09-15
Fix scalar bool args to SDFGs

orausch:


dace/dtypes.py
```diff
@@ -813,6 +813,39 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
 
+# Helper function to determine whether a global variable is a constant
+_CONSTANT_TYPES = [
```
orausch, line 817: This had to be moved because you overwrite the `bool` builtin below

tbennun (approved):

PR #371, opened 2020-09-24
Issues with redundant array transformations

alexnick83:
The redundant array transformations suffer from two issues:
- They do not check whether the edge subsets of the datanode/array to be removed are covered by the edge subset of the datanode/array that they will be connected to.
- Composing the above subsets in cases where they do not match exactly in size, returns erroneous ranges.


dace/transformation/dataflow/redundant_array.py
```diff
@@ -2,13 +2,84 @@
 """ Contains classes that implement a redundant array removal transformation.
 """
 
+import dace
```
tbennun, line 5: (style) please import the innermost module (`data as dt`, seems like?) instead of importing the entire package

dace/transformation/dataflow/redundant_array.py
```diff
@@ -2,13 +2,84 @@
...
+
+    # Find the src and dst subsets (deep-copy to allow manipulation)
+    if src_name == dataname:
+        src_subset = dcpy(edge.data.subset)
+        dst_subset = dcpy(edge.data.other_subset)
+    elif dst_name == dataname:
+        src_subset = dcpy(edge.data.other_subset)
+        dst_subset = dcpy(edge.data.subset)
+    else:
+        raise NotImplementedError
```
tbennun, line 42: I thought `edge.data.{src,dst}_subset` does the same?

dace/transformation/dataflow/redundant_array.py
```diff
@@ -158,13 +229,37 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
...
+            except NotImplementedError:
+                return False
+            # 2-b. Check where b1_subset covers b2_subset
+            if not b1_subset.covers(b2_subset):
+                return False
+            # 2-c. Validate subsets in memlet tree
+            # (should not be needed for valid SDGs)
+            path = graph.memlet_tree(e2)
+            for e3 in path:
+                if not e2 is e3:
```
tbennun, line 257:
```suggestion
                if e2 is not e3:
```

tests/python_frontend/identifiers_and_keywords_test.py
```diff
@@ -237,7 +237,7 @@ def test_keyword_lambda():
         keyword_lambda(A, B)
     except Exception as e:
         print(e)
-        return False
+        return True
```
tbennun, line 240: was this change intentional?

tbennun (changes requested):
minor changes and question

dace/transformation/dataflow/redundant_array.py
alexnick83: In one of the (many) earlier versions, I needed `dace.Memlet`. Is that top-level or not?

dace/transformation/dataflow/redundant_array.py
alexnick83: Is this being automatically set now? I still live in the old days ...

tests/python_frontend/identifiers_and_keywords_test.py
alexnick83: Yes, as we do not support lambda functions in the context that one is used in that particular test. The test was erroneously returning `False` because I initially thought I could support them with type annotations. This failed and then I just forgot to change the test. This brings up another point, returning `True` or `False` from those tests does not trigger `SUCCESS` or `ERROR`. This should be fixed in another PR.

dace/transformation/dataflow/redundant_array.py
tbennun: `from dace import memlet` (`as mmlt` or `as mm`) also works.

dace/transformation/dataflow/redundant_array.py
tbennun: Let's hope it works fine with everything :)

tbennun (approved):

PR #372, opened 2020-09-25
Disable Memcpy in HIP Kernel Code

fthaler:
Memcpy is supported by the hipcc compiler, but seems to block conversion from stack arrays into registers arrays in some cases.


tbennun (approved):
LGTM

Issue #369, opened 2020-09-23
Let can_be_applied throw exceptions


definelicht:
For the sake of our sanity, exceptions thrown by `can_be_applied` are lowered to warnings by default.

However, for debugging purposes (or for any other situation where you would _want_ the exception to be thrown), it should be possible to have it _actually_ throw the error.

We can add this as a configuration flag, either globally or per-transformation.

tbennun:
Fixed in #373

PR #373, opened 2020-09-25
Transformation ease-of-use API and tutorial

tbennun:
- [x] Refactor
- [x] `sdfgcc` entry point to optimization
- [x] `Transformation.apply_to` API
- [x] Tutorial
- [x] `match` can raise exceptions (#369)
- [x] `enumerate_matches` general pattern matching API


dace/transformation/pattern_matching.py
```diff
@@ -64,6 +64,23 @@ def type_match(node_a, node_b):
     return isinstance(node_a['node'], type(node_b['node']))
 
 
+def type_or_class_match(node_a, node_b):
+    """ Checks whether the node types of the inputs match.
```
definelicht, line 68: What exactly does it mean that "types of the inputs match"? Maybe this needs a bit of context?

definelicht (changes requested):

dace/transformation/pattern_matching.py
tbennun: Thanks, fixed.

definelicht (approved):

PR #390, opened 2020-10-08
Remove the VSCode interface

phschaad:
The VSCode interface was moved to the VSCode extension itself, so this removes it from DaCe again.


tbennun (changes requested):
Please remove from other meta-files (codecov etc.)

phschaad:
Don't think it should be anywhere else, can't grep it in any other meaningful place. Let me know if I overlooked something

tbennun (approved):

Issue #392, opened 2020-10-09
Generate code for duplicated NestedSDFG only once


TizianoDeMatteis:
It would be nice to have a way to avoid generating multiple times code for a nested SDFG that appears multiple times in the SDFG.

- It must be understood how to detect duplicated NestedDFG: for example, we can add another property to the SDFG, such as "unique_name" or something like this

- how to trigger this behavior: we may don't want to do this always

PR #393, opened 2020-10-09
Generate Duplicated NestedSDFGs only once

TizianoDeMatteis:
PR for #392 

First commit, so that we can iterate.

**Goal:** avoid generating 2+ times the same code for a NestedSDFGs that is used multiple times (which include also LibNodes after expansion)

**Current implementation**:
- we need to unequivocally identify SDFG. For the moment, I've added to the SDFG an additional property `unique_name` (type string, default empty)
- this is used in the code generator (CPU) to keep track of the already generated Nested SDFG. If we try to generate an already seen NestedSDFG, it will skip it
- there are two additional tests for checking that it works on CPU and FPGA (under the assumption that the topmost SDFG is scheduled on the CPU)

So, up to now, it is up to the user to specify the SDFG unique name. We would probably need something (in the configuration file?), to disable this.

Don't know why code coverage fails: the relative difference is 100%


dace/codegen/targets/cpu.py
```diff
@@ -1411,39 +1424,42 @@ def _generate_NestedSDFG(
...
+                                              memlet_references, sdfg_label),
+                sdfg, state_id, node)
 
         #############################
         # Generate function contents
-        self._frame.generate_constants(node.sdfg, nested_stream)
 
-        old_schedule = self._toplevel_schedule
-        self._toplevel_schedule = node.schedule
+        if not code_already_generated:
```
tbennun, line 1436: wouldn't you prefer to do `if code_already_generated: return`?

dace/codegen/targets/cpu.py
```diff
@@ -148,11 +152,10 @@ def generate_node(self, sdfg, dfg, state_id, node, function_stream,
 
         gen(sdfg, dfg, state_id, node, function_stream, callsite_stream)
 
+        self._locals.clear_scope(self._ldepth + 1)
         # Mark node as "generated"
         self._generated_nodes.add(node)
 
-        self._locals.clear_scope(self._ldepth + 1)
-
     def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
```
tbennun, line 159: That might not be a good idea

dace/codegen/targets/cpu.py
```diff
@@ -1456,7 +1472,8 @@ def _generate_NestedSDFG(
         ###############################################################
         # Write generated code in the proper places (nested SDFG writes
         # location info)
-        function_stream.write(global_code)
+        if not code_already_generated:
+            function_stream.write(global_code)
```
tbennun, line 1476: not sure why you'd want to put that in a condition, since above you skip the generation of the function

tbennun (changes requested):

dace/codegen/targets/cpu.py
TizianoDeMatteis: Thanks for spotting it. Some wrong editing!

dace/codegen/targets/cpu.py
TizianoDeMatteis: global_code is created in one of the ifs above

TizianoDeMatteis:
Also @tbennun : the `unique_name` property (set at the original SDFG name) is still used to generate the name for the function, otherwise it uses a combination of the name and various id to generate it. I don't like it too much.

definelicht:
@TizianoDeMatteis I discussed this with Tal, and I think there's a pretty clean solution that will benefit the whole codegen (not just FPGA). We generate nested SDFGs bottom up instead of top down:
- When we serialize the SDFG before code generation, we will also hash each nested SDFG, and build a call graph for nested SDFGs, where nested SDFGs are uniquely identified by the hash of their string representation.
- Instead of starting from the top-level SDFG, we traverse the nested SDFG call graph backwards (reverse toppological order), and generate the SDFGs in this order.
- Whenever we encounter a nested SDFG while traversing, we simply replace this with a function call and return (the function has already been generated at this point).

TizianoDeMatteis:
@definelicht @tbennun 

I didn't quite get your proposals:
1. What is the rationale behind using a reverse topological order?
2. How to traverse NestedSDFG using that order: this seems to be different than how code is currently generated
3. Where should be placed the code for the Nested SDFG? Some additional file that will be then included/compiled together

Independently from this, we should decide how to hash (symbols are included? access nodes?...)

tbennun:
@TizianoDeMatteis 
1. I think the meaning was more "bottom up in the nested SDFG hierarchy"
2. The difference is we generate the nested-most SDFGs first, and then record that we have (so that we don't generate them again)
3. Where we always did, but move the code to generate the nested SDFG's "body" to the frame code generator.
4. In the discussion Johannes and I had, we found that while it would be beneficial to ignore connector/array names inside the nested SDFG, right now in our standard use case they're all going to be the same anyway. Thoughts/counter-arguments?

TizianoDeMatteis:
@tbennun 
Regarding 2: this still seems a different ordering wrt to what is currently used. IIRC now code is generated by following the topological ordering of the SDFG, while in this proposal we should first generate innermost components (that is the bottom of NestedSDFG hierarchy). But, I will probably have more precise questions when I will work on this

Regarding 4. Ok, I think is that the main problem could be related to symbols. But that (the way in which hashing is done) could be easily changed later on.

Another minor thing: the way in which function names for nested SDFG is generated must be changed as well.

TizianoDeMatteis:
We should decide whether to have a configuration to enable/disable this behavior. Maybe be useful especially for FPGAs, where we explicitly want to generate duplicated code

dace/sdfg/sdfg.py
```diff
@@ -324,6 +323,19 @@ def from_json(cls, json_obj, context_info=None):
 
         return ret
 
+    def hash_sdfg(self) -> int:
+        '''
+        Returns an hash of the current SDFG, withou considering IDs and attributes name
+        :return:
+        '''
+        jsondict = self.to_json()  # No more nonstandard objects
```
tbennun, line 331: this would still not work with nested SDFGs in nested SDFGs. We need to probably walk into the entire key structure of the dictionary and delete all `sdfg_list_id` and `['attributes']['name']` keys

tbennun (changes requested):
One (not so) minor comment

dace/codegen/targets/cpu.py
```diff
@@ -1398,18 +1398,22 @@ def _generate_NestedSDFG(
...
-        # use the saved name to call it, otherwise save the hash and the associated name
-        hash = node.sdfg.hash_sdfg()
-        if hash in self._generated_nested_sdfg:
-            code_already_generated = True
-            sdfg_label = self._generated_nested_sdfg[hash]
-        else:
-            sdfg_label = "%s_%d_%d_%d" % (node.sdfg.name, sdfg.sdfg_id,
-                                          state_id, dfg.node_id(node))
-            self._generated_nested_sdfg[hash] = sdfg_label
+        unique_functions = Config.get('compiler', 'unique_functions')
```
tbennun, line 1401:
```suggestion
        unique_functions = Config.get_bool('compiler', 'unique_functions')
```

dace/config_schema.yml
```diff
@@ -167,6 +167,17 @@ required:
...
                     that the bitstream is present in the build directory, and
                     will crash if this is not the case.
+
+            unique_functions:
+                type: bool
+                default: true
+                title: Generate unique functions
+                description: >
+                    If set to false, for any Nested SDFG it is generated the
+                    corresponding code.
```
tbennun, line 177:
```suggestion
```

dace/config_schema.yml
```diff
@@ -167,6 +167,17 @@ required:
...
+                    If set to true, if  multiple, equal Nested SDFGs exist,
```
tbennun, line 178:
```suggestion
                    If set to true, if multiple nested SDFGs with equivalent contents exist,
```

tests/unique_nested_sdfg_test.py
```diff
@@ -63,9 +69,71 @@ def make_vecAdd_sdfg(sdfg_name: str, dtype=dace.float32):
...
+                                                             "0:size",
+                                                             num_accesses=n))
+
+    return vecAdd_parent_sdfg
+
+
 def make_nested_sdfg_cpu_single_state():
     '''
-    Build an SDFG with two nested SDFGs
+    Builds an SDFG with two identic nested SDFGs
```
tbennun, line 136:
```suggestion
    Builds an SDFG with two identical nested SDFGs
```

tbennun:
@TizianoDeMatteis very minor issues (documentation) remain, please make it ready-for-review once tests pass and comments have been addressed (seems like it works, right?)

tbennun:
@TizianoDeMatteis if it's done and passing all tests, make sure it's not WIP and re-tag me for review. looks like some minor documentation updates

tbennun (approved):

PR #398, opened 2020-10-13
Fix for integer (but not `int`) symbols

alexnick83:
Symbols that are of type `long long` or any other integer type apart from `int` are currently generated without the assumption `Integer=True`. The purpose of this PR is to fix this issue.


tbennun (approved):
Great!

PR #401, opened 2020-10-14
Change schedule inference for SDFG expansions

orausch:


tbennun:
Problem:

GPU_Device and nested SDFG node schedule behave oddly.

Example 1:
* Top-level / CPU_Multicore Map
  * Nested SDFG with GPU_Device
    * Default Map
Default turns into GPU_Device

Example 2:
* GPU_Device Map
  * Nested SDFG with Default -> GPU_Device (due to schedule inference)
    * Default Map
Default turns into GPU_ThreadBlock

Inconsistent!

We cannot remove nested SDFG schedule because it is used in FPGAs

tbennun:
After a meeting today we decided to try and remove `NestedSDFG.schedule` altogether and see what happens. This will be done in a separate PR.

Issue #97, opened 2020-01-17
Python frontend: constants / symbols do not appear in generated code


tbennun:
**To Reproduce**
```python
@dace.program
def subrange_of_subrange(A: dace.float32[2, 3, 4, 5], B: dace.float32[4]):
    i = 0
    j = 0
    k = 0
    B[:] = A[:, i, :, j][k, :]
```
`i`, `j`, and `k` do not appear in the generated code

alexnick83:
They currently appear, but not in the same scope:
```cpp
    {
        long long i;
        
        
        {
            long long __out;

            ///////////////////
            // Tasklet code (assign_55_4)
            __out = 0;
            ///////////////////

            i = __out;
        }
    }
    {
        long long j;
        
        
        {
            long long __out;

            ///////////////////
            // Tasklet code (assign_56_4)
            __out = 0;
            ///////////////////

            j = __out;
        }
    }
    {
        long long k;
        
        
        {
            long long __out;

            ///////////////////
            // Tasklet code (assign_57_4)
            __out = 0;
            ///////////////////

            k = __out;
        }
    }
    {
        
        

        dace::CopyND<float, 1, false, 2, 4>::template ConstDst<4, 1>::Copy(
        A + ((20 * i) + j), __tmp3, 60, 5);
    }
    {
        
        

        dace::CopyND<float, 1, false, 4>::template ConstDst<1>::Copy(
        __tmp3 + (4 * k), __tmp4, 1);
    }
    {
        
        

        dace::CopyND<float, 1, false, 4>::template ConstDst<1>::Copy(
        __tmp4, B, 1);
    }
```

The issue is that they are not in the same scope. We have already decided that i, j, k should be indirections, which will fix the issue. However, it is a good idea to keep the code-generation issue in mind as well.

tbennun:
Great, good to keep open then.

PR #405, opened 2020-10-15
Generate indirection for memlet subsets using scalar data (non-subscripted data names)

alexnick83:
Enforcing proper SDFG semantics (indirection) and fix for #97.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -259,18 +259,18 @@ def _disallow_stmt(visitor, node):
 ###############################################################
 
 
-def _subset_has_indirection(subset, PVisitor: 'ProgramVisitor' = None):
+def _subset_has_indirection(subset, pvisitor: 'ProgramVisitor' = None):
     for dim in subset:
         if not isinstance(dim, tuple):
             dim = [dim]
         for r in dim:
             if symbolic.contains_sympy_functions(r):
                 return True
-            if PVisitor:
+            if pvisitor:
                 for s in r.free_symbols:
                     try:
-                        name = PVisitor._visitname(str(s), None)
-                        if name in PVisitor.sdfg.arrays:
+                        name = pvisitor._visitname(str(s), None)
+                        if name in pvisitor.sdfg.arrays:
                             return True
                     except DaceSyntaxError:
                         continue
@@ -283,7 +283,7 @@ def add_indirection_subgraph(sdfg: SDFG,
                              dst: nodes.Node,
                              memlet: Memlet,
                              local_name: str,
-                             PVisitor: 'ProgramVisitor',
+                             pvisitor: 'ProgramVisitor',
                              output: bool = False):
     """ Replaces the specified edge in the specified graph with a subgraph that
         implements indirection without nested memlet subsets. """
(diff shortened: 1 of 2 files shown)


dace/frontend/python/newast.py
```diff
@@ -375,18 +396,20 @@ def add_indirection_subgraph(sdfg: SDFG,
                 # TODO: Make this work for nested for-loops
                 arr_rng = dace.subsets.Range([(a, a, 1) for a in access])
                 if output:
-                    arrname, _ = PVisitor._add_write_access(arr_name,
+                    arrname, rng = PVisitor._add_write_access(arr_name,
                                                             arr_rng,
                                                             target=None)
                 else:
-                    arrname, _ = PVisitor._add_read_access(arr_name,
+                    arrname, rng = PVisitor._add_read_access(arr_name,
```
tbennun, line 403: are you supposed to use `rng` somewhere? There's `arr_rng` below and the subset range that we create from the whole array

dace/frontend/python/newast.py
```diff
@@ -259,13 +259,21 @@ def _disallow_stmt(visitor, node):
 ###############################################################
 
 
-def _subset_has_indirection(subset):
+def _subset_has_indirection(subset, PVisitor: 'ProgramVisitor' = None):
```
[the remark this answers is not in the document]
tbennun: Also, since all the uses below use `self`, we should consider making it a method.

dace/frontend/python/newast.py
```diff
@@ -2535,13 +2576,33 @@ def _add_access(
         self.accesses[(name, rng, access_type)] = (var_name, squeezed_rng)
 
         inner_indices = set(non_squeezed)
+        # From TaskletTransformer
+        # inner_indices = set()
+        # for n, r in reversed(list(enumerate(squeezed_rng))):
+        #     if r == rng[n]:
+        #         inner_indices.add(n)
```
tbennun, line 2583:
```suggestion
```

dace/frontend/python/newast.py
```diff
@@ -2535,13 +2576,33 @@ def _add_access(
...
+            else:
+                self.outputs[var_name] = (dace.Memlet.simple(
+                    parent_name, rng), inner_indices)
+
+        # if access_type == 'r':
+        #     self.inputs[var_name] = (dace.Memlet.simple(parent_name,
+        #                                                 rng), inner_indices)
+        # else:
+        #     self.outputs[var_name] = (dace.Memlet.simple(parent_name,
+        #                                                  rng), inner_indices)
```
tbennun, line 2605:
```suggestion
```

tbennun (commented):
Some comments (and change suggestions that are not mandatory) before approval

dace/frontend/python/newast.py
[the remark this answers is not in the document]
alexnick83: On the second comment, no (not yet), because the method is also called by `TaskletTransformer.` Eventually, we need to refactor the frontend to have another class, that will contain the methods `_subset_has_indirection`, `add_indirection_subgraph`, `_add_(read/write)_access`, etc. `TaskletTransformer` and `ProgramVisitor` will inherit from this class. However, it is a major change, for which I don't currently have time.

dace/frontend/python/newast.py
alexnick83: I was going to use it, but I found `subsets::Range::from_array` better suited (for the time being).

tbennun (approved):

PR #417, opened 2020-10-17
Style cleanup #1: Transformations

tbennun:
- [x] More documentation in Transformations
- [x] Implement the `PatternNode` property type throughout pattern-matching transformations
- [x] Update tutorial


dace/properties.py
```diff
@@ -415,7 +415,8 @@ def initialize_properties(obj, *args, **kwargs):
         # Assert that there are no fields in the object not captured by
         # properties, unless they are prefixed with "_"
         for name, prop in obj.__dict__.items():
-            if name not in properties and not name.startswith("_"):
+            if (name not in properties and not name.startswith("_")
```
am-ivanov, line 418: It is still a bit ugly, can we completely avoid operation on strings here?

dace/sdfg/graph.py
```diff
@@ -207,15 +207,15 @@ def add_nodes_from(self, node_list):
         for node in node_list:
             self.add_node(node)
 
-    def node_id(self, node):
+    def node_id(self, node) -> int:
```
am-ivanov, line 210: What is the type of node?

dace/sdfg/graph.py
```diff
@@ -207,15 +207,15 @@ def add_nodes_from(self, node_list):
...
         """Returns a numeric ID that corresponds to the node index in the
            internal graph representation (unique)."""
         for i, n in enumerate(self.nodes()):
             if node == n:
                 return i
         raise NodeNotFoundError(node)
 
-    def edge_id(self, edge):
+    def edge_id(self, edge) -> int:
```
am-ivanov, line 218: type of edge

dace/sdfg/graph.py
```diff
@@ -451,7 +451,7 @@ def add_node(self, node):
     def add_nodes_from(self, node_list):
         raise PermissionError
 
-    def node_id(self, node):
+    def node_id(self, node) -> int:
```
am-ivanov, line 454: type of node

dace/transformation/dataflow/local_storage.py
```diff
@@ -10,17 +10,17 @@
...
 @make_properties
-class LocalStorage(transformation.Transformation, ABC):
+class LocalStorage(Transformation, ABC):
     """ Implements the Local Storage prototype transformation, which adds a
         transient data node between two nodes.
     """
 
-    _node_a = nodes.Node()
-    _node_b = nodes.Node()
+    node_a = PatternNode(nodes.Node)
```
am-ivanov, line 22: can we import a module instead of a class?

dace/transformation/dataflow/map_expansion.py
```diff
@@ -24,33 +24,33 @@ class MapExpansion(pm.Transformation):
...
+        map_entry = graph.node(candidate[MapExpansion.map_entry])
         return map_entry.map.get_param_num() > 1
 
     @staticmethod
-    def match_to_str(graph: dace.sdfg.graph.OrderedMultiDiConnectorGraph,
-                     candidate: Dict[dace.sdfg.nodes.Node, int]):
-        map_entry = graph.nodes()[candidate[MapExpansion._map_entry]]
+    def match_to_str(graph: dace.SDFGState, candidate: Dict[pm.PatternNode,
+                                                            int]) -> str:
+        map_entry = graph.nodes()[candidate[MapExpansion.map_entry]]
```
am-ivanov, line 47: can we use `node()` instead of `nodes()` here?

dace/transformation/pattern_matching.py
```diff
@@ -7,7 +7,7 @@
 import networkx as nx
 from networkx.algorithms import isomorphism as iso
 from typing import Dict, Iterator, List, Tuple, Type, Union
-from dace.transformation.transformation import Transformation
+from dace.transformation.transformation import Transformation, PatternNode
```
am-ivanov, line 10: import module instead of class

dace/transformation/transformation.py
```diff
@@ -9,20 +19,39 @@
 from dace.sdfg.graph import SubgraphView
```
am-ivanov, line 19: module instead of class

dace/transformation/transformation.py
```diff
@@ -55,7 +84,11 @@ def expressions():
         raise NotImplementedError
 
     @staticmethod
-    def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
+    def can_be_applied(graph: Union[SDFG, SDFGState],
```
am-ivanov, line 87: `graph: Union[SDFG, SDFGState]` is confusing. Can we have separate functions with different names for SDFG and SDFGState? can we replace `graph` with `sdfg` and `sdfg_state` depending on the concrete type?

dace/transformation/transformation.py
```diff
@@ -72,18 +105,31 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
...
+        Applies this transformation instance on the matched pattern graph.
+        :param sdfg: The SDFG to apply the transformation to.
+        :return: A transformation-defined return value, which could be used
+                 to pass analysis data out, or nothing.
+        """
+        raise NotImplementedError
+
     @staticmethod
-    def match_to_str(graph, candidate):
+    def match_to_str(graph: Union[SDFG, SDFGState],
```
am-ivanov, line 118: can we avoid union of types?

dace/transformation/transformation.py
```diff
@@ -164,12 +230,19 @@ def __lt__(self, other):
...
+    def _get_pattern_nodes(cls) -> Dict[str, 'PatternNode']:
+        """ 
+        Returns a dictionary of pattern-matching node in this transformation
+        subclass. Used internally for pattern-matching.
+        :return: A dictionary mapping between pattern-node name and its type.
+        """
+        return {
+            k: getattr(cls, k)
+            for k in dir(cls)
+            if isinstance(getattr(cls, k), PatternNode) or (k.startswith(
```
am-ivanov, line 243: There was a ruse in google style guide that suggests avoiding multiline conditions in comprehensions. Can we extract the condition into a method?

dace/transformation/transformation.py
```diff
@@ -272,7 +354,7 @@ def to_json(self, parent=None):
         }
 
     @staticmethod
-    def from_json(json_obj, context=None):
+    def from_json(json_obj, context=None) -> 'Transformation':
```
am-ivanov, line 357: what are the types?

dace/transformation/transformation.py
```diff
@@ -296,11 +378,65 @@ def from_json(json_obj, context=None):
...
+    class MyTransformation(Transformation):
+        some_map_node = PatternNode(nodes.MapEntry)
+        array = PatternNode(nodes.AccessNode)
+    ```
+
+    The two nodes can then be used in the transformation static methods (e.g.,
+    `expressions`, `can_be_applied`) to represent the nodes, and in the instance
+    methods to point to the nodes in the parent SDFG.
+    """
+    def __init__(self, nodeclass: Type[Union[nd.Node, SDFGState]]) -> None:
```
am-ivanov, line 399: can we avoid the union of types? For example with two classmethods `instance_node(node: nd.Node)`, `instance_state(state: SDFGState)` that returns instance of PatternNode?

dace/transformation/transformation.py
```diff
@@ -296,11 +378,65 @@ def from_json(json_obj, context=None):
...
+        """
+        Initializes a pattern-matching node.
+        :param nodeclass: The class of the node to match (can either be a state
+                          or a node type in a state).
+        """
+        self.node = nodeclass
+        self.match_instance: Optional[Transformation] = None
+
+    def __call__(self, sdfg: SDFG) -> Union[nd.Node, SDFGState]:
```
am-ivanov, line 408: can we avoid union in return values?

tutorials/transformations.ipynb
```diff
@@ -608,30 +608,30 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from dace.transformation.transformation import Transformation\n",
+    "from dace.transformation.transformation import PatternNode, Transformation\n",
```
am-ivanov, line 611: import module instead of class?

am-ivanov (changes requested):

dace/properties.py
tbennun: This is how properties work (@definelicht). We say that anything that starts with `self._` is a private field and thus should not be saved with the serialization etc. There is no way of avoiding it without breaking all the code we currently have.

dace/transformation/transformation.py
tbennun: Unfortunately not, because a transformation can be used on multiple states (and then graph is `SDFG`) or a single state. If you want, you can make a separate PR where Transformation is extended to `SinglestateTransformation` and `InterstateTransformation`.

dace/transformation/transformation.py
tbennun: see above.

dace/transformation/transformation.py
tbennun: This would make the code unnecessarily longer IMO.

dace/transformation/transformation.py
tbennun: That would make the syntax more confusing rather than what it looks like now (`PatternNode(class)`). If you want, I can just make it Type and it will be completely unclear what it might be.

dace/transformation/transformation.py
tbennun: Unfortunately not, for the same reasons as above

dace/properties.py
definelicht: Changing this would be a huge PR on its own, so if we want to consider this, it should be a PR of its own.

tbennun: OK then I'll resolve this for now and queue up another PR for this. We should consider existing projects like Python's built-in dataclasses and pydantic.

dace/transformation/transformation.py
am-ivanov: Longer is better if it is more readable, feel free to leave as is if not agree.

dace/sdfg/graph.py
```diff
@@ -207,15 +207,15 @@ def add_nodes_from(self, node_list):
         for node in node_list:
             self.add_node(node)
 
-    def node_id(self, node):
+    def node_id(self, node: Any) -> int:
```
am-ivanov, line 210: `Any` is no better than no type. Why it is so difficult to accept one concrete type (or union at least)?

dace/sdfg/graph.py
```diff
@@ -451,7 +451,7 @@ def add_node(self, node):
     def add_nodes_from(self, node_list):
         raise PermissionError
 
-    def node_id(self, node):
+    def node_id(self, node: Any) -> int:
```
am-ivanov, line 454: I don't like Any :)

dace/sdfg/graph.py
```diff
@@ -207,15 +207,15 @@ def add_nodes_from(self, node_list):
...
         """Returns a numeric ID that corresponds to the node index in the
            internal graph representation (unique)."""
         for i, n in enumerate(self.nodes()):
             if node == n:
                 return i
         raise NodeNotFoundError(node)
 
-    def edge_id(self, edge):
+    def edge_id(self, edge: Edge) -> int:
```
am-ivanov, line 218: why Edge is not inside the module (e.g. dace.sdfg,Edge)?

dace/transformation/transformation.py
am-ivanov: Yes, probably it should go in a separate PR

dace/transformation/transformation.py
```diff
@@ -107,29 +152,49 @@ def __init__(self, sdfg_id, state_id, subgraph, expr_index):
...
         self._subgraph = {expr.node_id(k): v for k, v in subgraph.items()}
-        self._subgraph_user = subgraph
+        self._subgraph_user = copy.copy(subgraph)
         self.expr_index = expr_index
 
+        # Ease-of-use API: Set new pattern-nodes with information about this
+        # instance.
+        for pname, pval in self._get_pattern_nodes().items():
+            # Create new pattern node from existing field
+            new_pnode = PatternNode(
```
am-ivanov, line 162: why PatternNode is not inside the module name?

dace/transformation/transformation.py
```diff
@@ -107,29 +152,49 @@ def __init__(self, sdfg_id, state_id, subgraph, expr_index):
...
+                pval.node if isinstance(pval, PatternNode) else type(pval))
```
am-ivanov, line 163: is it really required to use ugly `isinstance` here? Why we can't have one specific type there?

dace/transformation/transformation.py
```diff
@@ -107,29 +152,49 @@ def __init__(self, sdfg_id, state_id, subgraph, expr_index):
...
+            setattr(self, pname, new_pnode)
+
     @property
     def subgraph(self):
         return self._subgraph_user
 
-    def query_node(
-        self, sdfg: SDFG, pattern_node: Union[nd.Node, SDFGState]
-    ) -> Union[nd.Node, SDFGState]:
+    def apply_pattern(self, sdfg: SDFG) -> Union[Any, None]:
```
am-ivanov, line 177: Now it is worse than before :) I would say that `Any` is the worst (the same as no annotation), `Union` is better but is still a candidate for refactoring. Ideally, it would be nice to see a concrete return type.

dace/transformation/transformation.py
```diff
@@ -243,12 +324,12 @@ def apply_to(cls,
...
         # Apply to SDFG
-        instance.apply_pattern(sdfg)
+        return instance.apply_pattern(sdfg)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return type(self).__name__
 
-    def print_match(self, sdfg):
+    def print_match(self, sdfg: SDFG) -> str:
```
am-ivanov, line 332: SDFG without module name

dace/transformation/transformation.py
am-ivanov: Just Type or Any is definitely worse. I don't see why it will be more confusing with two classmethods. My main argument is that if you see call instance_node(x) somewhere in the code, you immediately understand what is the type of x. With the current interface, I would need to run a debugger to see what exactly is passed.

dace/transformation/transformation.py
am-ivanov: Probably another approach can be used here: make different PatternNode implementations that accept different types. Anyway, it seems a good candidate for refactoring (maybe in another PR).

dace/transformation/transformation.py
```diff
@@ -386,20 +525,20 @@ class SubgraphTransformation(object):
                            desc='Subgraph in transformation instance')
 
     def __init__(self,
-                 subgraph: Union[Set[int], SubgraphView],
+                 subgraph: Union[Set[int], gr.SubgraphView],
                  sdfg_id: int = None,
                  state_id: int = None):
-        if (not isinstance(subgraph, (SubgraphView, SDFG, SDFGState))
+        if (not isinstance(subgraph, (gr.SubgraphView, SDFG, SDFGState))
```
am-ivanov, line 531: SDFG and SDFGState without module name

dace/transformation/transformation.py
```diff
@@ -418,14 +557,15 @@ def __init__(self,
             self.sdfg_id = sdfg_id
             self.state_id = state_id
 
-    def subgraph_view(self, sdfg: SDFG) -> SubgraphView:
+    def subgraph_view(self, sdfg: SDFG) -> gr.SubgraphView:
```
am-ivanov, line 560: SDFG without module name

dace/transformation/transformation.py
```diff
@@ -418,14 +557,15 @@ def __init__(self,
...
         graph = sdfg.sdfg_list[self.sdfg_id]
         if self.state_id != -1:
             graph = graph.node(self.state_id)
-        return SubgraphView(graph, [graph.node(idx) for idx in self.subgraph])
+        return gr.SubgraphView(graph,
+                               [graph.node(idx) for idx in self.subgraph])
 
     @staticmethod
-    def can_be_applied(sdfg: SDFG, subgraph: SubgraphView) -> bool:
+    def can_be_applied(sdfg: SDFG, subgraph: gr.SubgraphView) -> bool:
```
am-ivanov, line 568: SDFG without module name

dace/transformation/transformation.py
```diff
@@ -446,13 +586,29 @@ def apply(self, sdfg: SDFG):
     @classmethod
     def apply_to(cls,
                  sdfg: SDFG,
-                 *where: Union[nd.Node, SDFGState, SubgraphView],
+                 *where: Union[nd.Node, SDFGState, gr.SubgraphView],
```
am-ivanov, line 589: SDFGState without module name

tests/transformations/maptoforloop_test.py
```diff
@@ -24,7 +24,7 @@ def test_map2for_overlap():
         expected[:, :, k + 1] = expected[:, :, k] + expected[:, :, k - 1]
 
     sdfg = map2for.to_sdfg()
-    sdfg.apply_transformations([MapExpansion, MapToForLoop])
+    assert sdfg.apply_transformations([MapExpansion, MapToForLoop]) == 2
```
am-ivanov, line 27: no module name before MapExpansion, MapToForLoop

am-ivanov (changes requested):

dace/sdfg/graph.py
tbennun: We can't, because `Graph` is a general data structure.

dace/sdfg/graph.py
tbennun: Same.

dace/sdfg/graph.py
tbennun: it exists in the same file (`dace.sdfg.graph.Edge`).

dace/transformation/transformation.py
tbennun: Because it is defined in the same file.

dace/transformation/transformation.py
tbennun: no, because we need to keep backwards compatibility to existing transformations :(

dace/transformation/transformation.py
tbennun: This is not the same function, but the diff makes it look like it is. apply_pattern always returned anything (as said in the documentation).

dace/transformation/transformation.py
tbennun: I won't be able to change everything in this PR, especially not SDFG and SDFGState which exist pretty much everywhere.

dace/transformation/transformation.py
tbennun: Same as above, I won't be able to fix all existing cases here.

dace/transformation/transformation.py
tbennun: Same as above, I won't be able to fix all existing cases here.

dace/transformation/transformation.py
tbennun: Same as above, I won't be able to fix all existing cases here.

PR #419, opened 2020-10-18
Convert all keys (including symbols) to strings for proper conversion…

andreaskuster:
While using `N = dace.symbol('N')` within the RTL codegen, i came accross the following json exception:

```
  File "/home/andreas/anaconda3/lib/python3.7/json/encoder.py", line 376, in _iterencode_dict
    raise TypeError(f'keys must be str, int, float, bool or None, '
TypeError: keys must be str, int, float, bool or None, not symbol
```
Since this happens within the json library, we have to ensure that the keys in the dictionary are already in a supported format. My suggestion is to convert all of them to strings before passing them to the `json.dumps()` method.

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -279,9 +279,11 @@ class SDFG(OrderedDiGraph):
         """
         tmp = super().to_json()
 
+        # Convert all keys to strings
+        constants_prop = {str(k): tmp['attributes']['constants_prop'][k] for k in tmp['attributes']['constants_prop']}
+
         # Ensure properties are serialized correctly
-        tmp['attributes']['constants_prop'] = json.loads(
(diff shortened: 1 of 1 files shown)


tbennun:
This is not a good solution IMO. There will always be a potential to introduce objects into the JSON and the solution in your case is to change the `specialize` method.

It also seems to create a sporadic failure in the CUDA test (see CI result above)

tbennun (approved):

andreaskuster:
> 
> 
> This is not a good solution IMO. There will always be a potential to introduce objects into the JSON and the solution in your case is to change the `specialize` method.
> 
> It also seems to create a sporadic failure in the CUDA test (see CI result above)

The main reason to do it that way was to keep the original type (symbol) within the constants dict, while ensuring correct json serialization. Unfortunately, this breaks some internal behaviour, so i changed it to convert the key right during the creation of the constants.

PR #422, opened 2020-10-20
Multi-kernel support for Intel

TizianoDeMatteis:


dace/codegen/CMakeLists.txt
```diff
@@ -532,6 +532,8 @@ if(DACE_ENABLE_INTELFPGA)
     string(REGEX REPLACE "kernel_(.+).cl" "\\1" DACE_KERNEL_NAME "${DACE_KERNEL_NAME}")
     set(DACE_AOC_KERNEL_FILES ${DACE_AOC_KERNEL_FILES} ${DACE_KERNEL_FILE})
 
+    # The output AOCX file will be named as the last OpenCL file given in input to the compiler
+    get_filename_component(DACE_AOC_OUTPUT_FILE ${DACE_KERNEL_FILE} NAME_WE)
```
definelicht, line 536: What about using the name of the top-level SDFG instead? Seems a bit more intuitive

tests/fpga/multiple_kernels.py
```diff
@@ -0,0 +1,164 @@
...
+    comp(a=a, b=b, c=c)
+
+    diff1 = ((ref_a + 1) - b) / b
+    diff2 = ((ref_a + 2) - c) / c
+    if diff1 <= 1e-5 and diff2 <= 1e-5:
+        print("==== Program end ====")
+    else:
+        print("==== Program Error! ====")
+
+    exit(0 if diff1 <= 1e-5 or diff2 <= 1e-5 else 1)
```
definelicht, line 164: Instead of having the same check twice, throw an exception if there's a program error

definelicht (changes requested):

dace/codegen/CMakeLists.txt
TizianoDeMatteis: If you call the compiler `aoc file1.cl file2.cl file3.cl`, the outputput will be named `file3.aocx`

definelicht: You can probably just use `${DACE_PROGRAM_NAME}`

TizianoDeMatteis: No, that does not work. CL files are not named using program name but using the state names.
For example, in the test, the program name is "multiple_kernels", but the two files (states) are "fpga_state_0.cl" and "fpga_state_1.cl"

definelicht: Yes, but aren't we talking about the name of the _output_ file? There is only one of them, can that not use the program name?

TizianoDeMatteis: That is done later, and, yes, is done using the Program name

```
COMMAND mv ${DACE_AOC_OUTPUT_FILE}.aocx  ${DACE_PROGRAM_NAME}_emulator.aocx
```

definelicht: Why this extra step of moving it instead of using the output name directly?

TizianoDeMatteis: Because you can not use `-o` if you have multiple input files

```
$ aoc fpga_state_0.cl fpga_state_1.cl -o out.aocx
Error: Cannot specify -o with multiple input files
```

definelicht: Ahaaaa I see: please write this exact problem in more detail in the comment in the CMake script, then it will be clear to the next person :-)

definelicht (approved):

tbennun (approved):

PR #426, opened 2020-10-23
Intel fpga symbol mapping

TizianoDeMatteis:
Handles symbol mapping for nested SDFG.

Currently, this was causing problems, as nested SDFGs are inlined into OpenCL Kernel. This was resulting in having the kernel with arguments named as the symbols defined in the parent SDFG, while the body of the nestedSDFG was using the inner symbols (hence compiler error).

This PR creates aliases between outer/inner symbols. 
This is supposed to change if we are going to have SDFG as-a-function also for Intel FPGA.


tests/fpga/symbol_mapping.py
```diff
@@ -0,0 +1,197 @@
+# Copyright 2019-2020 ETH Zurich and the DaCe authors. All rights reserved.
+
+# The scope of the test is to verify that code nested SDFGs with a unique name is generated only once
+# The nested SDFG compute vector addition on FPGA, with vectorization
```
definelicht, line 4: How do you check that it is only generated once? I can't see this in the test

definelicht (changes requested):

definelicht:
Why this behavior was different for Intel FPGA compared to the rest of DaCe in the first place?

tests/fpga/symbol_mapping.py
TizianoDeMatteis: My fault, wrong comment

tests/fpga/symbol_mapping.py
```diff
@@ -0,0 +1,196 @@
+# Copyright 2019-2020 ETH Zurich and the DaCe authors. All rights reserved.
+
+# The scope of the test is to verify that nestedSDFG mapped symbols are handled correctly
```
definelicht, line 3: Please elaborate a bit more on what should happen, and what was wrong before. You can even link the github issue if there is one (e.g., "see #12345").

definelicht (changes requested):

TizianoDeMatteis:
@definelicht I think it is because nested SDFGs are generated as functions. Then the symbol mapping is "handled" directly in the function signature. Xilinx backend defaults to CPU backend.

tests/fpga/symbol_mapping.py
TizianoDeMatteis: Mapping was not handled at all.

definelicht:
> @definelicht I think it is because nested SDFGs are generated as functions. Then the symbol mapping is "handled" directly in the function signature. Xilinx backend defaults to CPU backend.

Is there any way we could unite them by using the (relatively new) `arg.ocltype`?

PR #431, opened 2020-10-26
Fix bug in numpy broadcasting

orausch:
Broadcasting with dim 1 matching dim None previously resulted in the output shape having `None` inside of it. This PR fixes that.


tbennun:
Looks good, waiting for build to finish.

orausch:
Could take a while again :firetravis:

tbennun (approved):

PR #433, opened 2020-10-26
Numpy frontend fixes

orausch:


orausch:
I don't think I will support dicts; this would require some sort of constant folding pass to fold away the slices, and would amount to implementing a large subset of python semantics as an ast transformer.

tbennun (approved):

PR #434, opened 2020-10-27
Centralized memory allocation

tbennun:
- [ ] Centralized allocation management
    - [x] "Determine allocation scope" logic
    - [x] Usage of scope-based allocation in targets
    - [ ] Fix FPGA memory management
    - [x] Global memory access through state struct (emit reference at nested SDFG begin)

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -342,6 +342,9 @@ class CompiledSDFG(object):
             if not dtypes.is_array(arg) and isinstance(atype, dt.Array):
                 if isinstance(arg, list):
                     print('WARNING: Casting list argument "%s" to ndarray' % a)
+                elif arg is None:
+                    # None values are passed as null pointers
+                    pass
                 else:
                     raise TypeError(
                         'Passing an object (type %s) to an array in argument "%s"'
@@ -389,6 +392,9 @@ class CompiledSDFG(object):
             # List to array
             elif isinstance(arg, list) and isinstance(argtype, dt.Array):
                 arglist[index] = np.array(arg, dtype=argtype.dtype.type)
+            # Null pointer
+            elif arg is None and isinstance(argtype, dt.Array):
+                arglist[index] = ctypes.c_void_p(0)
 
         # Retain only the element datatype for upcoming checks and casts
         arg_ctypes = [t.dtype.as_ctypes() for t in argtypes]
--- a/dace/codegen/dispatcher.py
+++ b/dace/codegen/dispatcher.py
@@ -137,7 +137,7 @@ class TargetDispatcher(object):
         # type: Dict[dace.dtypes.InstrumentationType, InstrumentationProvider]
         self.instrumentation = {}
 
-        self._array_dispatchers = {
+        self._array_dispatchers : dict[dtypes.StorageType, target.TargetCodeGenerator] = {
         }  # Type: dtypes.StorageType -> TargetCodeGenerator
         self._map_dispatchers = {
         }  # Type: dtypes.ScheduleType -> TargetCodeGenerator
@@ -154,13 +154,13 @@ class TargetDispatcher(object):
         self._state_dispatchers = []  # [(predicate, dispatcher)]
         self._generic_state_dispatcher = None  # Type: TargetCodeGenerator
 
-        self._declared_vars = DefinedMemlets()
+        self._declared_arrays = DefinedMemlets()
         self._defined_vars = DefinedMemlets()
 
     @property
-    def declared_vars(self) -> DefinedMemlets:
+    def declared_arrays(self) -> DefinedMemlets:
         """ Returns a list of declared variables. """
-        return self._declared_vars
+        return self._declared_arrays
 
     @property
     def defined_vars(self) -> DefinedMemlets:
@@ -445,7 +445,7 @@ class TargetDispatcher(object):
         if declare and not allocate:
             self._array_dispatchers[datadesc.storage].declare_array(
                 sdfg, dfg, state_id, node, datadesc, function_stream,
-                declaration_stream, callsite_stream)
+                declaration_stream)
         elif allocate:
             self._array_dispatchers[datadesc.storage].allocate_array(
                 sdfg, dfg, state_id, node, datadesc, function_stream,
--- a/dace/codegen/dispatcher.py
+++ b/dace/codegen/dispatcher.py
@@ -69,7 +69,7 @@ class DefinedMemlets:
         # Search among globally defined variables (top scope), if not already visited
         # TODO: The following change makes it so we look in all top scopes, not
         # just the very top-level one. However, it we are in a nested SDFG,
-        # then we must limist the search to that SDFG and only. There is one
+        # then we must limit the search to that SDFG only. There is one
         # exception, when the data has Global or Persistent allocation lifetime.
         # Then, we expect it to be only in the very top-level scope.
         # if last_visited_scope != self._scopes[0]:
--- a/dace/codegen/dispatcher.py
+++ b/dace/codegen/dispatcher.py
@@ -10,7 +10,7 @@ from dace import config, data as dt, dtypes, nodes, registry
 from dace.codegen import exceptions as cgx, prettycode
 from dace.codegen.targets import target
 from dace.sdfg import utils as sdutil, SDFG, SDFGState, ScopeSubgraphView
-from typing import Tuple
+from typing import Dict, Tuple
 
 
 @registry.extensible_enum
@@ -51,7 +51,8 @@ class DefinedMemlets:
         except KeyError:
             return False
 
-    def get(self, name: str,
+    def get(self,
+            name: str,
             ancestor: int = 0,
             is_global: bool = False) -> Tuple[DefinedType, str]:
         last_visited_scope = None
@@ -68,7 +69,7 @@ class DefinedMemlets:
 
         # Search among globally defined variables (top scope), if not already visited
         # TODO: The following change makes it so we look in all top scopes, not
-        # just the very top-level one. However, it we are in a nested SDFG,
+        # just the very top-level one. However, ft we are in a nested SDFG,
         # then we must limit the search to that SDFG only. There is one
         # exception, when the data has Global or Persistent allocation lifetime.
         # Then, we expect it to be only in the very top-level scope.
@@ -137,8 +138,9 @@ class TargetDispatcher(object):
         # type: Dict[dace.dtypes.InstrumentationType, InstrumentationProvider]
         self.instrumentation = {}
 
-        self._array_dispatchers : dict[dtypes.StorageType, target.TargetCodeGenerator] = {
-        }  # Type: dtypes.StorageType -> TargetCodeGenerator
+        self._array_dispatchers: Dict[
+            dtypes.StorageType, target.TargetCodeGenerator] = {
+            }  # Type: dtypes.StorageType -> TargetCodeGenerator
         self._map_dispatchers = {
         }  # Type: dtypes.ScheduleType -> TargetCodeGenerator
         self._copy_dispatchers = {}  # Type: (dtypes.StorageType src,
@@ -159,12 +161,20 @@ class TargetDispatcher(object):
 
     @property
     def declared_arrays(self) -> DefinedMemlets:
-        """ Returns a list of declared variables. """
+        """ Returns a list of declared variables.
+        
+            This is used for variables that must have their declaration and
+            allocation separate. It includes all such variables that have been
+            declared by the dispatcher.
+        """
         return self._declared_arrays
 
     @property
     def defined_vars(self) -> DefinedMemlets:
-        """ Returns a list of defined variables. """
+        """ Returns a list of defined variables.
+        
+            This includes all variables defined by the dispatcher.
+        """
         return self._defined_vars
 
     @property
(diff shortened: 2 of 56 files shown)


dace/codegen/dispatcher.py
```diff
@@ -64,9 +67,23 @@ def get(self, name: str, ancestor: int = 0) -> Tuple[DefinedType, str]:
                 break
 
         # Search among globally defined variables (top scope), if not already visited
-        if last_visited_scope != self._scopes[0]:
-            if name in self._scopes[0][1]:
-                return self._scopes[0][1][name]
+        # TODO: The following change makes it so we look in all top scopes, not
+        # just the very top-level one. However, it we are in a nested SDFG,
+        # then we must limist the search to that SDFG and only. There is one
```
TizianoDeMatteis, line 72: typos: "limist", "and only"

dace/codegen/targets/fpga.py
```diff
@@ -690,12 +690,68 @@ def generate_scope(self, sdfg, dfg_scope, state_id, function_stream,
...
+        except KeyError:  # Array not declared yet
+            pass
+
+        allocname = cpp.ptr(dataname, nodedesc, sdfg)
+
+        if nodedesc.storage == dtypes.StorageType.FPGA_Global:
+
+            if self._in_device_code:
+
+                # TODO: Is this needed here?
```
TizianoDeMatteis, line 720: We added this safety check to verify that all global arrays (even in nested SDFGs) have been allocated before we reach this point (we are in device code). I would leave it, at least for the moment being

dace/codegen/targets/fpga.py
```diff
@@ -690,12 +690,68 @@ def generate_scope(self, sdfg, dfg_scope, state_id, function_stream,
...
+                if nodedesc not in self._allocated_global_arrays:
+                    raise RuntimeError("Cannot allocate global array "
+                                       "from device code: {} in {}".format(
+                                           node.label, sdfg.name))
+
+            else:
+                # TODO: Distinguish between read, write, and read+write
+                # Define buffer, using proper type
+                result_decl.write(
+                    "hlslib::ocl::Buffer <{}, hlslib::ocl::Access::readWrite> {};"
```
TizianoDeMatteis, line 730: It is a bit strange that you are actually allocating it, in the function that should only declare the array.
But I don't have a clear solution for this

dace/codegen/targets/framecode.py
```diff
@@ -467,6 +393,325 @@ def dispatch_state(state: SDFGState) -> str:
...
+
+                # TODO: The change below was done so that test_alloc_persistent
+                # passes (tests/codegen/allocation_lifetime_test.py)
+                definition = desc.as_arg(name=f'__{sdfg.sdfg_id}_{name}') + ';'
+                # definition = desc.as_arg(name=f'{name}') + ';'
+                self.statestruct.append(definition)
+
+                # self.to_allocate[top_sdfg].append(
+                #     (sdfg.sdfg_id, sdfg.node_id(state), node))
+                self.to_allocate[top_sdfg].append(
```
TizianoDeMatteis, line 490: Could you please describe (here or where `to_allocate` is defined), what's the meaning of the different tuple fields?

dace/codegen/targets/intel_fpga.py
```diff
@@ -1067,15 +1068,17 @@ def generate_memlet_definition(self, sdfg, dfg, state_id, src_node,
 
         result = ""
 
+        is_global = False
```
TizianoDeMatteis, line 1071: `is_global` means that this is `FPGA_Global`, or that it is allocated in the top-level scope?

tests/symbol_dependent_transients_test.py
```diff
@@ -0,0 +1,163 @@
...
+    sdfg_exec = sdfg.compile()
+    sdfg_exec(A=A, B=B, N=10)
+    B_ref = np.ndarray(10, dtype=np.float64)
+    for i in range(10):
+        tmp = A[2:-2, 2:-2, i:]
+        B_ref[i] = np.sum(tmp)
+    assert (np.allclose(B, B_ref))
+
+
+@pytest.mark.skip
```
TizianoDeMatteis, line 141: Does this pass?

dace/codegen/targets/fpga.py
alexnick83: Isn't this just the declaration code? The allocation (in `allocate_array`) appears to me to be as follows:
```python
                        result_alloc.write(
                            "{} = __state->fpga_context->Get()."
                            "MakeBuffer<{}, hlslib::ocl::Access::readWrite>"
                            "({}{});".format(allocname, nodedesc.dtype.ctype,
                                             memory_bank_arg,
                                             cpp.sym2cpp(arrsize)))
```

dace/codegen/targets/framecode.py
alexnick83: Added in lines ~450-460.

dace/codegen/targets/intel_fpga.py
alexnick83: The latter

dace/codegen/targets/fpga.py
TizianoDeMatteis: That's right, I missed that!

tests/symbol_dependent_transients_test.py
alexnick83: No, but this is probably because of the reasons we discussed. This must always be overapproximated for FPGAs. I am disabling it and I will think of a better test later.

alexnick83 (approved):

dace/codegen/targets/cpu.py
```diff
@@ -238,12 +293,20 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
         except KeyError:  # Array not allocated yet
             pass
 
+        # Check if array is already declared
+        declared = False
+        try:
+            self._dispatcher.declared_arrays.get(name)
+            declared = True  # Array was already declared in this or upper scopes
+        except KeyError:  # Array not declared yet
+            pass
```
tbennun, line 302: don't we have a method called `has` for this exact purpose?

dace/codegen/targets/cuda.py
```diff
@@ -332,17 +332,61 @@ def cmake_options():
...
-                       declaration_stream, allocation_stream):
+    def declare_array(self, sdfg, dfg, state_id, node, nodedesc,
+                      function_stream, declaration_stream):
+
+        # NOTE: We currently only support Arrays (not Views)
+        # that are dependent on non-free SDFG symbols.
+        assert (isinstance(nodedesc, dt.Array)
+                and not isinstance(nodedesc, dt.View) and any(
+                    str(s) not in sdfg.free_symbols.union(sdfg.constants.keys())
+                    for s in nodedesc.free_symbols))
```
tbennun, line 343: would be nice to change this to `if ....: raise TypeError` with an informative message

dace/codegen/targets/cuda.py
```diff
@@ -332,17 +332,61 @@ def cmake_options():
...
             pass  # The variable was not defined, we can continue
 
-        nodedesc = node.desc(sdfg)
+        # Check if array is already declared
+        declared = False
+        try:
+            self._dispatcher.declared_arrays.get(node.data)
+            declared = True  # Array was already declared in this or upper scopes
+        except KeyError:  # Array not declared yet
+            pass
```
tbennun, line 386: use `has`?

dace/codegen/targets/fpga.py
```diff
@@ -690,12 +690,67 @@ def generate_scope(self, sdfg, dfg_scope, state_id, function_stream,
...
-                       declaration_stream, allocation_stream):
+    def declare_array(self, sdfg, dfg, state_id, node, nodedesc,
+                      function_stream, declaration_stream):
+
+        # NOTE: We currently only support Arrays (not Views)
+        # that are dependent on non-free SDFG symbols.
+        assert (isinstance(nodedesc, dt.Array)
+                and not isinstance(nodedesc, dt.View) and any(
+                    str(s) not in sdfg.free_symbols.union(sdfg.constants.keys())
+                    for s in nodedesc.free_symbols))
```
tbennun, line 701: same as the cuda comment

dace/codegen/targets/fpga.py
```diff
@@ -710,7 +765,15 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
                 pass  # The variable was not defined,  we can continue
 
-        allocname = cpp.ptr(dataname, nodedesc)
+        # Check if array is already declared
+        declared = False
+        try:
+            self._dispatcher.declared_arrays.get(dataname)
+            declared = True  # Array was already declared in this or upper scopes
+        except KeyError:  # Array not declared yet
+            pass
```
tbennun, line 774: has

dace/codegen/targets/framecode.py
```diff
@@ -467,6 +393,317 @@ def dispatch_state(state: SDFGState) -> str:
...
+        node = scope if isinstance(scope, nodes.EntryNode) else None
+        cstate = scope if isinstance(scope, SDFGState) else state
+        csdfg = scope if isinstance(scope, SDFG) else sdfg
+
+        if desc.storage in dtypes.FPGA_STORAGES:
+            return sdscope.is_devicelevel_fpga(csdfg, cstate, node)
+        elif desc.storage in dtypes.GPU_STORAGES:
+            return sdscope.is_devicelevel_gpu(csdfg, cstate, node)
+
+        # return True
```
tbennun, line 430: the commented code in this function is confusing. could you please remove?

dace/codegen/targets/framecode.py
```diff
@@ -467,6 +393,317 @@ def dispatch_state(state: SDFGState) -> str:
...
+                else:
+                    raise TypeError
+
+            if curscope is None:
+                curscope = top_sdfg
+
+            # Check if array shape is dependent on non-free SDFG symbols
+            # NOTE: Only arrays supported now
+            # NOTE: We may also need to support views
+            # NOTE: Tuple is (SDFG, State, Node, declare, allocate deallocate)
```
tbennun, line 639:
```suggestion
            # NOTE: Tuple is (SDFG, State, Node, declare, allocate, deallocate)
```

dace/codegen/targets/framecode.py
```diff
@@ -467,6 +393,317 @@ def dispatch_state(state: SDFGState) -> str:
...
+
+            # Possibly confusing control flow below finds the first state
+            # and node of the data descriptor, or continues the
+            # arrays_recursive() loop
+            first_state_instance: int = None
+            first_node_instance: nodes.AccessNode = None
+            last_state_instance: int = None
+            last_node_instance: nodes.AccessNode = None
+            first = True
+            for id, state in enumerate(sdfg.nodes()):
```
tbennun, line 469: this should be a topological sort or DFS of the states. otherwise after transformations state IDs may be reversed or the order does not match the appearance order.

tbennun (commented):
@alexnick83 Mostly looks good, some comments. For one crucial comment I left I also pushed a test

dace/codegen/compiled_sdfg.py
```diff
@@ -230,7 +230,7 @@ def _try_parse_state_struct(self) -> Optional[Type[ctypes.Structure]]:
             field_str = field_str.strip()
 
             match_name = re.match(
-                r'(?:const)?\s*(.*)(?:\s+\*\s*|\s*\*\s+)([a-zA-Z_][a-zA-Z_0-9]*)$',
+                r'(?:const)?\s*(.*)(?:\s+\*\s*|\s*\*\s+\_\_restrict\_\_\s+)([a-zA-Z_][a-zA-Z_0-9]*)$',
```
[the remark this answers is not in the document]
alexnick83: The declaration of persistent memory is now created by `Data.as_arg` (see framecode.py, line 484), which adds the `__restrict__` keyword (and probably makes sense since, similarly to program arguments, we don't want it to alias). The change above makes sure that we can properly parse structs (in this case program states) that contain `__restrict__` pointers (see tests/parse_state_struct_test.py).

dace/codegen/dispatcher.py
```diff
@@ -120,7 +137,7 @@ def __init__(self, framecode):
         # type: Dict[dace.dtypes.InstrumentationType, InstrumentationProvider]
         self.instrumentation = {}
 
-        self._array_dispatchers = {
+        self._array_dispatchers : dict[dtypes.StorageType, target.TargetCodeGenerator] = {
```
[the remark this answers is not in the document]
alexnick83: I know that this is inconsistent with the project but all generic concrete collections are now deprecated in favor of the builtins (see, for example, https://docs.python.org/3/library/typing.html#typing.Dict).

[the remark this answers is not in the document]
tbennun: since Python 3.9 yes, but you can't use it with older versions of python. as long as we support 3.6+, we have to use Dict. I can't wait for this either.

dace/codegen/targets/framecode.py
alexnick83: I can actually see more potential problems here regarding the topological sort, for example, multiple paths. This would necessitate the use of realloc (with which we cannot use `DACE_ALIGN`?) but there all sorts of other issues if we use it. I will update the codegen to work with SDFGs similar to the example provided in the test but I think we should leave a more elaborate realloc-solution for a different PR. We already support a strict superset of SDFGs compared to master.

tbennun: ok, please leave a comment in the code then, so that we know something better needs to be done. Also, even if you reverse the states (so that they are in the same order as their ids) it seems to fail for an unrelated reason. Does it fail for you as well?

alexnick83: This is probably related to https://github.com/spcl/dace/pull/434/commits/360d9a6b08ef40c9c195c38415607bb19f070831

PR #436, opened 2020-10-27
Fix reductions when reducing along all axes and allow local SDFGs to be called

orausch:
Reductions rely on `Subset.pop`, but this doesn't work correctly when reducing along all axes. It instead reduces along the last axes.

This PR bypasses that issue.

### DIFF
--- a/dace/frontend/python/replacements.py
+++ b/dace/frontend/python/replacements.py
@@ -127,7 +127,15 @@ def _reduce(sdfg: SDFG,
                                            {})
         input_memlet = Memlet.simple(inarr, input_subset)
         output_shape = None
-        if axis is None:
+
+        # check if we are reducing along all axes
+        if axis is not None and len(axis) == len(input_subset.size()):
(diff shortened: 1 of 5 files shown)


dace/frontend/python/replacements.py
```diff
@@ -127,7 +127,15 @@ def _reduce(sdfg: SDFG,
                                            {})
         input_memlet = Memlet.simple(inarr, input_subset)
         output_shape = None
-        if axis is None:
+
+        # check if we are reducing along all axes
+        if axis is not None and len(axis) == len(input_subset.size()):
```
tbennun, line 132: Do you mean `is None or...`?

tbennun (commented):
One question

dace/frontend/python/replacements.py
orausch: No I mean `is not None`. The `is None` case is handled on L138

tbennun (approved):

Issue #437, opened 2020-10-28
Python Frontend: SDFG call bug with multiple Nested SDFGs


orausch:
When you call an SDFG from a dace program, the nested SDFG nodes are not connected correctly. In the following example, the connector `output` of the second nested SDFG is incorrectly connected to `out_tmp`, even though `output=output` is specified in the python code. The issue disappears when the output arrays of both nested SDFGs have different names.

```python
import copy

import numpy as np

import dace

def call_multiple_sdfgs():
    inparr = dace.float32[2, 2]
    axis = 1
    out_tmp_shape = inparr.shape
    out_tmp_dtype = inparr.dtype

    tmp_max_shape = list(copy.deepcopy(inparr.shape))
    tmp_max_shape.pop(axis)

    ##################
    # exp - max
    exp_minus_max = dace.SDFG("exp_minus_max")
    exp_minus_max.add_array("tmp_max", tmp_max_shape, inparr.dtype)
    exp_minus_max.add_array("original_input", inparr.shape, inparr.dtype)
    exp_minus_max.add_array("output", out_tmp_shape, out_tmp_dtype)
    exp_minus_max.add_state().add_mapped_tasklet(
        "_softmax_exp_",
        map_ranges={
            "__i" + str(i): "0:" + str(shape)
            for i, shape in enumerate(inparr.shape)
        },
        inputs={
            '__max':
                dace.Memlet.simple(
                    "tmp_max", ','.join("__i" + str(i) for i in range(len(inparr.shape))
                                        if i != axis)),
            '__x':
                dace.Memlet.simple(
                    "original_input", ','.join("__i" + str(i) for i in range(len(inparr.shape))))
        },
        code='__out = exp(__x - __max)',
        outputs={
            '__out':
                dace.Memlet.simple(
                    "output",
                    ','.join("__i" + str(i) for i in range(len(inparr.shape))))
        },
        external_edges=True)

    ##################
    # out_tmp / sum
    out_tmp_div_sum = dace.SDFG("out_tmp_div_sum")
    out_tmp_div_sum.add_array("out_tmp", inparr.shape, inparr.dtype)
    out_tmp_div_sum.add_array("tmp_sum", tmp_max_shape, inparr.dtype)
    out_tmp_div_sum.add_array("output", out_tmp_shape, out_tmp_dtype)

    out_tmp_div_sum.add_state().add_mapped_tasklet(
        "_softmax_div_",
        map_ranges={
[...]

Issue #438, opened 2020-10-28
Erroneous data-to-data copy with other_subset


alexnick83:
The bug can be reproduced by running `test_out_success` in `tests/transformations/redundant_copy_test.py`.
The test produces an SDFG with an edge from access node A to access node C. The edge has source subset `0, 0:3, 4` and destination subset `1, 2, 0:3, 4`.
However, if you uncomment the assertion on line 91, it will fail.
The compiled program appears to copy the correct subset of A, but on `C[0:3, 2, 0, 4]` instead of `C[1, 2, 0:3, 4]`.

alexnick83:
Linked to the wrong PR

tbennun:
Seems like it was erroneously reopened.

PR #439, opened 2020-10-28
Fix for #438

alexnick83:
This is a fix for issue #438.


orausch (approved):
Thanks for the quick fix!

orausch:
Also closes #437

PR #443, opened 2020-10-30
Switch to pytest, GitHub Actions

tbennun:


tests/python_frontend/binop_replacements_test.py
```diff
@@ -397,6 +397,7 @@ def mixed(A: dace.int64[M, N], B:dace.int64):
     return 5j + dace.complex64(M) + A[0, 0] + 32 + A[0, 1] + B + 2 + M + N
 
 
+@pytest.mark.skip
```
alexnick83, line 400: This test should be passing. What seems to be the issue?

alexnick83 (approved):

tests/python_frontend/binop_replacements_test.py
tbennun: I actually don't remember. We can try it in a separate PR

PR #455, opened 2020-11-05
Make scalar promotion match numpy behavior when scalars are known

### LINKED ISSUE #454: Scalar type promotion issue
Currently, the following fails compilation:
```python
import dace
import numpy as np

@dace
def add_one(A: dace.int32[10, 10], B: dace.int32[10, 10]):
    B[:] = A + 1

A = np.zeros((10, 10)).astype(np.int32)
B = np.zeros((10, 10)).astype(np.int32)
add_one.to_sdfg(strict=false)(A=A, B=B)
print(B)
```
In numpy this works:
```python
>>> a = np.zeros((2,2)).astype(np.int32)
>>> a + 1
array([[1, 1],
       [1, 1]], dtype=int32)
```

This is caused by the fact that the dtype of `A + 1` is inferred to be `int64`. In numpy it would be `int32` due to their scalar-array type promotion rules (which are quite ugly: https://github.com/numpy/numpy/issues/13591).

To sidestep this I think it would be best to implement the type constructors in the python frontend (i.e. `dace.int32`).

orausch:
Closes #454
Current exceptions are:
bitwise ops, pow, div, floordiv and boolean ops.

### DIFF
--- a/dace/frontend/python/replacements.py
+++ b/dace/frontend/python/replacements.py
@@ -830,7 +830,7 @@ def _sym_type(expr: Union[symbolic.symbol, sp.Expr]) -> dtypes.typeclass:
     return dtypes.DTYPE_TO_TYPECLASS[np.result_type(*typeclasses).type]
 
 
-def _convert_type(dtype1, dtype2, operator) -> Tuple[dace.dtypes.typeclass]:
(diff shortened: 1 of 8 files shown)


orausch:
Suceeded by #474

PR #456, opened 2020-11-06
Fixed wrong arguments to Python 3.6 version of subprocess.run()

carljohnsen:
The `capture_output` argument to `subprocess.run()` wasn't added until Python 3.7 [https://docs.python.org/3.7/library/subprocess.html#subprocess.run](url)

This patch reverts the call to be functional in Python 3.6

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -5,7 +5,6 @@ import os
 import re
 import shutil
 import subprocess
-from subprocess import PIPE
 from typing import Any, Dict, List
 import warnings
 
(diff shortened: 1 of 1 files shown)


dace/codegen/compiled_sdfg.py
```diff
@@ -5,6 +5,7 @@
 import re
 import shutil
 import subprocess
+from subprocess import PIPE
```
definelicht, line 8: Please omit the extra import and just use `subprocess.PIPE` in the code

definelicht (changes requested):

dace/codegen/compiled_sdfg.py
tbennun: I agree as well

tbennun (changes requested):
One change

tbennun (approved):

definelicht (approved):

PR #457, opened 2020-11-06
Optimize Transformations, Symbol Replacement, and Fix Bugs in Replacement-Based Transformations

tbennun:


dace/transformation/pattern_matching.py
```diff
@@ -63,8 +64,8 @@ def type_match(node_a, node_b):
     """
     if isinstance(node_b['node'], xf.PatternNode):
         return isinstance(node_a['node'], node_b['node'].node)
-    elif isinstance(node_a['node'], xf.PatternNode):
-        return isinstance(node_b['node'], node_a['node'].node)
```
definelicht, line 67: Is this intentional?

definelicht (approved):

definelicht:
Don't merge before you check my comment :-)

dace/transformation/pattern_matching.py
tbennun: Yes. The pattern-node is always the second one (node_b, I verified in the networkx source code). Since the code runs so often on the huge graphs, I decided to make sure it runs as fast as possible. Good catch! I renamed the parameters accordingly so it's less confusing.

PR #459, opened 2020-11-08
Fixes nested loop tracking and branch annotation in state propagation

phschaad:
- Corrects wrongful removal of iteration variables from the iteration variable stack by moving the iteration variable stack to the traversal's own state instead of keeping it global.
- Make branch detection correctly detect full merges in cases where there are two branches with one of the branch's only state itself being the branch construct's full merge state.


tbennun (approved):
LGTM once tests pass

PR #460, opened 2020-11-08
Python frontend dtype converters

alexnick83:
The purpose of this PR is to add the capability to convert out-of-place the datatype of `dace.Data,` or even constants and symbols. For example:
```python
@dace.program
def conversion(A: dace.int32[10]):
    return dace.int64(A)
```

### DIFF
--- a/dace/codegen/targets/framecode.py
+++ b/dace/codegen/targets/framecode.py
@@ -709,6 +709,13 @@ DACE_EXPORTED void __dace_exit_%s(%s)
 
         # Generate code
         ###########################
+        
+        # Keep track of allocated variables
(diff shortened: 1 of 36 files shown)


tbennun (approved):
Looks good. LGTM if passes tests.

alexnick83:
> Looks good. LGTM if passes tests.

Apparently, scalar promotion triggers with scalars and then fails because it tries to sympify "dace.int64(A)". The error is that "Symbol"  object doesn't have attribute "int64". How are we fixing this?

tbennun:
Maybe since it's an unsupported function/attribute we can decide not to promote this scalar

PR #461, opened 2020-11-09
Allows loop detection to correctly handle for loops with > 1 init edge

phschaad:
- Enable detection of for loop constructs with more than 1 init edge going in to the guard.
- Let `find_for_loop` return all states immidiately before the loop, as well as the last loop state, to generalize this functionality and move it out of the transformations using it (loop peeling, unroll).


tbennun (approved):
Looks good. Will merge if passes all tests.

PR #466, opened 2020-11-10
Augmented assignments to WCR

tbennun:


definelicht (approved):

definelicht:
@tbennun Can you review? I cannot request you, because it's your own PR :-)

dace/sdfg/validation.py
```diff
@@ -481,14 +481,15 @@ def validate_state(state: 'dace.sdfg.SDFGState',
...
-                )
+                pass
+                # raise InvalidSDFGEdgeError(
+                #     "Memlet creates an "
+                #     "invalid path (sink node %s should "
+                #     "be a data node)" % str(dst_node),
+                #     sdfg,
+                #     state_id,
+                #     eid,
+                # )
```
tbennun, line 492: well, now you're allowing a bad pattern to validate:
```
<tasklet>
    |
    v
\mapexit/
    |
    v
<tasklet>
```

tbennun: The difference here should be that we can end in an `nd.ExitNode` if and only if the `edge.data.is_empty()`

dace/transformation/dataflow/wcr_conversion.py
```diff
@@ -0,0 +1,108 @@
...
+from dace import registry, nodes, dtypes
+from dace.transformation import transformation
+from dace.sdfg import utils as sdutil
+from dace import SDFG, SDFGState
+
+
+@registry.autoregister_params(singlestate=True)
+class AugAssignToWCR(transformation.Transformation):
+    """
+    Converts a augmented assignment ("a += b", "a = a + b") into a tasklet
```
tbennun, line 13:
```suggestion
    Converts an augmented assignment ("a += b", "a = a + b") into a tasklet
```

tbennun (commented):
@definelicht validation needs to change a bit, otherwise some minor fixes

dace/sdfg/validation.py
definelicht: Makes sense! tbh I just didn't know what the right condition was, so I put this in the PR to trigger your input :kappa:

definelicht:
@tbennun Requesting review again, please pay special attention to the changes to `read_and_write_sets` in both `sdfg.py` and `state.py`

definelicht (approved):

PR #473, opened 2020-11-12
Fix GPU Expansions for Library Nodes

orausch:
Summary:
- [x] keep changes in infer types
- [x] GPUTransformSDFG should change top-level NestedSDFG and LibraryNode nodes to GPU_Persistent instead of GPU_Global

replaces #401


tbennun (approved):
This is nice! Elegant and may fix more than just library nodes (also almost fixes "recursive" GPUTransformSDFG)

PR #474, opened 2020-11-12
Improved result type inference

alexnick83:
The purpose of this PR is to overhaul casting and result type inference. More specifically:
- Constants are passed to `numpy.result_type` by value. Similarly, scalars and symbols are passed to `numpy.result_type` by a 'representative value'. This essentially assigns priority to `dace.data.Array` objects, effectively emulating the behavior of NumPy when performing operations among `ndarrays` and constants.
- Float/true division between integers now yields float64 (consistent with Python/NumPy). The interpretation of true division as integer division (C/C++ semantics) may still return through some sort of flag.
- Datatype inference of symbolic expressions is improved. The symbolic expression is evaluated by replacing symbols with a 'representative value'.
- USub on unsigned integers now returns signed integer with the same number of bytes. 
- The casting/result type method (`_result_type`) now supports operations with any number of arguments.


dace/frontend/python/replacements.py
```diff
@@ -789,23 +791,79 @@ def _is_op_bitwise(op: str):
 
 
 def _is_op_boolean(op: str):
-    if op in {'And', 'Or', 'Not', 'Eq', 'NotEq', 'Lt', 'LtE', 'Gt', 'GtE'}:
+    if op in {'And', 'Or', 'Not', 'Eq', 'NotEq', 'Lt', 'LtE', 'Gt', 'GtE'
+              'Is', 'NotIs'}:
```
tbennun, line 795: missing comma, will lead to 'GtEIs'

dace/frontend/python/replacements.py
```diff
@@ -789,23 +791,79 @@ def _is_op_bitwise(op: str):
...
+def _np_result_type(nptypes):
+    # Fix for np.result_type returning platform-dependent types,
+    # e.g. np.longlong
+    restype = np.result_type(*nptypes)
+    if restype.type not in dtypes.DTYPE_TO_TYPECLASS.keys():
+        for k in dtypes.DTYPE_TO_TYPECLASS.keys():
+            if k == restype:
+                restype = k
+                break
+    if restype in {int, float, complex}:
```
tbennun, line 822:
```suggestion
    if restype in (int, float, complex):
```

set is an overkill

dace/frontend/python/replacements.py
```diff
@@ -814,90 +872,122 @@ def _convert_type(dtype1, dtype2, operator) -> Tuple[dace.dtypes.typeclass]:
...
+            if operator == 'Div' and max(type1, type2) < 2:
+                # TODO: Leaving this here in case we implement a C/C++ flag
+                # max_bytes = max(dtype1.bytes, dtype2.bytes)
+                # if type1 == type2 and type1 == 0:  # Unsigned integers
+                #     result_type = eval('dace.uint{}'.format(8 * max_bytes))
+                # else:
+                #     result_type = eval('dace.int{}'.format(8 * max_bytes))
+                result_type = dace.float64
+                left_cast = dace.float64
+                right_cast = dace.float64
```
tbennun, line 918: we do have a flag like that - `Config.get_bool('compiler', 'default_data_types')`

dace/frontend/python/replacements.py
```diff
@@ -814,90 +872,122 @@ def _convert_type(dtype1, dtype2, operator) -> Tuple[dace.dtypes.typeclass]:
...
+                    result_type = eval('dace.uint{}'.format(8 * max_bytes))
+                else:
+                    result_type = eval('dace.int{}'.format(8 * max_bytes))
+                right_cast = dace.float64
+            # Power with base integer and exponent signed integer
+            elif (operator == 'Pow' and max(type1, type2) < 2 and
+                    dtype2 in signed_types):
+                result_type = dace.float64
+                left_cast = dace.float64
+                right_cast = dace.float64
```
tbennun, line 938: Anything (float,complex) with integer exponent can keep the original type, right? I don't want float32 ** int32 to be float64

dace/frontend/python/replacements.py
```diff
@@ -814,90 +872,122 @@ def _convert_type(dtype1, dtype2, operator) -> Tuple[dace.dtypes.typeclass]:
...
+                #     result_type = eval('dace.int{}'.format(8 * max_bytes))
+                result_type = dace.float64
+                left_cast = dace.float64
+                right_cast = dace.float64
+            # Floor division with at least one complex argument
+            elif operator == 'FloorDiv' and max(type1, type2) == 3:
+                raise TypeError("can't take floor of complex number")
+            # Floor division with at least one float argument
+            elif operator == 'FloorDiv' and max(type1, type2) == 2:
+                result_type = dace.float64
```
tbennun, line 924: Is this also true for C types? see above comment for flag

dace/frontend/python/replacements.py
```diff
@@ -814,90 +872,122 @@ def _convert_type(dtype1, dtype2, operator) -> Tuple[dace.dtypes.typeclass]:
...
+            # Floor division between integers
+            elif operator == 'FloorDiv' and max(type1, type2) < 2:
+                max_bytes = max(dtype1.bytes, dtype2.bytes)
+                if type1 == type2 and type1 == 0:  # Unsigned integers
+                    result_type = eval('dace.uint{}'.format(8 * max_bytes))
+                else:
+                    result_type = eval('dace.int{}'.format(8 * max_bytes))
+                right_cast = dace.float64
```
tbennun, line 932: same

dace/frontend/python/replacements.py
```diff
@@ -2039,6 +2149,11 @@ def _set_tasklet_params(ufunc_impl: Dict[str, Any],
     # Remove input connectors related to constants
     # and fix constants/symbols in the tasklet code
     for i, arg in reversed(list(enumerate(inputs))):
+        inp_conn = inp_connectors[i]
+        if casting and casting[i]:
+            repl = "{c}({o})".format(c=str(casting[i]).replace('::', '.'),
```
tbennun, line 2154: ugh... maybe we can add a `FIXME` comment here for later? (so that we have better functions for this)

tests/numpy/unop_test.py
```diff
@@ -14,7 +14,7 @@ def usubtest(A: dace.int64[5, 5], B: dace.int64[5, 5]):
 
 
 @dace.program
-def nottest(A: dace.bool[5, 5], B: dace.bool[5, 5]):
+def nottest(A: dace.bool_[5, 5], B: dace.bool_[5, 5]):
```
tbennun, line 17: I still don't understand the difference between dace.bool and dace.bool_. Could you please explain?

tbennun (changes requested):

dace/frontend/python/replacements.py
alexnick83: No, this flag is about how to interpret vague types like int, float, complex. This hypothetical flag would change operator behavior. For example '/' between integers is integer division in C/C++, but it is float division in Python.

dace/frontend/python/replacements.py
alexnick83: Yes, this should trigger only when both arguments are integers. It also looks like NumPy sets the result to float64 regardless of the arguments being int16 or int32. Make sense, because when converting to float you care about the resolution.

dace/frontend/python/replacements.py
alexnick83: My interpretation of C types strictly has to do with the number of bytes in the types, nothing else.

tests/numpy/unop_test.py
alexnick83: `dace.bool` is supposed to be `bool` or `numpy.bool`. However, this never works for arrays. Any array of type `np.bool` automatically becomes `np.bool_`. This is standard NumPy behavior because they cannot create arrays with types (Python int) that do not have a fixed number of bytes.

dace/frontend/python/replacements.py
alexnick83: I believe this is fine. However, I would find it interesting if I changed the connector type, and instead of validation throwing an error, the backend implemented the casting. This would open the way to have casting work with vectorization.

PR #480, opened 2020-11-15
Fix incompatibility issue from verilator 4.1xx upwards.

andreaskuster:
As Carl has pointed out, our DaCe/CMake implementation is broken for verilator >4.1xx, see below:
```
RuntimeError: Could not load library librtl_tasklet_scalar.so. Reason:
/home/carljohnsen/git/dace_rtl/.dacecache/rtl_tasklet_scalar/build/librtl_tasklet_scalar.so: undefined reference to `Vtop_0_0_0::~Vtop_0_0_0()'
/home/carljohnsen/git/dace_rtl/.dacecache/rtl_tasklet_scalar/build/librtl_tasklet_scalar.so: undefined reference to `Vtop_0_0_0::_eval_settle(Vtop_0_0_0__Syms*)'
/home/carljohnsen/git/dace_rtl/.dacecache/rtl_tasklet_scalar/build/librtl_tasklet_scalar.so: undefined reference to `Vtop_0_0_0::_eval_initial(Vtop_0_0_0__Syms*)'
/home/carljohnsen/git/dace_rtl/.dacecache/rtl_tasklet_scalar
```

I investigated this and found out that they now generate an additional cpp file that we have to include, namely: `Vtop_0_0_0__Slow.cpp`


andreaskuster:
I have set up a travis pipeline to check all verilator versions for compatibility issues (see https://github.com/andreaskuster/dace-verilator-compatibility-check)


The conclusion is that the problem already exists from `v4.036` up to the most recent one (see 
https://travis-ci.com/github/andreaskuster/dace-verilator-compatibility-check/builds/201576912)

As you can see, it is fixed for all version >4.028 by this PR: https://travis-ci.com/github/andreaskuster/dace-verilator-compatibility-check/builds/201577757

definelicht (approved):
I will trust you that this is what we need :-) Ideally we could have a test for both, but we can leave that as a nice-to-have for the future.

PR #481, opened 2020-11-16
Generic type hints for graphs

tbennun:


dace/sdfg/graph.py
```diff
@@ -6,7 +6,7 @@
 import networkx as nx
 from dace.dtypes import deduplicate
 import dace.serialize
-from typing import Any, List
+from typing import Any, Callable, Generic, Iterable, List, Sequence, TypeVar, Union
```
alexnick83, line 9: Question: are we allowing importation of the typing hints as above in the new code-guidelines? I am also importing them like this, but I would like confirmation on whether it is OK, or I will be asked in the future to switch to `typing.List`, etc.

tbennun: Yes. It is part of the Google style guidelines too. `typing` is allowed as-is (and some parts of `collections` I think). We also allow our own primitive types to be imported as-is for this purpose: `SDFG`, `SDFGState`, `Memlet`, `InterstateEdge`.

alexnick83 (approved):

PR #485, opened 2020-11-17
Add Verilator DaCe compatibility check github action

andreaskuster:
The github action tests all supported verilator version (currently >=v4.028) for compatibility with the DaCe RTL codegen software simulation. You can check it out here: https://github.com/andreaskuster/dace/actions/runs/368183076

The test can be triggered in the action section manually, see:
grafik

I would have liked to add a script to determine all relevant verilator versions, but unfortunately, i did not find a way to add the script below to the test matrix, which would automatically fetch all relevant versions.

`curl -s 'https://api.github.com/repos/verilator/verilator/tags' | jq -r 'map(select(.name >= "v4.028")) | [.[].name]'`


.github/workflows/verilator_compatibility.yml
```diff
@@ -0,0 +1,36 @@
+name: DaCe Verilator Compatibility Check
+on:
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: 'Reason for the trigger'
+        required: true
+        default: 'Check compatibility'
+  schedule:
+    - cron: '0 0 1 * *' # monthly
```
tbennun, line 10: Please remove

.github/workflows/verilator_compatibility.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+          architecture: 'x64'
+      - uses: BSFishy/pip-action@v1
+        with:
+          requirements: requirements.txt
+      - name: install dace
+        run: python3 -m pip install .
+      - run: python3 tests/rtl/simulation_test.py
```
tbennun, line 36: Please use `pytest -m "verilator"` for this

tbennun (changes requested):

.github/workflows/verilator_compatibility.yml
andreaskuster: Sure, i can remove this. I thought it would be a good idea to test the compatibility monthly. As soon as nobody is working on the Verilator integration anymore, there will be no more manual triggers and we might loose compatibility at some point without detecting this.

andreaskuster:
Thanks a lot for the feedback. I fixed all your suggestions.

tbennun:
@andreaskuster we can run it monthly also

andreaskuster:
> 
> 
> @andreaskuster we can run it monthly also

What do you prefer? I can also adjust the interval to more or less frequent triggers?

andreaskuster:
A added the monthly test again. It is ready for final review.

tbennun (approved):

PR #489, opened 2020-11-18
Fix visit_Name and type inference

TizianoDeMatteis:
Minor fixes in:
- intel FPGA backend: now it uses `memlet.dynamic` instead of `memlet.num_accesses` in understanding if a memory access must be referenced 
- type inference: if a malformed python expression is passed to type inference, it returns no inferred types rather than raising an exception


dace/codegen/tools/type_inference.py
```diff
@@ -27,7 +27,12 @@ def infer_types(code, symbols=None):
     symbols = symbols or {}
     inferred_symbols = {}
     if isinstance(code, str):
-        _dispatch(ast.parse(code), symbols, inferred_symbols)
+        try:
+            parsed_ast = ast.parse(str(code))
+        except SyntaxError:
+            # the input code is not valid Python
+            return inferred_symbols
```
definelicht, line 34: Why should this error be silent?

definelicht (changes requested):

definelicht (approved):

PR #492, opened 2020-11-21
Nested SDFG as function on Intel FPGA code (and unique naming)

TizianoDeMatteis:
With this PR we want to generate Nested SDFG as a function also for Intel FPGA.
This required to:
- generate code for nested SDFG similarly to how it is done for CPU. The entry point is the `generate_Nested_SDFG` function in the cpu codegen that will then call backend-specialized functions
- since we need to generate OpenCL code there are different things to consider: argument can not be passed as reference, streams are global object that can not be passed as arguments...

With this PR we also take care of:
- generate proper macro for stream accesses: now they refer to the "global stream" and not to local connectors
- generate unique names for FPGA kernels and functions, to avoid name clashing


dace/codegen/targets/cpu.py
```diff
@@ -820,14 +820,14 @@ def process_out_memlets(self,
...
                             state_id, node)
                     else:
+                        if isinstance(node, nodes.NestedSDFG):
+                            # This case happens with nested SDFG outputs,
+                            # which we skip since the memlets are references
+                            continue
                         try:
                             defined_type, _ = self._dispatcher.defined_vars.get(
                                 memlet.data)
                         except KeyError:  # The variable is not defined
```
definelicht, line 830: Looks like you can just remove this `try: except`

dace/codegen/targets/cpu.py
```diff
@@ -1425,11 +1425,11 @@ def _generate_NestedSDFG(
         # Arguments are input connectors, output connectors, and symbols
         codegen = self.calling_codegen
         memlet_references = codegen.generate_nsdfg_arguments(
-            sdfg, state_dfg, node)
+            sdfg, dfg,  state_dfg, node)
```
definelicht, line 1428:
```suggestion
            sdfg, dfg, state_dfg, node)
```

dace/codegen/targets/cpu.py
```diff
@@ -1452,7 +1452,7 @@ def _generate_NestedSDFG(
             nested_stream.write(local_code)
 
             # Process outgoing memlets with the internal SDFG
-            self.process_out_memlets(sdfg,
+            codegen.process_out_memlets(sdfg,
                                      state_id,
```
definelicht, line 1456: yapf!

dace/codegen/targets/cuda.py
```diff
@@ -2075,19 +2078,19 @@ def generate_node(self, sdfg, dfg, state_id, node, function_stream,
...
     def generate_nsdfg_call(self, sdfg, state, node, memlet_references,
                             sdfg_label):
         return self._cpu_codegen.generate_nsdfg_call(sdfg, state, node,
                                                      memlet_references,
                                                      sdfg_label)
 
-    def generate_nsdfg_arguments(self, sdfg, state, node):
-        result = self._cpu_codegen.generate_nsdfg_arguments(sdfg, state, node)
+    def generate_nsdfg_arguments(self, sdfg, dfg, state, node):
+        result = self._cpu_codegen.generate_nsdfg_arguments(sdfg,dfg,  state, node)
```
definelicht, line 2093:
```suggestion
        result = self._cpu_codegen.generate_nsdfg_arguments(sdfg, dfg, state, node)
```

dace/codegen/targets/fpga.py
```diff
@@ -1021,9 +1022,11 @@ def _generate_MapEntry(self, sdfg, dfg, state_id, node, function_stream,
             # Generate nested loops
             if not isinstance(node, dace.sdfg.nodes.PipelineEntry):
 
-                if is_innermost and not fully_degenerate:
+                if is_innermost and not fully_degenerate and not is_degenerate[0]:
```
definelicht, line 1025: I don't understand this condition: why not put it if only the outer loop is degenerate?

Actually it would make more sense if the condition was "if more than one loop is *not* degenerate".

tests/fpga/fpga_stencil.py
```diff
@@ -3,7 +3,7 @@
 import dace
 
 
-def make_sdfg(name="fpga_stencil_test", dtype=dace.float32, veclen=8):
+def make_sdfg(name="fpga_stcl_test", dtype=dace.float32, veclen=8):
```
definelicht, line 6: What happened here? :tizianoeat:

definelicht (changes requested):
Nice work!! Some minor things

tests/fpga/fpga_stencil.py
TizianoDeMatteis: That would generate a long name, longer than the fantastic kernel name limit for Intel

definelicht: Hahaha I forgot about that, okay then

dace/codegen/targets/fpga.py
TizianoDeMatteis: Because that was generating something like the following:
```
#pragma loop_coalesce
int index=0; //degenerate loop
for(....)
```
and intel compiler complains because there is a loop_coalesce but there is no loop. If some loop in the middle of the loop nest is generated then it would be ok.
```
# pragma loop_coalesce
for(....)
  for(....)
    int idx =0 ; //degenerate loop
```

dace/codegen/targets/cpu.py
TizianoDeMatteis: But that is already yapfed :/
Is there some yapf style guide?

dace/codegen/targets/fpga.py
definelicht: Ok, but then we need to move the coalesce pragma to the first non-degenerate loop instead of removing it completely, no?

dace/codegen/targets/cpu.py
definelicht: `state_id` and the following arguments are not indented properly, so it seems not. Are you running yapf 0.30 or newer? It should use the `.style.yapf` in the DaCe top-level directory.

dace/codegen/targets/fpga.py
```diff
@@ -1021,10 +1022,14 @@ def _generate_MapEntry(self, sdfg, dfg, state_id, node, function_stream,
             # Generate nested loops
             if not isinstance(node, dace.sdfg.nodes.PipelineEntry):
 
-                if is_innermost and not fully_degenerate:
-                    self.generate_flatten_loop_pre(result, sdfg, state_id, node)
+
 
                 for i, r in enumerate(node.map.range):
+
+                    if is_innermost and not fully_degenerate and not is_degenerate[i]:
```
definelicht, line 1029: Now it will put one at every loop, right? Is that we want?

definelicht (changes requested):

dace/codegen/targets/fpga.py
TizianoDeMatteis: yes, one per loop. I don't see how to generate it differently

definelicht (approved):

PR #497, opened 2020-11-27
Validate SDFG fields `parent`, `parent_nsdfg_node` and Nested SDFGs

fthaler:


dace/sdfg/validation.py
```diff
@@ -259,6 +262,8 @@ def validate_state(state: 'dace.sdfg.SDFGState',
                 state_id,
                 nid,
             )
+        if isinstance(node, nd.NestedSDFG):
+            validate_sdfg(node.sdfg)
```
tbennun, line 266: This already happens in `node.validate`

dace/sdfg/validation.py
```diff
@@ -28,6 +28,9 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG'):
         on failure.
     """
     try:
+        if sdfg.parent is not None and sdfg.parent_nsdfg_node not in sdfg.parent.nodes():
+            raise InvalidSDFGError("Invalid parent or parent_nsdfg_node", sdfg, None)
+
```
tbennun, line 33: Since we have the parent and the node, we could just pass them along to `validate_sdfg` (kwargs with None defaults) to avoid looping over the parent's nodes.

tbennun (changes requested):

PR #503, opened 2020-12-01
Various FPGA fixes: unroll maps, volatile accesses, naming, ivdep

TizianoDeMatteis:
This PR brings multiple fixes to fpga backend (mainly Intel)
- add unroll pragma on each loop generated from a multi dimensional map
- intel FPGA, does not create pointer as `volatile`
- intel FPGA: cmake, change the name of folder containing the reports according to dace program name
- Xilinx: generates dependency pragma considering for every nested loop, taking in consideration if it is degenerate or not
- Naming: changed the kernel naming, to have more meaningful file names. For intel, changed the way in which the names are generated (the compiler looks at the first 40 characters), and cut down the name to 42
- Intel FPGA: add generic pragma to ignore loop carried dependencies when possible, since the compiler is not able to track them correctly
- CPP unparse: deal with OpenCL vector data types


dace/codegen/targets/intel_fpga.py
```diff
@@ -519,17 +532,14 @@ def generate_module(self, sdfg, state, name, subgraph, parameters,
...
         # can also append text to the module. Longest seen so far is
         # "_cra_slave_inst", which is 15 characters, so we restrict to
         # 61 - 15 = 46, and round down to 42 to be conservative.
-        if len(module_function_name) > 42:
-            raise NameTooLongError(
-                "Due to a bug in the Intel FPGA OpenCL compiler, "
-                "kernel names cannot be longer than 42 characters:\n\t{}".
-                format(module_function_name))
+        # Therefore we cut down names longer than that
+        module_function_name = module_function_name[0:42]
```
definelicht, line 541: Then we need a check for collisions!

definelicht (changes requested):

dace/codegen/targets/intel_fpga.py
TizianoDeMatteis: isn't the `sdfg.id` unique?

definelicht: Ah, are you appending this _after_ you cut this off? Then you're right

TizianoDeMatteis: Yep It is prepended ` module_function_name = "mod_" + str(sdfg.sdfg_id) + "_" + name`

definelicht: Ok, great!

definelicht (approved):

dace/codegen/targets/intel_fpga.py
```diff
@@ -1141,7 +1151,8 @@ def unparse_tasklet(self, sdfg, state_id, dfg, node, function_stream,
                                        locals,
```
tbennun, line 1151: (comment about line 1090)
Since you're in the Intel FPGA codegen, don't you want to support dtypes.Language.OpenCL as a tasklet language? (and also change the exception raised from "CPU" to "Intel FPGA")

TizianoDeMatteis: thanks!

tbennun (approved):

PR #507, opened 2020-12-02
Modernize Systolic GEMM and fix issues along the way

definelicht:
Modernize the systolic GEMM sample, and fix a handful of issues that popped up along the way:
- Handling of `StreamArray` in Xilinx wrongly removed the index
- Stream buffer sizes was not being symbolically evaluated before checking if the value is compile-time fixed
- Processing element unrolling for Intel FPGA made a wrong assumption
- Intel codegen should actually use the memlet name when generating channel accesses
- Intel codegen was defining memlets accessing stream arrays as stream arrays, when they should be streams
- Added missing undefines causing warnings

### DIFF
--- a/dace/codegen/targets/intel_fpga.py
+++ b/dace/codegen/targets/intel_fpga.py
@@ -650,7 +650,6 @@ __kernel void \\
                 "#undef _DACE_FPGA_KERNEL_{}\n".format(module_function_name))
         self._dispatcher.defined_vars.exit_scope(subgraph)
 
-
     def generate_nsdfg_header(self, sdfg, state, state_id, node,
                               memlet_references, sdfg_label):
         # Intel FPGA needs to deal with streams
@@ -1007,8 +1006,7 @@ __kernel void \\
 
             if data_name is not None:
                 data_desc = sdfg.arrays[data_name]
-                if (isinstance(data_desc, dace.data.Stream)
-                        and memlet.num_accesses != 1):
+                if (isinstance(data_desc, dace.data.Stream) and memlet.dynamic):
                     callsite_stream.write("#undef {}".format(memlet_name), sdfg)
 
     def _generate_converter(self, is_unpack, ctype, veclen, sdfg,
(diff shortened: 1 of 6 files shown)


dace/codegen/targets/intel_fpga.py
```diff
@@ -538,40 +536,39 @@ def generate_module(self, sdfg, state, name, subgraph, parameters,
...
         # Add kernel call host function
-        if unrolled_loops == 0:
+        if unrolled_loop is None:
             host_body_stream.write(
                 "kernels.emplace_back(program.MakeKernel(\"{}\"{}));".format(
                     module_function_name, ", ".join([""] + kernel_args_call)
                     if len(kernel_args_call) > 0 else ""), sdfg, state_id)
         else:
             # We will generate a separate kernel for each PE. Adds host call
-            for ul in self._unrolled_pes:
```
TizianoDeMatteis, line 555: Can't we have nested maps that generates PE?

dace/codegen/targets/intel_fpga.py
```diff
@@ -257,8 +257,6 @@ def make_read(self, defined_type, dtype, var_name, expr, index, is_pack,
         if defined_type == DefinedType.Stream:
             read_expr = "read_channel_intel({})".format(expr)
         elif defined_type == DefinedType.StreamArray:
-            # remove "[0]" index as this is not allowed if the subscripted value is not an array
-            expr = expr.replace("[0]", "")
```
TizianoDeMatteis, line 261: Is this handled differently?
IIRC there was a case where a single stream was accessed as an array.
(But this wasn't the right place to fix it)

definelicht: This was a bad hack that covered up the actual problem, which was that we defined memlets `#define pipe pipes[i]` as a `StreamArray`, when it should be a `Stream`. This is fixed now, so this hack is no longer necessary :-)

dace/codegen/targets/intel_fpga.py
definelicht: Maybe, but the rest of the code you wrote assumes maximum one map :P

TizianoDeMatteis (approved):

PR #512, opened 2020-12-03
Intel Autorun Kernels

definelicht:
If kernels have no arguments that need to be passed from the host, generate them as `__attribute__((autorun))` kernels.

### DIFF
--- a/tests/intel_fpga/autorun.py
+++ b/tests/intel_fpga/autorun.py
@@ -1,0 +1,132 @@
+import argparse
+import dace
+import numpy as np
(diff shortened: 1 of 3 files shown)


tests/intel_fpga/autorun.py
```diff
@@ -0,0 +1,132 @@
+import argparse
```
TizianoDeMatteis, line 1: Can you add a small comment on what this test does?

TizianoDeMatteis (approved):

PR #513, opened 2020-12-03
Nested ufunc calls

alexnick83:
Allows nesting of ufunc calls, including their supporting methods. A few examples are the following:
- np.less(np.absolute(Z), 2.0)
- np.add.reduce(np.absolute(Z))
- np.add.accumulate(np.absolute(Z))
- np.add.outer(np.divmod(A))


tbennun (approved):
I don't know why we have them as a list of size 1 in the first place (I think it's because some functions can return None and others return nothing at all?)

alexnick83:
> I don't know why we have them as a list of size 1 in the first place (I think it's because some functions can return None and others return nothing at all?)

They return one or two outputs (e.g. divmod)

PR #514, opened 2020-12-03
Speed up Xilinx CI

definelicht:
Run HLS instead of the compilation flow creating an .xo file.
This should shave off about 25-33% of the Xilinx testing time.

Also update hlslib for this to work with newer versions of Vitis.


dace/runtime/include/dace/xilinx/stream.h
```diff
@@ -14,13 +14,11 @@ namespace dace {
...
  public:
-  FIFO() : stream_(capacity) {
+  FIFO() : stream_() {
     #pragma HLS INLINE
-    #pragma HLS STREAM variable=stream_ depth=capacity
   }
-  FIFO(char const *const name) : stream_(name, capacity) {
+  FIFO(char const *const name) : stream_(name) {
     #pragma HLS INLINE
-    #pragma HLS STREAM variable=stream_ depth=capacity
```
tbennun, line 23: What's happening here? Are streams unbounded now?

definelicht: Xilinx changed how you specify the stream depth, it's still being passed as a template parameter (see bottom of class).

tbennun (approved):

PR #519, opened 2020-12-07
Fix bug in state fusion

am-ivanov:


dace/transformation/interstate/state_fusion.py
```diff
@@ -26,6 +26,10 @@ def __init__(self, first_inputs: Set[str], first_outputs: Set[str],
         self.second_input_nodes = second_input_nodes
 
 
+def top_level_nodes(state: SDFGState):
+    return [k for k, v in state.scope_dict().items() if v is None]
```
tbennun, line 30: Please use `state.scope_children()[None]` instead

tbennun (changes requested):

tbennun (approved):

PR #522, opened 2020-12-08
IntelFPGA: gearboxing with multiple kernels

TizianoDeMatteis:
Gearbox definitions are code generated in an external file.
FPGA kernels have only forward declaration.
In this way, If the same converter is used by multiple FPGA kernels, we don't have compiler errors (symbols defined multiple times)

### DIFF
--- a/dace/codegen/targets/intel_fpga.py
+++ b/dace/codegen/targets/intel_fpga.py
@@ -68,7 +68,8 @@ class IntelFPGACodeGen(fpga.FPGACodeGen):
         if fpga_vendor.lower() != "intel_fpga":
             # Don't register this code generator
             return
-        self.generated_converters = set() #Keep track of generated converters
+        self.generated_converters = set(
+        )  # Keep track of generated converters to avoid multiple definition
         super().__init__(*args, **kwargs)
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


dace/codegen/targets/intel_fpga.py
```diff
@@ -68,6 +68,8 @@ def __init__(self, *args, **kwargs):
         if fpga_vendor.lower() != "intel_fpga":
             # Don't register this code generator
             return
+        self.generated_converters = set(
+        )  # Keep track of generated converters to avoid multiple definition
```
definelicht, line 72:
```suggestion
        # Keep track of generated converters to avoid multiple definition
        self.generated_converters = set()  
```

tests/intel_fpga_test.sh
```diff
@@ -123,7 +123,10 @@ run_all() {
     run_sample fpga/nested_sdfg_as_kernel nested_sdfg_kernels "\n"
 
     # Generating autorun kernels
-    run_sample intel_fpga/autorun autorun_test "\n" 
+    run_sample intel_fpga/autorun autorun_test "\n"
+
+    # Multiple gearboxing
+    run_sample fpga/multiple_veclen_conversions multiple_veclen_conversions "\n"
```
definelicht, line 129: Does this not work for Xilinx?

definelicht (changes requested):

definelicht (approved):

PR #523, opened 2020-12-08
Remove validation check that disallows valid cases

am-ivanov:


tbennun:
@am-ivanov Following the discussion today, could you please change to ignore all the rest of the memlet path checks if the memlets are empty? Then I'll approve it

PR #524, opened 2020-12-08
Add dependency support to environments

orausch:
This PR adds a dependency field to dace environments, which is a list of environments.

When generating the framecode, the initialization and exit code ordering respects the dependencies.
This is required for a change in daceml: https://github.com/spcl/daceml/pull/30

### DIFF
--- a/dace/__init__.py
+++ b/dace/__init__.py
@@ -15,6 +15,8 @@ from .sdfg.propagation import propagate_memlets_sdfg, propagate_memlet
 from .memlet import Memlet
 from .symbolic import symbol
 
+import dace.library
+
 # Run Jupyter notebook code
 from .jupyter import *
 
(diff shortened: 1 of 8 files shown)


dace/__init__.py
```diff
@@ -15,6 +15,8 @@
 from .memlet import Memlet
 from .symbolic import symbol
 
+import dace.library
```
tbennun, line 18: Is this necessary? it slows down the import time of dace

orausch: Not sure, but see the commit before "Small Fixes", it was failing CI. 

Also; this fixes long standing static checker errors in PyCharm

tbennun: well, something is adding 5 more minutes to the build time... probably not this, unsure

tbennun (approved):

PR #525, opened 2020-12-08
Intel FPGAs: mangle channel names and PE function names

TizianoDeMatteis:
- Since channels are defined as global variables, we need to deal with this when having multiple kernels.
- This add also some mangling for PEs functions

### DIFF
--- a/dace/codegen/targets/intel_fpga.py
+++ b/dace/codegen/targets/intel_fpga.py
@@ -182,36 +182,33 @@ DACE_EXPORTED void __dace_exit_intel_fpga({signature}) {{
 
         return [host_code_obj] + kernel_code_objs + other_code_objs
 
-    def set_mangled_channel_name(self, var_name, sdfg_id, state_id):
+    def set_mangled_channel_name(self, var_name, kernel_id):
         '''
         Memorize and returns  the mangled name of a global channel
-        The dictionary is organized as (var_name) : {(sdfg_id, state_id): mangled_name)
+        The dictionary is organized as (var_name) : {kernel_id: mangled_name)
         '''
-        key = (sdfg_id, state_id)
-        if key not in self.channel_mangle[var_name]:
+
+        if kernel_id not in self.channel_mangle[var_name]:
             existing_count = len(self.channel_mangle[var_name])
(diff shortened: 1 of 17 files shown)


dace/codegen/targets/intel_fpga.py
```diff
@@ -177,6 +181,31 @@ def get_generated_codeobjects(self):
...
+        '''
+
+        if kernel_id not in self.channel_mangle[var_name]:
+            existing_count = len(self.channel_mangle[var_name])
+            suffix = f"_{existing_count}" if existing_count > 0 else ""
+            mangled_name = f"{var_name}{suffix}"
+            self.channel_mangle[var_name][kernel_id] = mangled_name
+        return self.channel_mangle[var_name][kernel_id]
+
+    def get_mangled_channel_name(self, var_name, kernel_id):
```
definelicht, line 197: This naming confuses me a bit, since both `set` and `get` return the variable name. Why can't they be combined? Maybe one should be implemented in terms of the other?

definelicht (changes requested):

dace/codegen/targets/intel_fpga.py
TizianoDeMatteis: Changing the name would be ok? Like `set_mangled_name` -> `create_mangled_name`
The `set` actually creates the mangled name (if not already present) and returns it .
The get only returns it, it does not have to create it.
They are called in different situations

definelicht: Create is better, but it's still weird, since it's not _necessarily_ creating anything. What is the reason that they cannot be combined in a single function?

TizianoDeMatteis: Because:
- I don't want to create a new mangle if that does not exist
- I want to deal with the "alias", that when we are trying to resolve a channel name that was defined by a `#define`

definelicht: Ok, doesn't the check know if a variable is an alias or not? Anyway, it's not that important, I just didn't really understand the difference between the two functions. I like `create` and `get` more than `set` and `get`.

definelicht (changes requested):

definelicht (approved):

PR #528, opened 2020-12-09
Added DOT xilinx expansion

carljohnsen:


dace/libraries/blas/nodes/dot.py
```diff
@@ -155,6 +160,137 @@ def expansion(node, state, sdfg):
...
+
+@dace.library.expansion
+class ExpandDOTFPGAStreamingLinearReduction(ExpandTransformation):
+
+    environments = []
+
+    @staticmethod
+    def make_sdfg(
+            dtype,
+            partial_width,
```
definelicht, line 173: What exactly is the partial width, and how does the user choose this?

dace/libraries/blas/nodes/dot.py
```diff
@@ -435,13 +572,32 @@ class Dot(dace.sdfg.nodes.LibraryNode):
     dtype = dace.properties.TypeClassProperty(allow_none=True)
     vec_width = dace.properties.SymbolicProperty(allow_none=False, default=1)
 
-    def __init__(self, name, dtype=None, vec_width=1, *args, **kwargs):
+    partial_width = dace.properties.SymbolicProperty(allow_none=False,
```
definelicht, line 575: Can you add a help entry here to describe this

dace/libraries/blas/utility/reductions.py
```diff
@@ -0,0 +1,845 @@
...
+            buf_out.data, '0',
+            wcr_str=reductionLambda,
+        )
+    )
+
+
+
+
+
+def fpga_make_matrixPartialReduction(dtype, nTile, mTile, partialWidth, n, m, veclen, a, b):
```
definelicht, line 323: Please do a pass to use our naming convention for `variable_names` and `function_names` (i.e., search for `[a-z][A-Z]`).

dace/libraries/blas/utility/reductions.py
```diff
@@ -0,0 +1,845 @@
...
+    inner_sdfg.add_array('_buf_in_unroll', shape=[max(2, partialWidth)], dtype=dtype,
+        storage=dtypes.StorageType.FPGA_Local if partialWidth > 8 else dtypes.StorageType.FPGA_Registers
+    )
+    inner_sdfg.add_array('buf_out', shape=[max(2, partialWidth)], dtype=dtype,
+        storage=dtypes.StorageType.FPGA_Local if partialWidth > 8 else dtypes.StorageType.FPGA_Registers
+    )
+
+    inner_sdfg.add_array('_x_buf', shape=[mTile], dtype=dtype, storage=dtypes.StorageType.FPGA_Local)
+    inner_sdfg.add_array('vecBuf_x', shape=[veclen], dtype=dtype, storage=dtypes.StorageType.FPGA_Registers, transient=True)
+    inner_sdfg.add_array('memBuf_x', shape=[veclen], dtype=dtype, storage=dtypes.StorageType.FPGA_Registers, transient=True)
```
definelicht, line 643: Please run `yapf` (version 0.30.0 or higher) on all the code files to format them

tests/blas/nodes/dot_test.py
```diff
@@ -1,11 +1,73 @@
...
 
 import dace.libraries.blas as blas
 
+import dace.libraries.blas.utility.fpga_helper as streaming
+from dace.libraries.blas.utility import memory_operations as memOps
+from dace.transformation.interstate import GPUTransformSDFG
+
+from dace.libraries.standard.memory import aligned_ndarray
+
+from multiprocessing import Process, Queue
```
definelicht, line 20: This doesn't seem to be used for anything?

tests/blas/nodes/dot_test.py
```diff
@@ -1,11 +1,73 @@
...
+import random
+
 import dace
 from dace.memlet import Memlet
 
 import dace.libraries.blas as blas
 
+import dace.libraries.blas.utility.fpga_helper as streaming
+from dace.libraries.blas.utility import memory_operations as memOps
+from dace.transformation.interstate import GPUTransformSDFG
```
definelicht, line 16: This seems leftover

tests/blas/nodes/dot_test.py
```diff
@@ -1,11 +1,73 @@
+#!/usr/bin/env python3
+
 import numpy as np
+
 import argparse
 import scipy
+import random
```
definelicht, line 7: Unused

definelicht (changes requested):

dace/libraries/blas/nodes/dot.py
carljohnsen: It defines the width of the buffer for the inner reduction. In pseudo python code:
```
partial = [0 for _ in range(partial_width)]
for i in range(n//veclen):
    x_buf = x.pop() # shape = [veclen]
    y_buf = y.pop() # shape = [veclen]
    for j in range(veclen):
        partial[j % partial_width] += x_buf[j] + y_buf[j]

res = 0
for i in range(partial_width):
    res += partial[i]

return res
```

So, the inner most map reduces into a buffer of size `partial_width`. This is to reduce the width of the final reduction.

It is set by the user, as a parameter to the library node.

dace/libraries/blas/utility/reductions.py
carljohnsen: Fixed

dace/libraries/blas/utility/reductions.py
carljohnsen: Fixed

tests/blas/nodes/dot_test.py
carljohnsen: Removed

tests/blas/nodes/dot_test.py
carljohnsen: All but `import dace.libraries.blas.utility.fpga_helper as streaming` was removed

tests/blas/nodes/dot_test.py
carljohnsen: Removed

dace/libraries/blas/nodes/dot.py
carljohnsen: A `desc` has been added

dace/libraries/blas/nodes/dot.py
definelicht: So the size of this is hardware-dependent, right? It has to be larger than the latency of the accumulation operation. What I meant with "how does the user choose this": it seems hard for the user to know what to set this to.

Is 2 really enough for Xilinx? I thought you needed more like 8. I would rather set this to something conservative so we know it always works, and then let the user reduce it if they want

carljohnsen: It should be a performance/resource trade-off. For optimal performance, I'd assume it should be equal to `veclen` and then the final reduction should be a reduction tree.

definelicht (approved):

PR #532, opened 2020-12-10
AXPY: Single FPGA implementation for all use-cases

manuelburger:
Sinlge FPGA implementation supporting all use-cases:
- Intel/Xilinx
- Streaming/Arrays


dace/libraries/blas/nodes/axpy.py
```diff
@@ -130,24 +141,27 @@ def make_sdfg(dtype, veclen, n, a, buffer_size_x, buffer_size_y,
             'axpy_task', ['x_con', 'y_con'], ['z_con'],
             'z_con = {} * x_con + y_con'.format(a))
 
+        access_vol = '0' if streaming else 'i'
```
definelicht, line 144: Isn't this an index, not a volume? The `dace.Memlet.simple(x_in.data, "0")` API is actually deprecated in favor of `dace.Memlet(f"{x_in.data}[{access_vol}]")`

dace/libraries/blas/nodes/axpy.py
```diff
@@ -273,14 +314,14 @@ class Axpy(dace.sdfg.nodes.LibraryNode):
...
+        fpga: FPGA implementation optimized for both streaming and array based inputs
     """
 
     # Global properties
     implementations = {
         "pure": ExpandAxpyVectorized,
-        "fpga_stream": ExpandAxpyFPGAStreaming,
-        "IntelFPGA": ExpandAxpyIntelFPGAVectorized
+        "fpga": ExpandAxpyFPGAStreaming,
+        "Intel_FPGA_DRAM": ExpandAxpyIntelFPGAVectorized
```
definelicht, line 324: Do we still need this, or does your new version cover all cases?

manuelburger: Depends, @TizianoDeMatteis version uses unrolling if I understood correctly. My thing is vectorized with Dace vectors and runs on both vendors. Don't know if you want to keep both, maybe there are use cases?

dace/libraries/blas/nodes/axpy.py
manuelburger: Fixed and updated to new API

dace/libraries/blas/nodes/axpy.py
definelicht: Ok, let's keep it. Could you rename it so we remember, to something like "IntelFPGAUnrolling"

tbennun: The reason I renamed Intel_FPGA_DRAM to `IntelFPGA` is because it needs to have a standard name across the library nodes. all the others use IntelFPGA

definelicht: @manuelburger Rename this to "IntelFPGAUnrolling" then we're ready to merge

tests/blas/nodes/axpy_test.py
```diff
@@ -314,7 +334,7 @@ def intel_fpga_graph(veclen, precision, vendor, testCase="0"):
     return test_sdfg.compile()
 
 
-def _test_fpga(type, vendor):
+def test_fpga(type, vendor):
```
tbennun, line 337: This is not good for pytest. It finds these tests and fails to run. We have to ensure that it is either in a valid pytest format (with `@pytest.mark.parametrize`) or not start with `test_`. This is why I changed it to `_test_fpga`.

dace/libraries/blas/nodes/axpy.py
manuelburger: Updated the naming back to IntelFPGA and the test is now called `intel_fpga_unroll`

tests/blas/nodes/axpy_test.py
```diff
@@ -140,12 +146,13 @@ def pure_graph(veclen, precision, implementation="pure", testCase="0"):
...
 
-    test_sdfg.expand_library_nodes()
+    # test_sdfg.expand_library_nodes()
+    saxpy_node.expand(test_sdfg, test_state, vec_width=veclen)
 
     return test_sdfg.compile()
 
 
-def test_pure():
+def _test_pure():
```
tbennun, line 155: that would probably be fine to keep as test_pure() so that our CPU tests run it automatically

definelicht (approved):

tests/blas/nodes/axpy_test.py
```diff
@@ -337,7 +357,8 @@ def _test_fpga(type, vendor):
...
     if args.target == "intel_fpga" or args.target == "xilinx":
         _test_fpga("fpga", args.target)
-    elif args.target == "intel_fpga_dram":
-        _test_fpga("intel_fpga_dram", "intel_fpga")
+        _test_fpga("fpga_array", args.target)
+    elif args.target == "intel_fpga_unroll":
+        _test_fpga("intel_fpga_unroll", "intel_fpga")
     else:
-        test_pure()
+        _test_pure()
```
tbennun, line 364: same

tbennun (changes requested):
one small change

tests/blas/nodes/axpy_test.py
tbennun, line 155:
```suggestion
def test_pure():
```

tests/blas/nodes/axpy_test.py
tbennun, line 364:
```suggestion
        test_pure()
```

PR #534, opened 2020-12-13
GEMM  fpga vectorized

TizianoDeMatteis:
New FPGA sample, with vectorization


samples/fpga/gemm_systolic_vectorized.py
```diff
@@ -0,0 +1,430 @@
...
+import dace
+import numpy as np
+import pdb
+import select
+import sys
+
+N = dace.symbol("N")
+K = dace.symbol("K")
+M = dace.symbol("M")
+P = dace.symbol("P")
```
definelicht, line 17: You could also make `W` a symbol (we used to not support this, but it should work now)?

(if it doesn't work, forget about it)

samples/fpga/gemm_systolic_vectorized.py
```diff
@@ -0,0 +1,430 @@
...
+    sdfg.add_stream("C_pipe",
+                    vec_type,
+                    transient=True,
+                    shape=(P + 1, ),
+                    storage=dace.dtypes.StorageType.FPGA_Local)
+
+    make_read_A(state)
+    make_read_B(state, sdfg, vec_width)
+    make_compute(sdfg, state, vec_width)
+    make_write_C(state, sdfg, vec_width)
```
definelicht, line 350: Are you sure you implemented `gemm`? Shouldn't there be a `read_C`?

definelicht (changes requested):

samples/fpga/gemm_systolic_vectorized.py
TizianoDeMatteis: it's in the writer

definelicht: Interesting idea :O Could you update the name?

definelicht (approved):

PR #535, opened 2020-12-13
FPGA BLAS: GEMV tiles streamed by row

manuelburger:
Adds FPGA GEMV implementation for streaming tiles in by row and within tiles by row `(tiles, withinTiles)->(row, row)`. Status:
- Works for Xilinx
- Should in theory also work for Intel, but there seems to be a bug in the codegen, where instead of a `float*` only a `float` is passed in a function declaration and thus leads to compilation errors when trying to dereference.

- Side change, adjusted `transformation.py`s base function `expansion` to contain all three arguments.


manuelburger:
`GEMV` row-streamed, row-ordered now also supports array-based inputs.

PR #539, opened 2020-12-29
strip_mining: number_of_tiles

thobauma:
- added number_of_tiles property that tiles the map into the number of provided tiles
- added test: tiling_number_of_tiles_test


tbennun (changes requested):
Thanks! Please use an enum instead of another string property (this could cover ceilrange as well)

dace/transformation/dataflow/strip_mining.py
```diff
@@ -137,9 +137,11 @@ class StripMining(transformation.Transformation):
     new_dim_prefix = Property(dtype=str,
                               default="tile",
                               desc="Prefix for new dimension name")
-    tile_size = Property(dtype=str,
-                         default="64",
-                         desc="Tile size of strip-mined dimension")
+    tile_size_or_number = Property(
```
tbennun, line 140: The name now is hard to remember when using the API. Maybe we should keep it tile_size

tbennun: It will also create some compatibility issues with current users of dace, which use "tile_size" as a property already.

tbennun (changes requested):

dace/transformation/dataflow/strip_mining.py
thobauma: Makes sense, I on the fence about changing it. Should it still be used as the number of tiles? Otherwise we would need a new Property again.

tbennun (changes requested):
1. this needs to be merged with master (CI doesn't run)
2. did you check that all occurrences of the `ceilrange` property were replaced with `tiling_type = 'ceilrange'`?

thobauma:
I just merged it.
I changed the occurrence of `ceilrange`.

tbennun (approved):

PR #549, opened 2021-01-13
Pipeline scope: user can define additional variables 

### LINKED ISSUE #548: Allow the user to define additional symbols in Pipeline scope
The pipeline scope is mainly used for generating nice FPGA code.
In this context, it could be the case that the user needs additional scope-level variables, that are then managed by him itself.

A possible solution could be to have an additional parameter in the Pipeline constructor (a dictionary)

TizianoDeMatteis:
- Resolves #548 
- Resolves a bug in cppunparse: if a symbol is already defined, there is no need to re-define it (or perform type inference when needed). Before, this was causing the re-definition of already defined variables.

### DIFF
--- a/dace/codegen/cppunparse.py
+++ b/dace/codegen/cppunparse.py
@@ -321,7 +321,7 @@ class CPPUnparser:
                     target.id, self._indent):
 
                 # if the target is already defined, do not redefine it
-                if target.id not in self.defined_symbols:
+                if self.defined_symbols is None or target.id not in self.defined_symbols:
                     # we should try to infer the type
                     if self.type_inference is True:
                         # Perform type inference
--- a/dace/codegen/cppunparse.py
+++ b/dace/codegen/cppunparse.py
@@ -320,10 +320,10 @@ class CPPUnparser:
                 (ast.Subscript, ast.Attribute)) and not self.locals.is_defined(
                     target.id, self._indent):
 
-                # the target is not already defined: we should try to infer the type
-                if self.type_inference is True:
-                    # if the target type is already known, no need to perform type inference
-                    if target.id not in self.defined_symbols:
+                # if the target is already defined, do not redefine it
+                if target.id not in self.defined_symbols:
+                    # we should try to infer the type
+                    if self.type_inference is True:
                         # Perform type inference
                         # Build dictionary with symbols
                         def_symbols = {}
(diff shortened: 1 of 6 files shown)


dace/codegen/cppunparse.py
```diff
@@ -320,32 +320,35 @@ def _Assign(self, t):
...
+                        if self.language == dace.dtypes.Language.OpenCL and (
+                                inferred_type is not None
+                                and inferred_type.veclen > 1):
+                            # if the veclen is greater than one, this should be defined with a vector data type
+                            self.write("{}{} ".format(
+                                dace.dtypes._OCL_VECTOR_TYPES[
+                                    inferred_type.type], inferred_type.veclen))
+                        else:
+                            self.write(dace.dtypes._CTYPES[inferred_type.type] +
+                                       " ")
```
definelicht, line 348: How is this related? Maybe add a comment about it on the PR.

dace/sdfg/nodes.py
```diff
@@ -1050,19 +1053,24 @@ class Pipeline(Map):
         dtype=bool,
         default=True,
         desc="Whether to increment regular map indices during pipeline drain.")
+    additional_variables = Property(
+        dtype=dict,
+        desc="Additional variables, managed by the user inside the scope.")
```
definelicht, line 1058: Maybe call these `iterators` or `symbols`?

dace/sdfg/state.py
```diff
@@ -1446,6 +1445,10 @@ def add_pipeline(self,
             :param drain_size:    Number of iterations of draining phase.
             :param drain_overlap: Whether the draining phase overlaps with
                                   the "main" streaming phase of the loop.
+            :param additional_variables: a dictionary contanining the additional
+                                  variables that will be created for this scope and that are not
+                                  handled by the scope guards it self.
+                                  The dictionary is in the form 'variable_name' -> init_value
```
definelicht, line 1451:
```suggestion
            :param additional_variables: a dictionary containing additional
                                  symbols that will be created for this scope and that are not
                                  automatically managed by the scope code.
                                  The dictionary takes the form 'variable_name' -> init_value
```

tests/fpga/pipeline_scope.py
```diff
@@ -43,8 +43,8 @@ def make_sdfg(dtype, name="pipeline_test"):
     post_write = post_state.add_write("b")
     post_state.add_memlet_path(post_read,
                                post_write,
-                               memlet=dace.Memlet.simple(post_write,
```
definelicht, line 46: Could we take this opportunity to get rid of `Memlet.simple` in this sample? :-)

definelicht (changes requested):

dace/codegen/cppunparse.py
TizianoDeMatteis: Yes. It popped while I was changing the test (but then i changed it again). Will add a comment

definelicht (approved):

PR #561, opened 2021-01-29
BufferTiling, TilingWithOverlap, Copyright update

PanicSheep:
For BufferTiling to work MapTiling needs to "tile" dimensions with tile size 1. This can be achieved with its new property 'tile_trivial'.
Also TrivialMapElimination had to be improved.

### DIFF
--- a/dace/transformation/dataflow/buffer_tiling.py
+++ b/dace/transformation/dataflow/buffer_tiling.py
@@ -1,5 +1,5 @@
-# Copyright 2019-2020 ETH Zurich and the DaCe authors. All rights reserved.
-""" Contains classes that implement the OnTheFlyMapFusion transformation. """
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+""" Contains classes that implement the BufferTiling transformation. """
 
 from dace import registry
 from dace.sdfg import nodes
@@ -9,6 +9,14 @@ from dace.transformation.dataflow import MapTiling, MapTilingWithOverlap, MapFus
 
 @registry.autoregister_params(singlestate=True)
 class BufferTiling(transformation.Transformation):
+    """ Implements the buffer tiling transformation.
+
+        BufferTiling tiles a buffer that is in between two maps, where the preceding map
+        writes to the buffer and the succeeding map reads from it.
+        It introduces additional computations in exchange for reduced memory footprint.
+        Commonly used to make use of shared memory on GPUs.
+    """
+
     _map1_exit = nodes.MapExit(nodes.Map('', [], []))
     _array = nodes.AccessNode('')
     _map2_entry = nodes.MapEntry(nodes.Map('', [], []))
--- a/dace/transformation/dataflow/buffer_tiling.py
+++ b/dace/transformation/dataflow/buffer_tiling.py
@@ -1,0 +1,134 @@
+# Copyright 2019-2020 ETH Zurich and the DaCe authors. All rights reserved.
+""" Contains classes that implement the OnTheFlyMapFusion transformation. """
+
+from dace import registry
+from dace.sdfg import nodes
+from dace.sdfg import utils as sdutil
+from dace.transformation import transformation
+from dace.transformation.dataflow import MapTiling, MapTilingWithOverlap, MapFusion, TrivialMapElimination
+
+@registry.autoregister_params(singlestate=True)
+class BufferTiling(transformation.Transformation):
+    _map1_exit = nodes.MapExit(nodes.Map('', [], []))
+    _array = nodes.AccessNode('')
+    _map2_entry = nodes.MapEntry(nodes.Map('', [], []))
+
+    # Returns a list of graphs that represent the pattern
+    @staticmethod
+    def expressions():
+        return [
+            sdutil.node_path_graph(
+                BufferTiling._map1_exit,
+                BufferTiling._array,
+                BufferTiling._map2_entry,
+            )
+        ]
+
+    @staticmethod
+    def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
+        map1_exit = graph.nodes()[candidate[BufferTiling._map1_exit]]
+        map2_entry = graph.nodes()[candidate[BufferTiling._map2_entry]]
+
+        # Check that locally provided buffer has volume 1.
+        for edge in graph.in_edges(map1_exit):
+            if edge.data.volume != 1:
+                return False
+
+        # Check that all locally provided buffers have the same subset.
+        subsets = set(edge.data.subset for edge in graph.in_edges(map1_exit))
+        if len(subsets) != 1:
+            return False
+
+        # Check existence of buffers
+        buffers = graph.all_nodes_between(map1_exit, map2_entry)
+        if buffers is None:
+            return False
+
+        for buf in buffers:
+            # Check that buffers are AccessNodes.
+            if not isinstance(buf, nodes.AccessNode):
+                return False
+
+            # Check that buffers are transient.
+            if not sdfg.arrays[buf.data].transient:
+                return False
+            
+            # Check that buffers have exactly 1 input and 1 output edge.
+            if graph.in_degree(buf) != 1:
+                return False
+            if graph.out_degree(buf) != 1:
+                return False
+            
+            # Check that buffers are next to the maps.
+            if graph.in_edges(buf)[0].src != map1_exit:
+                return False
+            if graph.out_edges(buf)[0].dst != map2_entry:
+                return False
+
+            # Check that the data consumed is provided.
+            provided = graph.in_edges(buf)[0].data.subset
+            consumed = graph.out_edges(buf)[0].data.subset
+            if not provided.covers(consumed):
+                return False
+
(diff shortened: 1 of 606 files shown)


PanicSheep:
There are over 600 instances of "Copyright 2019-2020" or similar in the code base. How about I make a dedicated PR changing them all at once and leave it as is for now?

tbennun:
@PanicSheep I just meant in your new files. I will do so on some occasion but at least the new files should use the new dates.

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @staticmethod
+    def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
+        map1_exit = graph.nodes()[candidate[BufferTiling._map1_exit]]
+        map2_entry = graph.nodes()[candidate[BufferTiling._map2_entry]]
+
+        # Check that locally provided buffer has volume 1.
+        for edge in graph.in_edges(map1_exit):
+            if edge.data.volume != 1:
+                return False
```
tbennun, line 35: how do we know that this is the data we need? i.e., the one that connects to the array? you should add a check on the data name

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Check that all locally provided buffers have the same subset.
+        subsets = set(edge.data.subset for edge in graph.in_edges(map1_exit))
```
tbennun, line 38: same

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+        if len(subsets) != 1:
+            return False
+
+        # Check existence of buffers
+        buffers = graph.all_nodes_between(map1_exit, map2_entry)
+        if buffers is None:
+            return False
```
tbennun, line 45: this will never return None, because you matched it in the pattern. The check can be removed

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+
+        for buf in buffers:
```
tbennun, line 47: why don't you use the `array` pattern node here instead? you can also not match the transformation if there is any other node in between the two maps.

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+            if graph.out_edges(buf)[0].dst != map2_entry:
+                return False
+
+            # Check that the data consumed is provided.
+            provided = graph.in_edges(buf)[0].data.subset
+            consumed = graph.out_edges(buf)[0].data.subset
+            if not provided.covers(consumed):
+                return False
+
+            # Check that buffers occure only once in this state.
```
tbennun, line 74:
```suggestion
            # Check that buffers occur only once in this state.
```

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+            num_occurrences = len([
```
tbennun, line 75: why not check if it occurs only once in the entire SDFG?

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+    def apply(self, sdfg):
+        graph = sdfg.nodes()[self.state_id]
+        map1_exit = graph.nodes()[self.subgraph[self._map1_exit]]
+        map1_entry = graph.entry_node(map1_exit)
+        map2_entry = graph.nodes()[self.subgraph[self._map2_entry]]
+        buffers = graph.all_nodes_between(map1_exit, map2_entry)
+        # Situation:
+        # -> map1_entry -> ... -> map1_exit -> buffers -> map2_entry -> ...
+
+        lower_extents = tuple(b[0]-a[0] for a,b in zip(map1_entry.range.ranges, map2_entry.range.ranges))
```
tbennun, line 100: there's a function called `min_element` for subsets, which should be helpful here.

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+        upper_extents = tuple(a[1]-b[1] for a,b in zip(map1_entry.range.ranges, map2_entry.range.ranges))
```
tbennun, line 101: same, max_element

dace/transformation/dataflow/buffer_tiling.py
```diff
@@ -0,0 +1,134 @@
...
+""" Contains classes that implement the OnTheFlyMapFusion transformation. """
+
+from dace import registry
+from dace.sdfg import nodes
+from dace.sdfg import utils as sdutil
+from dace.transformation import transformation
+from dace.transformation.dataflow import MapTiling, MapTilingWithOverlap, MapFusion, TrivialMapElimination
+
+@registry.autoregister_params(singlestate=True)
+class BufferTiling(transformation.Transformation):
```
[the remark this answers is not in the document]
tbennun: Seems like it not only performs tiling, but also fusion.

dace/transformation/dataflow/tiling_with_overlap.py
```diff
@@ -0,0 +1,64 @@
...
+            return
+        if len(self.upper_overlap) == 0:
+            return
+        
+        graph = sdfg.nodes()[self.state_id]
+        map_entry = graph.nodes()[self.subgraph[self._map_entry]]
+
+        # Tile the map
+        self.tile_trivial = True
+        MapTiling.apply(self, sdfg)
```
tbennun, line 41: a very strange way of invoking a `super()` method

dace/transformation/dataflow/tiling_with_overlap.py
```diff
@@ -0,0 +1,64 @@
...
+
+@registry.autoregister_params(singlestate=True)
+@make_properties
+class MapTilingWithOverlap(MapTiling):
+    """ Implements the orthogonal tiling transformation with overlap.
+
+        Orthogonal tiling is a type of nested map fission that creates tiles
+        in every dimension of the matched Map. The overlap can vary in each
+        dimension and direction. It is added to each tile and the starting
+        and end points of the outer map are ajusted to account for the overlap.
```
tbennun, line 19:
```suggestion
        and end points of the outer map are adjusted to account for the overlap.
```

dace/transformation/dataflow/trivial_map_elimination.py
```diff
@@ -13,9 +13,10 @@
 class TrivialMapElimination(transformation.Transformation):
     """ Implements the Trivial-Map Elimination pattern.
 
-        Trivial-Map Elimination takes a map with a range
-        containing one element and removes the map.
-        Example: Map[i=0] -> nothing
+        Trivial-Map Elimination removes all dimensions containing only one
+        element from a map. If this applies to all ranges the map is removed.
+        Example: Map[i=0:I,j=7] -> Map[j=7]
```
tbennun, line 18: you mean `Map[i=0:I]`, right?

tbennun (changes requested):

dace/transformation/dataflow/buffer_tiling.py
PanicSheep: To be conform with [MapFusion](https://github.com/spcl/dace/blob/master/dace/transformation/dataflow/map_fusion.py#L112).

dace/transformation/dataflow/buffer_tiling.py
PanicSheep: You're right. I moved the check inside the buffer-loop.

dace/transformation/dataflow/buffer_tiling.py
PanicSheep: I removed this check. It tried to restrict BufferTiling to stencil-like programs, but I think this is no longer needed. Programs that don't satisfy this check are quite weird, but excluding them seems premature now. After all, MapTiling and MapFusion might be able to process them correctly.

dace/transformation/dataflow/buffer_tiling.py
PanicSheep: Because there can be multiple parallel AccessNodes between the maps. Like Smagorinsky for example.

dace/transformation/dataflow/buffer_tiling.py
PanicSheep: But maybe we should change both. But I'm not 100% sure though.

PanicSheep (commented):
I just replaced all the copyrights. It was faster.

dace/transformation/dataflow/buffer_tiling.py
PanicSheep: done

dace/transformation/dataflow/tiling_with_overlap.py
PanicSheep: I replaced it.

tbennun (approved):

Issue #570, opened 2021-02-08
Type Inference fails with pointers generated by OpenCL Keyword Remover


TizianoDeMatteis:
**Describe the bug**
If in a tasklet, there is an assignment in which the right-hand side is a connector and the corresponding memlet is dynamic, the OpenCL Keyword Remover, will update the code to be `target = *rhs`.
In a similar situation, type inference fails because it is not able to understand the type of "*rhs" 

**Expected behavior**
Considering the example above, assuming that the type of `rhs` is known, type inference should infer the type of target to be the same of `rhs`

PR #574, opened 2021-02-10
Sympy floor ceil

TizianoDeMatteis:
When a symbolic expression is converted to C++, the  `sympy_intdiv_fix` to simplify ceiling/floor expression.

This happens, for example, when the code for a map is generated. If map ranges are defined in terms of ceiling (for example), the code generated is ok as long as the range expression is not complicated (I don't how to define complicate it). For example:
- a map range defined as `"n": "0:ceiling(ONNX_num_filters/6) - 1"` is then codegenerated as `for (int n0 = 0; n0 < int_ceil(ONNX_num_filters, 6); n0 += 1)`, where the ceiling function as been replaced with `int_ceil` (FPGA backend)
- something like the following `"n": "0:ceiling((ONNX_in_features_x - 4)*(ONNX_in_features_y - 4)/576) - 1"` is instead left as it is, codegenerating also the `ceiling` function call, hence compiler error

This PR, matches also generic expressions at the numerator

Note: This also Solves #570.
This solution takes into account that the OCL Keyword Remover updates the AST node, by maintaining the same type: in this particular case, it is an ast.Name, where the id is something like *smtg.

Therefore, in the type inference, if we find a leading * we simply remove it before inferring the type.


dace/symbolic.py
```diff
@@ -516,6 +516,7 @@ def sympy_intdiv_fix(expr):
     b = sympy.Wild('b', properties=[lambda k: k.is_Symbol or k.is_Integer])
     c = sympy.Wild('c')
     d = sympy.Wild('d')
+    e = sympy.Wild('e', properties=[lambda k: isinstance(k, sympy.Basic)])
```
tbennun, line 519: Even integers are sympy.Basic (`isinstance(sp.sympify('1'), sp.Basic)` --> `True`). Would this always work?

alexnick83: Has this been resolved? We can probably add a test. If it is an issue, then I suggest to either do `isinstance(k, sympy.Expr)` or perhaps even better `isinstance(k, sympy.Basic) and not isinstance(k, sympy.Atom)`. 
From the SymPy documentation about the `Atom` class:
> A parent class for atomic things. An atom is an expression with no subexpressions.
>
> Examples
> 
> Symbol, Number, Rational, Integer, … But not: Add, Mul, Pow, …

alexnick83 (changes requested):
Looks ok to me apart from verifying whether `sympy.Basic` is an issue, as Tal suggested.

dace/symbolic.py
TizianoDeMatteis: It should be solved now, I've explicitly added to not match Atom expressions

tbennun (approved):
Approved if tests pass

PR #577, opened 2021-02-16
Various fixes for GPU code generation and miscellaneous things

tbennun:
- [x] Expand library nodes only after setting schedules
- [x] Fix API for GPU return values
- [x] New API for loading precompiled SDFGs
- [x] Fix longstanding variable shadowing in GPU batched matmul
- [x] Fix schedule type inference for top-level schedules
- [x] `InlineSDFG` creates reshapes when possible, applies in more cases
- [x] Add test for inout reshape in inline SDFG

### DIFF
--- a/dace/transformation/interstate/sdfg_nesting.py
+++ b/dace/transformation/interstate/sdfg_nesting.py
@@ -311,19 +311,6 @@ class InlineSDFG(transformation.Transformation):
                 sink_accesses.add(node)
 
         #######################################################
-        # Add nested SDFG into top-level SDFG
-
-        # Add nested nodes into original state
-        subgraph = SubgraphView(nstate, [
-            n for n in nstate.nodes()
-            if n not in (source_accesses | sink_accesses)
-        ])
-        state.add_nodes_from(subgraph.nodes())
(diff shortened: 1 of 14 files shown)


dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -260,6 +235,27 @@ def apply(self, sdfg):
...
 
+        # Access nodes that need to be reshaped
+        reshapes: Set(str) = set()
+        for aname, array in nsdfg.arrays.items():
+            if array.transient:
+                continue
+            edge = None
+            if aname in inputs:
+                edge = inputs[aname]
+                if len(array.shape) > len(edge.data.subset):
```
alexnick83, line 246: Why the reshape is happening when `len(array.shape) > len(edge.data.subset)` and not the opposite as well?

tbennun: In this case, there is definitely no unsqueeze possible. For the other case (whether it is an unsqueeze or reshape), the edge is kept for the _check_strides test below

alexnick83 (approved):

PR #581, opened 2021-02-18
RedundantArray fix and invocation casting warning fixes

tbennun:
- [x] Emit reshape instead of not removing array when shapes differ
- [x] Fix type warnings on SDFG invocation

### DIFF
--- a/dace/transformation/dataflow/redundant_array.py
+++ b/dace/transformation/dataflow/redundant_array.py
@@ -176,7 +176,8 @@ class RedundantArray(pm.Transformation):
 
         # Finally, remove in_array node
         graph.remove_node(in_array)
-        del sdfg.arrays[in_array]
+        if in_array.data in sdfg.arrays:
+            del sdfg.arrays[in_array.data]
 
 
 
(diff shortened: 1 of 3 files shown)


dace/transformation/dataflow/redundant_array.py
```diff
@@ -152,8 +147,21 @@ def gnode(nname):
...
-        in_array = gnode(RedundantArray._in_array)
-        out_array = gnode(RedundantArray._out_array)
+        in_array = self.in_array(sdfg)
+        out_array = self.out_array(sdfg)
+        in_desc = sdfg.arrays[in_array.data]
+        out_desc = sdfg.arrays[out_array.data]
+
+        # If arrays are not of the same shape, modify first array to reshape
+        # instead of removing it
+        if len(in_desc.shape) != len(out_desc.shape) or any(
```
alexnick83, line 157: Does this solve all issues?

alexnick83 (approved):

dace/transformation/dataflow/redundant_array.py
tbennun: It solves redundant arrays that we couldn't remove, for example anything that comes out of numpy with keepdims=True, but I'm not sure how it generalizes to different kinds of slices.

PR #585, opened 2021-02-22
Add free running kernel support

andreaskuster:


andreaskuster:
Open issues:
- [ ] we currently have to add the declaration of the dace_exit(..) function call manually and call it from the last tasklet
- [ ] there is no type like dace.vector, for streams, i.e. `dace.stream(dtype=dace.int32, size=1)`. As a temporary workaround, i used `dace.struct(name="stream", a=dace.int32)`

andreaskuster:
FPGA transformation did not seem to have helped a lot.

PR #586, opened 2021-02-22
Bugfix: resolves symbol name clashes

thobauma:
- resolved a bug with symbol name clashes in `_check_strides`
- added myself to the list of authors


tbennun:
@thobauma for some reason, one of the tests is crashing now...

tbennun (approved):

PR #588, opened 2021-02-22
Transformation functions as methods

tbennun:
- [x] Make `can_be_applied` and others methods
- [x] Ensure Transformation classes are always initialized with properties before matching

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -1947,15 +1947,15 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
         for xform, opts in zip(xforms, options):
             # Find only the first match
             try:
-                match = next(m for m in opt.get_pattern_matches(
-                    strict=strict, patterns=[xform], states=states))
+                match = next(m
+                             for m in opt.get_pattern_matches(strict=strict,
+                                                              patterns=[xform],
+                                                              states=states,
+                                                              options=[opts]))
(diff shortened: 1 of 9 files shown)


definelicht (commented):
Looks reasonable. Is this backwards compatible with the static methods?

Do we have a use case where we need to use properties in can_be_applied that we can test with?

tbennun:
> Looks reasonable. Is this backwards compatible with the static methods?
> 
> Do we have a use case where we need to use properties in can_be_applied that we can test with?

@definelicht It is backwards compatible (all transformations are still using `staticmethod`s), which also answers your second question: we don't have any case where we actually use it. Any ideas?

definelicht:
> 
> 
> > Looks reasonable. Is this backwards compatible with the static methods?
> > Do we have a use case where we need to use properties in can_be_applied that we can test with?
> 
> @definelicht It is backwards compatible (all transformations are still using `staticmethod`s), which also answers your second question: we don't have any case where we actually use it. Any ideas?

Not with existing transformations. I bumped into a need for this when looking at doing a gearboxing transformation, where the streaming memory module reads 512-bit vectors and gearboxes them into the N-bit vectors/values used by the kernel, which obviously shouldn't match if the kernel is already reading 512-bit vectors. But I'm not working on this right now :-/

tbennun:
In any case, it definitely can't hurt to have it.

definelicht (approved):

PR #590, opened 2021-02-22
Fix for BLAS dot library node

alexnick83:
The return value of `cblas_?dot` is supposed to be a scalar and not a pointer.
Fix by changing the OpenBLAS expansion tasklet code from `_result = cblas ...` to `_result[0] = cblas ...`.
`_result` is an array of size 1.

### DIFF
--- a/tests/library/blas_dot_test.py
+++ b/tests/library/blas_dot_test.py
@@ -125,8 +125,8 @@ def _test_dot(implementation, dtype, sdfg):
 
 
 def test_dot():
-    # _test_dot("32-bit pure SDFG", np.float32, make_sdfg("pure", dace.float32))
-    # _test_dot("64-bit pure SDFG", np.float64, make_sdfg("pure", dace.float64))
+    _test_dot("32-bit pure SDFG", np.float32, make_sdfg("pure", dace.float32))
+    _test_dot("64-bit pure SDFG", np.float64, make_sdfg("pure", dace.float64))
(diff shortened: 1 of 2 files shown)


tim0s (approved):
this should work - let's see if the tests pass :)

tbennun:
@alexnick83 shouldn't you change the connector type in the tasklet to scalar instead? This might fail in the cases where the output is a scalar

tbennun (changes requested):
see message

tests/library/blas_dot_test.py
```diff
@@ -52,7 +52,7 @@ def make_sdfg(implementation, dtype, storage=dace.StorageType.Default):
     state.add_memlet_path(dot_node,
                           result,
                           src_conn="_result",
-                          memlet=Memlet.simple(result, "0", num_accesses=-1))
+                          memlet=Memlet.simple(result, "0", num_accesses=1))
```
alexnick83, line 55: Major issue: If this goes back to -1 (memlet dynamic), code generation does not generate code (see dace/codegene/targets/cpu.py, line 837)

tbennun: it shouldn't be dynamic anyway - this is by design. The code generator does not output anything on its own if `dynamic=True`. Instead, this is done in the tasklet.

tbennun (approved):

PR #592, opened 2021-02-22
Stop redundant_array from creating non-Scope lifetime views, Fix MatMul

orausch:


tests/library/gemm_test.py
```diff
@@ -14,6 +14,22 @@
 N = dace.symbol('N')
 
 
+@pytest.mark.parametrize(('implementation', ),
+                         [('pure', ), ('MKL', ),
+                          pytest.param('cuBLAS', marks=pytest.mark.gpu)])
+def test_gemm_no_c(implementation):
+    Gemm.default_implementation = implementation
```
tbennun, line 21: please keep the old implementation and set it back. pytest does not create new python scripts when running.
Also, not everyone has MKL so you need to either mark it as such or have a try/except block

tbennun (changes requested):

tbennun (approved):
LGTM if passes

orausch:
Waiting on #590

orausch:
merged in #597

PR #593, opened 2021-02-23
Fix schedule inference

orausch:


dace/codegen/targets/cpu.py
```diff
@@ -215,6 +219,13 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
 
         # Compute array size
         arrsize = nodedesc.total_size
+        arrsize_bytes = arrsize * nodedesc.dtype.bytes
+        try:
+            size_too_big = bool(
```
tbennun, line 224: what you want is `(arrsize_bytes > max_stack) == True` (note the `==`!). It gives you True if it's definitely true, or False if it is false or indeterminate.
Also the try/catch would not work now, so you'll need to re-add the `issymbolic(arrsize, sdfg.constants)` clause too

tbennun (changes requested):

tbennun (approved):

PR #596, opened 2021-02-23
Support for constant negative indices

alexnick83:
Adds support to the Python frontend for accessing arrays with constant negative indices, e.g. `A[-2]`.
This PR does not include support for non-constant negative indices, e.g. `A[-i]`, where `i` is a map index variable or something similar.

### DIFF
--- a/dace/frontend/python/astutils.py
+++ b/dace/frontend/python/astutils.py
@@ -1,7 +1,6 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
 """ Various AST parsing utilities for DaCe. """
 import ast
-from tokenize import Number
 import astunparse
 from collections import OrderedDict
 import inspect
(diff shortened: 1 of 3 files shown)


dace/frontend/python/astutils.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
 """ Various AST parsing utilities for DaCe. """
 import ast
+from tokenize import Number
```
tbennun, line 4: shouldn't this be `numbers`?

dace/frontend/python/astutils.py
```diff
@@ -222,10 +223,14 @@ def astrange_to_symrange(astrange, arrays, arrname=None):
                 begin = symbolic.pystr_to_symbolic(0)
             else:
                 begin = symbolic.pystr_to_symbolic(unparse(begin))
+                if isinstance(begin, numbers.Number) and begin < 0:
```
tbennun, line 226: (applies to all other checks in the PR) you can check for "definitely under zero" by changing the check to `(begin < 0) == True` (note the `==` rather than `is`)

tbennun: This will also solve the currently-failing cases

tbennun (changes requested):

dace/frontend/python/astutils.py
alexnick83: ohdeer, pylance added this automatically.

tbennun (approved):
LGTM if tests pass

PR #599, opened 2021-02-24
Lapack lib nodes

alexnick83:
Adding library nodes for lapack getrf, getri, getrs, as well as numpy linalg functions inv and solve.


alexnick83:
Current work was absorbed by the mpi_lib_nodes branch and was merged into master. Obsolete.

PR #613, opened 2021-03-02
Fix sporadic errors in reloadable library test

tbennun:


dace/frontend/python/parser.py
```diff
@@ -286,7 +282,13 @@ def __call__(self, *args, **kwargs):
         # altering transformations such as Vectorization)
         binaryobj = sdfg.compile()
 
-        return binaryobj(**kwargs)
+        # Call SDFG
+        result = binaryobj(**kwargs)
+
+        # Enforce deletion and closure of compiled program
+        del binaryobj
```
orausch, line 289: `__del__` should be called anyway on function exit, not sure if this changes anything.

orausch (approved):

dace/frontend/python/parser.py
tbennun: Sometimes it does, from experience

PR #616, opened 2021-03-02
Improvements to Redundant Array transformations.

alexnick83:
The redundant array transformations should always work when the shape of the array to-be-removed is a subset of the shape of the array to-stay:
- RedundantSecondArray now checks that output shape is a subset of input shape in strict mode.
- RedundantArray now doesn't create Views when the input shape is a subset of the output shape. Instead, it uses subset composition, exactly the same way as RedundantSecondArray.

### DIFF
--- a/dace/transformation/dataflow/redundant_array.py
+++ b/dace/transformation/dataflow/redundant_array.py
@@ -43,47 +43,49 @@ def _validate_subsets(edge: graph.MultiConnectorEdge,
     # dimensionality of the subset and we pop or pad indices/ranges accordingly.
     # In that case, we also set the subset to start from 0 in each dimension. 
     if not src_subset:
-        src_subset = copy.deepcopy(dst_subset)
         if src_name:
             desc = arrays[src_name]
-            padding = len(desc.shape) - len(src_subset)
-            if padding != 0:
-                if padding > 0:
-                    if isinstance(src_subset, subsets.Indices):
-                        indices = [0] * padding + src_subset.indices
-                        src_subset = subsets.Indices(indices)
-                    elif isinstance(src_subset, subsets.Range):
-                        ranges = [(0, 0, 1)] * padding + src_subset.ranges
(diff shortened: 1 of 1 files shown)


dace/transformation/dataflow/redundant_array.py
```diff
@@ -154,28 +173,46 @@ def gnode(nname):
...
 
-        # If arrays are not of the same shape, modify first array to reshape
-        # instead of removing it
-        if len(in_desc.shape) != len(out_desc.shape) or any(
-                i != o for i, o in zip(in_desc.shape, out_desc.shape)):
+        # 1. Get edge e1 and extract subsets for arrays A and B
+        e1 = graph.edges_between(in_array, out_array)[0]
+        a1_subset, b_subset = _validate_subsets(e1, sdfg.arrays)
+
+        # If the memlet does not cover the remove array, create a view.
```
tbennun, line 180:
```suggestion
        # If the memlet does not cover the removed array, create a view.
```

dace/transformation/dataflow/redundant_array.py
```diff
@@ -154,28 +173,46 @@ def gnode(nname):
...
+            for e3 in path:
+                # 2-a. Extract subsets for array B and others
+                other_subset, a3_subset = _validate_subsets(
+                    e3, sdfg.arrays, dst_name=in_array.data)
+                # 2-b. Modify memlet to match array B.
+                e3.data.data = out_array.data
+                a3_subset.offset(a1_subset, negative=True)
+                if isinstance(b_subset, subsets.Indices):
+                    tmp = copy.deepcopy(b_subset)
+                    tmp.offset(a3_subset, negative=False)
```
tbennun, line 202: you can use `offset_new` for this, which creates an out-of-place (new) offsetted subset

tbennun (approved):

tbennun:
@alexnick83 approved in general, but two suggestions to apply

PR #620, opened 2021-03-03
LAPACK and MPI lib nodes

tim0s:
New PR for lapack and MPI library nodes


tests/numpy_bool_input_test.py
```diff
@@ -24,7 +24,7 @@ def test_numpy_bool_input():
     assert result[0]
 
     # test numpy.bool (which is just bool)
-    result = sdfg(in_bool=np.bool(True))
+    result = sdfg(in_bool=bool(True))
```
tbennun, line 27: do we need the `bool(True)` here or can we just use `True`

tests/library/lapack_getrs_test.py
```diff
@@ -0,0 +1,199 @@
...
+    np_dtype = getattr(np, dtype.to_string())
+
+    # this is what we are trying to do, using getrf (LU factorize the matrix a) and getrs (solve the system for b as rhs)
+    a1 = np.array([[1, 2], [3, 5]], dtype=np_dtype)
+    b1 = np.array([1, 2], dtype=np_dtype)
+    x = np.linalg.solve(a1, b1)
+
+    # verify if it works in numpy :)
+    if not np.allclose(np.dot(a1, x), b1):
+        raise ValueError("NumPy solve returned wrong result o_O")
```
tbennun, line 170:
```suggestion
        raise ValueError("NumPy solve returned wrong result")
```

tests/library/linalg_solve_test.py
```diff
@@ -0,0 +1,194 @@
...
+    sdfg.add_array("bout", out_shape, dtype)
+
+    ain = state.add_read("ain")
+    bin = state.add_read("bin")
+    bout = state.add_write("bout")
+
+    solve_node = Solve("solve")
+    solve_node.implementation = implementation
+
+    state.add_memlet_path(ain,
```
tbennun, line 52: could use `.add_edge` here

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+
+        if "commsize" in map_entry.map.range.free_symbols:
+            return False
+        if "Px" in map_entry.map.range.free_symbols:
+            return False
+        if "Py" in map_entry.map.range.free_symbols:
+            return False
+
+        # If the map iterators are used in the code of a Tasklet,
+        # then we cannot flatten them (currently).
```
tbennun, line 49: Doesn't `subgraph.replace` take care of replacing the iterators inside the tasklets too?

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+
+
+    def apply(self, sdfg: dace.SDFG):
+        graph = sdfg.nodes()[self.state_id]
+        map_entry = graph.nodes()[self.subgraph[self._map_entry]]
+        map_exit = graph.exit_node(map_entry)
+
+        sz = dace.symbol('commsize', dtype=dace.int32)
+
+        def _prod(sequence):
```
tbennun, line 125: can be imported from `dace.data` instead

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                for e in graph.edges_between(inp, map_entry):
+                    graph.add_edge(local_access, None, map_entry, e.dst_conn,
+                                   dace.Memlet.from_array(local_name, local_arr))
+                    graph.remove_edge(e)
+                for e in graph.out_edges(map_entry):
+                    if e.data.data == inp.data:
+                        e.data = dace.Memlet.simple(local_name, params[0])
+                continue
+
+            raise NotImplementedError
```
tbennun, line 204: I'd modify the code above to `if`, `elif`, and `else` in this case in order to avoid confusion.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+        for _, _, dst, _, m in graph.out_edges(map_exit):
+            if not isinstance(dst, nodes.AccessNode):
+                raise NotImplementedError
+            desc = dst.desc(sdfg)
+            if not isinstance(desc, data.Array):
+                raise NotImplementedError
+            try:
+                if list(desc.shape) != m.dst_subset.size_exact():
+                    # Second attempt
+                    # TODO: We need a solution for symbols not matching
```
tbennun, line 216: which symbols wouldn't match? Does it have to do with nested SDFGs?

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                               dace.Memlet.from_array(out.data, desc))
+                for e in graph.edges_between(map_exit, out):
+                    graph.add_edge(map_exit, e.src_conn, local_access, None,
+                                   dace.Memlet.from_array(local_name, local_arr))
+                    graph.remove_edge(e)
+                for e in graph.in_edges(map_exit):
+                    if e.data.data == out.data:
+                        e.data = dace.Memlet.simple(local_name, params[0])
+                continue
+            raise NotImplementedError
```
tbennun, line 251: same as the above comment

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+        params = [dace.symbol(p) for p in map_entry.map.params]
+        if len(params) != 2:
+            return False
+
+        if "commsize" in map_entry.map.range.free_symbols:
+            return False
+        if "Px" in map_entry.map.range.free_symbols:
+            return False
+        if "Py" in map_entry.map.range.free_symbols:
+            return False
```
tbennun, line 286: Same question as in the above transformation

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+        return map_entry.map.label + ': ' + str(map_entry.map.params)
+
+
+    def apply(self, sdfg: dace.SDFG):
+        graph = sdfg.nodes()[self.state_id]
+        map_entry = graph.nodes()[self.subgraph[self._map_entry]]
+        map_exit = graph.exit_node(map_entry)
+
+        sz = dace.symbol('commsize', dtype=dace.int32, integer=True, positive=True)
+        Px = dace.symbol('Px', dtype=dace.int32, integer=True, positive=True)
```
tbennun, line 359: to ensure no one else is using these, we could prepend two underscores.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+        Py = dace.symbol('Py', dtype=dace.int32, integer=True, positive=True)
+
+        def _prod(sequence):
```
tbennun, line 362: import from dace.data

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                for e in graph.edges_between(inp, map_entry):
+                    graph.add_edge(local_access, None, map_entry, e.dst_conn,
+                                   dace.Memlet.from_array(local_name, local_arr))
+                    graph.remove_edge(e)
+                for e in graph.out_edges(map_entry):
+                    if e.data.data == inp.data:
+                        e.data.data = local_name
+                continue
+
+            raise NotImplementedError
```
tbennun, line 466: if/elif/else

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+
+        map_entry.map.params = params
+        map_entry.map.range = subsets.Range(ranges)
+
+
+@registry.autoregister_params(singlestate=True, strict=False)
+class RedundantComm2D(pm.Transformation):
+    """ Implements the redundant array removal transformation, applied
+        when a transient array is copied to and from (to another array),
+        but never used anywhere else. """
```
tbennun, line 534: Please add a comment that it has to do with distributed.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+
+    in_array = pm.PatternNode(nodes.AccessNode)
+    gather = pm.PatternNode(nodes.Tasklet)
+    mid_array = pm.PatternNode(nodes.AccessNode)
+    scatter = pm.PatternNode(nodes.Tasklet)
```
tbennun, line 539: These should be the specific library nodes.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                           dace.Memlet.from_array(in_array, in_desc))
+
+        graph.remove_node(gather)
+        graph.remove_node(mid_array)
+        graph.remove_node(scatter)
+        graph.remove_node(out_array)
+
+
+@registry.autoregister_params(singlestate=True)
+class StencilOperation(pm.Transformation):
```
tbennun, line 614: Do we use this transformation?

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                                                            int]) -> str:
+        map_entry = graph.node(candidate[StencilOperation.map_entry])
+        return map_entry.map.label + ': ' + str(map_entry.map.params)
+
+    def apply(self, sdfg: dace.SDFG):
+        pass
+
+
+@registry.autoregister_params(singlestate=True)
+class OuterProductOperation(pm.Transformation):
```
tbennun, line 728: Same as above. I'd rather have those in separate files/PRs if we don't use them now.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                                                            int]) -> str:
+        map_entry = graph.node(candidate[OuterProductOperation.map_entry])
+        return map_entry.map.label + ': ' + str(map_entry.map.params)
+
+    def apply(self, sdfg: dace.SDFG):
+        pass
+
+
+@registry.autoregister_params(singlestate=True)
+class Reduction1Operation(pm.Transformation):
```
tbennun, line 816: Same as above. I'd rather have those in separate files/PRs if we don't use them now.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+    """ Detects reduction1 operations.
```
tbennun, line 817:
```suggestion
    """ Detects reduction operations.
```

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                                                            int]) -> str:
+        map_entry = graph.node(candidate[Reduction1Operation.map_entry])
+        return map_entry.map.label + ': ' + str(map_entry.map.params)
+
+    def apply(self, sdfg: dace.SDFG):
+        pass
+
+
+@registry.autoregister_params(singlestate=True)
+class ReductionNOperation(pm.Transformation):
```
tbennun, line 869: Same as above. I'd rather have those in separate files/PRs if we don't use them now.

tbennun (changes requested):
Cool PR. There is some extra stuff to filter out from here.

dace/codegen/targets/cpu.py
```diff
@@ -228,7 +228,9 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
 
         # Compute array size
         arrsize = nodedesc.total_size
-        arrsize_bytes = arrsize * nodedesc.dtype.bytes
+        arrsize_bytes = 42
```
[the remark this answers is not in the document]
alexnick83: The value `42` has no special meaning here (other than the well-known one). Furthermore, it is not possible to get the size for an opaque type (by definition of it being opaque). However, this is fine since the value is not used when generating code with opaque types. The issue was that at some point, the code generator was trying to access `arrsize_bytes`, resulting in an error. Thus the need for initializing `arrsize_bytes` with some value. This was apparently rectified in another PR, so I removed the (apparently) obsolete initialization line.

dace/libraries/mpi/utils.py
```diff
@@ -0,0 +1,74 @@
+import dace
+
+
+def MPI_DDT(dtype):
```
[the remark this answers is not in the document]
alexnick83: This is replying to all questions related to this file. I don't really want to touch this file further because there might be further developments soon (but it has to be merged so that the current progress becomes available).

dace/transformation/dataflow/map_distribution.py
```diff
@@ -0,0 +1,940 @@
...
+                       sdfg: dace.SDFG,
+                       strict: bool = False):
+
+        map_entry = graph.node(candidate[ElementWiseArrayOperation._map_entry])
+        map_exit = graph.exit_node(map_entry)
+        params = [dace.symbol(p) for p in map_entry.map.params]
+
+        if "commsize" in map_entry.map.range.free_symbols:
+            return False
+        if "Px" in map_entry.map.range.free_symbols:
```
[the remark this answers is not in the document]
alexnick83: These are fixed symbols for defining a 2D process grid. The check disallows the transformation from applying on an already distributed map, which should have these symbols in the range expressions.

dace/transformation/dataflow/map_distribution.py
alexnick83: Probably. I just want to only support clean dataflow and not stuff like dappy tasklets or tasklets with custom code at the moment.

dace/transformation/dataflow/map_distribution.py
alexnick83: I agree that we shouldn't define `prod` again and again, but we should import from some general utility file. Importing from `dace.data` is very strange.

dace/transformation/dataflow/map_distribution.py
alexnick83: It does not work in general. The issue is that the pattern might emerge only after expanding the library nodes, at which point you have tasklets and not library nodes anymore. You can lookup the commit history of the file where I have commented out an unsuccessful attempt at matching the library nodes.

dace/transformation/dataflow/map_distribution.py
alexnick83: No, it is part of the stuff that we left partially implemented for SC, i.e., if it goes now in another PR it will die.

dace/transformation/dataflow/map_distribution.py
alexnick83: Intended

tbennun (approved):

tbennun:
If merging, please squash and merge to avoid extra files being committed.

PR #623, opened 2021-03-03
Subset ambiguity in RedundantArray.

alexnick83:
RedundantArray sets `src_subset` and `dst_subset` instead of `subset` and `other_subset` to avoid ambiguity when the new edge is copying from/to the same array.


tbennun (approved):
Perfect, just did the same for other xforms.

PR #624, opened 2021-03-04
RedundantArray transformations and views

alexnick83:
The purpose of this PR is to enable RedundantArray transformations to remove redundant arrays when they are connected to a view.


tbennun (approved):

tests/transformations/view_redundant_array_test.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
```
tbennun, line 1:
```suggestion
# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
import numpy as np
```

tbennun (approved):

PR #625, opened 2021-03-04
Application of Redundant(Second)Array leads to data races

alexnick83:
The PR attempts to address an issue that occurs with the following code:
```python
@dace.program
def rw_data_race(A: dace.float32[10, 10]):
    A2 = A.copy()
    A[1:-1, 1:-1] = A2[:-2,:-2] + A2[2:,2:]
```
The SDFG with fused states but not redundant copies removed will approximately have the following flow: 
``` 
A --> A2 --> tmp1 --> Map Entry --> i, j -----> Tasklet --> i+1, j+1 --> Map Exit --> A
         --> tmp2 --> Map Entry --> i+2,j+2 --> Tasklet
```
If `tmp1` and `tmp2` are removed, then the only thing holding the "double buffering" scheme is `A2`. However, RedundantSecondArray will trigger and remove it as well. In this case, a data race is produced due to the offsets accesses to A inside the Map. This PR attempts to fix this issue.

### DIFF
--- a/dace/transformation/dataflow/map_fusion.py
+++ b/dace/transformation/dataflow/map_fusion.py
@@ -94,6 +94,7 @@ class MapFusion(transformation.Transformation):
         first_map_exit = graph.nodes()[candidate[MapFusion.first_map_exit]]
         first_map_entry = graph.entry_node(first_map_exit)
         second_map_entry = graph.nodes()[candidate[MapFusion.second_map_entry]]
+        second_map_exit = graph.exit_node(second_map_entry)
 
         for _in_e in graph.in_edges(first_map_exit):
             if _in_e.data.wcr is not None:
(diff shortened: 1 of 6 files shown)


dace/transformation/dataflow/redundant_array.py
```diff
@@ -260,11 +260,25 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
...
+            # RW, WR, or WW dependency.
+            # NOTE: Can we make relax this somehow?
+            accesses = [n for n in graph.nodes()
+                        if isinstance(n, nodes.AccessNode) and
+                        n.desc(sdfg) == in_desc]
+            if len(accesses) > 1:
+                if any(a.access in (
+                        dtypes.AccessType.ReadWrite,
+                        dtypes.AccessType.WriteOnly)
+                       for a in accesses):
```
tbennun, line 279: I'd check for in_degree instead, safer

tbennun (changes requested):

dace/transformation/dataflow/redundant_array.py
alexnick83: Good idea, then it doesn't depend on how it was set by the user or the frontend.

dace/transformation/dataflow/map_fusion.py
```diff
@@ -180,6 +181,53 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
...
+
+            output_accesses = [graph.memlet_path(e)[0].data.dst_subset
+                               for e in graph.in_edges(second_map_exit)
+                               if e.data.data in common_data]
+
+            # Compute output accesses with respect to first map's symbols
+            oacc_permuted = [
+                dcpy(a).replace({
+                    symbolic.pystr_to_symbolic(k): symbolic.pystr_to_symbolic(v)
+                    for k, v in params_dict.items()
```
tbennun, line 217: this may be bad with replacing symbols with the same name, e.g., `{i: j, j: i}`. For these cases we make a two-stage replacement where we replace the keys with temporary `__dacesym_{value}` and then replace the values with the final value.

dace/transformation/dataflow/redundant_array.py
```diff
@@ -154,6 +156,63 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
...
+            # write access. Therefore, there might be a RW, WR, or WW dependency.
+            accesses = [n for n in graph.nodes()
+                        if isinstance(n, nodes.AccessNode) and
+                        n.desc(sdfg) == out_desc and n is not out_array]
+            if len(accesses) > 0:
+                # We need to ensure that a data race will not happen if we
+                # remove in_array.
+                # First, we simplify the graph (see transient_reuse)
+                # Copy the whole graph
+                G = nx.MultiDiGraph()
```
tbennun, line 170: could we refactor this (graph creation part) to a separate function in dace.transformation.helpers or dace.sdfg.utils?

tests/transformations/mapfusion_data_races_test.py
```diff
@@ -0,0 +1,24 @@
...
+
+@dace.program
+def rw_data_race(A: dace.float32[10, 10], B: dace.float32[10, 10]):
+    tmp = A[:-1,:-1] + A[1:,1:]
+    B[1:-1, 1:-1] = tmp[:-1,:-1] + tmp[:-1,:-1]
+
+
+def test_rw_data_race():
+    sdfg = rw_data_race.to_sdfg(strict=True)
+    sdfg.apply_transformations_repeated([MapFusion])
```
tbennun, line 17:
```suggestion
    sdfg.apply_transformations_repeated(MapFusion)
```

tbennun (changes requested):
minor changes

dace/transformation/dataflow/map_fusion.py
alexnick83: Hmm, I copied this from the same transformation, a few lines above. I will check if I have to change both.

tbennun (approved):

Issue #575, opened 2021-02-13
RedundantSecondArray fails with dimension mismatch, missing SubgraphFusion


phschaad:
Removing redundant transients via `RedundantSecondArray` fails with a dimension mismatch, even if the dimensions would allow removal. Elements in question are `softmax__tmp7` and `softmax_tmp_max` with shapes `[H, B, SM]` and `[H, B, SM, 1]`, respectively:

![subgraph_fusion](https://user-images.githubusercontent.com/9193712/107848871-4fe9bc80-6df7-11eb-9873-428f83b74dae.gif)

This additionally makes it impossible to fuse the two shown maps, even though their ranges are the same in all dimensions.

### To reproduce:
This exact scenario can be found in the SDFG generated from BERT's encoder via [spcl/substation](https://github.com/spcl/substation/blob/master/substation/transformer_sdfg.py).

alexnick83:
@phschaad Can you test if the issue has been resolved in the latest master?

PR #626, opened 2021-03-05
Better subset composition

alexnick83:
This PR attempts to address issue #575. The fix is done through `Redundant(Second)Array` instead of `Range.compose`, because it becomes impossible to infer the correct composition when reaching the innermost Memlets, where subsets are `Indices`.

### DIFF
--- a/dace/libraries/standard/nodes/reduce.py
+++ b/dace/libraries/standard/nodes/reduce.py
@@ -89,6 +89,8 @@ class ExpandReducePure(pm.ExpandTransformation):
             nstate = nsdfg.add_state()
         # END OF INIT
 
+        keepdims = input_dims == output_dims
+
         # (If axes != all) Add outer map, which corresponds to the output range
         if len(axes) != input_dims:
             # Interleave input and output axes to match input memlet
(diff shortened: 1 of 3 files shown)


dace/libraries/standard/nodes/reduce.py
```diff
@@ -1,870 +1,870 @@
...
-
-    def __label__(self, sdfg, state):
-        return str(self).replace(' Axes', '\nAxes')
-
-    def validate(self, sdfg, state):
-        if len(state.in_edges(self)) != 1:
-            raise ValueError('Reduce node must have one input')
-        if len(state.out_edges(self)) != 1:
-            raise ValueError('Reduce node must have one output')
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
```
alexnick83, line 1: @tbennun There is no change in this file (reduce.py). It is marked as changed due to an issue with git and VS code on my WSL installation.

dace/transformation/dataflow/redundant_array.py
```diff
@@ -200,11 +250,36 @@ def apply(self, sdfg):
...
+                bset, popped = pop_dims(b_subset, b_dims_to_pop)
+
+        reduction = False
+        for e in graph.in_edges(in_array):
+            if isinstance(e.src, Reduce):
+                reduction = True
+
         # If the memlet does not cover the removed array, create a view.
-        if any(m != a for m, a in zip(a1_subset.size(), in_desc.shape)):
+        if reduction or any(m != a
```
alexnick83, line 273: @tbennun Due to the improvements in RedundantArray, we can now automatically remove the transient between a Reduce node and its output when `keepdims=True`. However, the Reduce node does not handle this case well (the produced code is wrong). I found how to fix it in the pure expansion, but I guess that the same problem exists in all other implementations. Therefore, I opted to put a special case here; just change the transient to a view.

dace/transformation/dataflow/redundant_array.py
```diff
@@ -200,11 +250,36 @@ def apply(self, sdfg):
...
+                            for m, a in zip(a1_subset.size(), in_desc.shape)):
+            view_strides = in_desc.strides
+            if (b_dims_to_pop and len(b_dims_to_pop) == len(out_desc.shape) -
+                    len(in_desc.shape)):
+                view_strides = [s for i, s in enumerate(out_desc.strides)
```
alexnick83, line 278: @tbennun If you want a sample where `view_strides` should not be `in_desc.strides`, look at conv2d (`[N, C_out] -> [N, i, j, C_out]`).

dace/transformation/dataflow/redundant_array.py
```diff
@@ -7,13 +7,13 @@
...
 import typing
 
-
 from dace import data, registry, subsets, dtypes, memlet as mm
 from dace.sdfg import nodes, SDFGState
 from dace.sdfg import utils as sdutil
 from dace.sdfg import graph
 from dace.transformation import transformation as pm, helpers
 from dace.config import Config
+from dace.libraries.standard import Reduce
```
tbennun, line 16: Please move this into the function, it's a very slow import

tbennun (changes requested):
The redundant array transformation is becoming a monster, but if all the tests pass then ok.

tbennun (approved):

PR #628, opened 2021-03-06
Intel fpga codegen pow

TizianoDeMatteis:
When a tasklet uses `pow`, it is code generated by using dace namespaces `dace::math::pow`
OpenCL does not support namespaces, therefore we intercept this in OpenCLDaceKeywordRemover before actually generating the code.

### DIFF
--- a/dace/codegen/targets/intel_fpga.py
+++ b/dace/codegen/targets/intel_fpga.py
@@ -1497,19 +1497,19 @@ class OpenCLDaceKeywordRemover(cpp.DaCeKeywordRemover):
 
         return ast.copy_location(updated, node)
 
-    def visit_BinOp(self, t):
-        if t.op.__class__.__name__ == 'Pow':
+    def visit_BinOp(self, node):
+        if node.op.__class__.__name__ == 'Pow':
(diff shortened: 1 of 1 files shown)


dace/codegen/targets/intel_fpga.py
```diff
@@ -1497,6 +1497,20 @@ def visit_Assign(self, node):
 
         return ast.copy_location(updated, node)
 
+    def visit_BinOp(self, node):
+        if node.op.__class__.__name__ == 'Pow':
```
definelicht, line 1501: Is pow the only math function we could encounter? Can't we look for any math function?

TizianoDeMatteis: No, is not the only one. But I was mimicking cppunparse, where Pow is a very special case and the `dace::math` string is printed out (cppunparse is called after OCL Keyword remover) https://github.com/spcl/dace/blob/b727e198e6fa8a8adf68b4c213365555ef1edef6/dace/codegen/cppunparse.py#L896

definelicht (approved):

PR #633, opened 2021-03-08
Nesting and Inline fixes

TizianoDeMatteis:
- When inlining, if inner/outer strides are equal, don't check for unsqueezing, that could add unneeded views
- When Nesting, removes from the parent SDFG the symbols that are defined in the nested one

### DIFF
--- a/dace/transformation/interstate/sdfg_nesting.py
+++ b/dace/transformation/interstate/sdfg_nesting.py
@@ -1055,3 +1055,17 @@ class NestSDFG(transformation.Transformation):
             outer_state.add_edge(
                 nested_node, val, arrnode, None,
                 memlet.Memlet.from_array(key, arrnode.desc(outer_sdfg)))
+
+        # Remove from the parent SDFG the symbols that are defined in the nested one
+        defined_syms = set()
+
+        for name, desc in nested_sdfg.arrays.items():
+            defined_syms.add(name)
+
+        for e in nested_sdfg.edges():
+            defined_syms |= set(e.data.new_symbols({}).keys())
+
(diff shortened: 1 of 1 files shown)


dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -72,9 +72,17 @@ def _check_strides(inner_strides: List[symbolic.SymbolicType],
         ostrides = [
             os for i, os in enumerate(outer_strides) if i not in dims_to_ignore
         ]
+
+        istrides = [
+            os for i, os in enumerate(inner_strides) if i not in dims_to_ignore
+        ]
```
tbennun, line 78: the dims to ignore are for the outer memlet. I don't see how this works here

TizianoDeMatteis: Maybe I can do the  same, but considering  the inner memlet?

dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -1049,3 +1055,17 @@ def apply(self, sdfg):
...
+        for name, desc in nested_sdfg.arrays.items():
+            defined_syms.add(name)
+
+        for e in nested_sdfg.edges():
+            defined_syms |= set(e.data.new_symbols({}).keys())
+
+        defined_syms |= set(nested_sdfg.constants.keys())
+
+        for s in defined_syms:
+            outer_sdfg.symbols.pop(s, None)
```
tbennun, line 1071: do we add them to the symbols of the nested SDFG a-la nest_state_subgraph? otherwise types can be erased

tbennun (approved):

PR #634, opened 2021-03-08
RedundantArray transformations copy WCR to new edge and memlet path

alexnick83:
What the title says pretty much. If the matched edge `A ----> B` has WCR, then the new edge `A0 ----> B` or `A ----> B0` (depending on which array is removed) and the corresponding memlet path must inherit the WCR.

I am also addressing in this PR a small bug in MapFusion, an issue with SDFG.free_symbols considering nested SDFG defined symbols to be undefined, and an issue where popped indices in RedundantArray are not pushed back in the correct order, an output shape issue with MatMul (GEMM), and an issue with MatMul output edge having WCR.

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -937,12 +937,6 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
         # Add free state symbols
         for state in self.nodes():
             free_syms |= state.free_symbols
-            # Add defined nested SDFG symbols
-            for n in state.nodes():
-                if isinstance(n, dace.nodes.NestedSDFG):
-                    for s in n.sdfg.symbols:
-                        if s not in n.symbol_mapping.keys():
-                            defined_syms.add(s)
 
         # Add free inter-state symbols
         for e in self.edges():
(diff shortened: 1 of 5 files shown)


dace/sdfg/sdfg.py
```diff
@@ -937,6 +937,12 @@ def free_symbols(self) -> Set[str]:
         # Add free state symbols
         for state in self.nodes():
             free_syms |= state.free_symbols
+            # Add defined nested SDFG symbols
+            for n in state.nodes():
+                if isinstance(n, dace.nodes.NestedSDFG):
+                    for s in n.sdfg.symbols:
+                        if s not in n.symbol_mapping.keys():
+                            defined_syms.add(s)
```
tbennun, line 945: That does not make sense to me. The free symbols in the nested SDFG, if not defined in symbol mapping, should not propagate upwards to the higher SDFG

tbennun (changes requested):
All approved but the free_symbols change

dace/sdfg/sdfg.py
alexnick83: Reverted.

tbennun (approved):

Issue #612, opened 2021-03-02
Fully Support views in FPGA


TizianoDeMatteis:
**Describe the bug**
With certain patterns, views do not produce right code for FPGAs.
Furthermore, views are often introduced by `InlineSDFG`, causing problems to execute various FPGA transformed SDFG

**To Reproduce**
For example, a simple SDFG with something like `ArrayA -> View -> ArrayB`, can not be transformed to FPGA

PR #636, opened 2021-03-09
Support FPGA view

TizianoDeMatteis:
This PR provides adds additional support for views in FPGA (solves #612):
- deals with views while nesting:
    - views are kept as transient
    - their storage type is the same as the viewed node
    - generates proper device code (for example, views with Global storage must not be generated on the host, or appear as kernel parameters)
- special attention for Intel FPGA codegeneration
-  adds tests for view on FPGA

### DIFF
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -398,8 +398,8 @@ class FPGACodeGen(TargetCodeGenerator):
         if isinstance(nodedesc, dace.data.View):
             return self._cpu_codegen.allocate_view(sdfg, dfg, state_id, node,
                                                    function_stream,
-                                                   callsite_stream,
-                                                   callsite_stream)
+                                                   declaration_stream,
+                                                   allocation_stream)
         elif isinstance(nodedesc, dace.data.Stream):
 
             if not self._in_device_code:
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -274,6 +274,7 @@ class FPGACodeGen(TargetCodeGenerator):
                     # transient (inner-level inputs/outputs are just connected
                     # to data in the outer layers, whereas transients can be
                     # independent).
+                    # Views are not nested global transients
                     if scope == subgraph or n.desc(scope).transient:
                         if scope.out_degree(n) > 0:
                             candidates.append((False, n.data, n.desc(scope)))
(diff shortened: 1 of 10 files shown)


dace/transformation/interstate/fpga_transform_state.py
```diff
@@ -77,7 +77,7 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
                     return False
 
                 # Streams cannot be unbounded on FPGA
-                if nodedesc.buffer_size < 1:
+                if ndedesc.buffer_size < 1:
```
definelicht, line 80:
```suggestion
                if nodedesc.buffer_size < 1:
```

definelicht (changes requested):
Great work! This is exactly what I imagined :-)

dace/codegen/targets/fpga.py
```diff
@@ -386,20 +391,28 @@ def generate_scope(self, sdfg, dfg_scope, state_id, function_stream,
...
+
         result_decl = StringIO()
         result_alloc = StringIO()
         nodedesc = node.desc(sdfg)
         arrsize = nodedesc.total_size
         is_dynamically_sized = dace.symbolic.issymbolic(arrsize, sdfg.constants)
-
         dataname = node.data
+
+        if not isinstance(nodedesc, dace.data.Stream):
```
tbennun, line 402: could you explain?

dace/transformation/interstate/fpga_transform_state.py
tbennun, line 80: oops!

tbennun (commented):
one typo and a question, but looks fine to me otherwise

dace/codegen/targets/fpga.py
TizianoDeMatteis: Part of the codegen relies on the fact that allocate_array will return if the array has been already allocated.
This was not done on the FPGA backend, so I added it.
There is this special case for Stream, because we need to keep track of the names of the streams across nested SDFG for Intel FPGA channel mangling (essentially we need to call allocate_stream for this)

I can also leave it like it was before, without returning if the array has been already allocated

dace/transformation/interstate/fpga_transform_state.py
TizianoDeMatteis: Thanks

dace/codegen/targets/fpga.py
tbennun: This makes sense, but a comment (with this content) would be welcome and helpful :)

TizianoDeMatteis: Will add it!

tbennun (approved):

definelicht (approved):

PR #640, opened 2021-03-10
Consistent symbols

alexnick83:
This PR attempts to address the fact that symbols are being converted to string and back to symbols again in the whole DaCe stack. This is a serious issue because important symbolic assumptions are lost. A characteristic example is the NPBench benchmark `stockham_fft` which has symbolic sizes with either powers or logarithms. This sample can only work with the array sizes being explicitly defined as positive integers and the loop-indices implicitly defined as non-negative integers.

### DIFF
--- a/dace/frontend/python/memlet_parser.py
+++ b/dace/frontend/python/memlet_parser.py
@@ -100,12 +100,12 @@ def _fill_missing_slices(das, ast_ndslice, array, indices):
             try:
                 if (rb < 0) == True:
                     rb += array.shape[indices[idx]]
-            except:
+            except (TypeError, ValueError):
                 pass
             try:
                 if (re < 0) == True:
                     re += array.shape[indices[idx]]
-            except:
+            except (TypeError, ValueError):
                 pass
             ndslice[idx] = (rb, re, rs)
             offsets.append(idx)
--- a/dace/frontend/python/memlet_parser.py
+++ b/dace/frontend/python/memlet_parser.py
@@ -72,6 +72,13 @@ def _ndslice_to_subset(ndslice):
                 if not is_tuple[i]:
                     ndslice[i] = (ndslice[i], ndslice[i], 1)
         return subsets.Range(ndslice)
+    
+
+def _parse_dim_atom(das, atom):
+    result = pyexpr_to_symbolic(das, atom)
+    if isinstance(result, data.Data):
+        return pystr_to_symbolic(astutils.unparse(atom))
+    return result
 
 
 def _fill_missing_slices(das, ast_ndslice, array, indices):
@@ -85,14 +92,21 @@ def _fill_missing_slices(das, ast_ndslice, array, indices):
     has_ellipsis = False
     for dim in ast_ndslice:
         if isinstance(dim, tuple):
-            rb = pyexpr_to_symbolic(das, dim[0] or 0)
-            re = pyexpr_to_symbolic(das, dim[1]
+            rb = _parse_dim_atom(das, dim[0] or 0)
+            re = _parse_dim_atom(das, dim[1]
                                     or array.shape[indices[idx]]) - 1
-            rs = pyexpr_to_symbolic(das, dim[2] or 1)
-            if (rb < 0) == True:
-                rb += array.shape[indices[idx]]
-            if (re < 0) == True:
-                re += array.shape[indices[idx]]
+            rs = _parse_dim_atom(das, dim[2] or 1)
+            # NOTE: try/except for cases where rb/re are not symbols/numbers
+            try:
+                if (rb < 0) == True:
+                    rb += array.shape[indices[idx]]
+            except:
+                pass
+            try:
+                if (re < 0) == True:
+                    re += array.shape[indices[idx]]
+            except:
+                pass
             ndslice[idx] = (rb, re, rs)
             offsets.append(idx)
             idx += 1
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -912,22 +912,30 @@ class TaskletTransformer(ExtNodeTransformer):
         if isinstance(node, ast.Name):
             actual_node = copy.deepcopy(node)
             actual_node.id = name
-            rng = dace.subsets.Range(
-                astutils.subscript_to_slice(actual_node, {
-                    **self.sdfg.arrays,
-                    **self.scope_arrays
-                })[1])
+            expr: MemletExpr = ParseMemlet(
+                self, {**self.sdfg.arrays, **self.scope_arrays, **self.defined},
+                actual_node)
+            rng = expr.subset
+            # rng = dace.subsets.Range(
+            #     astutils.subscript_to_slice(actual_node, {
+            #         **self.sdfg.arrays,
+            #         **self.scope_arrays
+            #     })[1])
         elif isinstance(node, ast.Subscript):
             actual_node = copy.deepcopy(node)
             if isinstance(actual_node.value, ast.Call):
                 actual_node.value.func.id = name
             else:
                 actual_node.value.id = name
-            rng = dace.subsets.Range(
-                astutils.subscript_to_slice(actual_node, {
-                    **self.sdfg.arrays,
-                    **self.scope_arrays
-                })[1])
+            expr: MemletExpr = ParseMemlet(
+                self, {**self.sdfg.arrays, **self.scope_arrays, **self.defined},
+                actual_node)
+            rng = expr.subset
+            # rng = dace.subsets.Range(
+            #     astutils.subscript_to_slice(actual_node, {
+            #         **self.sdfg.arrays,
+            #         **self.scope_arrays
+            #     })[1])
         elif isinstance(node, ast.Call):
             rng = dace.subsets.Range.from_array({
                 **self.sdfg.arrays,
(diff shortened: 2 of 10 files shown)


dace/sdfg/sdfg.py
```diff
@@ -157,8 +157,29 @@ def replace(self, name: str, new_name: str, replace_keys=True) -> None:
...
+            if new_idx + length < len(condition):
+                if re.match("[A-Za-z0-9_]+$", condition[new_idx + length]):
+                    matched = False
+            last_idx, new_idx = new_idx, condition.find(name, new_idx + length)
+            if matched:
+                new_condition += new_name
+                last_idx += length
+        if last_idx < len(condition):
+            new_condition += condition[last_idx:]
+        self.condition.as_string = new_condition
```
tbennun, line 182: shouldn't we just convert the condition to AST (we do this just above for assignments) and then use ASTFindAndReplace?

dace/symbolic.py
```diff
@@ -854,3 +854,43 @@ def persistent_load(self, pid):
...
+
+def inequal_symbols(a: Union[sympy.Expr, Any],
+                    b: Union[sympy.Expr, Any]) -> bool:
+    """ Compares 2 symbolic expressions and returns True if they are not equal.
+    """
+    if not isinstance(a, sympy.Expr) or not isinstance(b, sympy.Expr):
+        return a != b
+    else:
+        a, b = equalize_symbols(a, b)
+        return (a - b).simplify() != 0
```
tbennun, line 896: do we need the simplify here? why not ask `a != b`? please leave a comment in the code too (this is one of the things that can be easily forgotten).

dace/symbolic.py
```diff
@@ -854,3 +854,43 @@ def persistent_load(self, pid):
...
+        repldict = dict()
+        for name in common_names:
+            repldict[b_syms[name]] = a_syms[name]
+        b = b.subs(repldict)
+    return a, b
+
+
+def inequal_symbols(a: Union[sympy.Expr, Any],
+                    b: Union[sympy.Expr, Any]) -> bool:
+    """ Compares 2 symbolic expressions and returns True if they are not equal.
```
tbennun, line 890:
```suggestion
    """
    Compares 2 symbolic expressions and returns True if they are not equal.
```

dace/transformation/dataflow/redundant_array.py
```diff
@@ -49,10 +49,15 @@ def _validate_subsets(edge: graph.MultiConnectorEdge,
...
+                src_expr = src_subset.num_elements()
+                src_expr_exact = src_subset.num_elements_exact()
+                dst_expr = dst_subset.num_elements()
+                dst_expr_exact = dst_subset.num_elements_exact()
+                if (src_expr != dst_expr and symbolic.inequal_symbols(
+                        src_expr_exact, dst_expr_exact)):
                     raise ValueError(
-                        "Source subset is missing (dst_subset: {}, "
-                        "src_shape: {}".format(dst_subset, desc.shape))
+                        "Destination subset is missing (src_subset: {}, "
```
tbennun, line 59: what happened here with the exception?

dace/transformation/dataflow/redundant_array.py
```diff
@@ -187,15 +197,59 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
...
             # In strict mode, make sure the memlet covers the removed array
-            edge = graph.edges_between(in_array, out_array)[0]
-            subset = copy.deepcopy(edge.data.subset)
+            subset = copy.deepcopy(e1.data.subset)
             subset.squeeze()
             shape = [sz for sz in in_desc.shape if sz != 1]
             if any(m != a for m, a in zip(subset.size(), shape)):
                 return False
+
+            # NOTE: Library node check
```
tbennun, line 213: with this code, can we remove the reduce check below?

tbennun (changes requested):

dace/frontend/python/memlet_parser.py
```diff
@@ -85,14 +92,21 @@ def _fill_missing_slices(das, ast_ndslice, array, indices):
...
-            if (rb < 0) == True:
-                rb += array.shape[indices[idx]]
-            if (re < 0) == True:
-                re += array.shape[indices[idx]]
+            rs = _parse_dim_atom(das, dim[2] or 1)
+            # NOTE: try/except for cases where rb/re are not symbols/numbers
+            try:
+                if (rb < 0) == True:
+                    rb += array.shape[indices[idx]]
+            except:
```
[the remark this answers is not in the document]
alexnick83: I think it is ValueError?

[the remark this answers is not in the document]
alexnick83: No +/- operation defined for 'Data' and 'int', something like this was the issue

dace/frontend/python/newast.py
```diff
@@ -798,7 +809,7 @@ def _add_access(
             sym_rng = []
             for i, r in enumerate(rng):
                 for s, sr in self.symbols.items():
-                    if s in symbolic.symlist(r).keys():
+                    if s in symbolic.symlist(r).values():
```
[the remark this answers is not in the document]
alexnick83: Let me clarify. Originally, `ProgramVisitor.symbols` was `Dict[str, Range]`, where the key was the symbol name, similar to `SDFG.symbols`. To make symbols consistent, this was changed to `Dict[Symbol, Range]`. Now the key is the Symbol object itself. `symbolic.symlist(Range)` return `Dict[str, Symbol]`, where the key is the symbol name and the value is the Symbol object. In the above snipper, `s` is now the Symbol object, therefore, I have to look into the values of the dictionary.

dace/symbolic.py
alexnick83: I don't understand the suggestion? Is it to move the text one line below? Why?

dace/symbolic.py
alexnick83: From SymPy's documentation (https://docs.sympy.org/latest/tutorial/gotchas.html):
```python
a = (x + 1)**2
b = x**2 + 2*x + 1
simplify(a - b)
0
```

alexnick83: And yes we need the `simplify` to try to remove `ceiling`, `floor`, etc. when the assumptions allow it

dace/transformation/dataflow/redundant_array.py
alexnick83: If you mean about the message, it is obviously a copy-paste issue. It is exactly the same code as below, but the message should be different.

dace/transformation/dataflow/redundant_array.py
alexnick83: The reduce check is for a different issue. It is related to reductions with the `keepdims` keyword enabled. To generate the correct SDFG, we first generate the reduction ignoring `keepdims` and then we copy to the final output. We are not checking any relation between the input and the output. Also the new library node check does not ensure that the "intermediate" reduction output in the case of `keepdims=True` is not removed.

dace/symbolic.py
alexnick83: (`stockham_fft` is triggering all these issues)

dace/frontend/python/memlet_parser.py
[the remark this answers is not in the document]
tbennun: yes, I mean that this is an exception that is too wide (takes things e.g., KeyboardInterrupt as well). this is disallowed in our coding guidelines. I was proposing to replace it with `except (TypeError, ValueError):`.

dace/transformation/dataflow/redundant_array.py
alexnick83: Let me also clarify that the problem with `keepdims=True` is that our Reduce node expansions are not compatible with writing directly to the final keepdims-enabled output. And we have to many reduce expansions to fix them now.

dace/symbolic.py
tbennun: ok, thanks. please add a short comment that it simplifies the ceiling expressions etc.

dace/symbolic.py
tbennun: just a matter of consistent style.
```
"""
doc
:param x:
:param y:
:return:
:note:
:raises z:
"""
```

dace/transformation/dataflow/redundant_array.py
tbennun: yes

dace/transformation/dataflow/redundant_array.py
tbennun: ok

dace/sdfg/sdfg.py
alexnick83: I will give it a shot

alexnick83: Done

dace/symbolic.py
alexnick83: Done

dace/symbolic.py
alexnick83: Done

dace/transformation/dataflow/redundant_array.py
alexnick83: Done

dace/frontend/python/memlet_parser.py
```diff
@@ -85,14 +92,21 @@ def _fill_missing_slices(das, ast_ndslice, array, indices):
...
+                pass
+            try:
+                if (re < 0) == True:
+                    re += array.shape[indices[idx]]
+            except:
```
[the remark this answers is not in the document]
alexnick83: Done

dace/frontend/python/memlet_parser.py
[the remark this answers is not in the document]
alexnick83: Done

tbennun (approved):
LGTM

PR #645, opened 2021-03-15
WIP: AutoOpt Fusion

sancierra:
Enhance AutoOpt with SubgraphFusion. Includes all the tools necessary as well as extensions to Subgraph Fusion


tbennun:
Merged implicitly by adding it in parts.

PR #653, opened 2021-03-19
Fix for views in intel fpga

TizianoDeMatteis:
- fixes view for Intel FPGA (some cases were not covered)
- when there is a container-container copy, on the same container, generates dependencies pragma  when it is safe to assume that there is no dependency


definelicht (approved):
Make sure the `diode/webclient` update isn't doing something it shouldn't be doing

PR #654, opened 2021-03-19
MapFusion: Use FPGA registers if we are on FPGA

TizianoDeMatteis:


dace/transformation/dataflow/map_fusion.py
```diff
@@ -488,18 +498,27 @@ def fuse_nodes(self,
         # Add intermediate memory between subgraphs. If a scalar,
         # uses direct connection. If an array, adds a transient node
         if edge.data.subset.num_elements() == 1:
+            node_desc = access_node.desc(graph)
+
+            # If we are on FPGA, use FPGA registers
+            if node_desc.storage == dtypes.StorageType.FPGA_Global or node_desc.storage == dtypes.StorageType.FPGA_Local:
```
tbennun, line 504: add parentheses so that yapf can fix this

tbennun (changes requested):
I don't see why we need a separate case for FPGA registers

PR #656, opened 2021-03-22
Add try_parse_state_struct method to CompiledSDFG

orausch:


tests/parse_state_struct_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+    sdfg: dace.SDFG = persistent_transient.to_sdfg()
+    sdfg.apply_gpu_transformations()
+
+    A = np.random.randn(3, 3).astype(np.float32)
+    B = np.random.randn(3, 5).astype(np.float32)
+    compiledsdfg = sdfg.compile()
+    compiledsdfg.initialize()
+
+    state_struct = ctypes.cast(
```
tbennun, line 90: would be nice to have `compiledsdfg.get_state_struct()` that wraps around this

tests/parse_state_struct_test.py
```diff
@@ -0,0 +1,98 @@
...
+        compiledsdfg._libhandle,
+        ctypes.POINTER(compiledsdfg.try_parse_state_struct()))
+    transient_gpu_ptr = state_struct.contents.persistent_transient
```
tbennun, line 93: Please document what's happening here (fill state struct with data from B etc.)

tests/parse_state_struct_test.py
```diff
@@ -0,0 +1,98 @@
...
+                    ctypes.c_void_p(gpu_ptr),
+                    ctypes.c_void_p(numpy_array.__array_interface__["data"][0]),
+                    ctypes.c_size_t(size)))
+            if result.value != 0:
+                raise ValueError("host_to_gpu returned nonzero result!")
+
+    return CudaHelper()
+
+
+@pytest.mark.gpu
```
tbennun, line 71: we don't even need this test to be on the GPU, but assuming you want to test the cublasHandle existence :)
please confirm

orausch: Well there's no cublasHandle in this case because blas isn't used, but that's a good idea for a test.

I made the test on GPU because that's closer to how I want to use it in daceml

tbennun (approved):

PR #657, opened 2021-03-22
Remove ArrayInterface in favor of multiple naked pointers

definelicht:
This is part of the effort to make the code generator produce more "clean" C-code to aid the compilers, triggered by issues arising in Vitis HLS with certain patterns using `ArrayInterface`.

I'm not claiming that this fix is beautiful. Suggestions welcome!


dace/codegen/targets/cpu.py
```diff
@@ -189,14 +190,24 @@ def allocate_view(self, sdfg: SDFG, dfg: SDFGState, state_id: int,
...
                                            global_stream, allocation_stream)
 
+        # When emitting ArrayInterface, we need to know if this is a read or
+        # write variation
+        if viewed_dnode in dfg.source_nodes():
+            is_write = False
+        elif viewed_dnode in dfg.sink_nodes():
+            is_write = True
+        else:
+            is_write = None
```
TizianoDeMatteis, line 200: Could this occur? If this is set to `None` then it may raise an exception in `cpp.py`

TizianoDeMatteis (commented):
Only one minor comment, otherwise it looks good to me.
Thanks!

dace/codegen/targets/cpu.py
definelicht: I'm not sure, not in any SDFG we've written until now at least. A slightly more robust solution would be to compare all sink/source nodes against `view_dnode.data` instead?

TizianoDeMatteis:
@definelicht another thing comes up in my mind: shouldn't you deal with `_in` and `_out` also in `write_and_resolve_expr` (xilinx.py) ?

definelicht:
> 
> 
> @definelicht another thing comes up in my mind: shouldn't you deal with `_in` and `_out` also in `write_and_resolve_expr` (xilinx.py) ?

Is there a test for this? Doing conflict resolution on off-chip memory is really, really disgusting :D But I can see that it might happen in an unoptimized sample...

TizianoDeMatteis (approved):
Thanks!

PR #660, opened 2021-03-25
More ArrayInterface Fixes

definelicht:
Centralize generation of references to `ArrayInterface` to make sure we always follow the same procedure when possible.


TizianoDeMatteis (approved):
Looks good to me!

PR #673, opened 2021-04-01
Codegen tasklet pointer issue

tbennun:
- [x] Fix tasklet pointers and dynamic memlets in codegen (dereferencing conditions)
- [x] Allow users to define variables in tasklets via annotated statements (`x: dace.float32`)
- [x] Simpler API for library node expansions (`@register_expansion`) that allows using multiple files for library nodes and expansions without circular imports.
- [x] Make memlet propagation more strict in un-propagatable cases: when one dimension fails to propagate, it no longer affects other dimensions that don't need to propagate.


dace/codegen/targets/cpp.py
```diff
@@ -1118,7 +1118,8 @@ def visit_Name(self, node: ast.Name):
         if name not in self.memlets:
             return self.generic_visit(node)
         memlet, nc, wcr, dtype = self.memlets[name]
-        if isinstance(dtype, dtypes.pointer):
+        if (isinstance(dtype, dtypes.pointer)
+                and memlet.subset.num_elements() == 1):
             return ast.Name(id="(*{})".format(name), ctx=node.ctx)
```
orausch, line 1123: How will this interact with GPU tasklets that are not devicelevel (e.g. ORT or cuBLAS)?
Dereferencing the pointer of a GPU_Global input would produce an error, right?

tbennun: This is meant for Python tasklets, where the behavior is actually to dereference (as Python has no notion of pointers)

codecov.yml
```diff
@@ -13,4 +13,4 @@ coverage:
 
 codecov:
     notify:
-        after_n_builds: 7
+        after_n_builds: 8
```
definelicht, line 16: What is this magical number?

tests/codegen/dynamic_memlet_test.py
```diff
@@ -0,0 +1,81 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
```
definelicht, line 1: Please add a brief description of what this is testing

tests/codegen/dynamic_memlet_test.py
```diff
@@ -0,0 +1,81 @@
...
+def test_dynamic_memlets():
+    sdfg = dace.SDFG('test')
+    state = sdfg.add_state('state')
+    sdfg.add_array('out_arr1', dtype=dace.float64, shape=(3, 3))
+    sdfg.add_array('out_arr2', dtype=dace.float64, shape=(3, 3))
+    tasklet = state.add_tasklet('tasklet',
+                                inputs={},
+                                outputs={'o1', 'o2'},
+                                code="o1 = 1.0\no2=2*o1")
+    map_entry, map_exit = state.add_map('map', ndrange=dict(i="0:3", j="0:3"))
```
definelicht, line 15: Mixing single and double quotes ;-)

codecov.yml
tbennun: The number of build submitted to codecov.io before it sends a notification here on the GitHub PR thread

definelicht (approved):

PR #681, opened 2021-04-06
Memory bank assignment for nested SDFGs

definelicht:
- Move memory bank assignment to kernel generation instead of array allocation.
- Make sure that bank assignments are consistent
- Make sure that bank assignment is present in the outermost scope, where the node is allocated


TizianoDeMatteis (approved):
Looks good to me!

PR #682, opened 2021-04-06
Parallelize Xilinx tests

definelicht:
Translate `xilinx_test.sh` into `xilinx_test.py` so we can run `multiprocessing.starmap` on our tests.

Time for running Xilinx tests reduced from ~27 minutes to ~11 minutes.


definelicht:
@TizianoDeMatteis and @tbennun: thoughts before I sink more time into this?

tbennun:
> @TizianoDeMatteis and @tbennun: thoughts before I sink more time into this?

shouldn't we aim to use pytest (actually parallel pytest) for this?

definelicht:
Unfortunately I don't know pytest. Does this play nicely with the non-Python things we have to run?

definelicht:
> > @TizianoDeMatteis and @tbennun: thoughts before I sink more time into this?
> 
> shouldn't we aim to use pytest (actually parallel pytest) for this?

That would make this a much bigger task since we need to change every single test, so it depends on whether we want to do this now or `TODO(later)`?

tbennun:
Fine by me

TizianoDeMatteis:
+1 on doing this later

definelicht:
@TizianoDeMatteis By +1 you mean for doing *this* now, and maybe pytest later? Or for waiting with everything until later?

TizianoDeMatteis:
Doing it later in any case. But if you have done already some progress, it is of course fine by me

.github/workflows/python-package.yml
```diff
@@ -124,11 +124,11 @@ jobs:
     - name: Run Xilinx tests
       run: |
         export NOSTATUSBAR=1
-        export DACE_testing_single_cache=1
+        export DACE_testing_single_cache=0
         export COVERAGE_RCFILE=`pwd`/.coveragerc
         export PYTHON_BINARY="coverage run --source=dace --parallel-mode"
         . /opt/setupenv
-        ./tests/xilinx_test.sh 0
+        ./tests/xilinx_test.py
```
tbennun, line 131: I think we should explicitly call it with `python` here (I remember env being an issue in some systems)

tbennun (changes requested):

.github/workflows/python-package.yml
definelicht: Ok, done!

tbennun (approved):

PR #683, opened 2021-04-07
Support CodeNode->Tasklet and AccessNode->MapEntry memlets on FPGA

definelicht:
Just by removing an unnecessary reference in emitting code-to-code memlets, we support the most common cases of CodeNode->Tasklet in OpenCL also. `FPGATransformSDFG` has been relaxed accordingly.

By doing this, several minor issues came up due to the auto-optimization tests triggering `FPGATransformSDFG` in more scenarios, which have been fixed:
- [x] Added support for dynamic map bounds through input memlets to map entries, which was missing from the FPGA codegen.
- [x] Fixes multiple issues in `NestSDFG`, which had issues with transients that should not be promoted having their name changed, as well as "ghost" non-transient arrays that did not appear as data nodes in the dataflow.
- [x] Fixed references to `EntryNodes` in the auto-optimizer, which only work for `MapNodes`.
- [x] Fix wrong call to `sdfg.make_stream` (passing a scalar as shape).

### DIFF
--- a/dace/transformation/interstate/sdfg_nesting.py
+++ b/dace/transformation/interstate/sdfg_nesting.py
@@ -1053,21 +1053,23 @@ class NestSDFG(transformation.Transformation):
                         and not node.desc(nested_sdfg).transient):
                     if (state.out_degree(node) > 0):  # input node
                         arrname = node.data
+                        arrname_nested = f"__{arrname}_in"
                         if arrname not in inputs:
                             arrobj = nested_sdfg.arrays[arrname]
-                            nested_sdfg.arrays['__' + arrname + '_in'] = arrobj
                             outer_sdfg.arrays[arrname] = dc(arrobj)
-                            inputs[arrname] = '__' + arrname + '_in'
-                        node_data_name = '__' + arrname + '_in'
+                            nested_sdfg.arrays[arrname_nested] = arrobj
+                            inputs[arrname] = arrname_nested
+                        node_data_name = arrname_nested
                     if (state.in_degree(node) > 0):  # output node
                         arrname = node.data
+                        arrname_nested = f"__{arrname}_out"
                         if arrname not in outputs:
                             arrobj = nested_sdfg.arrays[arrname]
-                            nested_sdfg.arrays['__' + arrname + '_out'] = arrobj
                             if arrname not in inputs:
                                 outer_sdfg.arrays[arrname] = dc(arrobj)
-                            outputs[arrname] = '__' + arrname + '_out'
-                        node_data_name = '__' + arrname + '_out'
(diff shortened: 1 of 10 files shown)


dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -1077,18 +1079,33 @@ def apply(self, sdfg):
...
+                            if arrname not in transients:
+                                arrobj = nested_sdfg.arrays[arrname]
+                                outer_sdfg.arrays[arrname] = dc(arrobj)
+                                nested_sdfg.arrays[arrname_nested] = arrobj
+                                transients[arrname] = arrname_nested
+
+        # Catch data containers that we didn't find on any access nodes, and add
+        # them as inputs. This can happen when a scalar input is used on an
+        # interstate edge, and thus doesn't appear in the dataflow.
+        nested_data = set(
```
TizianoDeMatteis, line 1094: Since we are here...
In the npbench_fpga branch there is something similar: we keep track of input/output transients and we add them as input/output connectors (https://github.com/spcl/dace/blob/61748d925fa9218ebfab4787fced0d15f72e2d9a/dace/transformation/interstate/sdfg_nesting.py#L1168-L1184).

I think that with the current changes, it adds the transients only to the output transients, but not the input. If it makes sense, can you add that?

tests/xilinx_test.py
```diff
@@ -40,6 +40,11 @@
     ("tests/fpga/streaming_memory.py", "streamingcomp_1", True, True, []),
     ("tests/fpga/conflict_resolution.py", "fpga_conflict_resolution", True,
      False, []),
+    # This doesn't pipeline with Vitis 2020.1 for whatever reason (it pipelines
+    # with both 2019.2 and 2020.2), so just switch this back on once CI starts
+    # running 2020.2 or newer.
+    ("tests/transformations/mapfusion_fpga.py",
```
TizianoDeMatteis, line 46: can you add this also to Intel?

TizianoDeMatteis (changes requested):
Two minor comments, otherwise it looks good to me.
Thanks!

tests/xilinx_test.py
definelicht: It's already there!

definelicht: In fact, Intel is the reason why we needed this fix in the first place :-)

TizianoDeMatteis: Gotcha!

dace/transformation/interstate/sdfg_nesting.py
definelicht: Done in 3196301b

TizianoDeMatteis (approved):

PR #685, opened 2021-04-13
FPGA global to local

TizianoDeMatteis:
Global To Local Transformation for FPGA
( and some minor fixes for intel fpga that was not correctly handling scalars with nested sdfg)

### DIFF
--- a/dace/transformation/interstate/fpga_global_to_local.py
+++ b/dace/transformation/interstate/fpga_global_to_local.py
@@ -1,0 +1,76 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+""" Contains a transformation that changes the storage type of a Global FPGA variable to Local when certains
+    conditions are met.
+"""
+
+import networkx as nx
+
+from dace import registry, properties
+from dace.transformation import transformation
+from dace import symbolic, dtypes
+from dace.sdfg import nodes, trace_nested_access
+from dace import config
+
+@registry.autoregister
+@properties.make_properties
+class FPGAGlobalToLocal(transformation.Transformation):
+    """ Implements the FPGAGlobalToLoca transformation, which takes an entire
+        SDFG  and changes the storage type of a Global FPGA data container to Local in the following situation:
+        - data the is transient,
+        - and the data is not a transient shared with other states,
+        - and data has a compile-time known size. """
+
+    @staticmethod
+    def annotates_memlets():
+        return True
+
+    @staticmethod
+    def expressions():
+        # Match anything
+        return [nx.DiGraph()]
+
+    @staticmethod
+    def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
+        # It can always be applied
+        return True
+
+    @staticmethod
+    def match_to_str(graph, candidate):
+        return graph.label
+
+
+    def apply(self, sdfg):
+
+        count = 0
+
+        for sd, name, desc in sdfg.arrays_recursive():
+            if desc.transient and name not in sd.shared_transients(
+            ) and desc.storage == dtypes.StorageType.FPGA_Global:
+
+                # Get the total size, trying to resolve it to constant if it is a symbol
+                total_size = symbolic.resolve_symbol_to_constant(desc.total_size, sdfg)
+
+                if total_size is not None:
+                    desc.storage = dtypes.StorageType.FPGA_Local
+                    count = count + 1
+
+                    # update all access nodes that refers to this container
+                    for state in sd.states():
+                        for node, graph in state.all_nodes_recursive():
+                            if isinstance(node, nodes.AccessNode):
+                                trace = trace_nested_access(
+                                    node, graph, graph.parent)
+                                outer_node = None
+                                for node_trace, memlet_trace, state_trace, sdfg_trace in trace:
+                                    # Find the name of the accessed node in our scope
+                                    if state_trace == state and sdfg_trace == sd:
+                                        _, outer_node = node_trace
+                                        if outer_node is not None:
+                                            break
+
+                                if outer_node is not None and outer_node.data == name:
+                                    nodedesc = node.desc(graph)
+                                    nodedesc.storage = dtypes.StorageType.FPGA_Local
+
+        if config.Config.get_bool('debugprint'):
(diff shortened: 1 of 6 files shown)


dace/transformation/interstate/fpga_global_to_local.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@registry.autoregister
+@properties.make_properties
+class FPGAGlobalToLocal(transformation.Transformation):
+    """ Implements the FPGAGlobalToLoca transformation, which takes an entire
+        SDFG  and changes the storage type of a Global FPGA data container to Local in the following situation:
+        - data the is transient,
+        - and the data is not a transient shared with other states,
+        - and data has a compile-time known size. """
```
definelicht, line 22:
```suggestion
    """ Implements the FPGAGlobalToLocal transformation, which takes an entire
        SDFG and changes the storage type of a global FPGA data container to Local in the following situation:
        - the data is transient,
        - the data is not a transient shared with other states, and
        - the data has a compile-time known size. """
```

dace/transformation/interstate/fpga_global_to_local.py
```diff
@@ -0,0 +1,76 @@
...
+
+    @staticmethod
+    def expressions():
+        # Match anything
+        return [nx.DiGraph()]
+
+    @staticmethod
+    def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
+        # It can always be applied
+        return True
```
definelicht, line 35: Why are you not checking if it can be applied? If you don't want to repeat code, then you can make a helper function. I don't like having a transformation that can be "applied" but doesn't do anything

dace/transformation/interstate/fpga_global_to_local.py
```diff
@@ -0,0 +1,76 @@
...
+                total_size = symbolic.resolve_symbol_to_constant(
+                    desc.total_size, sdfg)
+
+                if total_size is not None:
+                    desc.storage = dtypes.StorageType.FPGA_Local
+                    count = count + 1
+
+                    # update all access nodes that refers to this container
+                    for state in sd.states():
+                        for node, graph in state.all_nodes_recursive():
```
definelicht, line 59: Why not use `sd.all_nodes_recursive()`?

dace/transformation/interstate/fpga_global_to_local.py
```diff
@@ -0,0 +1,76 @@
...
+    def match_to_str(graph, candidate):
+        return graph.label
+
+    def apply(self, sdfg):
+
+        count = 0
+
+        for sd, name, desc in sdfg.arrays_recursive():
+            if desc.transient and name not in sd.shared_transients(
+            ) and desc.storage == dtypes.StorageType.FPGA_Global:
```
definelicht, line 47: Is there any way we can make this apply to only a single array? I'm not sure the matching supports this. Maybe we can use a random node as our match?

dace/transformation/interstate/fpga_global_to_local.py
```diff
@@ -0,0 +1,76 @@
...
+                                for node_trace, memlet_trace, state_trace, sdfg_trace in trace:
+                                    # Find the name of the accessed node in our scope
+                                    if state_trace == state and sdfg_trace == sd:
+                                        _, outer_node = node_trace
+                                        if outer_node is not None:
+                                            break
+
+                                if outer_node is not None and outer_node.data == name:
+                                    nodedesc = node.desc(graph)
+                                    nodedesc.storage = dtypes.StorageType.FPGA_Local
```
definelicht, line 73: Don't we need to apply this to _all_ nodes in the trace, not just the outer one?

definelicht (changes requested):
A lot of complaints 🤠

dace/transformation/interstate/fpga_global_to_local.py
TizianoDeMatteis: Should this return True if at least one array can be converted to Local?

dace/transformation/interstate/fpga_global_to_local.py
TizianoDeMatteis: I think you are right. That was not a problem in the examples considered so far. Will fix this

dace/transformation/interstate/fpga_global_to_local.py
TizianoDeMatteis: Why do you want to do this? To apply it repeatedly?

dace/transformation/interstate/fpga_global_to_local.py
definelicht: Ideally it should match on a particular array that could be converted to local -- although maybe the best we can do is match a data node pointing to this array (@tbennun any better idea?), so you get N matches for N arrays that could be converted from glboal to local

dace/transformation/interstate/fpga_global_to_local.py
definelicht: So you don't have to promote everything if you don't want to. For example, one array could be constant-size, but **HUGE**, so you don't want to put it in local memory

TizianoDeMatteis: Makes sense. Not sure how it should like at the end: the matching returns the list of the potential matches and then it's up to the user to decide to which container apply the transformation?

definelicht: No, I was thinking of one match per container that can be converted :-)

TizianoDeMatteis: So that it can be used with apply_repeatedly?

definelicht: Yes if desired :-) Or someone can get all matches, and search for the one with the right array name in it programmatically

TizianoDeMatteis: I'm honestly struggling in understanding how to do this (get all the matches). Can you please point me to any transformation that does this so I can take some inspiration? :grin:

definelicht: Maybe [local storage](https://github.com/spcl/dace/blob/master/dace/transformation/dataflow/local_storage.py) is relevant?

PR #688, opened 2021-04-15
Fix Xilinx codegen trailing comma

komplexon3:
In part of the Xilinx codegen, a trailing comma is mistakingly generated in some cases (if "parameters" is empty), which breaks the syntax of the generated code.

This fix ensures, that no such trailing comma is generated.

Example of such a broken syntax:
`auto kernel_rtl_ma_0_1_26 = program.MakeKernel("rtl_ma_0_1_26_top", ).ExecuteTaskFork();`

### DIFF
--- a/dace/codegen/targets/xilinx.py
+++ b/dace/codegen/targets/xilinx.py
@@ -639,7 +639,7 @@ DACE_EXPORTED void __dace_exit_xilinx({sdfg.name}_t *__state) {{
             # Launch the kernel from the host code
             rtl_name = self.rtl_tasklet_name(rtl_tasklet, state, sdfg)
             host_stream.write(
-                f"  auto kernel_{rtl_name} = program.MakeKernel(\"{rtl_name}_top\", {', '.join([name for _, name, p, _ in parameters if not isinstance(p, dt.Stream)])}).ExecuteTaskFork();",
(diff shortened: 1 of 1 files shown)


dace/codegen/targets/xilinx.py
```diff
@@ -639,7 +639,7 @@ def generate_module(self, sdfg, state, name, subgraph, parameters,
             # Launch the kernel from the host code
             rtl_name = self.rtl_tasklet_name(rtl_tasklet, state, sdfg)
             host_stream.write(
-                f"  auto kernel_{rtl_name} = program.MakeKernel(\"{rtl_name}_top\", {', '.join([name for _, name, p, _ in parameters if not isinstance(p, dt.Stream)])}).ExecuteTaskFork();",
+                f"  auto kernel_{rtl_name} = program.MakeKernel(\"{rtl_name}_top\"{', '.join(['']+[name for _, name, p, _ in parameters if not isinstance(p, dt.Stream)])}).ExecuteTaskFork();",
```
definelicht, line 642: Run yapf on the code please!

definelicht (changes requested):

dace/codegen/targets/xilinx.py
komplexon3: I did. I assume you are referring to the missing spaces around "+"? I added them now but, rather surprisingly, pep8 actually doesn't require spaces around arithmetic operators.

https://hg.python.org/peps/rev/37af28ad2972

definelicht: It does, but it was inside the string format, so I guess it's not smart enough to parse this :-) Thanks!

pep

definelicht (approved):

PR #690, opened 2021-04-18
@dace.program performance optimizations

tbennun:
- [x] Cache compiled SDFG from called `@dace.program`s
- [x] Support for transforming DacePrograms directly (#691) and calling SDFGs with their original Python argument order
- [x] More robust and safer checks for "if the SDFG is already loaded, don't override/recompile"
- [x] Minor auto-optimization fix
- [x] Return values are allocated as `np.empty` rather than `np.zeros`


dace/frontend/python/parser.py
```diff
@@ -282,50 +248,99 @@ def __call__(self, *args, **kwargs):
...
 
+    def parse(self, *compilation_args, strict=None, save=False) -> SDFG:
+        """ 
+        Try to parse a DaceProgram object and return the `dace.SDFG` object
+        that corresponds to it.
+        :param function: DaceProgram object (obtained from the ``@dace.program``
+                        decorator).
+        :param compilation_args: Various compilation arguments e.g. dtypes.
+        :param strict: Whether to apply strict transformations or not (None
+                    uses configuration-defined value).
```
definelicht, line 272:
```suggestion
        :param strict: Whether to apply strict transformations or not (None
                       uses configuration-defined value). 
```

definelicht: I wish yapf could handle docstrings... There are [tools that can do this](https://github.com/myint/docformatter) (albeit separate), we could consider if it's worth adopting this at some point.

dace/frontend/python/parser.py
```diff
@@ -282,50 +248,99 @@ def __call__(self, *args, **kwargs):
...
+            sdfg.apply_strict_transformations()
+
+            # Split back edges with assignments and conditions to allow richer
+            # control flow detection in code generation
+            xfh.split_interstate_edges(sdfg)
+
+        # Save the SDFG. Skip this step if running from a cached SDFG, as
+        # it might overwrite the cached SDFG.
+        if not Config.get_bool('compiler', 'use_cache') and save:
+            sdfg.save(os.path.join('_dacegraphs', 'program.sdfg'))
```
definelicht, line 306: I know this is not new, but out of curiosity -- why do we always save the SDFG here? It's already saved in the `.dacecache`.

dace/transformation/auto/auto_optimize.py
```diff
@@ -166,17 +166,28 @@ def tile_wcrs(graph_or_subgraph: GraphViewType,
...
+
+                # Make transient only if there is one WCR/stream
+                if to_apply is not None:
+                    to_apply = None
+                    break
+
+                to_apply = (dataflow.StreamTransient, {},
+                            dict(tasklet=tasklet,
+                                 map_exit=mapexit,
+                                 outer_map_exit=outer_mapexit))
```
definelicht, line 190: This snuck in? ;-)

definelicht (approved):
Approving because comments are so minor, just murge when ready

dace/frontend/python/parser.py
tbennun: yeah no idea why yapf doesn't do this.

dace/frontend/python/parser.py
tbennun: we now don't :) the diff that for some reason doesn't show up is that the argument is now `save=False` 💯

dace/transformation/auto/auto_optimize.py
tbennun: 1. it's part of the PR description
2. it appeared on auto-optimized `@dace.program`s

dace/frontend/python/parser.py
definelicht: Ah, gr8

PR #692, opened 2021-04-21
Use pytest-xdist to run some pytest tests in parallel

definelicht:
- Use pytesdt-xdist to test in parallel.
- Introduce a new flag for generating the SDFG cache name with more options, allowing a random name based on the process ID to be used, avoiding clashes between parallel tests.
- Reorganize heterogeneous tests so they can potentially be run in parallel (if we add more runners).
- Rename a bunch of SDFGs to avoid name clashes even without using the unique cache name.

### DIFF
--- a/dace/config_schema.yml
+++ b/dace/config_schema.yml
@@ -740,10 +740,10 @@ required:
         title: Naming of cache entry 
         description: >
             Determine the name of the generated dacecache folder.
-            "name" uses the name of the SDFG, causing it to be overridden by
-            other programs using a similar SDFG name.
-            "hash" uses a mangled name using the hash of the SDFG, such that any
-            change to the SDFG will generate a different cache folder.
+            "name" uses the name of the SDFG directly, causing it to be
+            overridden by other programs using the same SDFG name.
+            "hash" uses a mangled name based on the hash of the SDFG, such that
+            any change to the SDFG will generate a different cache folder.
             "unique" uses a name based on the currently running Pythion process 
             at code generation time, such that no caching or clashes can happen
             between different processes or subsequent invokations of Python.
--- a/dace/config_schema.yml
+++ b/dace/config_schema.yml
@@ -744,9 +744,9 @@ required:
             other programs using a similar SDFG name.
             "hash" uses a mangled name using the hash of the SDFG, such that any
             change to the SDFG will generate a different cache folder.
-            "unique" uses a name based on the SDFGs current location in memory
+            "unique" uses a name based on the currently running Pythion process 
             at code generation time, such that no caching or clashes can happen
-            between different processes or subsequent invokations of Python..
+            between different processes or subsequent invokations of Python.
             "single" uses a single cache folder for all SDFGs, saving space and
             potentially build time, but disallows executing SDFGs in parallel
             and caching of more than one simultaneous SDFG.
--- a/dace/config_schema.yml
+++ b/dace/config_schema.yml
@@ -734,6 +734,23 @@ required:
         title: Debug printing
         description: Enable verbose printouts.
 
+    cache:
+        type: str 
+        default: name 
+        title: Naming of cache entry 
+        description: >
+            Determine the name of the generated dacecache folder.
+            "name" uses the name of the SDFG, causing it to be overridden by
+            other programs using a similar SDFG name.
+            "hash" uses a mangled name using the hash of the SDFG, such that any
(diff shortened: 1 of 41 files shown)


dace/config_schema.yml
```diff
@@ -734,6 +734,23 @@ required:
...
+        type: str 
+        default: name 
+        title: Naming of cache entry 
+        description: >
+            Determine the name of the generated dacecache folder.
+            "name" uses the name of the SDFG, causing it to be overridden by
+            other programs using a similar SDFG name.
+            "hash" uses a mangled name using the hash of the SDFG, such that any
+            change to the SDFG will generate a different cache folder.
+            "unique" uses a name based on the currently running Pythion process
```
tbennun, line 747:
```suggestion
            "unique" uses a name based on the currently running Python process 
```

dace/config_schema.yml
```diff
@@ -734,6 +734,23 @@ required:
...
+        title: Naming of cache entry 
+        description: >
+            Determine the name of the generated dacecache folder.
+            "name" uses the name of the SDFG, causing it to be overridden by
+            other programs using a similar SDFG name.
+            "hash" uses a mangled name using the hash of the SDFG, such that any
+            change to the SDFG will generate a different cache folder.
+            "unique" uses a name based on the currently running Pythion process 
+            at code generation time, such that no caching or clashes can happen
+            between different processes or subsequent invokations of Python.
```
tbennun, line 749:
```suggestion
            between different processes or subsequent invocations of Python.
```

tests/instrumentation_test.py
```diff
@@ -32,6 +32,7 @@ def onetest(instrumentation: dace.InstrumentationType, size=128):
     C = np.zeros([size, size], dtype=np.float64)
 
     sdfg: dace.SDFG = slowmm.to_sdfg()
+    sdfg._name = f"instrumentation_test_{instrumentation.name}"
```
tbennun, line 35:
```suggestion
    sdfg.name = f"instrumentation_test_{instrumentation.name}"
```

This should work too I think.

tests/transformations/loop_to_map_test.py
```diff
@@ -66,8 +68,8 @@ def make_sdfg(with_wcr, map_in_guard, reverse_loop, use_variable, assign_after):
                                 """\
 static std::mutex mutex;
 std::unique_lock<std::mutex> lock(mutex);
-std::ofstream of("loop_to_map_test.txt", std::ofstream::app);
-of << i << "\\n";""",
+std::ofstream of("{}", std::ofstream::app);
```
tbennun, line 71: maybe fstring here?

tests/transformations/loop_to_map_test.py
```diff
@@ -111,11 +119,6 @@ def apply_and_verify(sdfg, n):
 
     num_transformations = sdfg.apply_transformations(LoopToMap)
 
-    try:
-        os.remove("loop_to_map_test.txt")
```
tbennun, line 115: don't you want to still remove the temp file?

tbennun (changes requested):

tbennun:
Also, why is the coverage down?

definelicht:
> 
> 
> Also, why is the coverage down?

The coverage is not down in this commit, but something might have happened in moving the FPGAs to the other testing system? Could you take a look at the github workflow file, and see if something is missing?

tests/transformations/loop_to_map_test.py
definelicht: This is not necessary anymore, as we are using a unique `tempfile.NamedTemporaryFile` every time the function is called, which is cleaned up after the test.

dace/config_schema.yml
definelicht: why tf is it spelled "invoke", but "invocation"? English is stoopid

tests/transformations/loop_to_map_test.py
definelicht: Done in 43cc83a7c74d7625c20fbe108e65e6ba32ea7ea9

tests/instrumentation_test.py
definelicht: Done in 2ec3478f917352bdf1816155caf1da3262b6b0c8

tbennun (approved):

tbennun:
test-fpga doesn't have a "report coverage" clause

PR #693, opened 2021-04-22
Fpga: auto optimization

TizianoDeMatteis:
FPGA Specific Auto optimizations:
- global_to_local
- round-robin allocation of containers to memory Bank

This also contains some specific fixes for Intel FPGA codegen (nested SDFG arguments generation)


dace/transformation/auto/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+""" This module initializes the inter-state transformations package."""
```
tbennun, line 2: Comment should change

tests/fpga/auto_opt_fpga.py
```diff
@@ -0,0 +1,86 @@
...
+            B[i] = tmp + 1
+
+    A = np.random.rand(size).astype(np.float32)
+    B = np.random.rand(size).astype(np.float32)
+
+    sdfg = global_to_local.to_sdfg()
+    sdfg.apply_transformations([FPGATransformSDFG])
+
+    aopt.fpga_global_to_local(sdfg)
+
```
tbennun, line 30: I don't understand. shouldn't this be under `aopt.auto_optimize(sdfg, dace.DeviceType.FPGA)`?

tbennun (changes requested):

tests/fpga/auto_opt_fpga.py
TizianoDeMatteis: Sure, I can do that! But for the moment being I will just have those two specific autoopts, that are not applied for CPU/GPU

dace/transformation/auto/__init__.py
TizianoDeMatteis: :+1:

tests/fpga/auto_opt_fpga.py
TizianoDeMatteis: And @tbennun , now that I think about it, the SDFG  must have been already Transformed for FPGA (otherwise we find less optimization opportunities). Is there a way to check this?

tbennun: we already added "apply_gpu_transformations()" to the case where devicetype == GPU, so why not here as well?

tbennun: as I said in our previous discussion, we do some transformations before the gpu transform AND after, so it's really convenient to have control over this @TizianoDeMatteis

TizianoDeMatteis: > we already added "apply_gpu_transformations()" to the case where devicetype == GPU, so why not here as well?

I can not see it into auto_optimize. Or is it up to the user to do this before actually invoking auto-optimizations with devicetype==GPU?

tbennun: sorry, we probably have not merged it yet: https://github.com/sancierra/dace/blob/autoopt-fusion/dace/transformation/auto_optimize.py#L423

dace/transformation/auto/fpga.py
```diff
@@ -0,0 +1,66 @@
...
+                        trace = trace_nested_access(node, graph, graph.parent)
+
+                        for (_, candidate
+                             ), memlet_trace, state_trace, sdfg_trace in trace:
+                            if candidate is not None and candidate.data == name:
+                                nodedesc = node.desc(graph)
+                                nodedesc.storage = dtypes.StorageType.FPGA_Local
+
+    if config.Config.get_bool('debugprint'):
+        print(f'Applied {count} Global-To-Local.')
```
definelicht, line 45: Very minor: it would be nice to have a debugprint per array instead of only the aggregate

dace/transformation/auto/fpga.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def fpga_rr_interleave_containers_to_banks(sdfg: SDFG, num_banks: int = 4):
```
definelicht, line 48: I think this function assumes that you are calling it on the outermost SDFG, right? It could potentially create an invalid SDFG it was called on both an SDFG and a nested SDFG inside of it, because there could be a clash in bank assignments.
This does not necessarily matter (especially if it's primarily called from auto-optimization), but perhaps we should be on the safe side and do some checks, or maybe just make it clear in the docstring?

dace/transformation/auto/auto_optimize.py
```diff
@@ -306,33 +314,43 @@ def auto_optimize(sdfg: SDFG,
...
-    sdfg.apply_transformations_repeated(MapCollapse,
-                                        strict=True,
-                                        validate=False,
-                                        validate_all=validate_all)
-    for node, _ in sdfg.all_nodes_recursive():
-        if isinstance(node, nodes.MapEntry):
-            node.map.collapse = len(node.map.range)
+        # Set all library nodes to expand to fast library calls
+        set_fast_implementations(sdfg, device)
+        return SDFG
```
tbennun, line 325: you return the class and not the object

dace/transformation/auto/auto_optimize.py
```diff
@@ -306,33 +314,43 @@ def auto_optimize(sdfg: SDFG,
...
+    else:
```
tbennun, line 326: no need for the `else` since you have the return stmt

tbennun (changes requested):
two very minor changes from my side, otherwise looks good!

tbennun (approved):
LGTM

dace/transformation/auto/fpga.py
```diff
@@ -0,0 +1,81 @@
...
+    :param sdfg: The SDFG to operate on. It must be a top-level SDFG.
+    :param: num_banks: number of off-chip memory banks to consider
+    :returns: a list containing  the number of arrays allocated to each bank
+    :note: Operates in-place on the SDFG.
+    '''
+
+    if sdfg.parent is not None:
+        if config.Config.get_bool('debugprint'):
+            print(
+                'Interleave Containers can be applied only to a top-level SDFG.'
```
definelicht, line 69: Wait, but is this true? I think it's also valid to be in a nested SDFG, as long as the array is transient. We just want to make sure that this is not a connector coming from outside, which points to a different bank!

definelicht (changes requested):

dace/transformation/auto/fpga.py
```diff
@@ -0,0 +1,76 @@
...
+        :param: sdfg: The SDFG to operate on. It must be a top-level SDFG.
+        :param: max_size: maximum size (in bytes) that a container can have to be considered for
+            storage type change
+        :note: Operates in-place on the SDFG.
+    """
+
+    if sdfg.parent is not None:
+        if config.Config.get_bool('debugprint'):
+            print('Global-To-Local can be applied only to a top-level SDFG.')
+        return  # Not the top-most SDFG. Do nothing.
```
definelicht, line 24: Do we need this?

TizianoDeMatteis: You requested that

definelicht (approved):

PR #696, opened 2021-04-27
Give the Data class' property 'dtype' the correct enum choices

phschaad:
The property `dtype` of the class `Data` and its subclasses is a `typeclass` property. The class `typeclass` seems to be constrained to a set of predefined types which are listed in `dace.dtypes`. These predefined values are also used for reconstructing the `dtype` property from JSON. This creates a proper enum for these options and provides them to the `dtype` property as choices.

Benefit: This allows for easier editing of a data container's data type since it removes the chance of typos and the need for additional string parsing after user input.

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -45,7 +45,7 @@ class Data(object):
         Examples: Arrays, Streams, custom arrays (e.g., sparse matrices).
     """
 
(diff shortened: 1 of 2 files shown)


tbennun (approved):

phschaad:
Thank you!

Issue #697, opened 2021-04-30
Nested SDFG reuse: allow explicit control


TizianoDeMatteis:
Currently, the reuse of Nested SDFG is implicit: the user can enable the corresponding dace configuration flag (`unique_function`), and then all the NestedSDFGs that have the same hash, are codegenerated only once.

We would like to have a more explicit control on this.

PR #699, opened 2021-05-03
Explicit control over Nested SDFG codegen

TizianoDeMatteis:
To allow more control over Nester SDFG code generation, the user can now:
- use a "hash" based approach (implicit): nested SDFGs with the same hash are code generated only once
- or, use the "unique_name" of SDFG (explicit): SDFGs with the same unique_name are generated only once
- or "none" of the above: a separate function is produced for each Nested SDFG 

This is controlled by a property in `dace.conf`.
Solves #697


dace/config_schema.yml
```diff
@@ -205,12 +205,18 @@ required:
...
             unique_functions:
-                type: bool
-                default: true
+                type: str
+                default: hash
                 title: Generate unique functions
                 description: >
-                    If set to true, if multiple nested SDFGs with equivalent contents exist,
-                    the code is generated only once.
+                    Determine if and how to generate the code for equivalent nestedSDFGs:
```
tbennun, line 212:
```suggestion
                    Determine if and how to generate the code for equivalent NestedSDFGs:
```

dace/sdfg/sdfg.py
```diff
@@ -253,6 +253,10 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
     orig_sdfg = SDFGReferenceProperty(allow_none=True)
     transformation_hist = TransformationHistProperty()
 
+    unique_name = Property(dtype=str,
```
tbennun, line 256: I thought this was a property of the NestedSDFG node?

tbennun (changes requested):

definelicht:
To be honest this approach is a bit dubious to me -- what happens if two SDFGs have the same `unique_name`, but are different?

I was hoping for an approach where we can guarantee programmatically that two nested SDFG nodes have the same SDFG _reference_, and therefore are identical.

This is not a good API, but an example of how you could guarantee this:
```python
sdfg = dace.SDFG("top_level")
nested_sdfg = dace.SDFG("nested")
sdfg.register_nested_sdfg(nested_sdfg)
state = sdfg.add_state("foo")
nested_sdfg_node = state.add_nested_sdfg("nested")
```

TizianoDeMatteis:
If the `unique_name` is the same, but the SDFGs are different, it will result in undefined behavior. 
It's up to the programmer to deal properly with this if he wants more control over NSDFG code generation.

Regarding your proposed API @definelicht , what will happen with the graphical representation of the SDFG?
If in the SDFG I have a Nested SDFG which is "registered", it becomes a sort of placeholder (like Access Nodes)

definelicht:
> 
> 
> If the `unique_name` is the same, but the SDFGs are different, it will result in undefined behavior.
> It's up to the programmer to deal properly with this if he wants more control over NSDFG code generation.

This is what I'm worried about, apart from being dirty, it's also not a nice workflow and has potential for bugs. For example, if you load the SDFG from JSON, and the nested SDFGs exist as two separate entities in memory. Then, you apply a transformation to one of the nested SDFG. But if you are unlucky and the other is found first, the transformation has no effect. It would be much nicer if this was guaranteed to work by design, so you could never get two copies of the same SDFG.

> Regarding your proposed API @definelicht , what will happen with the graphical representation of the SDFG?
> If in the SDFG I have a Nested SDFG which is "registered", it becomes a sort of placeholder (like Access Nodes)

Nothing would happen with the graphical representation: There is a unique mapping from each nested SDFG node to which SDFG it represents, so it's easy to retrieve this and draw the same way as we do now.

tbennun:
@TizianoDeMatteis @definelicht 
Another much simpler option is to prepend the hash (in a way that is invisible to the user) to the unique_name.

definelicht:
> 
> 
> @TizianoDeMatteis @definelicht
> Another much simpler option is to prepend the hash (in a way that is invisible to the user) to the unique_name.

That doesn't address my primary concern, which is to expose an explicit way for the programmer to instantiate nested SDFGs

dace/codegen/targets/cpu.py
```diff
@@ -1570,30 +1570,36 @@ def _generate_NestedSDFG(
...
             else:
-                # Use the SDFG label, keep track if this has been already code generated
+                # Use the SDFG label to check if this has been already code generated.
+                # Check the hash of the formerly generated SDFG to check that we are not
+                # generating different SDFGs with the same name
                 if sdfg_label in self._generated_nested_sdfg:
                     code_already_generated = True
+                    if hash != self._generated_nested_sdfg[sdfg_label]:
+                        raise ValueError(
+                            f'Different Nested SDFGs have tha same unique name: {sdfg_label}'
```
tbennun, line 1599:
```suggestion
                            f'Different Nested SDFGs have the same unique name: {sdfg_label}'
```

tbennun (approved):
very minor change but otherwise approved

PR #702, opened 2021-05-04
Fix state fission, improved testing

TizianoDeMatteis:
State fission: Improved conditions on how to detect node to not remove.
Improved (and updated to new memlet API) the test as well


tbennun:
@TizianoDeMatteis There's an even more accurate fix here: https://github.com/spcl/dace/pull/645/files?file-filters%5B%5D=.py#diff-f76940db8b32cd3a0807a9d0b3b9fa8aa845dd60ff5c91cb6b6f292db860373b

PR #704, opened 2021-05-05
Use $PYTHON_BINARY to fix code coverage

definelicht:


tbennun:
maybe we still need to call this "coverage combine" that was removed one PR later?
https://github.com/spcl/dace/commit/83f38ae1206cf8ff58553a37d1035ef5cf36d914

definelicht:
> maybe we still need to call this "coverage combine" that was removed one PR later?
> [83f38ae](https://github.com/spcl/dace/commit/83f38ae1206cf8ff58553a37d1035ef5cf36d914)

Same error as before, "no data to combine" :-/
Could it be subprocess or multiprocessing that messes up coverage somehow?

In `intel_fpga_test.py`, for example, we use subprocess to call tests like this:

```python
        proc = sp.Popen(map(str, [sys.executable, path] + args),
                        cwd=TEST_DIR,
                        stdout=sp.PIPE,  
                        stderr=sp.PIPE,
                        encoding="utf-8") 
```

I could try changing this to use `importlib` to run them in the same interpreter as the process spawned by multiprocessing.

tbennun (approved):

PR #705, opened 2021-05-05
Arm SVE code generation

sscholbe:
The most relevant parts are _codegen.py_ and _unparse.py_, because they are responsible for generating the code.

* codegen.py: SDFG -> SVE
* unparse.py: AST -> SVE
* preprocess.py: Detect fused operations like Multiply-Add, is a NodeTransformer
* type_compatibility.py: Checking, that vectors are compatible and they can be represented in SVE (very frequently used in the unparser)
* util.py: Frequently used mappings (e.g. ops to SVE instructions, datatypes to SVE, etc.) and some utility functions

_vectorization.py_ is not relevant, because it just forces a vectorization for demonstration purposes.

_inference.py_ and _cppunparse.py_ are minimally tweaked versions of the original files, maybe at some point I can find a workaround, so I don't have to copy a file just to change for example an import.

The tests are mainly written for coverage (and a little bit of bug fixing).

Not all imports are fixed yet because some parts were copied from the tutorial.


tbennun (changes requested):

dace/codegen/targets/sve/unparse.py
```diff
@@ -0,0 +1,562 @@
...
+            raise NotImplementedError(f'Can not infer RHS of assignment ({astunparse.unparse(t.value)})')
+
+        is_new_variable = False
+
+        if lhs_type is None:
+            # The LHS could involve a variable name that was not declared (which is why inference fails)
+            if not isinstance(
+                    target,
+                    ast.Name) or target.id in self.get_defined_symbols():
+                # Either we don't assign to a name, or the variable name has already been declared (but infer still fails, i.e. something went wrong!)
```
tbennun, line 307:
```suggestion
                # Either we don't assign to a name, or the variable name 
                # has already been declared (but infer still fails, i.e. something 
                # went wrong!)
```

dace/codegen/targets/sve/unparse.py
```diff
@@ -0,0 +1,562 @@
...
+                # SVE register is possible (declare it as svXXX_t)
+                self.fill(util.TYPE_TO_SVE[rhs_type.type])
+                self.write(' ')
+                # Define the new symbol as vector
+                self.defined_symbols.update({target.id: rhs_type})
+            elif isinstance(rhs_type, dtypes.pointer):
+                raise util.NotSupportedError(
+                    'Defining pointers in Tasklet code not supported')
+
+            # Otherwise, the fallback will grab the case of a scalar, because the RHS is scalar, and the LHS is the same
```
tbennun, line 322:
```suggestion
            # Otherwise, the fallback will grab the case of a scalar,
            # because the RHS is scalar, and the LHS is the same
```

dace/codegen/targets/sve/unparse.py
```diff
@@ -0,0 +1,562 @@
...
+            self.write(', ')
+
+        # Close all except the last bracket (because the last entry isn't nested)
+        self.write(')' * (len(t.values) - 1))
+
+    def _Compare(self, t):
+        if len(t.ops) != 1:
+            # This includes things like  a < b <= c
+            raise NotImplementedError(
+                'Multiple comparisions at once not implemented')
```
tbennun, line 522:
```suggestion
                'Multiple comparisons at once not implemented')
```

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+
+
+def generate_footer_hook(self, sdfg: SDFG, global_stream: CodeIOStream,
+                        callsite_stream: CodeIOStream):
+    global old_generate_footer
+    if contains_any_sve(sdfg):
+        util.patch_ctypes()
+    res = old_generate_footer(self, sdfg, global_stream, callsite_stream)
+    util.restore_ctypes()
+    return res
```
tbennun, line 57: :(

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+
+
+@dace.registry.autoregister_params(name='sve')
+class SVEMap(TargetCodeGenerator):
```
tbennun, line 61: The name could be more descriptive (`SVECodeGenerator`)

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+    target_name = 'armv8'
+    title = 'sve'
+    language = 'cpp'
+
+    def get_load_stride(self, sdfg: SDFG, state: SDFGState, node: nodes.Node,
```
tbennun, line 66: nice! Can be made generic later on for transformations etc. :)

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+
+        """
+        # Hook the generate_header in DaCeCodeGenerator
+        global old_generate_footer, old_generate_header
+        if old_generate_footer is None:
+            old_generate_header = DaCeCodeGenerator.generate_header
+            old_generate_footer = DaCeCodeGenerator.generate_footer
+        DaCeCodeGenerator.generate_header = generate_header_hook
+        DaCeCodeGenerator.generate_footer = generate_footer_hook
+        """
```
tbennun, line 122: Why is this commented out? Where does the hooking happen?

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+        var_name = None
+        var_type = None
+        var_ctype = None
+
+        if output:
+            var_name = edge.dst_conn
+        else:
+            var_name = edge.src_conn
+
+        if is_interconnect:
```
tbennun, line 159: I would prefer if you didn't define new names where current ones exist. These are called code->code edges (because they are directly between two `CodeNode`s)

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+        ltype_size = loop_type.bytes
+
+        long_type = copy.copy(dace.int64)
+        long_type.ctype = 'int64_t'
+
+        self.counter_type = {
+            1: dace.int8,
+            2: dace.int16,
+            4: dace.int32,
+            8: long_type
```
tbennun, line 275: 💯

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+        # We should always be in an SVE scope
+        scope = util.get_sve_scope(sdfg, dfg, dst_node)
+        if scope is None:
+            raise NotImplementedError('Not in an SVE scope')
+
+        in_conn = dst_node.in_connectors[edge.dst_conn]
+
+        if isinstance(src_node, dace.nodes.Tasklet):
+            # Copy from tasklet is just copying the shared register
+            # HACK: We abuse defined_vars to get the C++ type of the shared register
```
tbennun, line 362: Not a hack, that's what defined_vars are for :) Remove `HACK:`

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+
+            # First part of the declaration is `type name`
+            load_lhs = '{} {}'.format(util.TYPE_TO_SVE[in_conn.type],
+                                      edge.dst_conn)
+
+            ptr_cast = ''
+            if in_conn.type == np.int64:
+                ptr_cast = '(int64_t*) '
+            elif in_conn.type == np.uint64:
+                ptr_cast = '(uint64_t*) '
```
tbennun, line 432: isn't this why you wrote the function in util?

dace/codegen/targets/sve/codegen.py
```diff
@@ -0,0 +1,588 @@
...
+                # If no WCR, we can directly store the vector (SVE register) in memory
+                # Determine the stride of the store and use a scatter load if applicable
+
+                stride = self.get_load_stride(sdfg, dfg, src_node, edge.data)
+
+                ptr_cast = ''
+                if out_conn.type == np.int64:
+                    ptr_cast = '(int64_t*) '
+                elif out_conn.type == np.uint64:
+                    ptr_cast = '(uint64_t*) '
```
tbennun, line 496: same as above comment

tbennun (changes requested):
Looks very good! Some comments remain but they are only minor :)

dace/codegen/tools/type_inference.py
```diff
@@ -355,17 +356,25 @@ def _BinOp(t, symbols, inferred_symbols):
...
+    inf_type = _dispatch(t.left, symbols, inferred_symbols)
+    any_vector = isinstance(inf_type, dtypes.vector)
     for o, e in zip(t.ops, t.comparators):
         if o.__class__.__name__ not in cppunparse.CPPUnparser.cmpops:
             continue
-        _dispatch(e, symbols, inferred_symbols)
+        inf_type = _dispatch(e, symbols, inferred_symbols)
+        any_vector = any_vector or isinstance(inf_type, dtypes.vector)
 
+    return dtypes.vector(dace.bool, -1) if any_vector else dtypes.bool
```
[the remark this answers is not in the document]
sscholbe: For the vector length, this is right.
In my opinion, the return type should still be a bool vector, because the result of a comparison is a boolean, right?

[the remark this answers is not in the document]
sscholbe: Also regarding the length, the inference is used everywhere in DaCe and the SVE_LEN symbol might pop up unexpectedly in other codes.

dace/codegen/targets/sve/codegen.py
sscholbe: There is no hooking happening anymore :)

dace/codegen/tools/type_inference.py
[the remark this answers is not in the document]
sscholbe: Makes sense with the vector length. In a vector comparision, the result type should be a bool vector because in the scalar case it a bool as well, right?

dace/codegen/tools/type_inference.py
```diff
@@ -360,23 +361,32 @@ def _BinOp(t, symbols, inferred_symbols):
...
+    if isinstance(inf_type, dtypes.vector):
+        vec_len = inf_type.veclen
     for o, e in zip(t.ops, t.comparators):
         if o.__class__.__name__ not in cppunparse.CPPUnparser.cmpops:
             continue
         inf_type = _dispatch(e, symbols, inferred_symbols)
-        any_vector = any_vector or isinstance(inf_type, dtypes.vector)
+        if isinstance(inf_type, dtypes.vector):
+            # Make sure all occuring vectors are of same size
+            if vec_len is not None and vec_len != inf_type.veclen:
```
tbennun, line 373: Nice

tests/codegen/sve/application_axpy_test.py
```diff
@@ -20,6 +21,7 @@ def multiplication(i):
         out = in_A * in_X + in_Y
 
 
+@pytest.mark.skip
```
tbennun, line 24: skip? Why not `@pytest.mark.sve` and add it to the markers in `pytest.ini` (as well as disable it by default in `.github/workflows`)

tests/codegen/sve/map_test.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+from dace import codegen
```
tbennun, line 2: why this change?

tbennun (changes requested):
One tiny change

tbennun (approved):

PR #706, opened 2021-05-05
Use module path instead of name only to identify environments

definelicht:


definelicht:
@alexnick83 Could you see if this helps?

tbennun (approved):

tbennun:
@definelicht actually, before we merge, can we consider using the full qualified name, as in here? https://github.com/spcl/dace/blob/master/dace/sdfg/nodes.py#L1172

Maybe someone would want to define multiple environments in one file etc.

definelicht:
> 
> 
> @definelicht actually, before we merge, can we consider using the full qualified name, as in here? https://github.com/spcl/dace/blob/master/dace/sdfg/nodes.py#L1172
> 
> Maybe someone would want to define multiple environments in one file etc.

Sure

tbennun (approved):

PR #714, opened 2021-05-13
Python 3.9 support

tbennun:


gronerl (approved):
Thanks! PR is compatible with GT4Py, no dependency issues.

PR #717, opened 2021-05-13
NumPy Frontend: implement broadcasted assignment

orausch:


tbennun (approved):

dace/transformation/dataflow/redundant_array.py
```diff
@@ -325,7 +325,7 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
         else:
             # Two views connected to each other
             if isinstance(in_desc, data.View):
-                return False
+                return True
```
alexnick83, line 328: Isn't this case handled by some new transformation that I saw recently (squeeze/unsqueeze view remove)? If not, we still should make it a separate transformation. RedundantArray is overloaded and it will be very difficult in the future to maintain.

dace/frontend/python/replacements.py
```diff
@@ -997,7 +997,20 @@ def _unop(sdfg: SDFG, state: SDFGState, op1: str, opcode: str, opname: str):
     return name
 
 
-def _broadcast_together(arr1_shape, arr2_shape):
+def _broadcast_to(target_shape, operand_shape):
```
alexnick83, line 1000: It would be nice if we can clarify what exactly is being supported here. For ufuncs, I made on purpose a new method to handle broadcast in order to understand how this works without potentially breaking expected/existing functionality.  I found that there are issues regarding symbols, e.g., which is greater; M or N. When it comes to symbolic sizes, what does `_broadcast_together` exactly support?. It would help with merging and removing duplicate functionality.

tests/numpy/assignment_test.py
```diff
@@ -36,6 +36,78 @@ def multiassign(B: dace.float64[1],
     assert B[0] == expected and C[1] == expected
 
 
+@compare_numpy_output(positive=True)
```
alexnick83, line 39: We really need to add tests with symbolic sizes to explore the limits of what we can handle.

alexnick83 (approved):
The PR is fine in general. There are mid/long-term issues that have to be addressed (read my comments). I imagine though that this is urgent and since it only adds functionality (doesn't seem to alter something) I am approving. However, we have to see how these mid/long-term issues, among many others, will be addressed. The bandwidth needed to maintain these is increasing sharply.

PR #718, opened 2021-05-16
Skip memlet propagation for LocalStorage transformation

noah95:
In a tiled map that doesn't divide evenly, a LocalStorage transformation should not trigger memlet propagation


tbennun (approved):
LGTM!

PR #721, opened 2021-05-17
Passing linker flags

definelicht:


dace/codegen/compiler.py
```diff
@@ -189,15 +189,17 @@ def configure_and_compile(program_folder,
...
     cmake_command.append("-DDACE_LIBS=\"{}\"".format(" ".join(libraries)))
 
-    # Override linker and linker arguments
-    if Config.get('compiler', 'linker', 'executable'):
-        cmake_command.append("-DCMAKE_LINKER=\"{}\"".format(
-            make_absolute(Config.get('compiler', 'linker', 'executable'))))
-    if Config.get('compiler', 'linker', 'args') is not None:
+    # Set linker and linker arguments, iff they have been specified (setting an
+    # empty string seems to cause issues on macOS).
+    cmake_linker = Config.get('compiler', 'linker', 'executable').strip()
```
tbennun, line 194: Wait, wouldn't it raise an exception if the config **is** None?

definelicht: I don't think it can be `None`? Doesn't `Config.get` always return a string?

definelicht: You are right that this can break if the variable is set to nothing at all, fixing

definelicht: Done in 659bdbe

tbennun (approved):

PR #722, opened 2021-05-17
FPGA Kernels

TizianoDeMatteis:
FPGA Kernel detection: we want to enable parallel execution of independent scopes/LibraryNodes/nested SDFGs within the same SDFG weakly connected component:
- when generating a state detects the kernels. This is done only at the top-most state. If the detected kernels are the same as the independent components, use them (PEs)
- we keep track of dependencies among kernels
- generates all the host code for running kernels of a state inside a single function: this helps with tracking dependencies without using more complicated mechanisms
- all kernels are executed asynchronously (ExecuteTaskFork). Then:
    - if a successive Kernel depends on a previous one, then the events of the predecessor are used for synchronization
    - otherwise, at the end of the state function we wait for the completion of all created kernels


dace/codegen/targets/intel_fpga.py
```diff
@@ -524,26 +542,51 @@ def generate_kernel_internal(self, sdfg, state, kernel_name, subgraphs,
...
+        needs_synch = len(predecessors) > 0
+
+        if needs_synch:
+            # Build a vector containing all the events associated with the kernels from which this one depends
+            kernel_deps_name = f"deps_{kernel_name}"
+            host_stream.write(f"std::vector<cl::Event > {kernel_deps_name};")
+            for pred in predecessors:
+                # concatenate events from predecessor kernel
+                host_stream.write(
+                    f"{kernel_deps_name}.insert({kernel_deps_name}.end(), {pred}_events.begin(), {pred}_events.end());"
```
definelicht, line 584: I think we should make this a constant size array instead of a vector, creating and appending a vector only creates unnecessary overhead when we anyway know the size at compile-time.

dace/codegen/targets/intel_fpga.py
```diff
@@ -524,26 +542,51 @@ def generate_kernel_internal(self, sdfg, state, kernel_name, subgraphs,
...
+        seperator = "/" * 59
+        host_stream.write(
+            f"\n{seperator}\n// Kernel: {kernel_name}\n{seperator}\n\n")
 
-    @staticmethod
-    def generate_host_function_prologue(sdfg, state, host_stream):
-        host_stream.write("std::vector<hlslib::ocl::Kernel> kernels;", sdfg,
-                          sdfg.node_id(state))
+        host_stream.write(
+            f"std::vector<hlslib::ocl::Kernel> {kernel_name}_kernels;", sdfg,
```
definelicht, line 556: Can we just make this a constant-sized array instead? We don't need to dynamic capabilities of `std::vector`

dace/codegen/targets/intel_fpga.py
```diff
@@ -558,24 +601,19 @@ def generate_host_function_epilogue(sdfg, state, host_stream):
...
-}""", sdfg, sdfg.node_id(state))
-
-    def generate_module(self, sdfg, state, name, subgraph, parameters,
-                        module_stream, host_header_stream, host_body_stream):
-
+                f"""\
+  std::vector<cl::Event> {kernel_name}_events;
+  for (auto &k : {kernel_name}_kernels) {{
+    {kernel_name}_events.emplace_back(k.ExecuteTaskFork({f'{kernel_deps_name}.begin(), {kernel_deps_name}.end()' if needs_synch else ''}));
+  }}
```
definelicht, line 610: Use constant size array that we can index into instead of a dynamic vector. We could even generate it as an initializer list:

```cpp
const std::array<cl::Event, 4> events = {
    kernels[0].ExecuteTaskFork(dependencies, dependencies + 3),
    kernels[1].ExecuteTaskFork(dependencies, dependencies + 3),
    kernels[2].ExecuteTaskFork(dependencies, dependencies + 3),
    kernels[3].ExecuteTaskFork(dependencies, dependencies + 3)
}
```

dace/codegen/targets/xilinx.py
```diff
@@ -2,6 +2,7 @@
 import collections
 import itertools
 import os
+import pdb
```
definelicht, line 5:
```suggestion
```

dace/codegen/targets/fpga.py
```diff
@@ -676,6 +851,162 @@ def deallocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
+        source_nodes = state.source_nodes()
+        max_kernels = default_kernel
+
+        # First step: assign a different Kernel ID
+        # to each source node which is not an AccessNode
+        for i, node in enumerate(source_nodes):
+            if isinstance(node, nodes.AccessNode):
+                continue
+
+            node._kernel = max_kernels
```
definelicht, line 888: Nooo... Please use an auxiliary data structure instead of secret fields on the graph, if at all possible. Can't we just keep a mapping in the codegen?

dace/codegen/targets/fpga.py
```diff
@@ -160,68 +163,242 @@ def on_target_used(self) -> None:
...
+        subgraphs = collections.defaultdict(
+            set)  # {kernel_id: {nodes in subgraph}}
+
+        # Go over the nodes and populate the kernels subgraphs
+        for node, node_state in graph.all_nodes_recursive():
+            if hasattr(node, "_kernel"):
+                if node._kernel in subgraphs:
+                    subgraphs[node._kernel].add(node)
+                else:
+                    subgraphs[node._kernel] = {node}
```
definelicht, line 193: I think you forgot to use defaultdict here? Then you don't need if branch

dace/codegen/targets/fpga.py
```diff
@@ -160,68 +163,242 @@ def on_target_used(self) -> None:
...
+                # AccessNodes can be read from multiple kernels, so
+                # check all out edges
+
+                start_nodes = [e.dst for e in node_state.out_edges(node)]
+                for n in start_nodes:
+                    if hasattr(n, "_kernel"):
+                        if n._kernel in subgraphs:
+                            subgraphs[n._kernel].add(node)
+                        else:
+                            subgraphs[n._kernel] = {node}
```
definelicht, line 206: Forgot to use defaultdict

dace/codegen/targets/fpga.py
```diff
@@ -160,68 +163,242 @@ def on_target_used(self) -> None:
...
+                    kernels_graph.add_edge(p, k)
+
+        subgraph_views = []
+        all_nodes = graph.nodes()
+
+        # Use topological sort to order kernels according to their dependencies
+        for kernel_id in nx.topological_sort(kernels_graph):
+            # Return the subgraph and the kernel id
+            subgraph_views.append((ScopeSubgraphView(
+                graph, [n for n in all_nodes if n in subgraphs[kernel_id]],
```
definelicht, line 229: Can't use use `subgraphs[kernel_id]` directly? Or do you need the ordering? (You might also be able to make this set a list instead, see other comment)

dace/codegen/targets/fpga.py
```diff
@@ -160,68 +163,242 @@ def on_target_used(self) -> None:
...
+        '''
+        from dace.sdfg.scope import ScopeSubgraphView
+
+        if not isinstance(graph, (dace.sdfg.SDFGState, ScopeSubgraphView)):
+            raise TypeError(
+                "Expected SDFGState or ScopeSubgraphView, got: {}".format(
+                    type(graph).__name__))
+
+        subgraphs = collections.defaultdict(
+            set)  # {kernel_id: {nodes in subgraph}}
```
definelicht, line 185: Why not use a list instead of a set? I don't think you can ever have duplicates, and then you can maintain the order for later

dace/codegen/targets/fpga.py
```diff
@@ -160,68 +163,242 @@ def on_target_used(self) -> None:
...
-        # Determine independent components
-        subgraphs = dace.sdfg.concurrent_subgraphs(state)
+        if not self._in_device_code:
+            kernels = []  # List of tuples (subgraph, kernel_id)
 
-        # Generate kernel code
-        shared_transients = set(sdfg.shared_transients())
+            # Start a new state code generation: reset previous dependencies if any
+            self._kernels_dependencies.clear()
+            self._kernels_names_to_id.clear()
```
definelicht, line 253: You can add the node->kernel mapping as an additional dictionary here

dace/codegen/targets/fpga.py
```diff
@@ -160,68 +163,242 @@ def on_target_used(self) -> None:
...
+
+                # Vitis HLS removes double underscores, which leads to a compilation
+                # error down the road due to kernel name mismatch. Remove them here
+                # to prevent this
+                while True:
+                    _kernel_name = kernel_name.replace("__", "_")
+                    if kernel_name == _kernel_name:
+                        break
+                    else:
+                        kernel_name = _kernel_name
```
definelicht, line 323: I know this is my own code, but now I realize how stupid it is, should be:

```python
kernel_name = re.sub(r"__+", "_", kernel_name)
```

dace/codegen/targets/fpga.py
```diff
@@ -287,7 +464,7 @@ def make_parameters(self, sdfg, state, subgraphs):
         # Build a dictionary of arrays to arbitrary data nodes referring to
         # them, needed to trace memory bank assignments and to pass to the array
         # allocator
-        data_to_node: Dict[str: dace.nodes.Node] = {}
+        data_to_node: Dict[str:dace.nodes.Node] = {}
```
definelicht, line 467:
```suggestion
        data_to_node: Dict[str, dace.nodes.Node] = {}
```

dace/codegen/targets/fpga.py
```diff
@@ -676,6 +851,162 @@ def deallocate_array(self, sdfg, dfg, state_id, node, function_stream,
                          callsite_stream):
         pass  # Handled by destructor
 
+    def compute_kernels(self, state: dace.SDFGState, default_kernel: int = 0):
```
definelicht, line 854: Maybe `partition_compute_kernels` or `annotate_compute_kernels`? Functions should preferably always start with a verb

dace/codegen/targets/fpga.py
```diff
@@ -676,6 +851,162 @@ def deallocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
+                                                 look_for_kernel_id=True)
+                        if kernel is not None:
+                            break
+                else:
+                    # Look at the successor nodes: because of the DFS visit, it may occur
+                    # that one of them has already an associated kernel ID. If this is the case, and
+                    # if the edge that connects this node with it is a tasklet-to-tasklet
+                    # edge, then we use that kernel ID. In all the other cases, we use a new one.
+
+                    # TODO: support more robust detection
```
definelicht, line 978: What are the current issues?

definelicht (changes requested):

dace/codegen/targets/fpga.py
TizianoDeMatteis: No, I need the ordering: when I spawn a kernel, all the predecessors must be already spawned. Kernel_id numbering does not reflect this

dace/codegen/targets/fpga.py
TizianoDeMatteis: There could be cases not covered by this. For example, I think that one should look also at the predecessors: if they are associated with a different kernel, and there is a tasklet-to-tasklet, maybe we don't want to generate a different kernel.
*Maybe* because I'm not sure...this could be further complicated as we want

dace/codegen/targets/fpga.py
TizianoDeMatteis: I don't like having to build another data structure for this, but I don't have an immediate, cleaner solution

dace/codegen/targets/fpga.py
TizianoDeMatteis: Because I can find the same node multiple times (e.g. Access Node)

dace/codegen/targets/fpga.py
definelicht: Put your concerns in the comment, currently it's not clear what is lacking in terms of robustness

dace/codegen/targets/fpga.py
definelicht: Why not make `subgraphs` a dictionary to lists instead of sets?

dace/codegen/targets/fpga.py
definelicht: I don't think you can? Aren't you iterating over `all_nodes_recursive`? This should never return the same node more than once

dace/codegen/targets/fpga.py
TizianoDeMatteis: How that can help here?

dace/codegen/targets/fpga.py
TizianoDeMatteis: You are right!

dace/codegen/targets/fpga.py
definelicht: How about flipping it, make `all_nodes = set(graph.nodes())`, and then do `n for n in subgraphs[kernel_id] if n in all_nodes` if subgraphs is a list, or if both are sets, just do `all_nodes & subgraphs[kernel_id]`

TizianoDeMatteis: Now I get. We were talking about two different things.

dace/codegen/targets/intel_fpga.py
TizianoDeMatteis: This would require assigning an increasing ID to all the modules (PEs), since the code for populating that vector is in `generate_module` (vendor specific, called by `fpga.py`)

definelicht: Aren't they ordered anyway?

dace/codegen/targets/fpga.py
TizianoDeMatteis: isn't compute a verb?

definelicht: I thought it was part of the noun "the compute kernels"?

definelicht (changes requested):

dace/codegen/targets/fpga.py
```diff
@@ -676,6 +849,177 @@ def deallocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
+        # disjoint paths that connect graph sinks and sources
+
+        scopes = state.scope_dict()
+
+        for e in state.dfs_edges(source_nodes):
+            if utils.unique_node_repr(state, e.dst) in self._node_to_kernel:
+                # Node has been already visited)
+                continue
+
+            def trace_back_edge(edge, state, look_for_kernel_id=False):
```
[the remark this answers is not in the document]
TizianoDeMatteis: Oh, I wasn't expecting this!!
What do you suggest? declare it not as nested function, but with a preceding underscore ?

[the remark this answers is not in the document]
definelicht: Since it's a fairly long function, that would be my suggestion, yes :-) But it's not a strict rule. You can also just put it outside the for loop.

dace/codegen/targets/intel_fpga.py
```diff
@@ -524,26 +541,51 @@ def generate_kernel_internal(self, sdfg, state, kernel_name, subgraphs,
...
-            sdfg, state_id)
         launch_async = Config.get_bool("compiler", "intel_fpga", "launch_async")
+
+        # Check if this kernel depends from other kernels
+        needs_synch = len(predecessors) > 0
+
+        if needs_synch:
+            # Build a vector containing all the events associated with the kernels from which this one depends
+            kernel_deps_name = f"deps_{kernel_name}"
+            host_stream.write(f"std::vector<cl::Event > {kernel_deps_name};")
```
definelicht, line 579:
```suggestion
            host_stream.write(f"std::vector<cl::Event> {kernel_deps_name};")
```

definelicht (changes requested):

definelicht (approved):

PR #730, opened 2021-05-24
Softmax transformations

am-ivanov:


tbennun:
Closed since it is handled in DaCeML.

PR #732, opened 2021-05-27
Improve handling of unregistered objects and custom/external extensions like transformations and library nodes

phschaad:
Provide a general deserialized representation for unregistered objects such that they can be re-serialized without any loss of information and handled without causing exceptions.
In particular: keep transformation histories intact when they contain unregistered transformations, and don't fail on deserializing SDFGs with unregistered LibraryNodes.

### DIFF
--- a/dace/serialize.py
+++ b/dace/serialize.py
@@ -139,8 +139,11 @@ def from_json(obj, context=None, known_type=None):
                         known_type.__name__)
 
     if t:
-        deserialized = _DACE_SERIALIZE_TYPES[t].from_json(obj, context=context)
-        if deserialized is None:
+        try:
+            deserialized = _DACE_SERIALIZE_TYPES[t].from_json(
+                obj, context=context
+            )
+        except Exception:
             deserialized = SerializableObject.from_json(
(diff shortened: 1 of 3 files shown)


tbennun (changes requested):
The Transformation constructor is great, but I'd opt to switch to a more general `SerializableObject` deserialized object such that it won't dive deeper into the JSON more than necessary, and would work with other things (unrecognized enums etc.).

dace/serialize.py
```diff
@@ -120,7 +139,12 @@ def from_json(obj, context=None, known_type=None):
                         known_type.__name__)
 
     if t:
-        return _DACE_SERIALIZE_TYPES[t].from_json(obj, context=context)
+        deserialized = _DACE_SERIALIZE_TYPES[t].from_json(obj, context=context)
+        if deserialized is None:
```
tbennun, line 143: hmm what if the value really is None? maybe we should look for raised exceptions? (for example, KeyError in _DACE_SERIALIZE_TYPES would also be supported etc.)

tbennun (approved):
LGTM

PR #739, opened 2021-06-07
NPBench related fixes

alexnick83:
This PR addresses issues related to mandelbrot2 and stockham_fft. The PR is made under the assumption that the method `symbolic.symstr` is converting symbolic expressions to cpp-compatible strings.

### DIFF
--- a/dace/codegen/targets/common.py
+++ b/dace/codegen/targets/common.py
@@ -4,6 +4,7 @@ from dace.sdfg import SDFG
 from dace.properties import CodeBlock
 from dace.codegen import cppunparse
 from typing import List, Optional, Set, Union
+import sympy
 import warnings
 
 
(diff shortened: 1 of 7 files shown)


dace/codegen/targets/common.py
```diff
@@ -38,8 +39,8 @@ def sym2cpp(s, arrayexprs: Optional[Set[str]] = None) -> Union[str, List[str]]:
     :return: C++-compilable expression or list thereof.
     """
     if not isinstance(s, list):
-        return cppunparse.pyexpr2cpp(symbolic.symstr(s, arrayexprs))
-    return [cppunparse.pyexpr2cpp(symbolic.symstr(d, arrayexprs)) for d in s]
+        return symbolic.symstr(s, arrayexprs)
+    return [symbolic.symstr(d, arrayexprs) for d in s]
```
tbennun, line 43: I'm a bit afraid of this change

tbennun (approved):

dace/codegen/targets/common.py
alexnick83: Well, there is no way to make this work otherwise. `symstr` return cpp strings and it is too deep to untangle (even uses a custom sympy printer to output cpp code). `symstr` also fixes min, max, floor, ceil, etc., so we cannot avoid using it. With the methods that we have now, this is the only way that produces cpp code with the above methods handled properly. The original code, took symbolic expressions, used `symstr` to make them (cpp) code and then tried to parse them with Python ast module. This fails whenever the generated cpp code is not valid Python code.

PR #740, opened 2021-06-07
General unroller

jnice-81:
Unrolled scheduler now supports unrolling maps anywhere in the SDFG, also if they contain nested SDFG's.
Adds 2 tests, one that checks nested unrolling with nested SDFG's, and one much simpler test, that unrolls a map, containing one tasklet.
The general concept of unrolling is to backup all the fields that might be affected by calls to replace, then replacing all the map parameters and generating the scope, followed by restoring the fields that were saved.


dace/sdfg/replace.py
```diff
@@ -24,6 +25,20 @@ def _replsym(symlist, symrepl):
                           if symbolic.issymbolic(dim) else dim)
     return symlist
 
+def deepreplace(subgraph: 'dace.sdfg.state.StateGraphView', name: str,
+                new_name: str):
```
definelicht, line 29: I'm a bit uncomfortable with this: The reason why we don't normally traverse into nested SDFGs when doing replacement is that symbols don't necessarily mean the same thing, `i` could refer to one thing in one SDFG, and something else in another.

Why did you need this? Perhaps we can find another solution, using `sdfg.constants` and `nested_sdfg.symbol_mapping` to solve this without doing the replacement in nested SDFGs? Or, at least, we should make sure that the symbol _does_ mean the same thing (by checking that it is not defined again in the nested SDFG and that the symbol mapping doesn't map it to something else).

tests/codegen/unroller_general_test.py
```diff
@@ -0,0 +1,107 @@
...
+                          src_conn="xout",
+                          )
+
+    return sdfg
+
+#code printing comments are left in case anyone would like to see the generated code at some point
+
+def test_unrolled_deeply_nested():
+    sdfg = create_deeply_nested_sdfg()
+    #print(Code(sdfg.generate_code()[0].code, language='cpp'))
```
definelicht, line 77:
```suggestion
```

tests/codegen/unroller_general_test.py
```diff
@@ -0,0 +1,107 @@
...
+
+    sdfg = ucopy.to_sdfg()
+    for node in sdfg.states()[0].nodes():
+        if (isinstance(node, dace.sdfg.nodes.MapEntry)):
+            node.schedule = dace.ScheduleType.Unrolled
+    return sdfg
+
+def test_unrolled_simple_map():
+    sdfg = create_simple_unrolled_sdfg()
+    #print(Code(sdfg.generate_code()[0].code, language='cpp'))
```
definelicht, line 99:
```suggestion
```

tests/codegen/unroller_general_test.py
```diff
@@ -0,0 +1,107 @@
...
+#code printing comments are left in case anyone would like to see the generated code at some point
+
+def test_unrolled_deeply_nested():
+    sdfg = create_deeply_nested_sdfg()
+    #print(Code(sdfg.generate_code()[0].code, language='cpp'))
+    passed = np.full((4, 100), 42.0, dtype=np.float32)
+    returns = np.zeros((4, 100), np.float32)
+    sdfg(x=passed, y=returns)
+    passed[1, 0] = 15.0
+    passed[0, 0] = 15.0
```
definelicht, line 82: What's on here, you're assigning some values after running the program?

tests/codegen/unroller_general_test.py
```diff
@@ -0,0 +1,107 @@
...
+                          topMapExit,
+                          yarr,
+                          memlet=mem.Memlet.from_array("y", sdfg.arrays["y"]),
+                          src_conn="xout",
+                          )
+
+    return sdfg
+
+#code printing comments are left in case anyone would like to see the generated code at some point
+
```
definelicht, line 74:
```suggestion
```

No need, it's easy to just check `.dacecache` and open the relevant code file, don't need to print it to the terminal :-)

dace/codegen/targets/unroller.py
```diff
@@ -8,7 +10,137 @@
...
 from dace.codegen import cppunparse
 from itertools import product
+from dace.sdfg import state
+import dace.subsets
+import dace.sdfg
+from dace.sdfg.replace import deepreplace
+from dace.sdfg import nodes as nd
+from dace import Config
 
+def backup_statescope_fields(
```
definelicht, line 20: Could you explain why we need these backup functions? We can we not just deep copy, change the deep-copied graph, then pass this to the dispatcher?

definelicht (changes requested):

tests/codegen/unroller_general_test.py
jnice-81: That's just convenience, since the sdfg is supposed to return exactly the same values that were passed with the only exception of the values at [1, 0], [0, 0]. In other words this generates the test values that should be returned. (Maybe I should copy it to a second variable for the sake of clarity)

dace/codegen/targets/unroller.py
jnice-81: The reason for this is that deepcopy doesn't work well with subscopes of the graph. There are functions and parts of the codegen, that access state data on the top-level (outside of the copied scope, e.g. on state or sdfg-level) which means you can still get the old objects for some type of calls. This then leads to all sorts of weird behavior.

jnice-81:
Do the failing tests here mean that there is something significantly wrong with this pull request?

dace/sdfg/replace.py
jnice-81: Assuming I would just add the value to sdfg.constants before generating it and afterwards removing them again, would this behave as expected? (I.e. use the constant if it's not defined, otherwise take whatever it is in the nested sdfgs)

definelicht:
> Do the failing tests here mean that there is something significantly wrong with this pull request?

Yes. You should be able to click the failing tests and see the errors. Hopefully they are all within the code you added, but if not, feel free to ask for advice on what went wrong.

dace/codegen/targets/unroller.py
definelicht: Ok, I understand. As usual, I'm not super comfortable with this, but also don't have an amazing alternative idea.

The only idea I have is whether we could limit the "backup" to be part of the replacement traversal only: When `replace` traverses the SDFG and finds an occurrences of the token that we are replacing, save a backup before replacing it, and return a tuple of replaces entities (similar to what you are doing now). Then, a separate "unreplace" ("replace inversion"? "rereplace"?) understands this format and reverts these changes. The advantage over what you are already doing is that we _only backup what we change_, so we don't risk backing up/restoring too much or too little.

Does this make any sense?

dace/sdfg/replace.py
definelicht: It _should_ work with nested SDFGs, I'm just saying that we need to be more careful there because of the name clash I mentioned above. If you fix the variable `w` in the outer SDFG, and then do a symbol mapping of, e.g., `{"w": "w"}` in the nested SDFG, does this still not produce a constant?

If not, then I would prefer the solution of adding them to `sdfg.constants` as you suggested.

jnice-81: I already implemented it now with constants. Should be fine. Copied from http://candcplusplus.com/c-variables-scopelocal-variables-and-global-variables:
"If a global and local variable have the same name and if we try to access the variable then note the local scope variable with the same name has a greater influence in that local region and so it will be used.However,if the local variable cease to exist then the global variable will be used instead.Consider the example below." Since dace always creates local regions whenever a variable is created, it should work out, but I will create a little addition to the test to verify it

dace/codegen/targets/unroller.py
jnice-81: Another suggestion: I'm not so sure if it is even so clever to change the actual sdfg during unrolling. In principle, we could do the same thing as with the constants in sdfgs, i.ee we create a local scope for each unrolled section (I believe thats already done in face), and then define a constexpr before each section? The only bad thing I could see is that something like nested unrolling would not work You could still have nested maps with unrolling, but you may not bind the range of the inner map to the variable defined by the outer map.

definelicht: If this works then yes absolutely, I'm not sure about Intel though, since OpenCL doesn't support `constexpr`. Do you want to try?

definelicht: It's definitely not as simple, I doubt you can even do shadowing of OpenCL constant variables. Honestly you're probably better off using `#define` and `#undef` or something nasty like this.

> how can I find out from the unroller if it is currently generating Intel FPGA code?

The dispatcher calls specific code generators, which then call other functions. Some of these other functions take the code generator currently running as an argument (typically a keyword argument `codegen`), but not necessarily all of them. In this case I don't see any easy way for you to know, since this is implemented as its own code generator, so by definition it's neither Xilinx or Intel.

Honestly, we might be better off holding off on Intel OpenCL support. If you can make it work for Xilinx and CPU, then we can leave this pain for another time. @TizianoDeMatteis what do you think?

dace/sdfg/replace.py
definelicht: So now we got rid of `deepreplace`? Did we even get rid of the backup stuff as well?

dace/codegen/targets/unroller.py
TizianoDeMatteis: Constants have always been problematic for Intel OpenCL: you can't do shadowing, and, we currently use`#define` to handle some strange situations. I'm also in favor of the `TODO(later)`

jnice-81: alright then I'l just do constexpr, add a TODO and get rid of the replacement/backup stuff

jnice-81:
Changes would again be ready to review. I play the a little bit dirty trick of adding the variables to the sdfg constants for the duration of the scope generation, even though they are actually not. This is required for HBM, because the magic index has to be evaluated during codegen.

dace/codegen/targets/unroller.py
```diff
@@ -23,31 +31,111 @@ def __init__(self, frame_codegen: DaCeCodeGenerator, sdfg: dace.SDFG):
...
+                            (node, True, copy.deepcopy(node.unique_name),
+                             copy.deepcopy(node.symbol_mapping),
+                             copy.deepcopy(node.sdfg.constants_prop)))
+                        node.unique_name = f"{node.unique_name}_{param}{paramval}"
+                    else:
+                        backup.append(
+                            (node, False, copy.deepcopy(node.sdfg.name),
+                             copy.deepcopy(node.symbol_mapping),
+                             copy.deepcopy(node.sdfg.constants_prop)))
+                        node.sdfg.name = f"{node.sdfg.name}_{param}{paramval}"
```
definelicht, line 81: Would it work to just always set _both_ `name` and `unique_name`, regardless of the configuration? It would clean up this whole function quite a lot

dace/codegen/targets/unroller.py
```diff
@@ -23,31 +31,111 @@ def __init__(self, frame_codegen: DaCeCodeGenerator, sdfg: dace.SDFG):
...
             while begin <= end:
                 l.append(begin)
                 begin += stride
             index_list.append(l)
 
+        sdfgconsts = sdfg.constants_prop
+        sdfg.constants_prop = copy.deepcopy(sdfgconsts)
+
         for indices in product(*index_list):
+            #backups = backup_statescope_fields(scope)
```
definelicht, line 110:
```suggestion
```

dace/codegen/targets/unroller.py
```diff
@@ -23,31 +31,111 @@ def __init__(self, frame_codegen: DaCeCodeGenerator, sdfg: dace.SDFG):
...
+                state_id,
+                function_stream,
+                callsite_stream,
+                skip_entry_node=True,
+                skip_exit_node=True,
+            )
             callsite_stream.write('}')
+            callsite_stream.write('}')
+            nsdfg_after_unroll(nsdfg_unroll_info)
+            #use_statescope_fields_backup(backups)
```
definelicht, line 139:
```suggestion
```

dace/codegen/targets/unroller.py
```diff
@@ -23,31 +31,111 @@ def __init__(self, frame_codegen: DaCeCodeGenerator, sdfg: dace.SDFG):
...
+                #dace.sdfg.replace(scope, str(param), str(index))
+                if nsdfg_unroll_info is None:
+                    nsdfg_unroll_info = nsdfg_prepare_unroll(
+                        scope, str(param), str(index))
+                else:
+                    nsdfg_prepare_unroll(scope, str(param), str(index))
+                callsite_stream.write(
+                    "constexpr %s %s = %s;\n" %
+                    ('long long', param,
+                     dace.codegen.targets.common.sym2cpp(index)), sdfg)
```
definelicht, line 123: Use f-string formatting:

```python
f"constexpr long long {param} {dace.codegen.targets.common.sym2cpp(index)}"
```

tests/codegen/unroller_general_test.py
```diff
@@ -0,0 +1,113 @@
...
+    )
+    state.add_memlet_path(
+        nsdfg_node,
+        topMapExit,
+        yarr,
+        memlet=mem.Memlet.from_array("y", sdfg.arrays["y"]),
+        src_conn="xout",
+    )
+
+    #sdfg.view()
```
definelicht, line 75:
```suggestion
```

definelicht (changes requested):

dace/codegen/targets/unroller.py
jnice-81: Theoretically yes, but I'm a bit afraid to do so, because I'm not so sure how unique_name is supposed to be used. However at the moment the only place that uses this at all is the nested sdfg generation, so it's probably fine.

jnice-81:
Again everything updated.

definelicht (approved):

definelicht:
Looks good for me, but I would just like for Tal to take a look before we merge

jnice-81:
By the way which script do I have to run to check everything in tests/codegen? Respectively which tests can I even run locally before declaring that it works?

definelicht:
I'll trigger the CI. You can test your own test locally, and then if anything else has broken you can debug it locally.

tbennun (changes requested):
Minor comments and a question

dace/codegen/targets/unroller.py
```diff
@@ -23,31 +30,86 @@ def __init__(self, frame_codegen: DaCeCodeGenerator, sdfg: dace.SDFG):
...
 
         # Register dispatchers
-        self._dispatcher.register_map_dispatcher(dace.ScheduleType.Unrolled, self)
+        self._dispatcher.register_map_dispatcher(dace.ScheduleType.Unrolled,
+                                                 self)
 
     def get_generated_codeobjects(self):
         return []
 
+    #TODO: Expand the unroller so it can also generate openCL code
```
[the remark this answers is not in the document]
jnice-81: Yes, both Tiziano and Johannes wanted to implement this at some later point when it is actually needed

dace/codegen/targets/unroller.py
```diff
@@ -23,31 +30,86 @@ def __init__(self, frame_codegen: DaCeCodeGenerator, sdfg: dace.SDFG):
...
+                    nsdfg_unroll_info = nsdfg_prepare_unroll(
+                        scope, str(param), str(index))
+                else:
+                    nsdfg_prepare_unroll(scope, str(param), str(index))
+                callsite_stream.write(
+                    f"constexpr long long {param} = {dace.codegen.targets.common.sym2cpp(index)};\n",
+                    sdfg)
+                sdfg.add_constant(param, int(index))
+
+            callsite_stream.write('{')
```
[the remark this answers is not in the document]
jnice-81: I'm actually not sure, but I don't think it's a problem, because you anyway have to be in a map to use this schedule type, which means you are inside function-code, with the exception of nested sdfg's, but they are handled seperately.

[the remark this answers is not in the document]
jnice-81: In fact I did not even change that, one curly brace was already there, I just added doing a second one

definelicht (approved):

tbennun (approved):

PR #741, opened 2021-06-08
Add enum properties and enums which are safe from custom values

phschaad:
If a value not registered in an `EnumProperty`'s choices is deserialized, it is deserialized as `Undefined` in the corresponding enum property's data type. The original value is remembered, such that it can be restored upon subsequent re-serialization. This means that custom/unregistered values can be used for enum properties without causing errors, while remaining intact through de-/re-serialization of the SDFG.


tbennun (approved):
Nice

PR #750, opened 2021-06-14
Append library node expansions to an SDFG's history

phschaad:


tbennun (changes requested):
I'd say this should happen after `apply` is complete.

phschaad:
That's a good point. Should I make the same change to all transformation applications? At the moment, `apply_pattern` appends before applying too.

tbennun:
@phschaad sure

dace/transformation/transformation.py
```diff
@@ -535,6 +537,39 @@ def apply(self, sdfg, *args, **kwargs):
...
+            'type': 'ExpandTransformation',
+            'transformation': type(self).__name__,
+            'classpath': nd.full_class_path(self),
+            **props
+        }
+
+    @staticmethod
+    def from_json(json_obj: Dict[str, Any],
+                  context: Dict[str, Any] = None) -> 'ExpandTransformation':
+        xform = pydoc.locate(json_obj['classpath'])
```
tbennun, line 552: what if it doesn't exist? does it become a serializableobject?

phschaad: Yes, exactly - it diverts to `Transformation.from_json`, which in turn fails and generates a SerializableObject

tbennun (approved):

PR #756, opened 2021-06-17
Source Mapping

benibenj:
Creates 2 mappings which are saved in the folder ".dacecache/functionName/map"
- Python source line -> list of SDFG nodes (sorted by priority)
- SDFG node -> C++ range of lines (from, to)

After the mapping is complete DACE sends the path to the cache, the name of the function, and the target name to the port DACE_port(if set as env variable) for the vscode extension to utilize


dace/codegen/codegen.py
```diff
@@ -98,6 +99,8 @@ def generate_code(sdfg) -> List[CodeObject]:
     # Before compiling, validate SDFG correctness
     sdfg.validate()
 
+    MapPython(sdfg.to_json(), sdfg.name)
```
tbennun, line 102: 1. `sdfg.to_json()` is an expensive operation, easier if `sdfg` is used directly.
2. operation happening in class constructor is a weird behavior
3. it ends up staying unused, as the object is not assigned to anything?

dace/codegen/codeobject.py
```diff
@@ -53,6 +54,9 @@ def __init__(self,
         self.linkable = linkable
         self.environments = environments or set()
 
+        if (language == "cpp") & (title == "Frame"):
+            MapCpp(code, name, self.target.target_name)
```
tbennun, line 58: same comment (2) from above. Better if a free function is used.

dace/codegen/codeobject.py
```diff
@@ -53,6 +54,9 @@ def __init__(self,
         self.linkable = linkable
         self.environments = environments or set()
 
+        if (language == "cpp") & (title == "Frame"):
```
tbennun, line 57: Not standard python. Should be `if language == 'cpp' and title == 'Frame':`

Also see coding guidelines at [CONTRIBUTING.md](https://github.com/spcl/dace/blob/master/CONTRIBUTING.md)

dace/frontend/python/newast.py
```diff
@@ -28,6 +29,7 @@
 from dace.properties import LambdaProperty, CodeBlock
 from dace.sdfg import SDFG, SDFGState
 from dace.symbolic import pystr_to_symbolic
+from dace.sourcemap import temporaryInfo
```
tbennun, line 32: please avoid importing functions/classes directly (we try to avoid this for any new code that gets added). See coding guidelines.

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
+import re
```
tbennun, line 1: Missing copyright, see guidelines

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+import json
+import os
+import socket
+from pathlib import Path
+from dace.config import Config
```
tbennun, line 6: direct import

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+    '''
+    path = ".dacecache/" + name
+
+    if not os.path.exists(path):
+        return
+
+    try:
+        os.remove(path + "/map/tmp.json")
+    except OSError:
+        print("removing error 1")
```
tbennun, line 79: does this happen? if so it should probably raise, no?

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+        remove_cache and
+        len(os.listdir(path)) == 1 and
+        len(os.listdir(path + "/map")) == 0
+    ):
+        if os.path.exists(path):
+            try:
+                os.rmdir(path + "/map")
+                os.rmdir(path)
+            except OSError:
+                print("removing error 2")
```
tbennun, line 92: same

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+                return      
+
+def send(data: json):
+    """ Sends a json object to the port given as the 
+        env variable DACE_port. If the port isn't set
+        we won't send anything
+            :param data, json object to send
+    """
+
+    if("DACE_port" not in os.environ):
```
tbennun, line 102:
```suggestion
    if 'DACE_port' not in os.environ:
```

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+                os.rmdir(path)
+            except OSError:
+                print("removing error 2")
+                return      
+
+def send(data: json):
+    """ Sends a json object to the port given as the 
+        env variable DACE_port. If the port isn't set
+        we won't send anything
+            :param data, json object to send
```
tbennun, line 99: indentation is strange, also not valid sphinx documentation

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+    HOST = socket.gethostname()
+    PORT = os.environ["DACE_port"]
+
+    data_bytes = bytes(json.dumps(data), "utf-8")
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.connect((HOST, int(PORT)))
+        s.sendall(data_bytes)
+
+
+class MapCreater:
```
tbennun, line 114:
```suggestion
class MapCreator:
```

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+    def __init__(self, name):
```
tbennun, line 115: what does this class do? needs docstring

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+            "/map/map_" + language + ".json"
+        )
+
+        with open(path, "w") as json_file:
+            json.dump(self.map, json_file, indent=4)
+
+        return os.path.abspath(folder)
+
+
+class MapCpp(MapCreater):
```
tbennun, line 135: same (creator, docstring)

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+    def __init__(self, code, name, target_name):
+        super(MapCpp, self).__init__(name)
+        self.code = code
+        self.mapper()
+        self.map['target'] = target_name
+
+        folder = self.save("cpp")
```
tbennun, line 142: code that has side-effects on `__init__` is generally bad practice

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+    def get_identifiers(self, line):
+        line_identifier = re.findall(
+            r'(\/\/\/\/__DACE:[0-9]:[0-9]:[0-9](,[0-9])*)',
+            line
+        )
+        # remove tuples
+        return [x or y for (x, y) in line_identifier]
+
+
+class MapPython(MapCreater):
```
tbennun, line 212: same comment as MapCpp

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+        return nodes
+
+    def get_identifiers(self, line):
+        line_identifier = re.findall(
+            r'(\/\/\/\/__DACE:[0-9]:[0-9]:[0-9](,[0-9])*)',
+            line
+        )
+        # remove tuples
+        return [x or y for (x, y) in line_identifier]
```
tbennun, line 209: strange behavior. Could you document with an example?

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+                    # SDFGs created with the API don't have tmp files
+                    if other_tmp is not None:
+                        remove_tmp(other_sdfg_name, True)
+                        ranges.append(
+                            (other_tmp["start_line"], other_tmp["end_line"])
+                        )
+
+                self.create_mapping(ranges)
+
+        self.save("py")
```
tbennun, line 250: side-effect in ctor

dace/sourcemap.py
```diff
@@ -0,0 +1,344 @@
...
+    def sdfg_debuginfo(self, graph, sdfg_id=0, state_id=0):
+        """ 
+        Returns an array of Json objects of sdfg nodes with their debuginfo
+        as also the sdfg, state and node id
+        """
+        if(sdfg_id is None):
+            sdfg_id = 0
+
+        mapping = []
+        for node in graph["nodes"]:
```
tbennun, line 286: definitely easier to do this on the original SDFG object rather than the json

tbennun (changes requested):
Good job. There are some critical points to fix regarding code style/behavior, and also please add yourself to the AUTHORS list in the file on the root folder.

dace/sourcemap.py
benibenj: Currently, the regex expression returns a list of 2 groups (list of tuples) where one element is empty. This is due to the identifiers also indicating edges ////DACE__0:0:0,2. I'm considering either ignoring edges, just taking the from node of the edge or I'll try to rewrite the regex expression to only return one group (I'm not very skilled in regex expressions)

dace/sourcemap.py
```diff
@@ -61,45 +68,34 @@ def get_tmp(name):
...
-    except OSError:
-        print("removing error 1")
-        return
-    
-    if (
-        remove_cache and
-        len(os.listdir(path)) == 1 and
-        len(os.listdir(path + "/map")) == 0
-    ):
+    os.remove(path + "/map/tmp.json")
```
tbennun, line 83: for compatibility, please use 
```suggestion
    os.remove(os.path.join(path, 'map', 'tmp.json'))
```

dace/sourcemap.py
```diff
@@ -61,45 +68,34 @@ def get_tmp(name):
...
+    if (remove_cache and len(os.listdir(path)) == 1
+            and len(os.listdir(path + "/map")) == 0):
         if os.path.exists(path):
-            try:
-                os.rmdir(path + "/map")
-                os.rmdir(path)
-            except OSError:
-                print("removing error 2")
-                return      
+            os.rmdir(path + "/map")
```
tbennun, line 88:
```suggestion
            os.rmdir(os.path.join(path, 'map'))
```

dace/sourcemap.py
```diff
@@ -111,82 +107,105 @@ def send(data: json):
         s.sendall(data_bytes)
 
 
-class MapCreater:
-    def __init__(self, name):
-        self.name = name
-        self.map = {}
+# TODO sdfg.build_folder()
```
tbennun, line 110: leftover? please fix

dace/sourcemap.py
```diff
@@ -111,82 +107,105 @@ def send(data: json):
...
+def save(language: str, name: str, map: dict) -> str:
+    """ Saves the mapping in the map folder of
+        the corresponding SDFG
+        :param language: used for the file name to save to: py -> map_py.json
+        :param name: name of the SDFG
+        :param map: the map object to be saved
+        :return: absolute path to the cache folder of the SDFG
+    """
+    folder = create_cache(name)
+    path = os.path.abspath(folder + "/map/map_" + language + ".json")
```
tbennun, line 120:
```suggestion
    path = os.path.abspath(os.path.join(folder, 'map', 'map_' + language + '.json'))
```

dace/sourcemap.py
tbennun: my comment would be just to add a comment above the list comprehension saying "This ignores edges of the format ..." or so

tbennun (changes requested):

tbennun (approved):

PR #762, opened 2021-06-21
Support for HBM on Xilinx FPGA's (pull-request 2)

jnice-81:
Add Support for HBM for Xilinx FPGA's

Changes:

sdfg/utils: 
- Added several utility functions for HBM-arrays/subsets to utils

targets/cpp: 
- Modifed ptr to generate distinct names for HBM-arrays. The function array_interface_variable is only called from ptr.
- Modified several functions in targets/cpp slighly to support HBM-arrays

targets/fpga:
- Modifed several of the functions which generate arguments/definitions and so on for variables to create multiple of those objects in case of HBM-arrays
- make_parameters recognizes and passes HBM and DDR in it's bank_assignments return value
- emit_copy has support for Host2Device, Device2Host, Device2Device and on Device copies from/to HBM-arrays and also between HBM and DDR. Writing directly from HBM to Streams is not yet supported (it can be done using tasklets with single bank memlets)
- copy_memory "disolves" the magic index for HBM-multibank memlets and only calls emit_copy with single bank HBM memlets
- found a possible old bug (maybe this is untested) at line 1423. Altered it to what seems to make sense there.

targets/xilinx:
- Modifed several of the functions which generate arguments/definitions and so on for variables to create multiple of those objects in case of HBM-arrays
- get_generated_codeobjects: HBM interfaces are declared

targets/cpu:
- Modifed calls to cpp.ptr in some locations
- Added NotImplementedException if nested sdfg's on cpu side deal with HBM-arrays. This will likely be supported in the next PR, but at the moment it seems to much like a nice to have.

sdfg/validation:
- Check that no HBM multibank memlets are attached to a tasklet
- Check that the bank/hbmbank location property is valid for the given array (correct type, correct arrayshape, ...)
- Check that all HBM-memlets have constant evaluatable indices

dace/data/*
- Added the value type of the locationproperty to str from pystr_to_symbolic. Needed because of the serialization for ranges

transformation/fpga_transform_state
- A few smaller modifications so that hbmbank is correctly set

tests/fpga/*
- Added several tests for all the mentioned changes above


dace/codegen/targets/fpga.py
```diff
@@ -1411,10 +1498,50 @@ def copy_memory(self, sdfg, dfg, state_id, src_node, dst_node, edge,
...
+                        mem.src_subset, sdfg)
+                if dst_is_hbm:
+                    bankbeg, bankend = utils.get_multibank_ranges_from_subset(
+                        mem.dst_subset, sdfg)
+                num_accessed_banks = bankend - bankbeg
+                oldmem = copy.deepcopy(mem)
+                for i in range(num_accessed_banks):
+                    src_index = oldmem.src_subset[0][0] + i
+                    dst_index = oldmem.dst_subset[0][0] + i
+                    #Support for ignoring the magic index if it's not required e.g on host
```
definelicht, line 1528:
```suggestion
                    # Support for ignoring the distributed index if it's not required, e.g., on the host
```

dace/codegen/targets/xilinx.py
```diff
@@ -258,15 +261,16 @@ def make_vector_type(dtype, is_const):
...
-                             is_output,
-                             with_vectorization,
-                             interface_id=None):
+    def make_kernel_argument(data: dt.Data,
+                             var_name: str,
+                             subset_info: Union[int, subsets.Subset],
+                             sdfg: SDFG,
+                             is_output: bool,
+                             with_vectorization: bool,
+                             interface_id: "Union[int, list[int]]" = None):
```
definelicht, line 270:
```suggestion
                             interface_id: Union[int, list[int]] = None):
```

dace/codegen/targets/xilinx.py
```diff
@@ -435,15 +439,39 @@ def generate_no_dependence_pre(kernel_stream,
...
+                                  is_array_interface=True)
+        if newname is not None:
+            newname = cpp.ptr(
+                var_name,
+                None,
+                None,
+                sdfg,
+                True,
+                self._dispatcher,
+                is_array_interface=(defined_type == DefinedType.ArrayInterface))
```
definelicht, line 474: Restructure this so you only have one call to `ptr`, and just set the arguments accordingly

dace/codegen/targets/xilinx.py
```diff
@@ -493,12 +538,12 @@ def generate_kernel_boilerplate_pre(self, sdfg, state_id, kernel_name,
                 interface_name = "gmem{}".format(num_mapped_args)
                 kernel_stream.write(
                     "#pragma HLS INTERFACE m_axi port={} "
-                    "offset=slave bundle={}".format(var_name, interface_name),
-                    sdfg, state_id)
+                    "offset=slave bundle={}".format(var_name, interface_name), )
```
definelicht, line 541: What happened here?

dace/transformation/interstate/fpga_transform_state.py
```diff
@@ -185,12 +187,17 @@ def apply(self, sdfg):
                         allow_conflicts=desc.allow_conflicts,
                         strides=desc.strides,
                         offset=desc.offset)
+                    if ("hbmbank" in desc.location):
+                        fpga_array[1].location["hbmbank"] = desc.location[
+                            "hbmbank"]
+                        desc.location.pop("hbmbank")
```
definelicht, line 193: Can you copy all of `location` instead?

tests/fpga/hbm_copy_fpga.py
```diff
@@ -0,0 +1,129 @@
...
+#A test checking copies involving HBM-arrays in some way
+
+
+def print_result(a, c, expect):
+    print("A:")
+    print(a)
+    print("C")
+    print(c)
+    print("E")
+    print(expect)
```
definelicht, line 17: This is not used. Maybe clean up this test a bit

tests/fpga/hbm_dynamic_memlets.py
```diff
@@ -0,0 +1,63 @@
...
+
+def create_dynamic_memlet_sdfg():
+    sdfg = dace.SDFG("dyn_memlet")
+    state: dace.SDFGState = sdfg.add_state("dyn_memlet")
+    xarr = state.add_array("x", [4, 10], dace.int32)
+    sdfg.arrays["x"].location["hbmbank"] = sbs.Range.from_string("0:4")
+    yarr = state.add_array("y", [4, 10], dace.int32)
+    sdfg.arrays["y"].location["hbmbank"] = sbs.Range.from_string("4:8")
+
+    hbmmapenter, hbmmapexit = state.add_map("hbmmap", dict(k="0:4"),
```
definelicht, line 18: Use underscores!

tests/fpga/hbm_vadd_fpga.py
```diff
@@ -0,0 +1,134 @@
...
+                               ndim=1,
+                               unroll_map_inside=False,
+                               sdfgname="vadd_hbm"):
+    N = dace.symbol("N")
+    M = dace.symbol("M")
+    S = dace.symbol("S")
+
+    sdfg = dace.SDFG(sdfgname)
+    state = sdfg.add_state('vadd_hbm', True)
+    shape = [bankcountPerArray, N]
```
definelicht, line 20: Follow naming convention `bank_count_per_array`

tests/fpga/hbm_validation.py
```diff
@@ -0,0 +1,96 @@
...
+from dace import subsets as sbs, dtypes, memlet as mem
+import dace
+import numpy as np
+from dace import subsets
+from dace.sdfg import nodes as nd
+
+#A test to check the changes to the validation required for the support for HBM
+
+
+def checkInvalid(sdfg, exceptiontype):
```
definelicht, line 13:
```suggestion
def assert_validation_failure(sdfg, exceptiontype):
```

tests/fpga/hbm_validation.py
```diff
@@ -0,0 +1,96 @@
...
+            if isinstance(node, nd.MapEntry):
+                node.map.schedule = dtypes.ScheduleType.Unrolled
+    sdfg.arrays["input"].location["hbmbank"] = subsets.Range.from_string("0:12")
+    sdfg.arrays["output"].location["hbmbank"] = subsets.Range.from_string(
+        "12:24")
+    sdfg.apply_fpga_transformations(validate=False)
+    sdfg.validate()
+
+
+def multitaskletTest():
```
definelicht, line 45:
```suggestion
def multi_tasklet():
```

We can't name stuff with "test" at the moment because pytest will pick.

Actually -- maybe you _could_ use PyTest for this? Does this even need the FPGA tools to run?

tests/fpga/xilinx_test.py
```diff
@@ -81,6 +81,15 @@
         "kernels_inside_component_2_1", "kernels_inside_components_0_1",
         "kernels_lns_inside_component_1", "multiple_kernels_multiple_states"
     ], True, False, []),
+    ("tests/fpga/hbm_validation.py", [], False, False, []),
```
definelicht, line 84: Maybe move to pytest if this doesn't depend on the FPGA tools?

tests/fpga/hbm_validation.py
jnice-81: No not really, just put it there because it logically belongs to FPGA. Can I just add it to the top level folder and name all functions I want to be tested with "test" at the end?

dace/codegen/targets/cpp.py
```diff
@@ -22,11 +22,82 @@
...
+                dataname, offset_cppstr))
+
+        if def_type == DefinedType.Scalar:
+            return "{}&{}".format(dt, expr)
+        else:
+            return dataname
+    else:
+        raise NotImplementedError("copy_expr not implemented "
+                                  "for connector type: {}".format(def_type))
+"""
```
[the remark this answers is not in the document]
jnice-81: Ahem this was my first bugfix. Forgot to remove it

tests/fpga/hbm_validation.py
jnice-81: No not really, just put it there because it logically belongs to FPGA. Can I just add it to the top level folder and name all functions I want to be tested with "test" at the end?

dace/codegen/targets/cpp.py
```diff
@@ -1374,11 +1487,18 @@ def array_interface_variable(var_name: str,
...
     else:
         # We might call this before the variable is even defined (e.g., because
         # we are about to define it), so if the dispatcher is not passed, just
         # return the appropriate string
-        return ptr_out if is_write else ptr_in
+        result = ptr_out if is_write else ptr_in
+    #Append the interface id, if provided
+    if interface_id is not None:
+        if isinstance(interface_id, tuple):
+            result = f"{result}_{interface_id[accessed_subset]}"
```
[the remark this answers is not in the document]
jnice-81: I think it's nicer to leave it here, because usually at the callsite you don't really know wheter the interface_id you are passing is an integer or a list, so you'd need to do this check there. Would it work to make those two parameters a tuple instead?

dace/codegen/targets/cpu.py
```diff
@@ -1139,24 +1140,28 @@ def memlet_definition(self,
...
+
+        try:
+            ptr = cpp.ptr(
+                memlet.data, desc, memlet.subset, sdfg, output,
+                self._dispatcher, 0, var_type == DefinedType.ArrayInterface
+                and not isinstance(desc, data.View))
+        except ValueError:
+            raise cgx.CodegenError(
+                "Only memlets accessing a single HBM-bank may be "
+                f"attached to a tasklet. See {str(memlet)}")
```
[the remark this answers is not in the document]
jnice-81: In ptr we dont know that it is because of a tasklet. But really this is verified in the validation, so I guess it should be save to just remove this seperate check

dace/codegen/targets/cpu.py
```diff
@@ -1139,24 +1140,28 @@ def memlet_definition(self,
...
 
-        # Special case: ArrayInterface, append _in or _out
-        _ptr = ptr
-        if var_type == DefinedType.ArrayInterface:
-            # Views have already been renamed
-            if not isinstance(desc, data.View):
-                ptr = cpp.array_interface_variable(ptr, output,
-                                                   self._dispatcher)
-        if expr != _ptr:
+        #TODO: May be dangerous like this. oldcode expected that ptr does not contain the arrayinterfacevariable stuff (why?)
```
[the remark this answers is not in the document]
jnice-81: Well all the tests work, so should not be a problem. I actually didn't get why it was like this in the first place

dace/codegen/targets/fpga.py
```diff
@@ -579,39 +587,74 @@ def make_parameters(self, sdfg, state, subgraphs):
...
                         trace_name = trace_node.data
                         trace_desc = trace_node.desc(trace_sdfg)
-                        if "bank" in trace_desc.location:
-                            trace_bank = trace_desc.location["bank"]
-                            if (bank is not None and bank != trace_bank):
+                        if "bank" in trace_desc.location or "hbmbank" in trace_desc.location:
+                            if "bank" in trace_desc.location and "hbmbank" in trace_desc.location:
+                                raise cgx.CodegenError(
+                                    "Found memory bank specifier "
+                                    f"with bank and hbmbank defined for {trace_name}"
```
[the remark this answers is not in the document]
jnice-81: I decided to do it with a dot as seperator e.g. "hbm.4:8" and "ddr.3", since at least the ranges are easier to read like this (I think).

dace/codegen/targets/cpu.py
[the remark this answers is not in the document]
definelicht: Let's remove it here, then. My problem was with `except ValueError`, as this is too broad in case something else throws a `ValueError`!

dace/sdfg/utils.py
```diff
@@ -1030,6 +1031,119 @@ def get_next_nonempty_states(sdfg: SDFG, state: SDFGState) -> Set[SDFGState]:
...
+def modify_subset_magic(array: dt.Data,
+                        subset: Union[sbs.Subset, list, tuple],
+                        change: int,
+                        force: bool = False):
+    """
+    Applies changes to magic indices from a subset if array is a HBM-array, otherwise
+    returns subset. subset is deepcopied before any modification to it is done.
+    :param change: magic index is set to this value, unless it's -1 in which case
+        the magic index is completly removed
+    :param force: Modify the first index even if this is not a HBM array
```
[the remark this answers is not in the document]
definelicht: What about moving the if-condition out, so we only call this function when we actually want to apply it? It's not very intuitive that it sometimes does nothing, depending on the inputs.

dace/codegen/targets/fpga.py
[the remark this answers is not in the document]
definelicht: I would prefer either without the dot `ddr3` or with indices `ddr[3]`. @TizianoDeMatteis what do you think?

dace/codegen/targets/fpga.py
```diff
@@ -579,39 +587,74 @@ def make_parameters(self, sdfg, state, subgraphs):
...
-                        if ("bank" not in outer_desc.location or
-                                str(outer_desc.location["bank"]) != str(bank)):
+                        okhbm, okbank = False, False
+                        if banktype == "HBM":
+                            okhbm = "hbmbank" in outer_desc.location and outer_desc.location[
+                                "hbmbank"] == bank
+                        if banktype == "DDR":
+                            okbank = "bank" in outer_desc.location and str(
+                                outer_desc.location["bank"]) == str(bank)
+                        if not (okhbm or okbank):
```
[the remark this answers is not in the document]
definelicht: Was this resolved?

dace/codegen/targets/fpga.py
```diff
@@ -775,23 +818,43 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
+                                    .format(allocname, nodedesc.dtype.ctype,
+                                            bankoffset + bank_index,
+                                            cpp.sym2cpp(arrsize)))
+                            elif memory_bank_arg_type == "DDR":
+                                result_alloc.write(
+                                    "{} = __state->fpga_context->Get()."
+                                    "MakeBuffer<{}, hlslib::ocl::Access::readWrite>"
+                                    "(hlslib::ocl::MemoryBank::bank{}, {});".
+                                    format(allocname, nodedesc.dtype.ctype,
+                                           bankoffset, cpp.sym2cpp(arrsize)))
```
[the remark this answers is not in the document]
definelicht: Wait, why can Intel not use the new API?

dace/codegen/targets/fpga.py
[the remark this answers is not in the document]
jnice-81: Yes.

dace/codegen/targets/fpga.py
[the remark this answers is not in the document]
jnice-81: It's only implemented and tested for Xilinx, since Intel doesn't have HBM... In principle one could probably just redirect from there to the old api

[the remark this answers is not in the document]
definelicht: Yes, we should do this. I didn't realize that Intel was not yet supported by the new API - it should be!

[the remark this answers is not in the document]
jnice-81: Ok I will create a small PR for this then

dace/sdfg/validation.py
```diff
@@ -160,7 +218,29 @@ def validate_state(state: 'dace.sdfg.SDFGState',
     for nid, node in enumerate(state.nodes()):
         # Node validation
         try:
-            node.validate(sdfg, state)
+            if isinstance(node, nd.NestedSDFG):
+                #Add constant's defined by unrolled maps temporary as sdfg-level
+                #constants with a dummy value and remove the corresponding symbol-bindings
+                #on the nsdfg
```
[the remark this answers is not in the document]
jnice-81: Would it be an option to add some dict argument to the validate call (default None), that can contain options for the validation? Then one could simply pass the scope local constants to nested sdfg's via this dict, instead of somehow adding them to the sdfg. Not so nice to change the validate interface just for this, but probably still nicer than modifying the SDFG?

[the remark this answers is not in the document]
definelicht: @tbennun Do you have an opinion here?

dace/sdfg/utils.py
```diff
@@ -1030,6 +1031,114 @@ def get_next_nonempty_states(sdfg: SDFG, state: SDFGState) -> Set[SDFGState]:
...
+            raise ValueError(
+                f"Failed to parse {val} as value for location['bank']")
+        split[0] = split[0].upper()
+
+        if (split[0] == "DDR" or split[0] == "HBM"):
+            return (split[0], split[1])
+        else:
+            raise ValueError(
+                f"{split[0]} is an invalid bank type for location['bank']. Supported are HBM and DDR."
+            )
```
definelicht, line 1137: In the error message, write specifically what format is expected, e.g., `f"Failed to parse memory bank specifier \"{val}\". Expected format <type>.<id> (e.g., ddr.3 or hbm.0:4)."`

dace/sdfg/validation.py
```diff
@@ -55,6 +58,46 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG'):
                     "storage type. Please use a different storage location." %
                     name, sdfg, None)
 
+            #Check for valid bank assignments
```
definelicht, line 61:
```suggestion
            # Check for valid bank assignments
```

dace/sdfg/validation.py
```diff
@@ -55,6 +58,46 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG'):
...
+            try:
+                bankassignment = sdutil.parse_location_bank(desc)
+            except ValueError as e:
+                raise InvalidSDFGError(
+                    f"Failed to parse location['bank'] for "
+                    f" array {name}: {str(e)}", sdfg, None)
```
definelicht, line 67: Don't catch and rethrow, just let the exception propagate (or if you insist, at least copy the error message directly from `e`).

dace/sdfg/validation.py
```diff
@@ -55,6 +58,46 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG'):
                     "storage type. Please use a different storage location." %
                     name, sdfg, None)
 
+            #Check for valid bank assignments
+            try:
+                bankassignment = sdutil.parse_location_bank(desc)
```
definelicht, line 63:
```suggestion
                bank_assignment = sdutil.parse_location_bank(desc)
```

Please go over the code and make sure you follow the formatting and naming conventions everywhere so I don't have to comment on each one :-)

dace/sdfg/validation.py
```diff
@@ -55,6 +58,46 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG'):
...
+                    f"Failed to parse location['bank'] for "
+                    f" array {name}: {str(e)}", sdfg, None)
+            if bankassignment is not None:
+                if bankassignment[0] == "DDR":
+                    try:
+                        tmp = int(bankassignment[1])
+                    except ValueError:
+                        raise InvalidSDFGError(
+                            "bank assignment must be convertible to int"
+                            f" for array {name}", sdfg, None)
```
definelicht, line 75: Just let the error propagate

dace/sdfg/validation.py
```diff
@@ -55,6 +58,46 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG'):
...
+                    try:
+                        tmp = subsets.Range.from_string(bankassignment[1])
+                    except SyntaxError:
+                        raise InvalidSDFGError(
+                            "locationproperty 'bank' must be convertible to subsets.Range"
+                            f" for array {name} since it uses HBM", sdfg, None)
+                    try:
+                        low, high = sdutil.get_multibank_ranges_from_subset(
+                            bankassignment[1], sdfg)
+                    except:
```
definelicht, line 86: Throw this error from inside the function. Never use `except` without catching a specific error

dace/sdfg/validation.py
```diff
@@ -370,8 +466,26 @@ def validate_state(state: 'dace.sdfg.SDFGState',
                 )
         ########################################
 
+    #Check that HBM Magic Indices are evaluatable
+
```
definelicht, line 470:
```suggestion
```

dace/sdfg/validation.py
```diff
@@ -370,8 +466,26 @@ def validate_state(state: 'dace.sdfg.SDFGState',
...
+        def validate_hbm_subset(magic_subset: subsets.Subset):
```
definelicht, line 471: Since this function is called `validate`, why not throw an exception from here directly?

Also, please get rid of "magic" here (and everywhere in your code, just search for "magic" and make sure nothing is left).

dace/transformation/interstate/fpga_transform_state.py
```diff
@@ -185,12 +187,14 @@ def apply(self, sdfg):
                         allow_conflicts=desc.allow_conflicts,
                         strides=desc.strides,
                         offset=desc.offset)
+                    fpga_array[1].location = desc.location
+                    desc.location.clear()
```
definelicht, line 191: Are you sure this works? I think `fpga_array[1].location = desc.location` assigns a pointer to the dictionary, so you will still clear it.

Why does it need to be deleted?

tests/fpga/hbm_copy_fpga.py
```diff
@@ -0,0 +1,118 @@
...
+from dace import subsets as sbs, dtypes, memlet as mem
+import dace
+import numpy as np
+from dace.dtypes import StorageType
+from dace.codegen.targets.fpga import _FPGA_STORAGE_TYPES
+
+# A test checking copies involving HBM-arrays in some way
+
+
+#helper MaKe_Copy that creates and appends states performing exactly one copy. If a provided
```
definelicht, line 11: Formatting

tests/fpga/hbm_copy_fpga.py
```diff
@@ -0,0 +1,118 @@
...
+        state = sdfg.add_state(is_start_state=True)
+    else:
+        state = sdfg.add_state_after(statebefore)
+
+    def mkarray(name, shape, storage, loc):
+        if (name in sdfg.arrays):
+            return sdfg.arrays[name]
+        isTransient = False
+        if (storage in _FPGA_STORAGE_TYPES):
+            isTransient = True
```
definelicht, line 36: Naming conventions (please check the full file)

tests/fpga/hbm_deeply_nested_fpga.py
```diff
@@ -0,0 +1,78 @@
...
+
+def create_deeply_nested_sdfg():
+    sdfg = dace.SDFG("deepnest_test")
+    state: dace.SDFGState = sdfg.add_state("init")
+    xarr = state.add_array("x", [4, 10], dace.float32)
+    sdfg.arrays["x"].location["bank"] = "hbm.0:4"
+    yarr = state.add_array("y", [4, 10], dace.float32)
+    sdfg.arrays["y"].location["bank"] = "hbm.4:8"
+
+    topMapEntry, topMapExit = state.add_map("topmap", dict(k="0:2"))
```
definelicht, line 18: Naming convention (check full file)

tests/fpga/hbm_dynamic_memlets.py
```diff
@@ -0,0 +1,63 @@
...
+                          memlet=mem.Memlet("x[1, 0]"),
+                          dst_conn="_dynbound")
+    sdfg.apply_fpga_transformations()
+    return sdfg
+
+
+def exec_dynamic_memlet_test():
+    sdfg = create_dynamic_memlet_sdfg()
+    x = np.zeros((4, 10), dtype=np.int32)
+    #has to be copied to sdfg
```
definelicht, line 51: Convention

definelicht (changes requested):

dace/codegen/targets/cpu.py
```diff
@@ -316,11 +316,12 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
-              or (nodedesc.storage == dtypes.StorageType.Register and
-                  ((symbolic.issymbolic(arrsize, sdfg.constants)) or
-                   ((arrsize_bytes > Config.get(
-                       "compiler", "max_stack_array_size")) == True)))):
+        elif (
+                nodedesc.storage == dtypes.StorageType.CPU_Heap or
+            (nodedesc.storage == dtypes.StorageType.Register and
+             ((symbolic.issymbolic(arrsize, sdfg.constants)) or
+              ((arrsize_bytes > Config.get("compiler", "max_stack_array_size"))
+               == True)))):
```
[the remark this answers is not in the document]
jnice-81: I don't know just installed it via pip, then invoked it directly via command line. Do I need some style file?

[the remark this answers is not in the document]
definelicht: The style file is already included in DaCe, but you need to make sure you have yapf version 0.30.0 or higher

dace/sdfg/validation.py
jnice-81: Are you sure that this is what we want here? The reason why it is converted to int is just to check if it is convertible. I would get confused if the validation would throw a cannot convert tmp = int(bank_assignment) to int or so without any info why this error is happening

definelicht: You're right, but then at least report the number, `"Expect integer identifier for bank assignment, got {bankassignment[1]}."`

dace/codegen/targets/cpp.py
```diff
@@ -216,41 +216,101 @@ def memlet_copy_to_absolute_strides(dispatcher,
...
-def ptr(name: str, desc: data.Data) -> str:
+def ptr(name: str,
+        desc: data.Data = None,
+        subset_info: Union[subsets.Subset, int] = None,
+        sdfg: dace.SDFG = None,
+        is_write: bool = None,
+        dispatcher=None,
+        ancestor: int = None,
+        is_array_interface: bool = False,
+        interface_id: Union[int, List[int]] = None) -> str:
```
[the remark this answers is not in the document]
jnice-81: No, because one can also have array interfaces that don't have an interface id appended. To be honest I don't really see a good way to save arguments here. 

It' s true that subset_info is just used for HBM, but I would like to keep this name because from the "logic" behind ptr, subset_info is just an optional parameter that contains any info on what will be accessed by this variable (can also be a whole subset not just the first index; and it's frequently also passed for arrays that don't use HBM, though it does not change anything then at the moment)

dace/codegen/targets/fpga.py
```diff
@@ -1894,9 +2008,18 @@ def generate_tasklet_postamble(self, sdfg, dfg, state_id, node,
...
                     and not cpp.is_write_conflicted(dfg, edge)
                     and self._dispatcher.defined_vars.has(edge.src_conn)):
+                accessed_subset = 0
+                if utils.is_hbm_array(datadesc):
+                    accessed_subset, high_check = utils.get_multibank_ranges_from_subset(
+                        edge.data.dst_subset or edge.data.subset, sdfg)
+                    if accessed_subset + 1 != high_check:
+                        raise cgx.CodegenError(
+                            "generate_tasklet_postamble was called on HBM memlet accessing multiple banks"
+                        )
```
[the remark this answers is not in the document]
jnice-81: In principle this is supposed to throw if the memlet for which it is called accesses multiple HBM banks, which it does not support. However I realized that this is only called for memlets attached to tasklets, and there this property is checked anyway, so I will remove it

dace/codegen/targets/cpu.py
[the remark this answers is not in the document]
definelicht: I also saw this in someone else's code, so if you have yapf 0.30+ then I guess this is just how it is :-(

dace/transformation/interstate/fpga_transform_state.py
jnice-81: No it does work, because python automatically duplicates objects pointed to by multiple references if their state diverges. (Tbh I don't fully get the exact rules for this either). It needs to be cleared, because desc will be the array used on the host, i.e. it does not have the same memory properties as the one on fpga.

dace/codegen/targets/cpp.py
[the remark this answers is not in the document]
definelicht: Can you see a potential use case outside of HBM for `subset_info`?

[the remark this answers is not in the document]
jnice-81: Not really ;) Ok I will rename it.

dace/codegen/targets/fpga.py
```diff
@@ -1411,10 +1485,50 @@ def copy_memory(self, sdfg, dfg, state_id, src_node, dst_node, edge,
...
+                                    callsite_stream)
+
         # Emit actual copy
-        self._emit_copy(sdfg, state_id, src_node, src_storage, dst_node,
-                        dst_storage, dst_schedule, edge, state_dfg,
-                        function_stream, callsite_stream)
+        if do_default_copy:
+            self._emit_copy(sdfg, state_id, src_node, src_storage, dst_node,
+                            dst_storage, dst_schedule, edge, state_dfg,
+                            function_stream, callsite_stream)
```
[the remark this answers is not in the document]
jnice-81: Did it now via a list, but not 100% sure that this is nicer. I don't want to somehow handle normal copies in the section where HBM copies are handled at the moment, because this would require handling quite a few special conditions in that part, which will make it harder to read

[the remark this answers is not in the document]
definelicht: Ah I see, I didn't consider that the HBM one is in a loop. Anyway, I'm okay with either solution then 🤷

dace/codegen/targets/fpga.py
```diff
@@ -474,7 +484,7 @@ def make_parameters(self, sdfg, state, subgraphs):
         external_streams: Set[tuple[bool, str, dt, dict[str, int]]] = set()
 
         # Mapping from global arrays to memory interfaces
-        bank_assignments: Dict[str, str] = {}
+        bank_assignments: Dict[str, (str, Union[int, subsets.Range])] = {}
```
tbennun, line 487:
```suggestion
        bank_assignments: Dict[str, Tuple[str, Union[int, subsets.Range]]] = {}
```

dace/codegen/targets/fpga.py
```diff
@@ -762,36 +794,45 @@ def allocate_array(self, sdfg, dfg, state_id, node, function_stream,
...
-                                f"hlslib::ocl::MemoryBank::bank{bank}, ")
-
-                        # Define buffer, using proper type
+                    # TODO: Distinguish between read, write, and read+write
+                    self._allocated_global_arrays.add(node.data)
+                    memory_bank_arg_count = 1
+                    bank_offset = -1
+                    storage_type_str = "hlslib::ocl::StorageType::DDR"  #DDR to use unspecified memory
+
+                    #Fix bankassignments if present
```
tbennun, line 803:
```suggestion
                    # Fix bank assignments if present
```

dace/codegen/targets/xilinx.py
```diff
@@ -194,6 +196,7 @@ def get_generated_codeobjects(self):
                            XilinxCodeGen,
                            "Xilinx",
                            target_type="device"))
+            bank_assignment_code.clear()
```
tbennun, line 199: why is the `clear()` here necessary?

dace/sdfg/utils.py
```diff
@@ -1030,6 +1032,119 @@ def get_next_nonempty_states(sdfg: SDFG, state: SDFGState) -> Set[SDFGState]:
     return result
 
 
+def is_hbm_array(array: dt.Data):
```
tbennun, line 1035: aren't these more FPGA-codegen-related and should be there? (e.g., `dace/codegen/targets/fpga/utils.py`)

tbennun (changes requested):
Overall looks good, apart from some API design decisions that should be changed.

dace/codegen/targets/xilinx.py
jnice-81: Because we are creating the bankassignments for each kernel seperately. If clear is missing here, then we will add all assignments from the previous kernels to the kernels generated later (applies only for SDFGs with multiple states running on the SDFG

dace/sdfg/utils.py
jnice-81: moved it to targets/fpga_helper/fpga_utils (fpga_helper so it does not clash with fpga.py)

dace/codegen/targets/xilinx.py
jnice-81: *on the FPGA

dace/codegen/targets/cpp.py
```diff
@@ -50,24 +50,28 @@ def copy_expr(
...
     dt = ""
 
-    expr = ptr(dataname, datadesc)
+    def_type, _ = dispatcher.defined_vars.get(data_name)
 
-    def_type, _ = dispatcher.defined_vars.get(dataname)
+    expr = ptr(
+        data_name,
+        data_desc,
+        hbm_bank,
```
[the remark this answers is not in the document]
jnice-81: To be honest I would not know how to do that easily. data_name and data_desc are definitively required as well as sdfg (one could save the data_desc by requiring sdfg to be always set). hbm_bank info is used as from the outside and changes often between calls. The actual reason why is_write, dispatcher, is_array_interface, ancestor and interface_id are there as well is because ptr was merged with array_interface_variable function (beeing seperate functions lead to problems since suddenly the order of calls to ptr and array_interface_variable matters, and even more because it's not so clear why ptr and array_interface_variable should be seperate functions, since they do the same thing in principle - generate the final name for a variable). It was allowed to call array_interface_variable when the variable was not yet defined, so dispatcher may be None, even if is_array_interface is True. Similarly one may have is_array_interface set to True with or without setting interface_id, they are independent. 
Guess the only way to reduce parameters I see would be to either remove is_array_interface, is_write and dispatcher and ancestor and expect some new parameter array_interface_info or so that is a tuple of is_write, dispatcher and ancestor (feels a bit forced), or to accept that ptr gets to complicated when merged with array_interface_variable and to split them again.

dace/sdfg/state.py
```diff
@@ -730,7 +730,7 @@ class SDFGState(OrderedMultiDiConnectorGraph[nd.Node, mm.Memlet],
 
     location = DictProperty(
         key_type=str,
-        value_type=symbolic.pystr_to_symbolic,
```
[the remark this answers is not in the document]
jnice-81: Changing back to symbolic doesn't directly work here because all value objects need to have the same type, which is not possible if one also wants to store the memorytype (since symbolic for the memorytype doesn't make much sense). I will change the code to use the 2 keys, but can we just keep string as type here?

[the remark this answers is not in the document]
jnice-81: Just realized that I mixed something here: I'd like to keep string type for the locationproperty of arrays, not for state. The reason I changed the type of the state locationproperty to string was because it lead to serialization errors for some of the HBM-tests. (I guess not all the values that are stored in there are symbols; e.g. the "is_FPGA_Kernel" is a boolean, not a symbol)

dace/codegen/targets/cpp.py
```diff
@@ -110,59 +114,61 @@ def memlet_copy_to_absolute_strides(dispatcher,
...
-            )
+            dst_expr = copy_expr(dispatcher,
+                                 sdfg,
+                                 dst_node.data,
+                                 memlet,
+                                 is_write=True,
+                                 offset=memlet.other_subset,
+                                 relative_offset=False,
+                                 packed_types=packed_types,
+                                 hbm_bank=memlet.other_subset)
```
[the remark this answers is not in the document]
jnice-81: Actually I did not change that, the only new parameter is hbm_bank

dace/codegen/targets/cpp.py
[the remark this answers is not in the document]
jnice-81: I readded the array_interface_variable function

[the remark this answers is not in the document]
jnice-81: Final solution is an fpga specific ptr in fpga_utils, that is called whenever we are dealing with fpga arrays. It's not part of the fpga-codegen because that's annoying/ugly to implement in cpp.

jnice-81:
I added .run/ to the gitignore. I'm aware that this technically does not belong here, feel free to throw it out again.

dace/sdfg/state.py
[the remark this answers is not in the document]
jnice-81: Anyway I will change it back here, since this definitely does not belong in this PR

tbennun (changes requested):
minor changes this time around. @definelicht next (final) review is yours

dace/codegen/targets/fpga_helper/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+from .fpga_utils import (
+    is_hbm_array,
+    iterate_hbm_multibank_arrays,
+    modify_distributed_subset,
+    get_multibank_ranges_from_subset,
+    parse_location_bank,
+    ptr,
+)
```
definelicht, line 9: Why does this need a separate folder and files? Can't we just put all of this in `fpga.py`?

dace/codegen/targets/fpga_helper/fpga_utils.py
```diff
@@ -0,0 +1,194 @@
...
+            cps[0] = (change, change, 1)
+    elif isinstance(subset, list) or isinstance(subset, tuple):
+        if isinstance(subset, tuple):
+            cps = list(cps)
+        if change == -1:
+            cps.pop(0)
+        else:
+            cps[0] = change
+        if isinstance(subset, tuple):
+            cps = tuple(cps)
```
definelicht, line 72: Couldn't this be anything iterable? I think you can replace all of this with just:
```python
cps = list(cps)  # Works for any iterable type
if change == -1:
    cps = cps[1:]
else:
    cps[0] = change
```
The input type would then just be `Iterable`.

definelicht (changes requested):

dace/codegen/targets/fpga.py
```diff
@@ -67,6 +68,186 @@ def is_fpga_kernel(sdfg, state):
...
+        if change == -1:
+            cps.pop([0])
+        else:
+            cps[0] = (change, change, 1)
+    elif isinstance(subset, Iterable):
+        cps = list(cps)  # Works for any iterable type
+        if change == -1:
+            cps = cps[1:]
+        else:
+            cps[0] = change
```
definelicht, line 128: This seems inconsistent, in the first part you change it to a tuple `(change, change, 1)`, in the second to `change`? Probably only one is correct?

You also don't really need the `isinstance(subset, Iterable)`, I would have an `else` and convert to a list directly (it'll fail anyway if it doesn't work). Whatever you prefer.

Actually, couldn't this whole thing be replaced with just (with no `isinstance` check):
```python
cps = list(subset)
if change == -1:
  return cps[1:]
else:
  cps[0] = (change, change, 1)
  return cps
```

definelicht (changes requested):

dace/codegen/targets/fpga.py
jnice-81: just assigning (change, change, 1) always will not work. The function is used in multiple contexts, to update shapes as well as subsets. Probably the clean way to go would be to actually have two different functions.

jnice-81: Ok, instead I know completly threw out the support for shapes, since this is anyway only used in one place. So now this actually does except Subset as type in all cases.

definelicht (approved):

tbennun (approved):

PR #763, opened 2021-06-22
Serialize start state and strict mode for HoistState

tbennun:


acalotoiu (approved):
it does solve the issue with hoist state, and also fixes a serious serialization issue.

PR #764, opened 2021-06-24
fixing prune connectors bug

acalotoiu:
added check to not remove arrays from sdfgs unless all connectors are removed.


tbennun (approved):
LGTM if all tests pass

PR #765, opened 2021-06-26
Python frontend: Better support for arguments

tbennun:
- [x] Default/keyword arguments
- [x] Variable-length arguments (`*args`, `**kwargs`, only in JIT mode)
- [x] Optional arguments (None inputs) for arrays
- [x] Arbitrary constant arguments that get evaluated at call time (JIT mode only)

### DIFF
--- a/dace/codegen/targets/common.py
+++ b/dace/codegen/targets/common.py
@@ -39,8 +39,8 @@ def sym2cpp(s, arrayexprs: Optional[Set[str]] = None) -> Union[str, List[str]]:
     :return: C++-compilable expression or list thereof.
     """
     if not isinstance(s, list):
-        return symbolic.symstr(s, arrayexprs)
-    return [symbolic.symstr(d, arrayexprs) for d in s]
+        return cppunparse.pyexpr2cpp(symbolic.symstr(s, arrayexprs))
+    return [cppunparse.pyexpr2cpp(symbolic.symstr(d, arrayexprs)) for d in s]
 
 
 def codeblock_to_cpp(cb: CodeBlock):
--- a/dace/symbolic.py
+++ b/dace/symbolic.py
@@ -831,7 +831,7 @@ class DaceSympyPrinter(sympy.printing.str.StrPrinter):
 
     def _print_Symbol(self, expr):
         if expr.name == 'NoneSymbol':
-            return 'None'
+            return 'nullptr'
         return super()._print_Symbol(expr)
 
     def _print_Pow(self, expr):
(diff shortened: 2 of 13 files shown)


dace/codegen/targets/common.py
```diff
@@ -39,8 +39,8 @@ def sym2cpp(s, arrayexprs: Optional[Set[str]] = None) -> Union[str, List[str]]:
     :return: C++-compilable expression or list thereof.
     """
     if not isinstance(s, list):
-        return symbolic.symstr(s, arrayexprs)
-    return [symbolic.symstr(d, arrayexprs) for d in s]
+        return cppunparse.pyexpr2cpp(symbolic.symstr(s, arrayexprs))
```
alexnick83, line 42: AFAIK `symbolic.symstr` is still returning valid C++ code. Therefore, it is very likely that passing the output of this method to `cppunparse.pyexpr2cpp` will fail. I will add a test.

dace/symbolic.py
```diff
@@ -820,6 +829,11 @@ def _print_Infinity(self, expr):
     def _print_NegativeInfinity(self, expr):
         return '-INFINITY'
 
+    def _print_Symbol(self, expr):
+        if expr.name == 'NoneSymbol':
+            return 'None'
```
alexnick83, line 834: Shouldn't this be `NULL` or `nullptr`? The description of the class is: `Several notational corrections for integer math and C++ translation`.

tests/python_frontend/lambda_test.py
```diff
@@ -0,0 +1,108 @@
...
+    A = np.random.rand(20)
+    B = np.random.rand(20)
+    C = np.random.rand(20)
+    lamb1(A, B, C)
+    assert np.allclose(A, B + C)
+
+
+if __name__ == '__main__':
+    test_inline_lambda_tasklet()
+    # test_inline_lambda_scalar()
```
alexnick83, line 104: Shall I assume that these do not work still and are left for another PR?

alexnick83 (changes requested):
I am pretty sure that bringing back the call to `pyexpr2cpp` is a problem. `symstr` returns a C++-code string, not a Python-code string. If you want to change this then you need a PR that will create separate methods/classes. I will check if it is possible to test whether the string is valid Python code or not without crashing.

tests/python_frontend/lambda_test.py
tbennun: yes

alexnick83 (approved):

PR #766, opened 2021-06-28
Fix stream allocation scoping

komplexon3:
There was an issue where streams would be allocated globally (to a state) and locally. This should not happen. The expected behaviour is to never allocate streams locally to a scope.

This PR fixes this issue by never including streams in the scope transient analysis.

### DIFF
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -451,6 +451,7 @@ class FPGACodeGen(TargetCodeGenerator):
                     # Global data is passed from outside the kernel
                     global_data_parameters.add(
                         (is_output, dataname, desc, interface_id))
+                # Second check needed because we want all streams allocated top-level locally
                 elif dataname in shared_data or isinstance(desc, dt.Stream):
                     # Add the data as a parameter to this PE
                     subgraph_parameters[subgraph].add(
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -451,7 +451,7 @@ class FPGACodeGen(TargetCodeGenerator):
                     # Global data is passed from outside the kernel
                     global_data_parameters.add(
                         (is_output, dataname, desc, interface_id))
-                elif dataname in shared_data:
+                elif dataname in shared_data or isinstance(desc, dt.Stream):
                     # Add the data as a parameter to this PE
                     subgraph_parameters[subgraph].add(
                         (is_output, dataname, desc, interface_id))
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -844,7 +844,8 @@ def separate_maps(state, dfg, schedule):
 def _transients_in_scope(sdfg, scope, scope_dict):
     return set(
         node.data for node in scope_dict[scope.entry if scope else scope]
-        if isinstance(node, nd.AccessNode) and sdfg.arrays[node.data].transient)
+        if isinstance(node, nd.AccessNode) and sdfg.arrays[node.data].transient
+            and not isinstance(sdfg.arrays[node.data], dt.Stream)) # streams should always be state global
 
 
 def local_transients(sdfg, dfg, entry_node):
--- a/tests/stream_allocation_scope_test.py
+++ b/tests/stream_allocation_scope_test.py
@@ -1,0 +1,98 @@
+
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+""" Test for an issue where streams would be allocated globally (to a state) and locally.
+    Expected behavior is to never allocate streams locally to a scope. """
+import dace
+
+def test_stream_allocation_scope():
+    sdfg = dace.SDFG("stream_allocation")
+    N = dace.symbol("N")
+    P = dace.symbol("P")
+    sdfg.add_array("A", [N], dace.float64)
+    sdfg.add_array("B", [N], dace.float64)
+    sdfg.add_stream("AtoB",
+                    dace.float64,
+                    transient=True,
+                    shape=(P, ),
+                    storage=dace.dtypes.StorageType.FPGA_Local
+                    )
+    state = sdfg.add_state()
+    a = state.add_read("A")
+    b = state.add_write("B")
+    read_AtoB = state.add_read("AtoB")
+    write_AtoB = state.add_write("AtoB")
+
+    read_entry, read_exit = state.add_map(
+        "read_A", {"n": "0:N/P"},
+        schedule=dace.ScheduleType.FPGA_Device)
+
+    unroll_read_entry, unroll_read_exit = state.add_map(
+        "unroll_read", {"p": "0:P"},
+        schedule=dace.ScheduleType.FPGA_Device,
+        unroll=True)
+
+    write_entry, write_exit = state.add_map(
+        "write_B", {"n": "0:N/P"},
+        schedule=dace.ScheduleType.FPGA_Device)
+
+    unroll_write_entry, unroll_write_exit = state.add_map(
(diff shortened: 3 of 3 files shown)


definelicht:
I think the approach is still not the right one. What we want is to make sure that _FPGA streams_ are always allocated in exactly one place, which is at the beginning of kernel generation. My intuition would be that we have to:
- Make sure that _all_ streams are allocated when [generating the FPGA kernel](https://github.com/spcl/dace/blob/62ae0f0a43352af1c88dea4d457dac392203e790/dace/codegen/targets/fpga.py#L298-L307).
- Never allocated local streams _when inside an FPGA kernel_.

komplexon3:
Okay - I found something interesting. The place where the stream allocation "would" be suppressed [here],(https://github.com/spcl/dace/blob/62ae0f0a43352af1c88dea4d457dac392203e790/dace/codegen/targets/fpga.py#L703) it is explicitly **not** suppressed because of something with the Intel FPGA backend. Can someone explain to me why this is? According to @TizianoDeMatteis 's commit message, it has to do with allowing views on Intel FPGAs.

definelicht:
> Okay - I found something interesting. The place where the stream allocation "would" be suppressed [here],(
> 
> https://github.com/spcl/dace/blob/62ae0f0a43352af1c88dea4d457dac392203e790/dace/codegen/targets/fpga.py#L703
> 
> ) it is explicitly **not** suppressed because of something with the Intel FPGA backend. Can someone explain to me why this is?

Honestly I already find it a bit weird that we are suppressing these allocations here. Have you checked the callsite of where `allocate_array` is being called from in the case that is causing problems for you? Could you post the call stack here?

tbennun:
> I think the approach is still not the right one. What we want is to make sure that _FPGA streams_ are always allocated in exactly one place, which is at the beginning of kernel generation. My intuition would be that we have to:
> 
> * Make sure that _all_ streams are allocated when [generating the FPGA kernel](https://github.com/spcl/dace/blob/62ae0f0a43352af1c88dea4d457dac392203e790/dace/codegen/targets/fpga.py#L298-L307).
> * Never allocate local streams _when inside an FPGA kernel_.

exactly this. This PR should be changed to focus on making sure the FPGA codegen allocates all streams in one place.

komplexon3:
@definelicht This would be the stack trace (made it throw an exception if an already defined stream reaches this bit of code).

It is called by `dispatch_allocate`. That's the same code allocation "top-level local" data. 
```
Traceback (most recent call last):
  File "tests/stream_allocation_scope_test.py", line 98, in <module>
    test_stream_allocation_scope()
  File "tests/stream_allocation_scope_test.py", line 91, in test_stream_allocation_scope
    assert sdfg.generate_code()[2].clean_code.count("dace::SetNames(AtoB") == 1
  File "/home/widmmarc/k3-dace/dace/sdfg/sdfg.py", line 2249, in generate_code
    program_code = codegen.generate_code(sdfg)
  File "/home/widmmarc/k3-dace/dace/codegen/codegen.py", line 178, in generate_code
    used_environments) = frame.generate_code(sdfg, None)
  File "/home/widmmarc/k3-dace/dace/codegen/targets/framecode.py", line 552, in generate_code
    states_generated = self.generate_states(sdfg, global_stream,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/framecode.py", line 463, in generate_states
    cft.as_cpp(self.dispatcher.defined_vars, sdfg.symbols), sdfg)
  File "/home/widmmarc/k3-dace/dace/codegen/control_flow.py", line 195, in as_cpp
    expr += elem.as_cpp(defined_vars, symbols)
  File "/home/widmmarc/k3-dace/dace/codegen/control_flow.py", line 126, in as_cpp
    expr += self.dispatch_state(self.state)
  File "/home/widmmarc/k3-dace/dace/codegen/targets/framecode.py", line 440, in dispatch_state
    self._dispatcher.dispatch_state(sdfg, state, global_stream, stream)
  File "/home/widmmarc/k3-dace/dace/codegen/dispatcher.py", line 310, in dispatch_state
    satisfied_dispatchers[0].generate_state(sdfg, state,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 211, in generate_state
    self.generate_kernel(sdfg, state, kernel_name, subgraphs,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 1399, in generate_kernel
    self.generate_kernel_internal(sdfg, state, kernel_name, subgraphs,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/xilinx.py", line 798, in generate_kernel_internal
    self.generate_modules(sdfg, state, kernel_name, subgraphs,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 1450, in generate_modules
    self.generate_module(sdfg, state, module_name, subgraph,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/xilinx.py", line 732, in generate_module
    self._dispatcher.dispatch_subgraph(sdfg,
  File "/home/widmmarc/k3-dace/dace/codegen/dispatcher.py", line 351, in dispatch_subgraph
    self.dispatch_scope(v.map.schedule, sdfg, scope_subgraph,
  File "/home/widmmarc/k3-dace/dace/codegen/dispatcher.py", line 403, in dispatch_scope
    self._map_dispatchers[map_schedule].generate_scope(
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 510, in generate_scope
    self.generate_node(sdfg, dfg_scope, state_id,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 1043, in generate_node
    getattr(self, method_name)(sdfg, dfg, state_id, node,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 1345, in _generate_MapEntry
    self._dispatcher.dispatch_allocate(sdfg, dfg, state_id, child, None,
  File "/home/widmmarc/k3-dace/dace/codegen/dispatcher.py", line 424, in dispatch_allocate
    dispatcher.allocate_array(sdfg, dfg, state_id, node, function_stream,
  File "/home/widmmarc/k3-dace/dace/codegen/targets/fpga.py", line 543, in allocate_array
    raise Exception
   ```

I will also need to ensure that **all** streams are allocated once. At the moment, only data shared between two or more subgraphs are allocated "top-level locally". (Missed that bit before.) This should be rather straightforward - just look for all streams in the subgraphs and allocate it when it allocates the data shared between subgraphs.

definelicht:
As a first step, make sure that all these streams are allocated where we _want_ them to be allocated (i.e., at the beginning of generating an FPGA kernel). Then we can see if we need to prevent them from being reallocated later, but we might not need to, because of that code you linked.

komplexon3:
Now all streams are allocated "top-level locally". The FPGA codegen now allocates data at that level if it shared between two subgraphs (status-quo) **or** if it is a stream.

There is still the issue with them being declared locally due to that explicit check mentioned before.https://github.com/spcl/dace/blob/62ae0f0a43352af1c88dea4d457dac392203e790/dace/codegen/targets/fpga.py#L703
@TizianoDeMatteis It was added by this [commit](https://github.com/spcl/dace/commit/f71cbdcaaad43fd9512b7be144a27793fb834601). Could you explain to me why this check is needed? I am not familiar with the concept of views in DaCe.

TizianoDeMatteis:
> @TizianoDeMatteis It was added by this [commit](https://github.com/spcl/dace/commit/f71cbdcaaad43fd9512b7be144a27793fb834601). Could you explain to me why this check is needed? I am not familiar with the concept of views in DaCe.

We need to allocate anyway streams to keep track of their names across the SDFGs hierarchy (a stream may have a certain name in the top-level scope, then another one in a nested SDFG and so on...)

PR #769, opened 2021-06-28
Allow manually setting the default build folder of SDFGs

tbennun:


dace/sourcemap.py
```diff
@@ -49,12 +50,13 @@ def tmp_location(name: str) -> str:
...
     """
-    return os.path.abspath(os.path.join(".dacecache", name, "map", "tmp.json"))
+    build_folder = Config.get('default_build_folder')
+    return os.path.abspath(os.path.join(build_folder, name, "map", "tmp.json"))
 
 
 def temporaryInfo(name: str, data):
     """ Creates a temporary file that stores the json object
-        in the map folder (.dacecache/{SdfgName}/map)
+        in the map folder ({build_folder}/{sdfg_name}/map)
```
definelicht, line 59: Is this some sort of reference or formatting string? If not, I would use `<build folder>/<SDFG name>/map` so it doesn't look like a fstring gone wrong 😛

definelicht (changes requested):
No test that tries changing it from the default?

dace/sourcemap.py
tbennun: yeah I tried to keep the old docs intact. will fix

tests/custom_build_folder_test.py
```diff
@@ -0,0 +1,26 @@
...
+    with dace.config.set_temporary('default_build_folder', value='.mycache'):
+        A = np.random.rand(20)
+        assert np.allclose(customprog(A), A + 1)
+
+        # Ensure the folder was created
+        assert os.path.isdir('.mycache')
+
+        # Ensure build folder matches
+        sdfg = customprog.to_sdfg()
+        assert '.mycache' in sdfg.build_folder
```
definelicht, line 22: Use a `tempfile.TemporaryDirectory`? 😜

definelicht (changes requested):

definelicht (approved):

PR #773, opened 2021-06-29
Codegen debugging mapping

benibenj:


dace/sourcemap.py
```diff
@@ -210,12 +210,18 @@ def create_cpp_map(code: str, name: str, target_name: str):
     # We send a message to the IDE in that case to notify the user of
     # restricted features.
     if Config.get('cache') != 'hash':
+        codegen_debug = True if Config.get('compiler',
+                                           'codegen_lineinfo') else False
```
benibenj, line 214: Could Config.get('compiler','codegen_debug') be None? otherwise I can remove the if/else

dace/sourcemap.py
```diff
@@ -237,15 +244,24 @@ def __init__(self, code: str, name: str, target_name: str):
...
-        """ For each line of code retrieve the corresponding nodes
-            and map the nodes to this line
+    def mapper(self, codegen_debug: bool = False):
+        """ For each line of code retrieve the corresponding identifiers
+            and create the mapping
         """
+        cpp_pattern = re.compile(
+            r'(\/\/\/\/__DACE:[0-9]+:[0-9]+:[0-9]+(,[0-9])*)')
+        codegen_pattern = re.compile(
+            r'(\/\/\/\/__CODEGEN;([A-z]:)?(\/|\\)([A-z0-9-_+]+(\/|\\))*([A-z0-9]+\.[A-z0-9]+);[0-9]+)'
```
benibenj, line 256: I precompile the RE patterns now so it's only done once. I should have done this before

dace/sourcemap.py
tbennun: Please use `Config.get_bool`. This always ensures either true or false

dace/sourcemap.py
tbennun: nice :)

tbennun: shouldn't this be part of `self`, though (as a field)? then you wouldn't need to pass it along to any method

benibenj: Sure, could do this, does it make a difference? Just more readable code

tbennun: yes, more readable code and fewer arguments (shorter diff too :) )

tbennun (approved):

PR #778, opened 2021-06-29
Classes, Fields and Closures

tbennun:
Support for closures (limited to arrays and constants), which includes fields and globals.

After this PR, the python frontend will extract a "reduced closure" of used elements that live outside the dace context, and include them into the SDFG call.

A dace function (program/method) is defined by its contents, as well as its **arguments** and **closure** (which is composed of **arrays** and **constants):
- [x] Disallow numpy view arguments
- [x] Support arrays in dace.program closure
- [x] Support nested objects as part of the closure, supported by returning a dictionary from the frontend along with the SDFG
    - [x] Handle different objects with same field names
    - [x] Native support for compiling `dataclass` objects with type annotation
- [x] Fix/refactor dace program cache to `CachedDaceProgram` objects with information about the closure constants and argument/closure types
    - [x] Re-evaluate closures upon call
    - [x] Cache should have configurable size


tests/blas/nodes/ger_test.py
```diff
@@ -165,4 +174,4 @@ def test_ger_fpga():
     args = parser.parse_args()
 
     run_ger(args.target, args.N, args.M, args.tile_size_x, args.tile_size_y,
-            args.alpha, args.veclen, args.eps)
+            args.alpha, args.veclen, args.eps)
```
alexnick83, line 177:
```suggestion
            args.alpha, args.veclen, args.eps)

```

alexnick83 (approved):
LGTM. I assume all the `np.copies` added to the tests are there because now it is explicitly not allowed to pass NumPy views to top-level DaCe programs?

tbennun:
> LGTM. I assume all the `np.copies` added to the tests are there because now it is explicitly not allowed to pass NumPy views to top-level DaCe programs?

yes, exactly.

PR #781, opened 2021-07-04
Consecutive Py and C++ mapping

benibenj:


dace/sourcemap.py
```diff
@@ -176,34 +177,27 @@ def create_py_map(sdfg):
...
-                    'made_with_api': True,
-                })
-        else:
-            tmp['src_files'] = sourceFiles
-            temporaryInfo(sdfg.name, tmp)
+        return {
+            'build_folder': folder,
+            'src_files': sourceFiles,
+            'made_with_api': True if tmp is None else False,
+        }
```
tbennun: now that I think of it, this is a very javascript thing to do. In Python one would normally make an object (dataclass in this case would make a lot of sense)

tbennun (changes requested):

tbennun (approved):

PR #785, opened 2021-07-08
Array access indirection

alexnick83:
The PR changes how the indirection of array accesses in Python code maps to their SDFG representation. Originally, an indirection `A[x[i]]` would result in an "indirection tasklet" with inputs the whole array `A` and the indirection index `x[i]`, and output the indirect access `A[x[i]]`. Under the new scheme, the indirection index `x[i]` is first converted to a symbol `__sym_xyz`, whose value is set in an interstate edge. Then, the indirect access becomes `A[__sym_xyz]`, which is represented in the SDFG as a regular slice. The PR generalizes this new approach for the following use cases:
- Scalar indirection (`a` is scalar data): `A[a]`
- Indirection in multiple dimensions (`a`, `b`, and `c` are scalar data): `A[a, b, c]`
- Range indirection (`a`, `b` are scalar data): `A[a:b]`
- Nested indirection (`a` is scalar data, `A`, `B` are arrays): `A[B[a]]`
- The above indirections but using array elements (`i` is symbol or scalar data): `A[x[i]`, `A[x[i], x[i+1]]`, `A[x[i]:x[i+1]]`, `A[x[x[i]]`


dace/frontend/python/newast.py
```diff
@@ -4188,9 +4215,26 @@ def visit_Call(self, node: ast.Call):
...
+                                        symbolic.symbol(str(sym)):
+                                        symbolic.symbol('__dacesym_' + str(sym))
+                                    }, sym, '__dacesym_' + sym)
+                        for sym, symvalue in mapping.items():
+                            if str(sym) != str(symvalue):
+                                sd.replace_properties(
+                                    newarr, {
+                                        symbolic.symbol('__dacesym_' + str(sym)):
+                                        symbolic.pystr_to_symbolic(symvalue)
+                                    }, '__dacesym_' + str(sym), symvalue)
```
definelicht, line 4231: Why does this need to be done in two steps, and does it need to be over two passes over the mapping? Can you not consolidate this into a single loop?

dace/frontend/python/newast.py
```diff
@@ -4661,6 +4705,111 @@ def visit_Compare(self, node: ast.Compare):
...
+                    other_subset.size(), arrobj.dtype, arrobj.storage)
+            wnode = self.last_state.add_write(tmp,
+                                              debuginfo=self.current_lineinfo)
+            self.last_state.add_nedge(
+                rnode, wnode,
+                Memlet.simple(array,
+                              expr.subset,
+                              num_accesses=expr.accesses,
+                              wcr_str=expr.wcr,
+                              other_subset_str=other_subset))
```
definelicht, line 4759: `Memlet.simple` is deprecated

dace/frontend/python/newast.py
```diff
@@ -4661,6 +4705,111 @@ def visit_Compare(self, node: ast.Compare):
...
+            return tmp
+
+    def _parse_subscript_slice(self, s: ast.AST, multidim: bool = False):
```
definelicht, line 4762: Put return type and/or docstring

definelicht:
It seems there is also an extension to `loop_to_map` in here! This should have been a separate PR :-/

tests/library/gemm_test.py
```diff
@@ -86,7 +86,7 @@ def run_test(implementation,
     C_str = "None" if C_shape is None else (
         str(C_shape[0]) if len(C_shape) == 1 else f"{C_shape[0]}_{C_shape[1]}")
     sdfg_name = f"{implementation}_{M}_{N}_{K}_{complex}_{transA}_{transB}_{alpha}_{beta}_{C_str}".replace(
-        ".", "_dot_").replace("+", "_plus_").replace("(", "").replace(")", "")
+        ".", "_dot_").replace("+", "_plus_").replace("-", "_minus_").replace("(", "").replace(")", "")
```
definelicht, line 89: This should just be a regex at this point ;-)

tests/python_frontend/indirections_test.py
```diff
@@ -0,0 +1,408 @@
...
+        y[i] = vals @ x[cols]
+
+    return y
+
+
+def test_spmv():
+
+    with dc.config.set_temporary('compiler',
+                                 'allow_view_arguments',
+                                 value=True):
```
definelicht, line 358: Why is this needed?

dace/frontend/python/newast.py
alexnick83: This is documented here: https://github.com/spcl/dace/blob/a654383d173881a981f34549c03cd90cc6db5594/dace/symbolic.py#L916

The code you commented on is not new btw, but it is the one that existed in master until yesterday. @tbennun refactored it to the `symbolic.safe_replace` method. However, as noted in the comment in the following lines, I had to revert the change because it doesn't work (yet) properly in the context of the `visit_Call` method.

tests/python_frontend/indirections_test.py
alexnick83: Because the default value of this new compilation flag breaks the test unless I make explicit copies of all the data, which is unnecessary and undesirable in this case.

tests/library/gemm_test.py
```diff
@@ -83,10 +84,9 @@ def run_test(implementation,
         C_shape = [s if isinstance(s, int) else replace_map[s] for s in C_shape]
 
     # unique name for sdfg
-    C_str = "None" if C_shape is None else (
-        str(C_shape[0]) if len(C_shape) == 1 else f"{C_shape[0]}_{C_shape[1]}")
-    sdfg_name = f"{implementation}_{M}_{N}_{K}_{complex}_{transA}_{transB}_{alpha}_{beta}_{C_str}".replace(
-        ".", "_dot_").replace("+", "_plus_").replace("(", "").replace(")", "")
+    sdfg_name = random.choices(string.ascii_lowercase)[0]
+    sdfg_name += ''.join(random.choices(string.ascii_lowercase + string.digits,
+                         k=19))
```
definelicht, line 89: Haha now it's even crazier! If you just want a mangled name, why not use `Config.set("cache", "unique")` or `Config.set("cache", "hash")`? You can still get the build folder with `sdfg.build_folder` if you need to

dace/frontend/python/newast.py
alexnick83: Imagine that you are updating the shape of an array given by `(5*M + N,)` and that the replacement dictionary is `{M: N, N: M}`, i.e., the shape must become `(5*N + M,)`. If you did it in a single loop with two replacements, then starting with `M` you get `5*M + N -> 5*__dace__sym_M + N -> 5* N + N`. Maybe I misunderstand what you mean.

definelicht (commented):
Ok from my side, but what's up with the coverage -- why did it go down so much?

dace/frontend/python/newast.py
```diff
@@ -4188,9 +4215,26 @@ def visit_Call(self, node: ast.Call):
                     if mapping is not None:
                         # Two-step replacement (N -> __dacesym_N --> mapping[N])
                         # to avoid clashes
-                        symbolic.safe_replace(
-                            mapping,
-                            lambda m: sd.replace_properties_dict(newarr, m))
+                        for sym, symvalue in mapping.items():
```
tbennun, line 4218: :( one of the reasons I refactored it is because some of the replacements don't actually work (e.g., when replacing something with a constant float value).
Is there a way to reuse the existing function and add an extra boolean argument?

dace/frontend/python/newast.py
```diff
@@ -4662,6 +4706,117 @@ def visit_Compare(self, node: ast.Compare):
...
+                    other_subset.size(), arrobj.dtype, arrobj.storage)
+            wnode = self.last_state.add_write(tmp,
+                                              debuginfo=self.current_lineinfo)
+            self.last_state.add_nedge(
+                rnode, wnode,
+                Memlet(f'{array}[{expr.subset}]->{other_subset}',
+                       volume=expr.accesses,
+                       wcr=expr.wcr))
+            return tmp
+
```
tbennun, line 4760: there was a good chance to make this into a view here. Should we do it in another PR?

dace/frontend/python/newast.py
```diff
@@ -4725,80 +4882,18 @@ def visit_Subscript(self, node: ast.Subscript):
                                   'Type "%s" cannot be sliced' % arrtype)
 
         # Visit slice contents
-        if isinstance(node.slice, ast.Constant):  # 1D index (since Python 3.9)
```
tbennun, line 4728: nice

dace/transformation/interstate/loop_to_map.py
```diff
@@ -202,7 +218,127 @@ def apply(self, sdfg: sd.SDFG):
...
+                    memlet.Memlet.from_array(name, sdfg.arrays[name]))
+            for name in write_set:
+                w = new_body.add_write(name)
+                new_body.add_edge(
+                    cnode, name, w, None,
+                    memlet.Memlet.from_array(name, sdfg.arrays[name]))
+
+            # Fix SDFG symbols
+            for sym, dtype in nsymbols.items():
+                if sym in sdfg.free_symbols:
```
tbennun, line 336: I would first collect the prior `sdfg.free_symbols` above this whole thing, and then check for the new ones (`sdfg.free_symbols - old_freesyms`). This is much safer w.r.t. actually free symbols.

dace/transformation/interstate/state_elimination.py
```diff
@@ -154,11 +154,146 @@ def apply(self, sdfg):
                     break
             else:
                 # If removed assignment does not appear in any other edge,
-                # remove symbol
+                # replace and remove symbol
+                if assignments_to_consider[varname] in sdfg.symbols:
+                    sdfg.replace(varname, assignments_to_consider[varname])
```
tbennun, line 159: safe replace of all symbols at the same time?

dace/transformation/interstate/state_elimination.py
```diff
@@ -154,11 +154,146 @@ def apply(self, sdfg):
...
+    1. We define inter-state edge assignments to be "SymbolAlias" when a symbol
+    has is set equal to another symbol or (potentially promotable) scalar.
+    2. We define inter-state edge assignment promotion as moving the assignment
+    to the previous inter-state edge according to the topological order, when
+    this is unambiguous.
+    3. SymbolAliasPromotion moves SymbolAlias assignments to the previous
+    inter-state edge according to the topological order.
+
+    The purpose of this transformation is to iteratively move symbolic aliases
+    together, so that true duplicates can be easily removed.
```
tbennun, line 185: I cannot understand what this transformation does from this docstring. I would first make it start with the second paragraph ("Iteratively moves..."), and explain what is a symbolic alias.

tests/python_frontend/indirection_with_scalars_test.py
```diff
@@ -35,9 +33,80 @@ def test_nested_scalar_indirection():
...
     ref = A[:, 0, :, 0]
-    assert(np.array_equal(B, ref))
+    assert (np.array_equal(B, ref))
+
+
+def test_array_element_scalar_indirection():
+    @dc.program
+    def nested_arrayindex_indirection(A: dc.float64[20, 10],
+                                      indices: dc.int32[2]):
+        start = indices[0]
```
tbennun, line 43: great tests! :)

tests/python_frontend/indirections_test.py
```diff
@@ -0,0 +1,408 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+import dace as dc
+import numpy as np
+
+
+@dc.program
+def indirection_scalar(A: dc.float32[10]):
```
tbennun, line 7: nice and comprehensive :)

tbennun (changes requested):
I really like some of the refactors here. The only thing that really bugs me here is that we moved back from `safe_replace`, which I refactored especially for that place in newast.py.

dace/frontend/python/newast.py
alexnick83: No, I can try to fix `replace_properties_dict` to ensure that `repl_dict` is indeed symbolic.

dace/transformation/interstate/state_elimination.py
alexnick83: The problem is that safe replace mixes strings and symbols, while SDFG.replace/replace_dict methods require only strings. I will do it with a custom lambda but I think you will see that it is becoming a bit confusing.

tbennun (approved):

PR #786, opened 2021-07-11
check in scal2sym promotion - fail gracefully

acalotoiu:
added check to remove candidate if any tasklet that should be made into an interstate edge could not be parsed as Python from C. Code translating C to Python refactored into separate function as it is used in two instances.

### DIFF
--- a/dace/sdfg/analysis/scalar_to_symbol.py
+++ b/dace/sdfg/analysis/scalar_to_symbol.py
@@ -172,6 +172,13 @@ def find_promotable_scalars(sdfg: sd.SDFG) -> Set[str]:
                                     cstr) is None:
                             candidates.remove(candidate)
                             continue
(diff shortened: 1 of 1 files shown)


dace/sdfg/analysis/scalar_to_symbol.py
```diff
@@ -172,6 +172,13 @@ def find_promotable_scalars(sdfg: sd.SDFG) -> Set[str]:
                                     cstr) is None:
                             candidates.remove(candidate)
                             continue
+                        newcode = translate_cpp_tasklet_to_python(cstr)
+                        try:
+                            parsed_ast = ast.parse(str(newcode))
```
tbennun, line 177: No need for the cast to string here.

tbennun (approved):

PR #790, opened 2021-07-13
Instrumentation of FPGA kernels

definelicht:
Register FPGA as a way to instrument states, using the DaCe instrumentation flow to report kernel execution times as events.

TODO: Report kernel names in the instrumentation output instead of arbitrary nodes (need @phschaad's input on this).


dace/codegen/targets/fpga.py
```diff
@@ -369,17 +371,51 @@ def generate_state(self, sdfg: dace.SDFG, state: dace.SDFGState,
...
+    cl::Event const &event = all_events[{i}];
+    event.getProfilingInfo(CL_PROFILING_COMMAND_START, &event_start);
+    event.getProfilingInfo(CL_PROFILING_COMMAND_END, &event_end);
+    if (event_start < first_start) {{
+        first_start = event_start;
+    }}
+    if (event_end > last_end) {{
+        last_end = event_end;
+    }}
+    __state->report.add_completion("{module_name}", "FPGA", event_start, event_end, {sdfg.sdfg_id}, {state_id}, {state.node_id(sg.nodes()[0])});{print_str}
```
TizianoDeMatteis, line 414: Is it ok if this is called multiple time with the same module_name?

TizianoDeMatteis (commented):
It looks good to me, thanks!

I have one general comment, not necessarily a request for changes: would it make sense to have also the option to enable wall-clock timing, instead of only FPGA profiling?
We used this for profiling some of the npbench kernels, where FPGA kernels were called tons of times and we did want to measure also the overhead of this

dace/codegen/targets/fpga.py
```diff
@@ -1842,45 +1878,47 @@ def get_kernel_name(val):
 
         self._allocated_global_arrays = set()
 
+    def _module_name(self, subgraph, state):
+        """
+        Generate the name of an FPGA module produced from the given subgraph.i
```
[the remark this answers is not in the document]
definelicht: Whoops, thanks!

definelicht:
> I have one general comment, not necessarily a request for changes: would it make sense to have also the option to enable wall-clock timing, instead of only FPGA profiling?

Yeah, that's a good idea. I will check if this is equivalent to using the "normal" profiler, otherwise I'll just add an extra report.

dace/codegen/targets/fpga.py
definelicht: You mean if the full kernel is invoked multiple times, or if there are multiple modules with the same name? I believe it will aggregate results reported for the same module (although that's currently based on a random node from the subgraph, rather than the name of the module itself, until I find a way to change this).

TizianoDeMatteis: I meant if the kernel is invoked multiple times (if multiple modules have the same name maybe it explodes earlier)

definelicht: That's fine, and an expected use case of the instrumentation flow :-)

TizianoDeMatteis (approved):
LGTM

tbennun (approved):
a bit of copypasta in intel/xilinx for the header but otherwise ok

PR #797, opened 2021-07-21
MapUnroll transformation

definelicht:


dace/transformation/dataflow/map_unroll.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def apply(self, sdfg):
+
+        from dace.transformation.dataflow import TrivialMapElimination
+
+        state = sdfg.nodes()[self.state_id]
+        map_entry = state.nodes()[self.subgraph[MapUnroll._map_entry]]
+        map_exit = state.exit_node(map_entry)
+
+        # Collect all nodes in this weakly connected component
```
alexnick83, line 61: If I understand correctly, this is extracting the SDFGState's connected component that the Map resides in. Might me worth to make it a helper method (not necessarily in this PR, just mark it for later). Networkx has a method but only for undirected graphs

dace/transformation/dataflow/map_unroll.py
```diff
@@ -0,0 +1,178 @@
...
+            if isinstance(node, nodes.NestedSDFG):
+                nested_sdfgs[node.sdfg] = node.sdfg.to_json()
+
+        # Check for local memories that need to be replicated
+        shared_transients = set(sdfg.shared_transients())
+        # Extend this list with transients that are used outside this subgraph
+        for node in state.nodes():
+            if isinstance(node, nodes.AccessNode) and node not in seen:
+                shared_transients.add(node.data)
+        local_memories = []
```
alexnick83, line 95: This might also be a good candidate for a helper method.

alexnick83 (approved):
Minor comments regarding marking some algorithms as candidates for being re-implemented as helper methods (in another PR).

PR #798, opened 2021-07-21
Unroll PEs in FPGA Codegen

definelicht:
Unroll maps with schedule `Unrolled` as part of the FPGA codegen in order to detect them as processing elements.

This is potentially fishy, since we are applying a transformation during code generation (!!) that modifies the SDFG. It is, however, a neat way of avoiding manually handling this in the FPGA codegen when detecting and generating modules.

### DIFF
--- a/dace/codegen/compiler.py
+++ b/dace/codegen/compiler.py
@@ -24,14 +24,15 @@ from dace.codegen.targets.target import make_absolute
 T = TypeVar('T')
 
 
-def generate_program_folder(sdfg,
+def generate_program_folder(sdfg_json,
                             code_objects: List[CodeObject],
                             out_path: str,
                             config=None):
     """ Writes all files required to configure and compile the DaCe program
         into the specified folder.
 
-        :param sdfg: The SDFG to generate the program folder for.
+        :param sdfg_json: The previously serialized JSON of the SDFG to generate
+                          the program folder for.
         :param code_objects: List of generated code objects.
         :param out_path: The folder in which the build files should be written.
         :return: Path to the program folder.
@@ -88,13 +89,16 @@ def generate_program_folder(sdfg,
     else:
         Config.save(os.path.join(out_path, "dace.conf"))
 
-    if sdfg is not None:
+    if sdfg_json is not None:
         # Save the SDFG itself and its hash
-        hash = sdfg.save(os.path.join(out_path, "program.sdfg"), hash=True)
-        filepath = os.path.join(out_path, 'include', 'hash.h')
-        contents = f'#define __HASH_{sdfg.name} "{hash}"\n'
+        with open(os.path.join(out_path, "program.sdfg"), "w") as fp:
+            fp.write(dace.serialize.dumps(sdfg_json))
+        hash = sdfg_json["attributes"]["hash"]
+        filepath = os.path.join(out_path, "include", "hash.h")
+        contents = (
+            f'#define __HASH_{sdfg_json["attributes"]["name"]} "{hash}"\n')
         if not identical_file_exists(filepath, contents):
-            with open(filepath, 'w') as hfile:
+            with open(filepath, "w") as hfile:
                 hfile.write(contents)
 
     return out_path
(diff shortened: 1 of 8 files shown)


definelicht:
@tbennun and @TizianoDeMatteis I'm not really asking for a code review (yet), but rather for you feedback on whether this is a really bad idea (calling a transformation as part of the codegen), or if you think this is an acceptable solution?

It means that the `program.sdfg` saved in `.dacecache` will not look like the SDFG that was compiled, for example, because it will have the unroll transformation applied to it.

The advantage of this approach is that the completely standard FPGA kernel detection can analyze the unrolled components as modules without any further changes.

definelicht:
Maybe @alexnick83 can also give his opinion on whether this is a terrible approach or if it's fine?

alexnick83:
> Maybe @alexnick83 can also give his opinion on whether this is a terrible approach or if it's fine?

We already do library node expansion in the codegen, so it is not unprecedented. From a workflow perspective, wouldn't it make more sense for this to be part of `SDFG.apply_fpga_transformations`?

definelicht:
That's a good point. The reason to execute this in the codegen is that it is based off a specific schedule (`dtypes.ScheduleType.Unrolled`), so it's not really an optional transformation, but rather part of the codgen of translating the map into code 🤔

alexnick83:
The main issue that I see is the one you have already identified; the SDFG stored in `.dacecache` is not the code-generated SDFG. I see two potential solutions here:
- Do some kind of special exception (perhaps part of the compilation API) that replaces the "original" SDFG with the "unrolled" one.
- Save both SDFGs in `.dacecache`

TizianoDeMatteis:
Following Alex's comment, I see this as another kind of "expansion" that we do while we generate the code.
As long as it is clear what we do, I can live with this (especially if I think about what we should do to avoid it)

definelicht:
@alexnick83 What do you think of this solution to the ==-2=-309=2-039854=1253

dace/codegen/compiler.py
```diff
@@ -24,14 +24,15 @@
 T = TypeVar('T')
 
 
-def generate_program_folder(sdfg,
+def generate_program_folder(sdfg_json,
```
tbennun, line 27: that is strange. Care to explain?

dace/sdfg/sdfg.py
```diff
@@ -1769,12 +1769,16 @@ def compile(self, output_file=None) -> \
         # Fill in scope entry/exit connectors
         sdfg.fill_scope_connectors()
 
+        # Serialize SDFG before generating code, since the codegen is allowed
+        # to mutate the SDFG internally
+        sdfg_json = sdfg.to_json(hash=True)
```
tbennun, line 1774: That is definitely not something we want. Since the SDFG with the materialized default storage/schedules and line information is created during code generation, that's the one we want to save.

tbennun (changes requested):
I have a critical issue with the PR

dace/sdfg/sdfg.py
definelicht: The reason why I implemented this is exactly the opposite: as we discussed above, I would think that the user expects to see the same SDFG that they edited and saved, not one with mysterious changes done by the codegen!

That's why we are changing this in the first place: otherwise `program.sdfg` will be the version where the map has been unrolled into the SDFG, which is probably not what the user expects to see just because they put the unroll schedule on a map.

dace/codegen/compiler.py
definelicht: Same as other discussion

dace/sdfg/sdfg.py
```diff
@@ -1756,6 +1758,9 @@ def compile(self, output_file=None) -> \
 
         # Clone SDFG as the other modules may modify its contents
         sdfg = copy.deepcopy(self)
```
tbennun, line 1760: told you we deepcopy :)

tbennun (approved):

PR #804, opened 2021-07-30
Convert between vector types on FPGA

definelicht:
Allow copying from one length vector to another length vector in FPGA codes in order to do memory buffering with gearboxing in device code.

### DIFF
--- a/dace/libraries/standard/nodes/gearbox.py
+++ b/dace/libraries/standard/nodes/gearbox.py
@@ -3,7 +3,7 @@
 Provides a library node that converts from a stream of type
 vector(vector(dtype, w0)) to a stream of type vector(dtype, w1), or vice versa.
 This is useful for achieving efficient memory reads on Xilinx FPGAs, where
-modules accessing memories should always read or write512-bit vectors, which
+modules accessing memories should always read or write 512-bit vectors, which
 then potentially need to be narrowed down to the vector width of the
 computational kernel.
 
@@ -29,9 +29,7 @@ class ExpandGearbox(dace.transformation.ExpandTransformation):
 
         if is_pack:
             vtype = out_desc.dtype
-            etype = in_desc.dtype
         else:
-            etype = out_desc.dtype
             vtype = in_desc.dtype
 
         sdfg = dace.SDFG("gearbox")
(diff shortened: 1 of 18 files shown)


definelicht:
@tbennun This actually seems to work on the device side, but on the CPU side I cannot instantiate a ``dace::vec<dace::vec<float, 4>, 4>``, because it seems all the vector types are hardcoded.

dace/libraries/standard/nodes/gearbox.py
```diff
@@ -0,0 +1,165 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+"""
```
tbennun, line 2: Documentation should move to the node itself. This goes into vscode too so it's nice to have it there

dace/libraries/standard/nodes/gearbox.py
```diff
@@ -0,0 +1,165 @@
...
+        (in_edge, in_desc, out_edge, out_desc, is_pack,
+         gear_factor) = node.validate(parent_sdfg, parent_state)
+
+        if is_pack:
+            vtype = out_desc.dtype
+        else:
+            vtype = in_desc.dtype
+
+        sdfg = dace.SDFG("gearbox")
+        in_desc_inner = copy.copy(in_desc)
```
tbennun, line 36: I think it should be deepcopy because of location and stuff

dace/transformation/interstate/fpga_transform_state.py
```diff
@@ -62,7 +62,8 @@ def can_be_applied(graph, candidate, expr_index, sdfg, strict=False):
                     return False
 
                 # Cannot allocate FIFO from CPU code
-                if sdict[node] is None:
+                if sdict[node] is None and (not nodedesc.transient or node.data
```
tbennun, line 65: sneaky change?

tbennun (changes requested):
Minor changes

dace/libraries/standard/nodes/gearbox.py
definelicht: Ok! I don't use VSCode so I don't see it. I'll put it in the node

dace/transformation/interstate/fpga_transform_state.py
definelicht: Ye, I can revert it for this PR since it's not actually used, but it's unnecessarily restrictive right now, not allowing anything with a stream to be transformed

definelicht: Reverted

dace/libraries/standard/nodes/gearbox.py
definelicht: Done in da422396

dace/libraries/standard/nodes/gearbox.py
definelicht: Resolved in e6d1044d

tests/half_cudatest.py
```diff
@@ -80,6 +80,6 @@ def halftest(A: dace.float16[N], mask: dace.int32[N]):
 
 
 if __name__ == '__main__':
-    test_relu()
```
tbennun, line 83: uhh why?

tbennun (changes requested):

tests/half_cudatest.py
definelicht: whoops, reverted

tbennun (approved):

PR #806, opened 2021-08-02
Hbm support 1 bank

jnice-81:
Adds support for leaving the distributed subset index away when an array is on HBM, but only uses one single bank


dace/codegen/targets/fpga.py
```diff
@@ -69,6 +69,7 @@ def is_fpga_kernel(sdfg, state):
             return False
     return True
 
+
 def is_hbm_array(array: dt.Data):
```
definelicht, line 73: This is still used in other places, right?

dace/codegen/targets/fpga.py
```diff
@@ -81,6 +82,20 @@ def is_hbm_array(array: dt.Data):
         return False
 
 
+def is_hbm_array_with_distributed_index(array: dt.Data):
+    """
+    :return: True if this array is placed on HBM and has a 'fake' first
```
definelicht, line 87:
```suggestion
    :return: True if this array is placed on HBM and has an extra first 
```

dace/codegen/targets/fpga.py
jnice-81: Yes there are like 2 or 3 locations

definelicht (approved):

PR #808, opened 2021-08-04
Hbm bank split transform

jnice-81:
Adds the HbmBankSplit Transformation, which allows to distribute/gather arrays to many HBM-banks. It's intended to be used for Host-Device copies, but works in principle everywhere.

### DIFF
--- a/dace/transformation/dataflow/hbm_bank_split.py
+++ b/dace/transformation/dataflow/hbm_bank_split.py
@@ -1,0 +1,197 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+from typing import Any, Dict, Iterable, List, Tuple, Union
+
+from dace import data, dtypes, properties, registry
+from dace.sdfg import utils
+from dace.transformation import transformation
+from dace.sdfg import nodes as nd
+from dace import SDFG, SDFGState, memlet
+import functools
+
+
+@registry.autoregister_params(singlestate=True)
+@properties.make_properties
+class HbmBankSplit(transformation.Transformation):
+    """
+    A transformation that allows to split an array and distribute on
+    an array with one dimension more, or the reverse operation. Works in principle
+    with arbitrary arrays, but it's real use case is to distribute data on many HBM-banks.
+    Matches any 2 AccessNodes connected by any edge, if the dimensionality of the two accessed
+    arrays differ by exactly one. The sizes of the arrays have to be large enough with
+    respect to the split executed, but this is not verified. While it is allowed to use symbolics 
+    for the shapes of the array, it is expected that each dimension is divisable by the number
+    of splits specified.
+
+    Examples:
+    Distribute: Suppose for example we copy from A to B, where A has shape [100, 100] and B shape
+    [10, 100, 10]. We can distribute A in that case to B using the transformation by setting
+    split_array_info=[1, 10]. A will then be divided along it's second dimension into 10 parts
+    of size [100, 10] and distributed on B.
+    Gather: Suppose A has shape [4, 50, 50] and B has shape [100, 100]. If one leaves 
+    split_array_info set to None and applies the transformation, it will try to split 
+    equally in all dimensions. In this case this works by setting split_array_info=[2, 2]
+    since 2**2 = 4.
+    Therefore A[0] will be copied to B[0:50, 0:50], A[1] to B[0:50, 50:100], A[2] to B[50:100, 0:50] and
+    A[3] to B[50:100, 50:100].
+
+    Note that simply turning the AccessNodes for the arrays in the above examples would
+    have lead to the inverse operation, i.e. the gather would become a distribute and 
+    the other way around.
+    """
+
+    _src_node = nd.AccessNode("")
+    _dst_node = nd.AccessNode("")
+
+    # dtype=List[int]
+    split_array_info = properties.Property(
+        dtype=List,
+        default=None,
+        allow_none=True,
+        desc="Describes how many times this array is split in each dimension, "
+        "where the k-th number describes how many times dimension k is split. "
+        "If the k-th number is 1 this means that the array is not split in "
+        "the k-th dimension at all. "
+        "If None, then the transform will try to split equally in each dimension."
+    )
+
+    default_to_storage = properties.Property(
+        dtype=dtypes.StorageType,
+        default=dtypes.StorageType.CPU_Heap,
+        allow_none=False,
+        desc=
+        "The storage type of involved arrays will be set to the value of this property if "
+        "they have Default storage type. ")
+
+    def _get_split_size(self, virtual_shape: Iterable,
+                        split_count: List[int]) -> List[int]:
+        """
+        :return: the shape of a part-array on one HBMbank
+        """
+        new_shape_list = []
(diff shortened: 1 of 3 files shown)


dace/transformation/dataflow/hbm_bank_split.py
```diff
@@ -0,0 +1,197 @@
...
+
+        plain_array = lambda array: isinstance(
+            array, data.Array) and not isinstance(array, data.View)
+
+        if not plain_array(src_array):
+            raise ValueError(
+                f"{src.data} must be of type array and mustn't be a view")
+        if not plain_array(dst_array):
+            raise ValueError(
+                f"{dst.data} must be of type array and mustn't be a view")
```
definelicht, line 95: Shouldn't you just return false here?

dace/transformation/dataflow/hbm_bank_split.py
```diff
@@ -0,0 +1,197 @@
...
+    A transformation that allows to split an array and distribute on
+    an array with one dimension more, or the reverse operation. Works in principle
+    with arbitrary arrays, but it's real use case is to distribute data on many HBM-banks.
+    Matches any 2 AccessNodes connected by any edge, if the dimensionality of the two accessed
+    arrays differ by exactly one. The sizes of the arrays have to be large enough with
+    respect to the split executed, but this is not verified. While it is allowed to use symbolics 
+    for the shapes of the array, it is expected that each dimension is divisable by the number
+    of splits specified.
+
+    Examples:
```
definelicht, line 25: You don't describe what splitting the array actually means: What happens in the SDFG when you apply this transformation?

tests/fpga/hbm_bank_split_test.py
```diff
@@ -0,0 +1,55 @@
...
+        xform.split_array_info = [3, 2, 4]
+        xform.apply(sdfg)
+    a = np.random.uniform(0, 100, [120, 100, 100]).astype(np.int32)
+    sdfg(a=a, b=b)
+    assert np.allclose(a[80:120, 50:100, 75:100], b[23])
+    assert np.allclose(a[0:40, 50:100, 75:100], b[7])
+    assert np.allclose(a[40:80, 0:50, 25:50], b[9])
+
+
+if __name__ == "__main__":
```
definelicht, line 52: I'm missing an example of why this transformation is useful in practice, it's not so easy to understand just from looking at it. Or maybe it needs to be explained in the docstring why you need this

dace/transformation/dataflow/hbm_bank_split.py
```diff
@@ -0,0 +1,197 @@
...
+        # create the copy-subgraph
+        ndrange = dict()
+        usable_params = []
+        for i in range(ndim):
+            usable_params.append(f"i{i}")
+        for i in range(ndim):
+            ndrange[usable_params[i]] = f"0:{split_info[i]}"
+        graph.remove_edge_and_connectors(graph.edges_between(src, dst)[0])
+        copy_map_enter, copy_map_exit = graph.add_map(
+            "copy_map", ndrange, dtypes.ScheduleType.Unrolled)
```
definelicht, line 168:
```suggestion
            "hbm_bank_split", ndrange, dtypes.ScheduleType.Unrolled)
```

tests/fpga/hbm_bank_split_test.py
jnice-81: The actual use case is this: Suppose you implement say e.g. dot to use HBM, hence you distribute both inputs across many banks, and your inputs (on device) are maybe of shape [16, N / 16]. But likely you would like to initialize this from just 2 1D arrays of shape [N]. Now here this transformation gets helpful, because you can simply connect the 2 arrays by an edge and apply - the host array will be distributed automatically.

Then there are situations, e.g when computing A^T*x where it's also usefull to not split along dimension 0. Here for example if A is split along rows (the same column is always on 1 bank) one could avoid reducing over multiple banks, so it would make sense to split along dimension 1 in that case. (It's because if A is split along columns you would actually get multiple y vectors which you have to add up to get the final result for A^T*x)

definelicht:
Could we use this transformation for splitting to different DDR banks too?
Based on some recent discussion, I think we're gonna want to extend the work you did to also work on DDR banks.

jnice-81:
> Could we use this transformation for splitting to different DDR banks too?
> Based on some recent discussion, I think we're gonna want to extend the work you did to also work on DDR banks.

Sure, actually the transformation has nothing to do with HBM specifically, it just matches on directly connected Accessnodes, where one of the arrays has 1 dimension more than the other.

definelicht:
What do you think about renaming this to `BankSplit` then?

dace/transformation/dataflow/hbm_bank_split.py
```diff
@@ -0,0 +1,197 @@
...
+    # dtype=List[int]
+    split_array_info = properties.Property(
+        dtype=List,
+        default=None,
+        allow_none=True,
+        desc="Describes how many times this array is split in each dimension, "
+        "where the k-th number describes how many times dimension k is split. "
+        "If the k-th number is 1 this means that the array is not split in "
+        "the k-th dimension at all. "
+        "If None, then the transform will split the first dimension."
```
TizianoDeMatteis, line 56: if None, the arrey will be splitted in as many parts as `shape[0]`?

TizianoDeMatteis (dismissed):

dace/transformation/dataflow/hbm_bank_split.py
jnice-81: Exactly. I will update the docstring.

dace/transformation/dataflow/bank_split.py
```diff
@@ -0,0 +1,194 @@
...
+            return False
+        if not plain_array(dst_array):
+            return False
+
+        # same dimensions means HBM-array needs 1 dimension more
+        collect_src = len(src_array.shape) - 1 == len(dst_array.shape)
+        distribute_dst = len(src_array.shape) + 1 == len(dst_array.shape)
+        if collect_src:
+            try:
+                tmp = int(src_array.shape[0])
```
definelicht, line 102: Use `symbolic.evaluate(src_array.shape[0], sdfg.constants)` so we can use constants here also. Instead of `except:`, catch the specific exception thrown by `symbolic.evaluate` if it cannot evaluate it to a constant

definelicht (changes requested):

dace/transformation/dataflow/bank_split.py
```diff
@@ -0,0 +1,195 @@
...
+            return False
+
+        # same dimensions means HBM-array needs 1 dimension more
+        collect_src = len(src_array.shape) - 1 == len(dst_array.shape)
+        distribute_dst = len(src_array.shape) + 1 == len(dst_array.shape)
+        if collect_src:
+            try:
+                tmp = symbolic.evaluate(src_array.shape[0], sdfg.constants)
+            except TypeError:
+                return False
```
definelicht, line 105: Sorry I gave you a suboptimal solution, this can just be:

```suggestion
        if collect_src and symbolic.issymbolic(src_array.shape[0], sdfg.constants):
            return False
```

definelicht (changes requested):

PR #811, opened 2021-08-05
FPGA Kernel detection fixes

TizianoDeMatteis:
This fix tries to detect local buffers on an SDFG path, and, if this is the case, then the path should not be in a different kernel

### DIFF
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -39,6 +39,11 @@ _FPGA_STORAGE_TYPES = {
     dtypes.StorageType.FPGA_Registers, dtypes.StorageType.FPGA_ShiftRegister
 }
 
+_FPGA_LOCAL_STORAGE_TYPES = {
+    dtypes.StorageType.FPGA_Local, dtypes.StorageType.FPGA_Registers,
+    dtypes.StorageType.FPGA_ShiftRegister
+}
+
 
 def vector_element_type_of(dtype):
     if isinstance(dtype, dace.pointer):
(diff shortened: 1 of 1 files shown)


dace/codegen/targets/fpga.py
```diff
@@ -1293,13 +1294,19 @@ def increment(kernel_id):
                             kernel = self._node_to_kernel[succ_edge_dst_repr]
                             break
                     else:
-                        kernel = max_kernels
-                        if (isinstance(e.dst, nodes.AccessNode) and isinstance(
-                                sdfg.arrays[e.dst.data], dt.View)):
-                            # Skip views
-                            pass
+                        # Trace this edge forward: if it lands to something that has a kernel id and
+                        # along the path there is at least one local buffer, then reuse that kernel id
```
definelicht, line 1298:
```suggestion
                        # Trace this edge forward: if it finds something that has a kernel id and
                        # there is at least one local buffer along the way, then reuse that kernel id
```

dace/codegen/targets/fpga.py
```diff
@@ -1348,12 +1362,43 @@ def _trace_back_edge(self, edge, state, look_for_kernel_id=False):
...
-            return False
+            return contains_compute and contains_only_global_buffers
         else:
             src_repr = utils.unique_node_repr(state, curedge.src)
             return self._node_to_kernel[
                 src_repr] if src_repr in self._node_to_kernel else None
 
+    def _trace_forward_edge(self, edge, state: dace.SDFGState):
+        '''
+        Given ad edge, this traverses the edges forward.
```
definelicht, line 1373:
```suggestion
        Given an edge, this traverses the edges forward.
```

dace/codegen/targets/fpga.py
```diff
@@ -1348,12 +1362,43 @@ def _trace_back_edge(self, edge, state, look_for_kernel_id=False):
...
+        It can be used either for:
+        - understanding if along the forward path there is a local buffer,  and
+        - returning the the kernel_id of a successor if any
```
definelicht, line 1376: Add the return value to the docstring and the type to the signature

definelicht (approved):

PR #813, opened 2021-08-08
Hbm fix to many interfaces

jnice-81:
Fixes the problem that dace generates distinct memory interfaces for all banks of HBM multibank arrays even when not used by a specific subgraph. This is achieved by storing the bank together with the interface id for each subgraph in make_parameters, and later only generating the mentioned interfaces for each module.


dace/codegen/targets/xilinx.py
```diff
@@ -495,10 +495,10 @@ def generate_kernel_boilerplate_pre(self, sdfg, state_id, kernel_name,
                         memory_bank[1], sdfg)
                 else:
                     lowest_bank_index = int(memory_bank[1])
-                for bank in fpga.iterate_hbm_multibank_arrays(
-                        data_name, data, sdfg):
+                for bank, if_id in fpga.iterate_hbm_interface_ids(
```
definelicht, line 498: Don't use `if` as an abbreviation, this is confusing :-) Just write out `interface_id`

tests/fpga/hbm_multiple_interface_test.py
```diff
@@ -0,0 +1,73 @@
...
+                          dst_conn="_x2")
+    state.add_memlet_path(t1,
+                          m1_out,
+                          acc_write1,
+                          memlet=memlet.Memlet("a[0, 1]"),
+                          src_conn="_y1")
+
+    sdfg.apply_fpga_transformations()
+    sdfg.apply_transformations(InlineSDFG)
+    sdfg.apply_transformations(MapUnroll)
```
definelicht, line 51: You should assert that the right number of transformations are actually applied here, i.e.:
```python
assert sdfg.apply_transformations(InlineSDFG) == 1
```

definelicht (changes requested):

definelicht (approved):

PR #815, opened 2021-08-10
Kernel instrumentation fix for single/multiple kernels

definelicht:
Kernel instrumentation for FPGA kernels was wrong for Xilinx kernels that had multiple processing elements.


TizianoDeMatteis (approved):
Looks good to me

PR #819, opened 2021-08-13
Fix FPGA profiling units

definelicht:
There was a mixup between the nanoseconds reported by OpenCL and microseconds expected by the profiler.


phschaad (approved):
LGTM

PR #822, opened 2021-08-15
Rot90

alexnick83:
This PR adds replacement methods for NumPy flip and rot90 methods.  
**NOTE**: The methods return copies of the original data.


tbennun (approved):
Looks good. We need to improve our subset stride generation

PR #823, opened 2021-08-16
Tasklet fusion

alexnick83:
This PR implements "simple" TaskletFusion, i.e. fusion of 2 Tasklets directly connected with each other.

### DIFF
--- a/dace/transformation/dataflow/tasklet_fusion.py
+++ b/dace/transformation/dataflow/tasklet_fusion.py
@@ -42,9 +42,11 @@ class PythonRHSExtractor(ast.NodeVisitor):
             self.assignments.add(node.targets[0].id)
 
 
-@registry.autoregister_params(singlestate=True, strict=True)
+@registry.autoregister_params(singlestate=True, strict=False)
 class SimpleTaskletFusion(pm.Transformation):
     """ Fuses two connected Tasklets.
+        It is recommended that this transformation is used on Tasklets that
+        contain only simple assignments.
     """
 
     _t1 = pm.PatternNode(nodes.Tasklet)
--- a/dace/transformation/dataflow/tasklet_fusion.py
+++ b/dace/transformation/dataflow/tasklet_fusion.py
@@ -1,0 +1,362 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+""" Contains classes that fuse Tasklets """
+
+from dace.sdfg.utils import consolidate_edges
+from copy import deepcopy
+from numbers import Number
+from typing import Any, Dict, List
+import ast
+import dace
+import re
+import sympy
+from dace import data, dtypes, registry, subsets, symbolic
+from dace.sdfg import nodes
+from dace.sdfg import utils as sdutil
+from dace.sdfg.graph import OrderedMultiDiConnectorGraph
+from dace.transformation import transformation as pm
+from dace.transformation.subgraph.helpers import subgraph_from_maps
+from functools import reduce
+
+
+class TaskletReplace(ast.NodeTransformer):
+    """
+    Replaces connector names in Tasklet code.
+    """
+    def __init__(self, repl_dict: Dict[str, str]) -> None:
+        """
+        Initializes AST transformer.
+        :param repl_dict: Replacement dictionary.
+        """
+        self.repl_dict = repl_dict
+
+    def visit_Name(self, node: ast.Name) -> Any:
+        # Replace connector name
+        if node.id in self.repl_dict:
+            node.id = self.repl_dict[node.id]
+        return self.generic_visit(node)
+
+
+class TaskletExtract(ast.NodeVisitor):
+    def __init__(self):
+        self.assignments = set()
+    
+    def visit_Assign(self, node: ast.Assign):
+        if isinstance(node.targets[0], ast.Name):
+            self.assignments.add(node.targets[0].id)
+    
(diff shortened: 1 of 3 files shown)


dace/transformation/dataflow/tasklet_fusion.py
```diff
@@ -0,0 +1,226 @@
...
+
+        # Avoid circles
+        t1_dst = set()
+        for e in graph.out_edges(t1):
+            t1_dst.add(e.dst)
+        t2_src = set()
+        for e in graph.in_edges(t2):
+            t2_src.add(e.src)
+        if len(t1_dst.intersection(t2_src)):
+            return False
```
definelicht, line 84: Isn't this guaranteed to never happen by DaCe semantics?

dace/transformation/dataflow/tasklet_fusion.py
```diff
@@ -0,0 +1,226 @@
...
+        if isinstance(node.targets[0], ast.Name):
+            self.assignments.add(node.targets[0].id)
+
+
+@registry.autoregister_params(singlestate=True, strict=False)
+class SimpleTaskletFusion(pm.Transformation):
+    """ Fuses two connected Tasklets.
+        It is recommended that this transformation is used on Tasklets that
+        contain only simple assignments.
+    """
```
definelicht, line 50: Put an example of the before and after code (including memlets) here to help explain what happens

definelicht (changes requested):

dace/transformation/dataflow/tasklet_fusion.py
alexnick83: The transformation always fuses the second tasklet (`t2`) to the first one (`t1`). The check is there to enforce a certain ordering of fusions so that a circle is not produced due to this design. A simple example is the following:

Assume that the tasklet subgraph contains the edges:
- `t1 -> t2`
- `t1 -> t3`
- `t3 -> t2`

If we fuse `t2` to `t1`, then the `t3 -> t2` edge will become `t3 -> t1`, producing a cycle. The subgraph can of course be fused. We just have to fuse first `t3` to `t1`.

definelicht (commented):
Looks good to me, maybe Tal also wants to take a look, since they worked on it in DaCeML

dace/transformation/dataflow/tasklet_fusion.py
```diff
@@ -0,0 +1,294 @@
...
+import ast
+import dace
+import re
+from dace import dtypes, registry
+from dace.sdfg import nodes
+from dace.sdfg import utils as sdutil
+from dace.transformation import transformation as pm
+
+
+class ConnectorRenamer(ast.NodeTransformer):
```
tbennun, line 14: isn't this ASTFindReplace from dace.frontend.python.astutils?

dace/transformation/dataflow/tasklet_fusion.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def visit_Name(self, node: ast.Name) -> Any:
+        # Rename connector
+        if node.id in self.repl_dict:
+            node.id = self.repl_dict[node.id]
+        return self.generic_visit(node)
+
+
+class PythonRHSExtractor(ast.NodeVisitor):
+    """ Extracts assignments' RHS in Tasklet code.
```
tbennun, line 31: lhs?

dace/transformation/dataflow/tasklet_fusion.py
```diff
@@ -0,0 +1,294 @@
...
+        ```
+        `e1: [s1], [sc1], t1, '__in1', [m1]`
+        `e2: [s2], [sc2], t1, '__in2', [m2]`
+        `e5: [s3], [sc3], t1, '__in3', [m3]`
+        `e6: [s4], [sc4], t1, '__in4', [m4]`
+        `e7: t1, '__out_1', [d1], [dc1], [m5]`
+    """
+
+    _t1 = pm.PatternNode(nodes.Tasklet)
+    _t2 = pm.PatternNode(nodes.Tasklet)
```
tbennun, line 121: no need for the preceding underscores (the interface is nicer without them)

dace/transformation/dataflow/tasklet_fusion.py
```diff
@@ -0,0 +1,294 @@
...
+                       strict: bool = False):
+
+        t1 = graph.node(candidate[SimpleTaskletFusion._t1])
+        t2 = graph.node(candidate[SimpleTaskletFusion._t2])
+
+        # Tasklets must be of the same language
+        if t1.language != t2.language:
+            return False
+
+        # Avoid circles
```
tbennun, line 144:
```suggestion
        # Avoid cycles
```

tbennun (changes requested):
minor comments, looks ok otherwise

dace/transformation/dataflow/tasklet_fusion.py
alexnick83: It seems that yes, but `ASTFindReplace` does more things that are not needed. Hmm, not sure if should change it or not.

alexnick83: My tests do not pass after doing the change. I think `ASTFindReplace` returns after doing a single `ast.Name` replacement, while `ConnectorRenamer` will replace all `ast.Names`. I am not sure if this is intended or a bug but I don't think it concerns this PR.

tbennun (approved):

PR #830, opened 2021-08-21
SpMV optimzation

alexnick83:
This PR introduces a tutorial on applying a few simple optimizations to SpMV. It further includes a couple of new transformations and updates to existing ones:
- TrivialTaskletElimination (new): removes trivial Tasklets with code of the form `out = inp`.
- PruneConnectors (updated): takes into account symbol mappings.
- WarpTiling (updated): when there is wcr, it creates a transient and sets it to the (initial) value of the output in a previous state.
- AugAssignToWCR (updated): it now works with Python Tasklets.
- (helper) `extract_map_dims` (updated): it now works with Maps that have only one dimension.
- LoopToMap (updated): it now has an option to explicitly select the iteration variable.
- InlineSDFG (updated): it will now split weakly connected components that have been split from the nested SDFG graph due to the transformation to a previous or next State.
- StartStateElimination (new): it removes the start state of a nested SDFG when it is empty.


tutorials/spmv.ipynb
```diff
@@ -0,0 +1,418 @@
+{
```
tbennun, line 1: paths can be sanitized (replace `alexnick` with `user`)

tutorials/spmv.ipynb
```diff
@@ -0,0 +1,418 @@
...
+ "cells": [
+  {
```
tbennun, line 3: Missing the respective MKL and CUSPARSE results IMO

dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -541,6 +541,10 @@ def apply(self, sdfg: SDFG):
                     '(reconnecting inputs)')
             state.add_edge(edge.src, edge.src_conn, node, edge.dst_conn,
                            edge.data)
+            # Fission state if necessary
```
tbennun, line 544: what happens if the inlined SDFG is inside a scope (e.g., two connected components in a map)? would it automatically not create two WCCs?

Also, what if there were already more than one WCC in the graph before inlining? Would it fission states that shouldn't be?

tbennun (commented):
overall looks good. tutorial could use some more numbers for comparison and I have a few questions on the inlinesdfg change

dace/transformation/interstate/sdfg_nesting.py
alexnick83: This detects the case where a weakly connected group of nodes in the parent state (A) that was connected to the nested SDFG ends up disconnected from the subgraph of the nested SDFG (B) after inlining. This may create a data race since (A) may be executed in parallel with (B) or in the wrong order after the transformation. The solution is to move (A) to a separate state before (after) if (A) was input (output) to the nest SDFG.

Your first question may be a real issue. Therefore, the current check may fail in some cases. It might be better to check that none of the nodes in (B) are part of (A).

I do not think that this is an issue. If there is a group (A) detected, it will be moved in a separate state from the rest of the nodes, regardless of how may other WCCs exist in the state (after the transformation).

tbennun (approved):

tbennun:
@alexnick83 please squash before merging to avoid multiple copies of the Jupyter notebook (including unsanitized paths, for example)

PR #831, opened 2021-08-23
Hbm transform

jnice-81:
Adds HbmTransform, a transformation that helps to transform SDFGs to use HBM.  Additionally adds a function to arrays which can update it's shape and a usability function to get all innermost edges, given an edge (could also be moved to the transformation). 
can_apply has some shared code with the can_apply function of FPGATransformState, therefore this was made a function in fpga, which is called by both.


TizianoDeMatteis (approved):
This looks good to me

Sorry for the long (unnecessary) delay in dealing with this.

PR #832, opened 2021-08-24
Add support for unrolled map around RTL tasklet

carljohnsen:
This pull request adds the functionality to wrap a tasklet targeting hardware in an unrolled map, allowing for easy replication of an RTL tasklet.


dace/codegen/targets/cpu.py
```diff
@@ -1109,6 +1109,9 @@ def memlet_definition(self,
         elif var_type in [DefinedType.Stream, DefinedType.StreamArray]:
             if not memlet.dynamic and memlet.num_accesses == 1:
                 if not output:
+                    index = cpp.cpp_offset_expr(desc, memlet.subset)
+                    if isinstance(desc, data.Stream) and desc.is_stream_array():
+                        expr = f"{memlet.data}[{index}]"
```
definelicht, line 1114: You can move the computation of index into the branch, since it's not needed otherwise

dace/codegen/targets/fpga.py
```diff
@@ -716,6 +716,7 @@ def make_parameters(self, sdfg, state, subgraphs):
             data_to_node.update(
                 {node.data: node
                  for node in subgraph.nodes() if isinstance(node, dace.sdfg.nodes.AccessNode)})
+            rtl_subgraph = any([isinstance(node, nodes.RTLTasklet) for node in subgraph.nodes()])
```
definelicht, line 719: At first glace this looks like a graph object, not a bool, better to rename:
```suggestion
            is_rtl_subgraph = any([isinstance(node, nodes.RTLTasklet) for node in subgraph.nodes()])
```

dace/codegen/targets/fpga.py
```diff
@@ -745,15 +741,9 @@ def make_parameters(self, sdfg, state, subgraphs):
             for n in subgraph.sink_nodes():
                 # Check if the node is connected to an RTL tasklet, in which
                 # case it should be an external stream
-                dsts = [e.dst for e in state.out_edges(n)]
-                srcs = [e.src for e in state.in_edges(n)]
-                tasks = [t for t in dsts + srcs if isinstance(t, dace.nodes.Tasklet)]
-
-                external = any([t.language == dtypes.Language.SystemVerilog for t in tasks])
+                external = rtl_subgraph
```
definelicht, line 744:
```suggestion
                is_external = rtl_subgraph
```

dace/codegen/targets/rtl.py
```diff
@@ -96,8 +97,17 @@ def copy_memory(self, sdfg: sdfg.SDFG, dfg: state.StateSubgraphView, state_id: i
...
+                arr = sdfg.arrays[edge.data.data]
+                if isinstance(arr, data.Array):
+                    line: str = "{}* {} = &{}[0];".format(dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn,
+                                                          edge.src.data)
+                elif isinstance(arr, data.Scalar):
+                    line: str = "{} {} = {};".format(dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn,
+                                                     edge.src.data)
+        elif isinstance(edge.src, nodes.MapEntry) and isinstance(edge.dst, nodes.Tasklet):
+            # TODO works for one kernel, but might break with several. Also only handles if the scope and tasklet are directly connected.
+            self.n_unrolled = symbolic.evaluate(edge.src.map.range[0][1] + 1, sdfg.constants)
```
definelicht, line 109: Why does it break with multiple kernels? If it's because of the `self.n_unrolled`, you can make this a dict instead or something, so you can assign it differently for different kernels

dace/codegen/targets/rtl.py
```diff
@@ -213,196 +225,218 @@ def generate_rtl_inputs_outputs(self, buses, scalars):
...
+        ]
 
     def generate_cpp_internal_state(self, tasklet):
         internal_state_str = " ".join(
             ["{}=0x%x".format(var_name) for var_name in {
                 **tasklet.in_connectors,
                 **tasklet.out_connectors
             }])
-        internal_state_var = ", ".join(
+        internal_state_var = ", ".join(  # TODO fix
```
definelicht, line 348: Fix what? ;-)

dace/codegen/targets/rtl.py
```diff
@@ -548,19 +592,20 @@ def unparse_tasklet(self, sdfg: sdfg.SDFG, dfg: state.StateSubgraphView, state_i
                 raise NotImplementedError('Only RTL codegen for Xilinx is implemented')
         else:  # not hardware_target
             # generate verilator simulation cpp code components
-            inputs, outputs = self.generate_cpp_inputs_outputs(tasklet)
-            valid_zeros, ready_zeros = self.generate_cpp_zero_inits(tasklet)
+            inputs, outputs = self.generate_cpp_inputs_outputs(
+                tasklet, buses)  # TODO i don't know if i want to do it differently.
```
definelicht, line 596: Elaborate!

dace/codegen/targets/xilinx.py
```diff
@@ -802,9 +840,11 @@ def generate_kernel_internal(self, sdfg: dace.SDFG, state: dace.SDFGState, kerne
                                                entry_stream)
 
         for is_output, name, node, _ in external_streams:
+            # TODO find better way to extract num kernels, rather than the shape of the streams
+            num_kernels = dace.symbolic.evaluate(node.shape[0], sdfg.constants)
```
definelicht, line 844: Didn't you already put this `n_unrolled` thing?

dace/codegen/targets/xilinx.py
```diff
@@ -802,9 +840,11 @@ def generate_kernel_internal(self, sdfg: dace.SDFG, state: dace.SDFGState, kerne
...
             self._dispatcher.defined_vars.add_global(name, DefinedType.Stream, node.ctype)
-            if name not in self._stream_connections:
-                self._stream_connections[name] = [None, None]
+            #if name not in self._stream_connections:
+            #    self._stream_connections[name] = [None, None]
```
definelicht, line 847: Remove if no longer necessary?

tests/rtl/simulation_test.py
```diff
@@ -360,8 +477,129 @@ def test_multi_tasklet():
...
-    test_tasklet_scalar()
-    test_tasklet_parameter()
-    test_tasklet_vector()
-    test_multi_tasklet()
+    #test_multi_tasklet()
+    #test_tasklet_array()
+    #test_tasklet_map()
+    #test_tasklet_parameter()
+    #test_tasklet_scalar()
+    test_tasklet_vector_add()
```
definelicht, line 604: Should this be commented out?

definelicht (changes requested):

dace/codegen/targets/cpu.py
carljohnsen: True!

dace/codegen/targets/fpga.py
carljohnsen: This has been implemented.

dace/codegen/targets/fpga.py
carljohnsen: This has been implemented.

dace/codegen/targets/rtl.py
carljohnsen: Fix the debug statements :) this should have been fixed now.

dace/codegen/targets/rtl.py
carljohnsen: It was a comment to an old way of doing things. This new version uses buses, which already contains all of the collected information needed.

dace/codegen/targets/xilinx.py
carljohnsen: Only in `rtl.py`, and this is in `xilinx.py`. However, I'm thinking this TODO should just be removed, since this is only used for generating the configuration file, which does not allow for array of streams. So array of streams should be unrolled in the configuration file. Maybe it should be renamed, since there doesn't necessarily have to be multiple kernels? So something like `num_streams`?

dace/codegen/targets/xilinx.py
carljohnsen: Removed.

tests/rtl/simulation_test.py
carljohnsen: No :) Fixed.

dace/codegen/targets/rtl.py
carljohnsen: The thought arose when implementing: this could fail if there was multiple kernels, since `n_unrolled` is bound to the codegen, although everything inside might not be unrolled. I can't think of an example, but converting to a dictionary sounds like a safe move.

dace/codegen/targets/xilinx.py
definelicht: Okay, if you're getting `num_streams` then this line makes perfect sense :-)

dace/codegen/targets/rtl.py
carljohnsen: This has been implemented.

dace/codegen/targets/xilinx.py
carljohnsen: This has been changed.

dace/codegen/targets/rtl.py
```diff
@@ -423,48 +463,54 @@ def unparse_tasklet(self, sdfg: sdfg.SDFG, dfg: state.StateSubgraphView, state_i
...
+                    raise NotImplementedError('Array input for RTL hardware* targets not implemented')
                 else:
-                    buses[edge.dst_conn] = (False, total_size, vec_len)
+                    buses[edge.dst_conn] = (edge.data.data, False, total_size, vec_len, edge.data.volume)
             elif isinstance(arr, data.Stream):
-                buses[edge.dst_conn] = (False, total_size, vec_len)
+                buses[edge.dst_conn] = (edge.data.data, False, total_size, vec_len, edge.data.volume)
             elif isinstance(arr, data.Scalar):
-                scalars[edge.dst_conn] = (False, total_size * 8)
+                scalars[edge.dst_conn] = (False, total_size)  #(edge.data.data, False, total_size, 1, edge.data.volume)
```
definelicht, line 491:
```suggestion
                scalars[edge.dst_conn] = (False, total_size)
```

definelicht (approved):

PR #833, opened 2021-08-25
Stencil detection

alexnick83:
This PR implements the StencilDetection transformation, which detects Maps that perform stencil operations and substitutes them with StencilNodes.


alexnick83:
I don't have any plans to pursue this in the near future. The transformation itself could be useful (after being updated to work with CFGs) to lift explicit SDFGs to a stencil representation (also needs work; the existing node is more tuned to FPGA development). However, I don't know whether such a workflow (lifting to library nodes and then applying transformations to the nodes rather than to the explicit SDFG) would make sense for anyone currently working on stencils. I would say close it because even if someone is interested, I don't think there is a special algorithm that may be lost; it just detects access with different offsets on the same scope, as far as I recall.

PR #834, opened 2021-08-25
Support for Vitis 2021.1

definelicht:
Fixed some messy pragmas in FPGA copies that were causing errors with new Vitis.

Vitis 2021.1 manages to pipeline some reduction patterns, so we can assert II=1 in these cases.


tbennun (approved):
LGTM

PR #836, opened 2021-08-28
Inline SDFGs with multiple states

tbennun:
The new strict transformation finds nested SDFGs that exist as the only component of the state (only node + input/output access nodes), and for now only cases where views or subsets would not be used. It can inline a loop within a nested SDFG, for example.


tbennun:
@acalotoiu @alexnick83 seems like there is some (unrelated?) FPGA test failing, so please ignore it

acalotoiu (approved):
LGTM.
Tests with polybench were succesful.

PR #839, opened 2021-09-01
Sanitization of subsets in LoopToMap

alexnick83:
When comparing read and write subsets in LoopToMap for the purpose of finding potential data races, we are only interested in the indices or ranges of the subsets that depend on the loop iteration variable. This PR updates `LoopToMap.can_be_applied`  so that it removes any indices/ranges that do not depend on the iteration variable before the data-race check, resolving cases where the transformation failed to apply but should.

### DIFF
--- a/dace/transformation/interstate/loop_to_map.py
+++ b/dace/transformation/interstate/loop_to_map.py
@@ -50,6 +50,28 @@ def _check_range(subset, a, itersym, b, step):
     return found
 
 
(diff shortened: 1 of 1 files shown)


definelicht (commented):
Looks good to me, can we add a dummy index to the existing test to check that it still applies?

tim0s (commented):
Makes sense to me. Maybe "intersects" should really be named "maybe_intersects" if it is pessimistic (not in this PR though).

PR #840, opened 2021-09-01
Split interstate edges of all SDFGs before LoopToMap

alexnick83:
In the auto-optimizer, before applying LoopToMap to all levels of the SDFGs, we must split the interstate edges to all levels of the SDFG as well (not only the top-level SDFG). This PR also updates LoopToMap slightly to work with multi-state loop bodies that contain double-linked chains of nested loops. Furthermore, the PR changes the multi-state inline transformation to non-strict.


tbennun (approved):
LGTM

PR #843, opened 2021-09-08
NumPy Copy/View Semantics

alexnick83:
This PR amends slicing of DaCe Arrays in the Python frontend so that it follows NumPy rules. For example, simple slicing should generate DaCe Views.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -409,6 +409,8 @@ def _subset_has_indirection(subset, pvisitor: 'ProgramVisitor' = None):
         if not isinstance(dim, tuple):
             dim = [dim]
         for r in dim:
+            if not symbolic.issymbolic(r):
+                continue
             if symbolic.contains_sympy_functions(r):
                 return True
             if pvisitor:
@@ -1172,7 +1174,7 @@ class TaskletTransformer(ExtNodeTransformer):
             arr_type = type(parent_array)
         if arr_type == data.Scalar:
             self.sdfg.add_scalar(var_name, dtype)
-        elif arr_type == data.Array:
+        elif arr_type in (data.Array, data.View):
             self.sdfg.add_array(var_name, shape, dtype, strides=strides)
         elif arr_type == data.Stream:
             self.sdfg.add_stream(var_name, dtype)
@@ -1272,6 +1274,9 @@ class TaskletTransformer(ExtNodeTransformer):
         else:
             raise NotImplementedError
 
+        if isinstance(rng, subsets.Indices):
+            rng = subsets.Range.from_indices(rng)
+
         return rng
 
     def _update_names(self,
@@ -1324,12 +1329,19 @@ class TaskletTransformer(ExtNodeTransformer):
                     if squeezed_rng is not None:
                         # TODO: Fix for `contains_sympy_functions`
                         # not liking ints
-                        memlet.subset = subsets.Range([
-                            (symbolic.pystr_to_symbolic(b),
-                             symbolic.pystr_to_symbolic(e),
-                             symbolic.pystr_to_symbolic(s))
-                            for b, e, s in squeezed_rng.ranges
-                        ])
+                        if isinstance(squeezed_rng, subsets.Indices):
+                            memlet.subset = subsets.Range([
+                                (symbolic.pystr_to_symbolic(i),
+                                 symbolic.pystr_to_symbolic(i), 1)
+                                for i in squeezed_rng.indices
+                            ]) 
+                        else:   
+                            memlet.subset = subsets.Range([
+                                (symbolic.pystr_to_symbolic(b),
+                                symbolic.pystr_to_symbolic(e),
+                                symbolic.pystr_to_symbolic(s))
+                                for b, e, s in squeezed_rng.ranges
+                            ])
                     if self.nested and _subset_has_indirection(rng):
                         memlet = dace.Memlet.simple(memlet.data, rng)
                     if connector in self.inputs or connector in self.outputs:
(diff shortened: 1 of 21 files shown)


dace/frontend/python/newast.py
```diff
@@ -1561,6 +1580,9 @@ def parse_program(self, program: ast.FunctionDef, is_tasklet: bool = False):
             if vname.startswith('__return'):
                 if isinstance(self.sdfg.arrays[arrname], data.View):
                     # In case of a view, make a copy
+                    # NOTE: If we are at the top level SDFG (not always clear),
+                    # and it is a View of an input array, can we return a NumPy
+                    # View directly?
```
tbennun, line 1585: probably not, a non-transient array acts as a view

dace/frontend/python/replacements.py
```diff
@@ -4305,19 +4353,21 @@ def reshape(pv: 'ProgramVisitor',
...
+                                    total_size=desc.total_size,
+                                    may_alias=desc.may_alias,
+                                    alignment=desc.alignment,
+                                    find_new_name=True)
 
     # Register view with DaCe program visitor
-    pv.views[newarr] = arr
+    aset = subsets.Range.from_array(desc)
+    vset = subsets.Range.from_array(newdesc)
+    pv.views[newarr] = (arr, Memlet(f'{arr}[{aset}]->{vset}'))
```
tbennun, line 4370: would be cleaner and safer to use `Memlet(data=arr, subset=aset, other_suset=vset)`

dace/frontend/python/replacements.py
```diff
@@ -4410,17 +4463,19 @@ def flat(pv: 'ProgramVisitor', sdfg: SDFG, state: SDFGState, arr: str) -> str:
...
+                                        allow_conflicts=desc.allow_conflicts,
+                                        total_size=totalsize,
+                                        may_alias=desc.may_alias,
+                                        alignment=desc.alignment,
+                                        find_new_name=True)
         # Register view with DaCe program visitor
-        pv.views[newarr] = arr
+        aset = subsets.Range.from_array(desc)
+        vset = subsets.Range.from_array(newdesc)
+        pv.views[newarr] = (arr, Memlet(f'{arr}[{aset}]->{vset}'))
```
tbennun, line 4478: same

tests/numpy/copies_and_views_test.py
```diff
@@ -0,0 +1,82 @@
+import dace
```
tbennun, line 1: copyright missing

tests/numpy/copies_and_views_test.py
```diff
@@ -0,0 +1,82 @@
...
+    is_a_copy.f(ref)
+    assert (np.allclose(val, ref))
+
+
+if __name__ == '__main__':
+    test_set_by_view()
+    test_set_by_view_1()
+    test_set_by_view_2()
+    test_set_by_view_3()
+    test_is_a_copy()
```
tbennun, line 82: newline missing

tbennun (changes requested):
cool PR, minor changes only

dace/frontend/python/newast.py
```diff
@@ -1172,7 +1174,7 @@ def _add_access(
             arr_type = type(parent_array)
         if arr_type == data.Scalar:
             self.sdfg.add_scalar(var_name, dtype)
-        elif arr_type == data.Array:
+        elif arr_type in (data.Array, data.View):
```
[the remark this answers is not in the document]
alexnick83: The `_add_access` methods are adding outer-level accesses to nested scopes/SDFGs, i.e., they are creating connectors. Therefore, such arrays will be pointers (and not copies) in C/C++ code anyway. Do you think that there is any reason (potentially for some kind of future use) to make connectors to Views be Views themselves? Will this even work properly at the moment?

dace/frontend/python/newast.py
alexnick83: I am considering two cases here that might be interesting to users (and probably difficult to implement properly, if possible at all):

The first case is returning a View of an input. For example:
```python
def foo(x):
    return x[1:5]
    
A = np.arange(10)
B = foo(A)
assert (B.base is A)
```

The second case is returning a View of transient data:
```python
def foo():
    x = np.arange(10)
    return x[1:5]

B = foo()
assert (B.base is not B)
```

Any work on such cases will have to be done on another PR. However, do you think that there would be any benefit in trying to support such codes or is it better to discourage the users from such coding practices?

dace/frontend/python/replacements.py
alexnick83: I thought that we actually preferred the string representation but it is easy to swap.

dace/frontend/python/newast.py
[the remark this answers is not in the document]
tbennun: Sounds good

dace/frontend/python/newast.py
tbennun: No, we do want that behavior possible and we can assume that a copy will be made, as long as it's documented

dace/frontend/python/replacements.py
tbennun: No, we prefer the raw objects as much as possible (faster performance, less ambiguity). The strings are mostly for convenience.

tbennun (approved):
LGTM!

tests/symbol_dependent_transients_test.py
```diff
@@ -144,6 +172,27 @@ def test_symbol_dependent_pinned_array():
     assert (np.allclose(B, B_ref))
 
 
+@pytest.mark.skip  # @pytest.mark.gpu
+def test_symbol_dependent_gpu_view():
+    # NOTE: This test cannot produce the correct result since the input
+    # data of the reduction are not contiguous and cub:reduce doesn't support
```
tbennun, line 178: can't we force expansion to be pure here? we do this in some of the library states with a `with` statement

alexnick83: Maybe the test should just change. It is not that important at this time.

PR #844, opened 2021-09-15
Missing check in kernel detection

TizianoDeMatteis:


definelicht (approved):
I can confirm that this fixes the issue in StencilFlow!

PR #845, opened 2021-09-21
Nested inout connector

alexnick83:
Fixes the case where data used in a nested call should be both input and output connector to the nested SDFG.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -4147,40 +4147,48 @@ class ProgramVisitor(ExtNodeVisitor):
                     inputs[k] = v
             # Unset parent inputs/read accesses that
             # turn out to be outputs/write accesses.
-            # TODO: Is there a case where some data is both input and output?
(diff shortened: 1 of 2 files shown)


dace/frontend/python/newast.py
```diff
@@ -4147,40 +4147,48 @@ def visit_Call(self, node: ast.Call):
...
                 aname = memlet.data
                 rng = memlet.subset
                 access_value = (aname, rng)
                 access_key = inverse_dict_lookup(self.accesses, access_value)
+                # NOTE: `memlet in inputs.values()` doesn't work because
+                # it only looks at the subset and not the data
+                # isinput = memlet in inputs.values()
+                isinput = False
+                for other in inputs.values():
+                    if memlet == other and memlet.data == other.data:
```
tbennun, line 4160: need to check that `other` is not a tasklet first

tbennun (changes requested):

tbennun (approved):

PR #847, opened 2021-09-27
FPGA GEMM Implementation based on FPGA'20 Paper

definelicht:
Translate the HLS code from [the repo](https://github.com/spcl/gemm_hls) into DaCe.

### DIFF
--- a/dace/codegen/targets/intel_fpga.py
+++ b/dace/codegen/targets/intel_fpga.py
@@ -240,8 +240,9 @@ DACE_EXPORTED void __dace_exit_intel_fpga({sdfg.name}_t *__state) {{
             whether this is a global variable or not
         """
         vec_type = self.make_vector_type(dtype, False)
-        if buffer_size > 1:
-            depth_attribute = " __attribute__((depth({})))".format(buffer_size)
+        if buffer_size != 1:
+            depth_attribute = " __attribute__((depth({})))".format(
+                cpp.sym2cpp(buffer_size))
         else:
             depth_attribute = ""
         if cpp.sym2cpp(array_size) != "1":
@@ -443,7 +444,7 @@ DACE_EXPORTED void __dace_exit_intel_fpga({sdfg.name}_t *__state) {{
 #pragma unroll
 for (int u_{name} = 0; u_{name} < {size} - {veclen}; ++u_{name}) {{
   {name}[u_{name}] = {name}[u_{name} + {veclen}];
-}}\n""".format(name=var_name, size=arr_size, veclen=dtype.veclen)
+}}\n""".format(name=var_name, size=arr_size, veclen=cpp.sym2cpp(dtype.veclen))
         # Then do write
         res += self.make_write(defined_type, dtype, var_name, write_expr, index,
                                read_expr, wcr, is_unpack, packing_factor)
@@ -1268,6 +1269,8 @@ __kernel void \\
             self._other_codes["converters"] = CodeIOStream()
         converter_stream = self._other_codes["converters"]
 
+        veclen = cpp.sym2cpp(veclen)
+
         if is_unpack:
             converter_name = "unpack_{dtype}{veclen}".format(dtype=ctype,
                                                              veclen=veclen)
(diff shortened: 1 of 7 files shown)


samples/fpga/gemm_systolic_vectorized.py
```diff
@@ -77,21 +81,19 @@ def make_copy_to_host_state(sdfg, vec_width=1):
...
     return state
 
 
-def make_read_A(state):
+def make_read_A(sdfg, state, vtype):
 
     # A is stored with plain data type
     entry, exit = state.add_map("read_A", {
-        "n0": "0:N/P",
+        "n0": "0:N//TN",
```
TizianoDeMatteis, line 93: Could you explain somewhere the meaning of the different symbols (in particular TN, TM and W)?

samples/fpga/gemm_systolic_vectorized.py
```diff
@@ -104,332 +106,605 @@ def make_read_A(state):
...
 @click.argument("vector-width", type=int)
+@click.option("--dtype", default="float32")
+@click.option("--tile-size-n",
+              type=int,
+              default=None,
+              help="Must be a multiple of the number of processing elements.")
+@click.option("--tile-size-m",
+              type=int,
+              default=None,
+              help="Must be a multiple of the vector size.")
```
TizianoDeMatteis, line 646: Should it also divide M?

samples/fpga/gemm_systolic_vectorized.py
```diff
@@ -104,332 +106,605 @@ def make_read_A(state):
...
+@click.argument("N", type=int)
+@click.argument("K", type=int)
+@click.argument("M", type=int)
+@click.argument("num-pes", type=int)
 @click.argument("vector-width", type=int)
+@click.option("--dtype", default="float32")
+@click.option("--tile-size-n",
+              type=int,
+              default=None,
+              help="Must be a multiple of the number of processing elements.")
```
TizianoDeMatteis, line 642: Should it also divide N?

tests/fpga/map_unroll_processing_elements_test.py
```diff
@@ -18,28 +18,36 @@ def test_map_unroll_processing_elements():
...
+    A[:] = np.random.rand(N, K).astype(dace.float32.type)
+    B[:] = np.random.rand(K, M).astype(dace.float32.type)
+    C[:] = np.random.rand(N, M).astype(dace.float32.type)
+
     C_regression = A @ B + C
 
-    sdfg(A=A, B=B, C=C, N=32, K=32)
-    diff = np.linalg.norm(C_regression - C) / float(32 * 32)
+    sdfg(A=A, B=B, C=C, N=N, M=M, K=K)
+    diff = np.linalg.norm(C_regression - C) / float(N * M)
```
TizianoDeMatteis, line 50: I think this is not robust. Better to use allclose or divide by the norm of the reference value (https://www.netlib.org/lapack/lug/node75.html)

dace/codegen/targets/intel_fpga.py
```diff
@@ -233,15 +233,22 @@ def create_mangled_module_name(self, module_name, kernel_id):
...
         vec_type = self.make_vector_type(dtype, False)
-        if buffer_size > 1:
-            depth_attribute = " __attribute__((depth({})))".format(buffer_size)
+        minimum_depth = Config.get("compiler", "fpga", "minimum_fifo_depth")
+        buffer_size = evaluate(buffer_size, sdfg.constants)
+        if minimum_depth:
+            minimum_depth = int(minimum_depth)
+            if minimum_depth > buffer_size:
+                buffer_size = minimum_depth
+        if buffer_size != 1:
```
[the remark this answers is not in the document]
definelicht: The feature I added here does not change the default behavior we had before, which is not setting the attribute at all. The DaCe config option defaults to `None`, so it will only change anything if explicitly set. Is this okay?

dace/codegen/targets/intel_fpga.py
```diff
@@ -1691,7 +1700,9 @@ def visit_Name(self, node):
              or defined_type == DefinedType.StreamArray) and memlet.dynamic):
             # Input memlet, we read from channel
             # we should not need mangle here, since we are in a tasklet
-            updated = ast.Name(id="read_channel_intel({})".format(node.id))
+            updated = ast.Call(func=ast.Name(id="read_channel_intel"),
```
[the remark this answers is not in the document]
definelicht: I had code that looked like this:

```python
if some_condition:
  some_tmp_var = read_channel_intel(connector_name)
  do_something_with_tmp_var(some_tmp_var)
```

This did not compile, because type inference does not understand that `read_channel_intel` returns the type of the channel object. With the update, type inference recognizes this case, and returns the type of the channel!

dace/codegen/targets/intel_fpga.py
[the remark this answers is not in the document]
TizianoDeMatteis: Yep

samples/fpga/gemm_systolic_vectorized.py
definelicht: I've added a comment giving an overview of all symbols used in b2922290

samples/fpga/gemm_systolic_vectorized.py
definelicht: Added to both the CLI help and to the overview comment at the top of the file

samples/fpga/gemm_systolic_vectorized.py
definelicht: Added to both the CLI help and to the overview comment at the top of the file

tests/fpga/map_unroll_processing_elements_test.py
definelicht: Fixed in 32efb3b6 to use same verification as other matmuls

PR #848, opened 2021-09-28
Add schedules to the labels of map nodes

phschaad:
Without this, SDFV needs to re-calculate the node label on each draw call


tbennun (changes requested):
Please also update the submodule so that we never have a version where the map has a label that is too long

phschaad:
Closed as discussed, not necessary

PR #850, opened 2021-09-29
Data Width Conversion Upgrade

definelicht:
Also support converting between vectors of different sizes with a vector length that is a multiple length of each other, e.g., `vector(dace.float32, 16) <-> vectr(dace.float32, 4)`.

### DIFF
--- a/dace/libraries/standard/nodes/gearbox.py
+++ b/dace/libraries/standard/nodes/gearbox.py
@@ -16,73 +16,240 @@ class ExpandGearbox(dace.transformation.ExpandTransformation):
         (in_edge, in_desc, out_edge, out_desc, is_pack,
          gear_factor) = node.validate(parent_sdfg, parent_state)
 
-        if is_pack:
-            vtype = out_desc.dtype
-        else:
-            vtype = in_desc.dtype
+        is_elementwise = in_desc.dtype.base_type == out_desc.dtype.base_type
 
-        sdfg = dace.SDFG("gearbox")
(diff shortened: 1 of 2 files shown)


dace/libraries/standard/nodes/gearbox.py
```diff
@@ -16,73 +16,240 @@ def expansion(node: "Gearbox", parent_state: dace.SDFGState,
...
+                dst_conn=buffer_name_inner,
+                memlet=dace.Memlet(f"{buffer_name}[0:{large_veclen}]"))
+            state.add_memlet_path(
+                nested_sdfg_node,
+                exit,
+                buffer_write,
+                src_conn=buffer_name_inner,
+                memlet=dace.Memlet(f"{buffer_name}[0:{large_veclen}]"))
+
+        else:  # Not elementwise, one side is a vector of vectors
```
TizianoDeMatteis, line 209: Do we have cases where this actually happens?

TizianoDeMatteis (approved):
Looks good to me.
I've just left a comment (more a curiosity than an actual comment)

dace/libraries/standard/nodes/gearbox.py
definelicht: It's two different styles of programming it -- honestly the vector of vectors is a bit more representative of what we do (since it enforces that it's a multiple). The problem is that it doesn't work with Intel, which is why I needed to implement the other version also...

PR #853, opened 2021-09-30
Fix torch tensor support

orausch:
Past changes broke the ability to pass torch tensors into SDFGs, this PR introduces a fix.

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -42,6 +42,18 @@ def create_datadescriptor(obj):
         return Array(dtype=dtype,
                      strides=tuple(s // obj.itemsize for s in obj.strides),
                      shape=obj.shape)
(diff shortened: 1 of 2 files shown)


dace/data.py
```diff
@@ -42,6 +42,18 @@ def create_datadescriptor(obj):
         return Array(dtype=dtype,
                      strides=tuple(s // obj.itemsize for s in obj.strides),
                      shape=obj.shape)
+    # special case for torch tensors. Maybe t __array__ could be used here for a more
```
tbennun, line 45:
```suggestion
    # special case for torch tensors. Maybe __array__ could be used here for a more
```

tbennun (changes requested):

tbennun (approved):

PR #854, opened 2021-10-01
Self-copy fixes

tbennun:
In transformations and memlet creation

### DIFF
--- a/dace/memlet.py
+++ b/dace/memlet.py
@@ -362,11 +362,21 @@ class Memlet(object):
             # Cannot initialize yet
             return
 
-        is_data_src = True
+        is_data_src = False
+        is_data_dst = False
+        if isinstance(path[0].src, AccessNode):
+            if path[0].src.data == self._data:
+                is_data_src = True
         if isinstance(path[-1].dst, AccessNode):
             if path[-1].dst.data == self._data:
-                is_data_src = False
-        self._is_data_src = is_data_src
+                is_data_dst = True
+        if is_data_src and is_data_dst:
+            # In case both point to the same array,
+            # prefer existing setting or is_data_src=True
+            if self._is_data_src is None:
+                self._is_data_src = True
+        else:
+            self._is_data_src = is_data_src
 
         # If subset is None, fill in with entire array
         if (self.data is not None and self.subset is None):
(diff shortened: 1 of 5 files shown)


dace/memlet.py
```diff
@@ -362,11 +362,21 @@ def try_initialize(self, sdfg: 'dace.sdfg.SDFG',
...
+        is_data_dst = False
+        if isinstance(path[0].src, AccessNode):
+            if path[0].src.data == self._data:
+                is_data_src = True
         if isinstance(path[-1].dst, AccessNode):
             if path[-1].dst.data == self._data:
-                is_data_src = False
-        self._is_data_src = is_data_src
+                is_data_dst = True
+        if is_data_src and is_data_dst:
```
alexnick83, line 373: Is this potentially also causing us issues, i.e., do we need to know with great confidence the data source?

dace/memlet.py
```diff
@@ -547,7 +557,10 @@ def _label(self, shape):
             result += ' (CR: %s)' % wcrstr
 
         if self.other_subset is not None:
-            result += ' -> [%s]' % str(self.other_subset)
+            if self._is_data_src is False:
+                result += ' <- [%s]' % str(self.other_subset)
```
alexnick83, line 561: Quite smart.

dace/transformation/dataflow/redundant_array.py
```diff
@@ -1326,21 +1326,22 @@ def gnode(nname):
                 # A -- (0, a:b)/(c:c+b) --> V -- (c+d)/None --> others
                 # A -- (0, a+d)/None --> others
                 e3.data.data = in_array.data
+                e3.data._is_data_src = True
```
alexnick83, line 1329: I see now; so it is indeed an issue with whether we can really tell which is the data source.

alexnick83: Btw, this formula might be also needed in the other transformations too. I would put a note for now.

alexnick83 (approved):
LGTM. However, this PR does not address the general issue where self-copies (in frontend code) should produce actual data copies with a transient (and not views) in the SDFG. Right?

dace/transformation/dataflow/redundant_array.py
tbennun: when we add the edge it should be set automatically. in self-copies it was ambiguous though and it reset it. This line may not be necessary

PR #872, opened 2021-10-28
[FPGA] MemoryBuffering Transformation

Walon1998:
Extends the StreamingMemory transformation to support memory buffering. 
The MemoryBuffering transformation allows reading/writing data from memory
using a wider data format (e.g. 512 bits), and then convert it on the fly to the right data type used by the computation


dace/transformation/dataflow/streaming_memory.py
```diff
@@ -6,10 +6,12 @@
 import networkx as nx
 import warnings
 
+import dace
```
TizianoDeMatteis, line 9: It is better to import just what you need

dace/transformation/dataflow/streaming_memory.py
```diff
@@ -200,6 +216,86 @@ def can_be_applied(graph: SDFGState,
...
+
+            # Last stride has to be one
+            if strides[-1] != 1:
+                return False
+
+            vector_size = int(self.memory_buffering_target_bytes /
+                              desc.dtype.bytes)
+            strides.pop()  # Remove last element since we already checked it
+
+            # Other strides has to be divisiable by vector size
```
TizianoDeMatteis, line 243: typo

dace/transformation/dataflow/streaming_memory.py
```diff
@@ -263,13 +359,100 @@ def apply(self, sdfg: SDFG) -> nodes.AccessNode:
...
+
+                arrname = str(self.access(sdfg))
+
+                # Add gearbox
+                total_size = edge.data.volume
+                vector_size = int(self.memory_buffering_target_bytes /
+                                  desc.dtype.bytes)
+
+                if isinstance(sdfg.arrays[dnode.data].shape[-1], dace.symbol):
+                    warnings.warn(
```
TizianoDeMatteis, line 373: Maybe we should try to resolve the symbol instead and fail if this is not the case. @definelicht what do you think?

tests/fpga/streaming_memory_test.py
```diff
@@ -237,6 +313,380 @@ def test_streaming_and_composition():
     return sdfg
 
 
+@xilinx_test()
```
TizianoDeMatteis, line 316: What happens if you test this (and the following) also with intel_fpga?

Walon1998: Timed out: [vec_add, mat_add, tensor_add, multistream, multistream_with_deps, map_order]
Compiler failure: [mapnests]
Pass: [vec_add_1, vec_add_1_symbolic, mat_add_symbol]

Walon1998:
Compiler error if the following vector types are used: float16, int8, int64

Walon1998:
I think that some of the code I added could be replaced by the vectorization transformation once it is complete.

PR #875, opened 2021-10-31
InlineSDFG fixes

alexnick83:
Assortment of fixes to InlineSDFG that specifically target issues with nested edges between access nodes corresponding to connectors of the nested SDFG.


tbennun (approved):
Looks ok. I got scared by the line at the test that removes InlineSDFG from strict transformations but it makes sense because it is only used locally.

alexnick83:
> Looks ok. I got scared by the line at the test that removes InlineSDFG from strict transformations but it makes sense because it is only used locally.

The issues manifest when RedundantSomething transformations are applied before InlineSDFG. Therefore, I think I will keep the test as-is so that we notice if there is any regression in the future.

PR #882, opened 2021-11-11
Support transient fields with strict-mode disabled, better nested error handling

tbennun:


alexnick83 (commented):
It is a bit unclear to me what this PR does. Is it removing transient variables from closures? Also, there is some exception handling code, which I assume tries to return the original exception when there multiple ones.

tbennun:
@alexnick83 precisely. Some transient variables were considered to be part of the closure, and I added the exception handling improvement this morning, which also has to do with nested functions.

alexnick83 (approved):

PR #887, opened 2021-11-28
Fix types in code generator and map range

tbennun:
- [x] Add `const` to nested SDFG type if array->scalar and read only
- [x] Fix scalar types becoming int32 if they are both arguments and variables in map range


tests/numpy/assignment_test.py
```diff
@@ -111,3 +133,4 @@ def test_assign_squeezed(A: dace.float32[3, 5, 10, 20, 13],
     test_broadcast4()
     test_broadcast5()
     test_assign_wild()
+    test_annotated_assign_type()
```
alexnick83, line 136: missing newline

alexnick83 (approved):
LGTM, just add the missing newline (force-merge though)

PR #889, opened 2021-11-29
Fix sequential reduction expansion

tbennun:
* Work with multiple dimensions
 * Work with arbitrary WCRs


orausch (approved):
Seems to fix my issues :+1:

PR #892, opened 2021-12-01
[FPGA] MemoryBuffering Transformation

Walon1998:
Extends the StreamingMemory transformation to support memory buffering.
The MemoryBuffering transformation allows reading/writing data from memory
using a wider data format (e.g. 512 bits), and then convert it on the fly to the right data type used by the computation.

Note: vector types using float16 give compiler error

- [x] Implement MemoryBuffering Transformation


dace/transformation/dataflow/streaming_memory.py
```diff
@@ -170,6 +192,95 @@ def can_be_applied(self, graph: SDFGState,
...
+            strides = list(desc.strides)
+
+            # Last stride has to be one
+            if strides[-1] != 1:
+                return False
+
+            vector_size = int(self.memory_buffering_target_bytes / desc.dtype.bytes)
+            strides.pop()  # Remove last element since we already checked it
+
+            # Other strides have to be divisiable by vector size
```
TizianoDeMatteis, line 218: typo (divisible)

TizianoDeMatteis (changes requested):
Some minor requests.

You mentioned that this is not working for certain data types: what is exactly the problem?

dace/transformation/dataflow/streaming_memory.py
```diff
@@ -246,14 +436,48 @@ def apply(self, state: SDFGState, sdfg: SDFG) -> nodes.AccessNode:
...
+            new_strides: List = list(sdfg.arrays[arrname].strides)
+
+            for i in range(len(new_strides)):
+                if i == len(new_strides) - 1:  # Skip last dimension since it is always 1
+                    continue
+                new_strides[i] = new_strides[i] / vector_size
+            sdfg.arrays[arrname].strides = new_strides
+
+            # Change subset int the post state
+            for e in sdfg.states()[-1].edges():
```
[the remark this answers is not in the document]
Walon1998: This is necessary because otherwise a memlet subset out-of-bounds exception will occur since the data container is now vectorized but the subset is still the old one.

Walon1998:
> Some minor requests.
> 
> You mentioned that this is not working for certain data types: what is exactly the problem?

I think some things have been fixed, now it is just the "float16" data type that is failing. There is a test in "streaming_memory_test.py" that you can comment out to see the problem. I'm not 100% sure, but I think something is going wrong when creating the vector type. There is some template programming going on that seems to cause the issue.

tests/fpga/streaming_memory_test.py
```diff
@@ -235,3 +979,34 @@ def test_streaming_and_composition():
     test_streaming_composition(None)
     test_streaming_composition_mapnests(None)
     test_streaming_and_composition(None)
+
```
TizianoDeMatteis, line 982: For the sake of CI testing times, can you reduce the number of tests?

You can leave them here, and decorate the test with `@pytest.mark.skip(reason="<put the reason here> ")`.
For example you can skip tests that use the same base type but different bitwidth (e.g., keep `int32` but skip `int8`/`int_16` ...)

TizianoDeMatteis (changes requested):
This looks good now! Thanks

I have just one minor comment about testing extensions. It is good that you covered all the cases, but for the sake of CI-testing time we may want to reduce the number of tests

TizianoDeMatteis (approved):

PR #903, opened 2021-12-17
Multi-State Inline SDFG as Strict Transformation

tbennun:
- [x] Turn multi-state inline into a strict transformation again, fix uncovered bugs in other transformations.
- [x] Faster computations of predecessors and validation for many-state SDFGs
- [x] Print progress of transformations if process takes over 5 seconds


dace/sdfg/graph.py
```diff
@@ -701,10 +701,10 @@ def remove_edge(self, edge: Edge[EdgeT]):
         del self._edges[t]
 
     def in_degree(self, node):
-        return len(self._nodes[node][0])
+        return self._nx.in_degree(node)
```
alexnick83, line 704: Since which nx version are these methods available? We should also update the requirements.

alexnick83 (approved):
Please check whether we need to update the minimum nx version. LGTM otherwise.

dace/sdfg/graph.py
tbennun: I checked, the methods have existed since at least 1.11, so it's fine :)
https://networkx.org/documentation/networkx-1.11/reference/generated/networkx.DiGraph.in_degree.html?highlight=in_degree#networkx.DiGraph.in_degree

PR #906, opened 2021-12-20
Optimizing DaCe parsing / codegen for large graphs and some fixes

tbennun:
Fixes:
- [x] Fix dereferencing issues in subscript expressions
- [x] Fix control flow not generating `continue;` on loop back-edge
- [x] Nicer exception messages on parsing, configurable
- [x] Support the built-in `abs` in type inference
- [x] `total_size` fix when strides and shape are present
- [x] Fix symbol type inference by running topologically on state machine

Optimizations:
- [x] Reduce the asymptotic complexity of loop detection
- [x] Do not recompute free symbols on "frozen" SDFGs during code generation
- [x] Cache more calls to `sympy`
- [x] Faster instantiation of transformations
- [x] Faster inlining
- [x] Vastly improved performance of `replace` methods


alexnick83 (approved):
LGTM

PR #915, opened 2022-01-03
Class-based interface for transformations

tbennun:
Replaces the decorator-based interface, which is hard to document and code-complete well.

The PR also updates all transformations to use the PatternNode syntax, and updates the `can_be_applied` and `apply` signatures.


dace/transformation/transformation.py
```diff
@@ -1,67 +1,55 @@
...
+    * ``SingleStateTransformation``: Patterns are limited to a single SDFG state.
+    * ``MultiStateTransformation``: Patterns are given on a subgraph of an SDFG state machine.
+    A pattern-matching must extend at least one of those two classes.
+  * Subgraph Transformations (extending SubgraphTransformation): Transformations that can operate on arbitrary
+    subgraphs.
+  * Library node expansions (extending ExpandTransformation): An internal class used for tracking how library nodes
+    are expanded.
+
+Some transformations are included in the SDFG Dataflow Coarsening pass. In order to declare a transformation as part
+of the dataflow coarsening pass, it should also extend the ``DataflowCoarseningTransformation`` mixin.
```
alexnick83, line 17: With such names, we will soon need another increase in line width :p

tbennun: We have another PR replacing it with `simplify` :)

dace/transformation/transformation.py
```diff
@@ -351,81 +351,204 @@ def print_match(self, sdfg: SDFG) -> str:
...
+            :param sdfg: The SDFG in which the match was found (equal to ``graph``).
+            :param permissive: Whether transformation should run in permissive mode.
+            :return: True if the transformation can be applied.
+        """
+        pass
+
+
+class DataflowCoarseningTransformation:
+    """
+    Mixin that includes the given PatternTransformation in the SDFG Dataflow Coarsening pass. This transformation
```
alexnick83, line 498: Isn't a "mixin" basically an interface? This class has no methods, therefore, it is not an interface. Wouldn't it be better to just make it a boolean class attribute (in PatternTransformation)?

dace/transformation/testing.py
```diff
@@ -68,7 +68,9 @@ def _optimize_recursive(self, sdfg: SDFG, depth: int):
                 print('    ' * depth, type(match).__name__, '- ', end='', file=self.stdout)
 
                 tsdfg: SDFG = new_sdfg.sdfg_list[match.sdfg_id]
-                match.apply(tsdfg)
+                tgraph = tsdfg.node(match.state_id) if match.state_id >= 0 else tsdfg
+                match._sdfg = tsdfg
```
alexnick83, line 72: Why do we need to set the (private) attribute manually here, since it is anyway passed through the `apply` method? Alternatively, if we must do this, wouldn't it be better to make the attribute public?

dace/transformation/dataflow/map_distribution.py
```diff
@@ -172,7 +160,7 @@ def _prod(sequence):
 
             elif isinstance(desc, data.Array):
 
-                local_name, local_arr = sdfg.add_temp_transient([(desc.total_size) // sz],
+                local_name, local_arr = sdfg.add_temp_transient([sympy.floor(desc.total_size / sz)],
```
alexnick83, line 163: In the past, we had issues using sympy's floor/ceiling methods. We probably don't have tests that check cases where this might be an issue (i.e., when the numbers don't divide evenly). This is something to keep in mind for future reference.

dace/transformation/dataflow/map_distribution.py
```diff
@@ -330,22 +312,15 @@ def can_be_applied(graph: dace.SDFGState,
...
+        map_entry = self.map_entry
         map_exit = graph.exit_node(map_entry)
 
         sz = dace.symbol('commsize', dtype=dace.int32, integer=True, positive=True)
         Px = dace.symbol('Px', dtype=dace.int32, integer=True, positive=True)
         Py = dace.symbol('Py', dtype=dace.int32, integer=True, positive=True)
 
-        def _prod(sequence):
-            return reduce(lambda a, b: a * b, sequence, 1)
+        from dace.data import _prod
```
alexnick83, line 323: Maybe we should finally move this to some generic utility submodule? I don't link defining this in multiple files, but importing it from `data` makes no sense.

alexnick83 (approved):
I left some notes, but they can be discussed/addressed in subsequent PRs.

dace/transformation/testing.py
tbennun: That's because the optimizer deepcopies the SDFG.

dace/transformation/transformation.py
tbennun: Yes, but the plan is to perhaps make more such Passes that transformations can participate in. In the new PR it's renamed to be `SimplifyPass`

dace/transformation/testing.py
tbennun: Good point though.

dace/transformation/dataflow/map_distribution.py
tbennun: This started raising exceptions for some reason with Integer-class `total_size`. I also don't know why it started raising now.

dace/transformation/dataflow/map_distribution.py
tbennun: Agreed.

Issue #902, opened 2021-12-16
FPGA: wrong generated data name/type when data is accessed on an interstate edge


TizianoDeMatteis:
**Describe the bug**

There are two different flavors of this bug.

The first one occurs with the Xilinx codegen when in an SDFG, an interstate edge a variable takes the value of an input array:
- in Xilinx, the array interface name gets mangled to avoid clashes with input/output data (done by  `fpga_ptr` in `fpga.py`)
- however, state transition assignments are generated somewhere else (in control_flow.py) where it has no clue of this
- in the generated code, the array name does not match the kernel argument name

The second one occurs with Intel FPGA, in the following situation:
- an interstate edge uses a variable which is a scalar (e.g., `t<TSTEPS`), which is passed as an argument to a nested SDFG
- due to how IntelFPGA codegen work, the argument is passed as a pointer
- but `control_flow.py` does not know this since it generates the code as plain cpp

There is no easy (clean) solution to this. 
This would be the opportunity to restructure how data names are resolved (it can be backend dependent), maybe having a general resolve_name that is called from everywhere in the entire codegen (Johannes TM)

**To Reproduce**
The Nussinov  polybench kernel (NPBench version) present the first issue
Heat3D and Seidel2D (without streaming mem) for IntelFPGA the second one

definelicht:
To solve issues with array interfaces once and for all, it was proposed to do a preprocessing step of the SDFG where we actually replace the array access with a separate read and write. This way, the code generation can run completely normally, with no annoying special cases for renaming array interfaces.

PR #925, opened 2022-01-13
Xilinx preprocess interstate edges

TizianoDeMatteis:
Preprocessing pass for Xilinx: it goes over interstate edges and replaces names of the accessed arrays with their mangled ones (if any).

The nussinov polybench test trigger this behavior.
Partially solves #902

### DIFF
--- a/dace/codegen/targets/cpp.py
+++ b/dace/codegen/targets/cpp.py
@@ -516,9 +516,7 @@ def cpp_array_expr(sdfg,
 
     if with_brackets:
         if fpga.is_fpga_array(desc):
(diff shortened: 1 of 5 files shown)


tbennun (changes requested):
The PR is ok, but please add a test that would fail under the previous system

tbennun (approved):

TizianoDeMatteis:
Needs more working.
It would be good to have a transformation pass that removes ArayInterface completely.
This would require to:
- detect in/out arrays and duplicate respective access nodes
- detect wcr-ed arrays, remove wcr so that there are the respective in/out access nodes

PR #933, opened 2022-01-27
Various Python preprocessing fixes

tbennun:
-[x] Deep copy in `dace.unroll`

### DIFF
--- a/dace/dtypes.py
+++ b/dace/dtypes.py
@@ -788,7 +788,7 @@ class constant:
     """
     @staticmethod
     def __descriptor__():
-        raise ValueError('All constant arguments must be provided in order ' 'to compile the SDFG ahead-of-time.')
+        raise ValueError('All constant arguments must be provided in order to compile the SDFG ahead-of-time.')
 
 
 ####### Utility function ##############
@@ -802,6 +802,7 @@ def ptrtocupy(ptr, inner_ctype, shape):
     import cupy as cp
     umem = cp.cuda.UnownedMemory(ptr, 0, None)
     return cp.ndarray(shape=shape, dtype=inner_ctype, memptr=cp.cuda.MemoryPointer(umem, 0))
+
 
 class callback(typeclass):
     """ Looks like dace.callback([None, <some_native_type>], *types)"""
--- a/dace/frontend/python/astutils.py
+++ b/dace/frontend/python/astutils.py
@@ -356,6 +356,33 @@ def negate_expr(node):
     return ast.fix_missing_locations(newexpr)
 
 
+def copy_tree(node: ast.AST) -> ast.AST:
+    """
+    Copies an entire AST without copying the non-AST parts (e.g., constant values).
+    A form of reduced deepcopy.
+    :param node: The tree to copy.
+    :return: The copied tree.
+    """
+    class Copier(ast.NodeTransformer):
+        def visit_Num(self, node):
+            # Ignore n
+            return ast.copy_location(ast.Num(n=node.n), node)
+
+        def visit_Constant(self, node):
+            # Ignore value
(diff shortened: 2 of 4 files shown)


dace/frontend/python/replacements.py
```diff
@@ -527,11 +526,21 @@ def _simple_call(sdfg: SDFG, state: SDFGState, inpname: str, func: str, restype:
...
+    else:
+        inparr = sdfg.arrays[inpname]
+
     if restype is None:
-        restype = sdfg.arrays[inpname].dtype
-    outname, outarr = sdfg.add_temp_transient(inparr.shape, restype, inparr.storage)
-    num_elements = reduce(lambda x, y: x * y, inparr.shape)
+        restype = inparr.dtype
+    outname, outarr = sdfg.add_temp_transient_like(inparr)
+    num_elements = data._prod(inparr.shape)
```
alexnick83, line 543: :(

dace/sdfg/sdfg.py
```diff
@@ -1537,14 +1537,21 @@ def add_temp_transient(self,
                               total_size=total_size,
                               may_alias=may_alias)
 
-    def add_temp_transient_like(self, desc: dt.Array, dtype=None, debuginfo=None):
+    def add_temp_transient_like(self, desc: Union[dt.Array, dt.Scalar], dtype=None, debuginfo=None):
         """ Convenience function to add a transient array with a temporary name to the data
             descriptor store. """
         debuginfo = debuginfo or desc.debuginfo
         dtype = dtype or desc.dtype
+        if isinstance(desc, dt.Scalar):
```
alexnick83, line 1545: Useful

dace/frontend/python/replacements.py
```diff
@@ -600,6 +609,16 @@ def _log(pv: 'ProgramVisitor', sdfg: SDFG, state: SDFGState, input: str):
     return _simple_call(sdfg, state, input, 'log')
 
 
+@oprepo.replaces('math.floor')
+def _floor(pv: 'ProgramVisitor', sdfg: SDFG, state: SDFGState, input: str):
```
alexnick83, line 613: I think we need special handling for those. `_simple_call` is going to create output that has the same dtype as the input. However, I quickly checked with Python 3.8 and this is not correct for floor and ceil:
```
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from math import floor, ceil
>>> floor(0.5)
0
>>> type(floor(0.5))
<class 'int'>
>>> type(ceil(0.5))
<class 'int'>
>>> 
```

alexnick83 (approved):
Overall LGTM. Check the two comments I left (can be addressed in separate PR).

dace/frontend/python/replacements.py
tbennun: :)

dace/frontend/python/newast.py
```diff
@@ -1686,7 +1690,7 @@ def _parse_map_inputs(self, name: str, params: List[Tuple[str, str]],
                         args = ','.join([str(a) for a in expr.args])
                         if arr in self.variables:
                             arr = self.variables[arr]
-                        if arr not in self.sdfg.arrays:
+                        if not isinstance(arr, str) or arr not in self.sdfg.arrays:
```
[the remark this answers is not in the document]
tbennun: the reason is that arr is always a string _before_ this if, because its source is `expr.func.__name__`. After that it can become any value due to self.variables

dace/frontend/python/replacements.py
tbennun: Fixed and pushed

PR #934, opened 2022-01-29
Support using constants as data descriptors

tbennun:
- [x] Support using constants as data descriptors
- [x] Support `numpy.array`

### DIFF
--- a/dace/frontend/python/replacements.py
+++ b/dace/frontend/python/replacements.py
@@ -3,6 +3,7 @@ import dace
 
 import ast
 import copy
+from copy import deepcopy as dcpy
 import itertools
 import warnings
 from functools import reduce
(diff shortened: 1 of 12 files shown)


dace/frontend/python/replacements.py
```diff
@@ -102,6 +104,64 @@ def _define_streamarray(pv: 'ProgramVisitor',
...
+    else:  # From literal / constant
+        if dtype is None:
+            arr = np.array(obj, copy=copy, order=order, subok=subok, ndmin=ndmin)
+        else:
+            npdtype = dtype.as_numpy_dtype()
+            arr = np.array(obj, npdtype, copy=copy, order=order, subok=subok, ndmin=ndmin)
+        desc = data.create_datadescriptor(arr)
+
+    # Set extra properties
+    desc.transient = True
```
TizianoDeMatteis, line 144: So literal arrays are always transient?

TizianoDeMatteis (approved):
Cool, thanks!

dace/frontend/python/replacements.py
tbennun: @TizianoDeMatteis yes, since they are internal, they can always be removed/modified/relayouted/etc.

PR #935, opened 2022-01-29
Handle `else` clause in for/while loops

tbennun:


alexnick83 (approved):
This LGTM, but I can think of a way that may break the semantics of for/else. Consider a simple for-loop followed by an else-clause that doesn't use the same data as the for-loop. Therefore, there is no actual dataflow between the for-loop and the else. Now consider the case where the for-loop can be converted to a Map. In such a case, the for-loop control flow is replaced by a state with a Map that can be fused with the states of the else-clause, allowing both the Map and the else to be executed concurrently. This sounds benign unless the else-clause is using code with side effects. It is good to keep in mind.

tbennun:
@alexnick83 Thanks! I think it cannot happen as it stands, because:
1. for loops will not be able to convert to maps if there is a break statement (because of the jump to the postloop state)
2. if there is no break statement in the loop, it would mean that the else clause will always be evaluated, which is fine with us.
3. if the else clause has side effects that may affect the loop (e.g., modifying the iteration variable), it will anyway happen after the loop has finished

Otherwise, it seems to be working just as well (or badly) as how we handle if/else branches.

PR #939, opened 2022-02-03
Multiple clock signals in the RTL codegen

carljohnsen:
This pull request adds support for multiple clock signals in the RTL back-end. These multi clocked designs can be both be simulated in software, emulated and run in hardware targeting Xilinx boards. 

For software simulation, the Verilator driver now supports multiple clock signals, which are discretely simulated, based on the requested frequencies. This is done by having a "time to flip" counter for each clock signal, which determines which signal should be flipped next. The simulation as a whole, and thus also the external buses, are driven by the primary clock signal, which has index 0. For every full cycle of this signal, the transactions on the external buses are handled, both for input and output.  This change also simulates time more accurately, given that we now know how many nanoseconds we simulate between each flip of a signal.

For the Xilinx back-end, the multiple clock signals are provided to the RTL Tasklets through the Vitis shell. 

In order to use multiple clock signals, the DaCe configuration entry `compiler/xilinx/frequency` must be set accordingly. The values are in the format `clock_id:frequency` separated by an escaped bar for each clock signal, all enclosed in quotes. E.g. `"0:250\|1:500"` specifies two clock signals, which are 250 and 500 MHz respectively, with the 250 MHz being the primary clock signal.


dace/config_schema.yml
```diff
@@ -461,7 +461,10 @@ required:
                         title: Target frequency for Xilinx kernels
                         description: >
                             Target frequency when compiling kernels for Xilinx.
-                            Will not necessarily be achieved in practice.
+                            Will not necessarily be achieved in practice. To
```
TizianoDeMatteis, line 464: Can you also add that frequencies must be expressed in MHz?

dace/codegen/targets/rtl.py
```diff
@@ -744,10 +792,12 @@ def unparse_tasklet(self, sdfg: sdfg.SDFG, dfg: state.StateSubgraphView, state_i
     RTL_HEADER = """\
 module {name}
 {parameters}
-( input  ap_aclk   // convention: ap_aclk clocks the design
-, input  ap_areset // convention: ap_areset resets the design
-, input  ap_start  // convention: ap_start indicates a start from host
-, output ap_done   // convention: ap_done tells the host that the kernel has finished
+( input  ap_aclk     // convention: ap_aclk clocks the design, specifically the external ports
+, input  ap_areset   // convention: ap_areset resets the design
+, input  ap_aclk_2   // convention: ap_aclk_2 is the secondary clock, which can be used inside
```
TizianoDeMatteis, line 797: From this, it seems that we are fully supporting only two clock cycles.
If this is really the case, I think this is enough for our intended use scenarios, but it should be clear from documentation/config schema

samples/rtl/axpy.py
```diff
@@ -0,0 +1,241 @@
...
+# add symbol
+N = dace.symbol('N')
+
+# add sdfg
+sdfg = dace.SDFG('axpy')
+
+# add state
+state = sdfg.add_state('device_state')
+
+# add parametr
```
TizianoDeMatteis, line 18: typo

samples/rtl/axpy.py
```diff
@@ -0,0 +1,241 @@
+# Copyright 2019-2022 ETH Zurich and the DaCe authors. All rights reserved.
+#
+# This sample shows the AXPY BLAS routine. It is implemented through Xilinx IPs in order to utilize floating point
+# operations. It is intended for running hardware_emulation or hardware xilinx targets.
+
+import dace
+import numpy as np
+
+# add symbol
```
TizianoDeMatteis, line 9: I would have this as a `make_sdfg` function, where parameters (e.g., veclen) are passed as arguments

samples/rtl/axpy.py
```diff
@@ -0,0 +1,241 @@
...
+sdfg = dace.SDFG('axpy')
+
+# add state
+state = sdfg.add_state('device_state')
+
+# add parametr
+veclen = 1
+sdfg.add_constant('VECLEN', veclen)
+sdfg.add_constant('DATA_WIDTH', 64)
+sdfg.add_constant('WORD_WIDTH', 32)
```
TizianoDeMatteis, line 22: What are these used for?

samples/rtl/axpy_double_pump.py
```diff
@@ -71,412 +71,361 @@
...
-                buffer_size=32,
+def make_sdfg():
+    # add sdfg
+    sdfg = dace.SDFG('axpy_double_pump')
+
+    # add state
+    state = sdfg.add_state('device_state')
+
+    # add parametr
+    veclen = 2
```
TizianoDeMatteis, line 82: I would have this as an argument and explain what is its meaning here (inward/outward story)

samples/rtl/axpy_double_pump.py
```diff
@@ -71,412 +71,361 @@
...
+    # validate sdfg
+    sdfg.validate()
+
+    return sdfg
 
 ######################################################################
 
 if __name__ == '__main__':
+    # Configuration has to use multiple clocks in this sample, and target hardware.
+    dace.config.Config.set('compiler', 'xilinx', 'frequency', value='"0:300\\|1:600"')
```
TizianoDeMatteis, line 413: I would reset these changes (there is something for doing this in dace.config)

tests/rtl/hardware_test.py
```diff
@@ -340,6 +341,44 @@ def test_hardware_add42_single():
...
+    # init data structures
+    N = dace.symbol('N')
+    N.set(16)
+    a = np.float32(0)  #np.random.rand(1)[0].astype(np.float32)
+    x = np.random.rand(N.get()).astype(np.float32)
+    y = np.random.rand(N.get()).astype(np.float32)
+    result = np.zeros((N.get(), )).astype(np.float32)
+
+    # Build the SDFG
+    sdfg = axpy.make_sdfg()
```
TizianoDeMatteis, line 366: I would also test with another vect. width

TizianoDeMatteis (changes requested):
Looks good to me, but I have some minor comments

tests/rtl/hardware_test.py
```diff
@@ -340,6 +341,44 @@ def test_hardware_add42_single():
...
+    spec.loader.exec_module(axpy)
+
+    # Configuration has to use multiple clocks in this sample.
+    old_freq = dace.config.Config.get('compiler', 'xilinx', 'frequency')
+    dace.config.Config.set('compiler', 'xilinx', 'frequency', value='"0:300\\|1:600"')
+
+    # init data structures
+    N = dace.symbol('N')
+    N.set(16)
+    a = np.float32(0)  #np.random.rand(1)[0].astype(np.float32)
```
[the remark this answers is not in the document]
carljohnsen: This is a temporary fix, due to the problem of scalars being overwritten by the host as zeros after passing the arguments to the kernel. But, I'll add that as a comment, so that it makes more sense.

dace/config_schema.yml
carljohnsen: This has been added!

dace/codegen/targets/rtl.py
carljohnsen: Right! Everything else has been set up for generating more than 2, so this has also been changed to generate multiple now.

samples/rtl/axpy.py
carljohnsen: This has been fixed.

samples/rtl/axpy.py
carljohnsen: This has been changed

samples/rtl/axpy.py
carljohnsen: They were old constants used in a previous version of the samples. They aren't used so they have been removed.

samples/rtl/axpy_double_pump.py
carljohnsen: This has been implemented.

tests/rtl/hardware_test.py
carljohnsen: This has been implemented.

tests/rtl/hardware_test.py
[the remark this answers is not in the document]
carljohnsen: If it is a free symbol, wont it be passed to the kernel as an argument upon calling the kernel? In regular OpenCL terms. In that case, it would result in the same problem. It could be a constant, so be hardcoded within the generated code, which should work just fine?

samples/rtl/axpy_double_pump.py
carljohnsen: Manually, like the new version is, or is there a specific function for doing this?

TizianoDeMatteis: There is a function to be used with `with` (`set_temporary`).
If that does not work, then manually

tests/rtl/hardware_test.py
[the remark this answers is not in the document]
TizianoDeMatteis: Yes it is passed as argument. Is the xilinx runtime that is causing this? Plain axpy works fine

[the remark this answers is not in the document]
carljohnsen: I've isolated the issue to a vectorization width of 16. The host overwrites the registers with 0's before calling. With a vectorization width of 8, the scaler registers don't get overwritten, so the tests run correctly. I'll investigate some more, but for now the tests run, and no longer pass on 0's

samples/rtl/axpy_double_pump.py
carljohnsen: I can't seem to find it anywhere, but it sounds like a more elegant solution! Do you have an example? Otherwise, it's been implemented manually.

TizianoDeMatteis (approved):
LGTM, Thanks!

PR #948, opened 2022-02-09
Remove scalars from the SDFG init signature.

orausch:
Adds SDFG.init_signature that is used for code generation.


orausch:
FPGA failure doesn't look related

tbennun (approved):
Thanks! Not happy about the missing free symbols. maybe just have it optional as a keyword argument, like it was with `arglist`?

tbennun (approved):

PR #955, opened 2022-02-18
Dynamic memlet propagation

tbennun:
- [x] Memlets propagated from dynamically executed states are dynamic
- [ ] ~~Memlet propagation: reduce range based on evaluatable conditionals~~

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -25,7 +25,7 @@ from dace.frontend.python import nested_call, replacements, preprocessing
 from dace.frontend.python.memlet_parser import (DaceSyntaxError, parse_memlet, pyexpr_to_symbolic, ParseMemlet,
                                                 inner_eval_ast, MemletExpr)
 from dace.sdfg import nodes
-from dace.sdfg.propagation import propagate_memlet, propagate_subset
+from dace.sdfg.propagation import propagate_memlet, propagate_subset, propagate_states
 from dace.memlet import Memlet
 from dace.properties import LambdaProperty, CodeBlock
 from dace.sdfg import SDFG, SDFGState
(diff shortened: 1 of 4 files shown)


dace/frontend/python/newast.py
```diff
@@ -1781,9 +1789,11 @@ def _add_dependencies(self,
         # Parse internal node inputs and indirect memory accesses
         if inputs:
             for conn, v in inputs.items():
-                if v is None:  # Input already handled outside
+                inner_state, memlet, inner_indices = v
+                if memlet is None:  # Input already handled outside
                     continue
-                if isinstance(v, nodes.Tasklet):
+
+                if isinstance(memlet, nodes.Tasklet):
```
alexnick83, line 1796: Maybe `memlet` is not the best name choice here?

alexnick83: `node` or something similar?

dace/frontend/python/newast.py
```diff
@@ -1236,6 +1236,11 @@ def _views_to_data(state: SDFGState, nodes: List[dace.nodes.AccessNode]) -> List
                     if (pyname and dtypes.validate_name(pyname) and pyname not in self.sdfg.arrays):
                         self.sdfg.replace(arrname, pyname)
 
+        propagate_states(self.sdfg)
```
alexnick83, line 1239: This is probably fine but I wonder whether we are doing duplicate operations. For example, in the `_add_dependencies` method, the symbols are iteratively propagated anyway and perhaps we could set there the `memlet.dynamic` property. The cost might be negligible anyway, but it would be good to note this down for future optimizations, in case it is not.

alexnick83 (approved):
LGTM, but please rename `memlet` because it is confusing (rename and force merge since the tests pass).

PR #958, opened 2022-02-23
Issue909 fixes

TizianoDeMatteis:


dace/transformation/interstate/state_fusion.py
```diff
@@ -168,6 +169,11 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
         first_state: SDFGState = self.first_state
         second_state: SDFGState = self.second_state
 
+        from dace.codegen.targets.fpga import is_fpga_kernel # avoid circular import
+        # Do not fuse FPGA and NON-FPGA states
+        if is_fpga_kernel(sdfg, first_state) != is_fpga_kernel(sdfg, second_state):
```
tbennun, line 174: please move this check later (latest in the method), it is expensive and there are cases that should be checked before that

tbennun (changes requested):

tbennun (approved):

PR #959, opened 2022-02-24
Data instrumentation

tbennun:
- [x] Save
- [x] Restore
- [x] GPU support
- [x] SDFG API
- [x] Support for symbolic and runtime-defined sizes


dace/codegen/dispatcher.py
```diff
@@ -143,17 +143,17 @@ class TargetDispatcher(object):
...
         from dace.codegen.targets import framecode as fc
+        from dace.codegen import instrumentation
 
         self.frame: fc.DaCeCodeGenerator = framecode
         self._used_targets: Set[target.TargetCodeGenerator] = set()
         self._used_environments = set()
 
-        # type: Dict[dace.dtypes.InstrumentationType, InstrumentationProvider]
-        self.instrumentation = {}
+        self.instrumentation: Dict[Union[dtypes.InstrumentationType, dtypes.DataInstrumentationType],
```
phschaad, line 152: Hm, I'm not a huge fan of having to do these unions with two so semantically and syntactically similar types. Semantiaclly, wouldn't it make sense to also include the `Save` and `Restore` from `DataInstrumentation` as e.g. `Instrumentation.Data_Save` and `Instrumentation.Data_Restore`? What would be an argument against that? That these operations would only have an effect on AccessNodes and not e.g. Tasklets, if chosen?

dace/codegen/instrumentation/provider.py
```diff
@@ -1,21 +1,22 @@
...
 
 
 @make_registry
 class InstrumentationProvider(object):
     """ Instrumentation provider for SDFGs, states, scopes, and memlets. Emits
         code on event. """
     @staticmethod
-    def get_provider_mapping() -> Dict[InstrumentationType, Type['InstrumentationProvider']]:
+    def get_provider_mapping(
+    ) -> Dict[Union[InstrumentationType, DataInstrumentationType], Type['InstrumentationProvider']]:
```
phschaad, line 13: Dito previous comment about Instumentation vs DataInstrumentation.

dace/dtypes.py
```diff
@@ -147,6 +145,16 @@ class InstrumentationType(aenum.AutoNumberEnum):
     FPGA = ()
 
 
+@undefined_safe_enum
+@extensible_enum
+class DataInstrumentationType(aenum.AutoNumberEnum):
```
phschaad, line 150: Dito previous comments.

dace/sdfg/nodes.py
```diff
@@ -217,6 +217,10 @@ class AccessNode(Node):
     debuginfo = DebugInfoProperty()
     data = DataProperty(desc="Data (array, stream, scalar) to access")
 
+    instrument = EnumProperty(dtype=dtypes.DataInstrumentationType,
```
phschaad, line 220: Goes into the same as the previous comments on this: the `instrument` property is now of two different types depending on what node it belongs to. If (and only if) the two were to be merged, it would maybe make sense to make `instrument` a general property for all Node (or even element) types.

phschaad (changes requested):
Small comment design wise / alternative design suggestion, otherwise LGTM

dace/codegen/dispatcher.py
tbennun: The argument is that you may select the wrong kind of instrumentation on access nodes or tasklets. I prefer the comboboxes to be sanitized.

dace/codegen/instrumentation/provider.py
tbennun: same answer :)

dace/codegen/dispatcher.py
phschaad: Ok I see, that makes sense.

phschaad (approved):

PR #961, opened 2022-03-01
Tuning API

tbennun:
- [x] Better support for integer powers and square root special cases
- [x] Refactor data instrumentation and cutout API
- [x] Transformation tuning API
- [x] GPU support for arbitrary-stride device copies via CopyToMap
- [x] codegen: allow library nodes to specify which files' headers should include files
- [x] Add `dace.data.make_array_from_descriptor` for easy creation of arrays mimicking data descriptor allocation scheme
- [x] Faster and more flexible `np.rot90` replacement
- [x] Native support for cupy arrays and CUDA-aware MPI
- [x] GPU codegen support for thread specialization via nested GPU_Device maps


dace/frontend/python/parser.py
```diff
@@ -653,6 +656,8 @@ def load_sdfg(self, path: str, *args, **kwargs) -> None:
         # Update SDFG cache with the SDFG (without a compiled version)
         self._cache.add(cachekey, sdfg, None)
 
+        return sdfg, cachekey
+
     def load_precompiled_sdfg(self, path: str, *args, **kwargs) -> None:
```
alexnick83, line 661: This should also have the output type `None` removed from the signature (or updated to reflect what is actually returned.

dace/frontend/python/replacements.py
```diff
@@ -456,24 +456,50 @@ def _numpy_rot90(pv: 'ProgramVisitor', sdfg: SDFG, state: SDFGState, arr: str, k
...
     else:  # k == 3
-        res = _transpose(pv, sdfg, state, arr, axes_list)
-        state = pv._add_state(state.label + '_b')
-        return _numpy_flip(pv, sdfg, state, res, axes[1])
+        to_flip = [axes[0]]
+        transpose = True
+
+    arr_copy, narr = sdfg.add_temp_transient_like(desc)
+
+    shape_list = list(narr.shape)
```
alexnick83, line 479: Couldn't we have updated `numpy.flip` instead here? Are the semantics different?

dace/frontend/python/replacements.py
```diff
@@ -4371,3 +4406,83 @@ def _inv(pv: 'ProgramVisitor', sdfg: SDFG, state: SDFGState, inp_op: str):
     state.add_memlet_path(chlsky_node, out, src_conn="_b", memlet=Memlet.from_array(*out_arr))
 
     return out_arr[0]
+
+
+# CuPy replacements
```
alexnick83, line 4411: Since CuPy is supposed (?) to follow NumPy syntax, maybe we should make a wrapper around all current NumPy replacements and automatically support the CuPy versions instead of writing new replacements? I would leave this for another PR though.

dace/symbolic.py
```diff
@@ -855,12 +896,28 @@ def _print_Symbol(self, expr):
...
+            if float(exponent) == 0.5:
+                return f'dace::math::sqrt({base})'
+        except ValueError:
+            pass
+
+        # Special case for integer powers
         try:
             int_exp = int(exponent)
-            assert (int_exp > 0)
+            if int_exp == 0:
```
alexnick83, line 910: Shouldn't this use the new `ipow` runtime method?

dace/transformation/dataflow/prune_connectors.py
```diff
@@ -116,7 +116,7 @@ def apply(self, state: SDFGState, sdfg: SDFG):
                         break
 
 
-class PruneSymbols(pm.SingleStateTransformation, pm.SimplifyPass):
+class PruneSymbols(pm.SingleStateTransformation):
```
alexnick83, line 119: Was it found that PruneSymbols removes symbols that it shouldn't?

dace/transformation/interstate/state_elimination.py
```diff
@@ -646,3 +646,47 @@ def apply(self, _, sdfg: SDFG):
         b: SDFGState = self.state_b
         edge = sdfg.edges_between(a, b)[0]
         edge.data.condition = CodeBlock("1")
+
+
+class FalseConditionElimination(transformation.MultiStateTransformation):
+    """
+    If a state transition condition is always true, removes condition from edge.
```
alexnick83, line 653: Is the docstring correct? The transformation name sounds like it handles conditions that are never satisfied.

dace/transformation/subgraph/__init__.py
```diff
@@ -5,4 +5,7 @@
 from .reduce_expansion import ReduceExpansion
 from .expansion import MultiExpansion
 from .subgraph_fusion import SubgraphFusion
-from .stencil_tiling import StencilTiling
\ No newline at end of file
+from .stencil_tiling import StencilTiling
+from .on_the_fly_map_fusion import OnTheFlyMapFusion
+from .map_fusion import SubgraphOTFFusion
+from .composite import CompositeFusion
```
alexnick83, line 11: Missing newline

alexnick83 (changes requested):
In general, it looks good. I have minor questions and remarks. If it is not too much work, could we get as part of this PR some document somewhere explaining a bit the tuning API? There are also some terms, e.g., on-the-fly recomputation, which I am not sure what they mean in the context of sugraph fusion. It would be great if they could be explained somewhere so that it makes future developments easier.

dace/__init__.py
```diff
@@ -9,6 +9,7 @@
 from .frontend.python.ndloop import ndrange
 from .frontend.operations import *
 
+from . import data, subsets
```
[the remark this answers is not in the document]
tbennun: You're right, but previously you would have to import it explicitly, i.e., `import dace.data`

dace/codegen/compiled_sdfg.py
```diff
@@ -280,6 +280,7 @@ def initialize(self, *args, **kwargs):
     def finalize(self):
         if self._exit is not None:
             self._exit(self._libhandle)
+            self._initialized = False
```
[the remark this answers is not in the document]
tbennun: :) Yes, I'm sure this explains some of the segfaults some people were getting before. We rarely reran multiple compiled SDFGs with different parameters before the tuning API.

dace/codegen/instrumentation/data/data_report.py
```diff
@@ -5,12 +5,9 @@
 import os
 
 from dace import dtypes, SDFG
+from dace.data import ArrayLike  # Type hint
```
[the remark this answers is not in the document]
tbennun: not sure about how Python imports work, but it may work without that line in init

dace/codegen/cppunparse.py
```diff
@@ -855,24 +855,48 @@ def _BinOp(self, t):
...
-            else:
-                self.write("dace::math::pow(")
-                self.dispatch(t.left)
-                self.write(", ")
-                self.dispatch(t.right)
-                self.write(")")
+                        self.write(f", {power})")
+                    self.write(")")
+                    return
+                elif power is not None and float(power) == 0.5 or float(power) == -0.5:  # Square root
```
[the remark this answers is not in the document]
tbennun: It definitely did not in the CUDA case. This provided a very nice performance boost on certain kernels that used `** 0.5`

dace/codegen/targets/cuda.py
```diff
@@ -134,12 +137,60 @@ def _emit_sync(self, codestream: CodeIOStream):
             DACE_CUDA_CHECK({backend}DeviceSynchronize());'''.format(backend=self.backend))
 
     def preprocess(self, sdfg: SDFG) -> None:
+        # Find GPU<->GPU strided copies that cannot be represented by a single copy command
```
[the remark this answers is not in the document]
tbennun: I moved the transformation into the code generation because it really is the final step before generating the copy code. Access node -> access node patterns appear in SDFGs and should be natively supported by the representation for as long as we could keep them. The copy-to-map transformation makes the schedule explicit and may inhibit further transformations such as redundant array removal.

dace/codegen/targets/cpp.py
```diff
@@ -116,52 +117,59 @@ def memlet_copy_to_absolute_strides(dispatcher, sdfg, memlet, src_node, dst_node
     copy_shape = memlet.subset.size_exact()
     src_nodedesc = src_node.desc(sdfg)
     dst_nodedesc = dst_node.desc(sdfg)
+    src_expr, dst_expr = None, None
 
     if memlet.data == src_node.data:
-        src_expr = copy_expr(dispatcher, sdfg, src_node.data, memlet, is_write=False, packed_types=packed_types)
+        if dispatcher is not None:
```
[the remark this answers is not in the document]
tbennun: yes: https://github.com/spcl/dace/pull/961/files#diff-dd2ffda4dcc84412c0d916ed43368ae62dc89935e8dab51237e8c39be8244dacR148

in the gpu codegen we reuse this function to get the copy shape and strides without actually generating expressions. I tried to refactor it at first but ended up with more copypasta than a simple if condition.

dace/frontend/python/replacements.py
tbennun: The semantics are the same, but we couldn't have updated `flip` because then the maps would not be fused. This was added because the CUDA compiler also had a bug when we tried running the two (flip/transpose) as separate kernels. Fusing them both makes rot90 faster, and works around that bug (which even appears in `-O0` and was reported).

dace/frontend/python/replacements.py
tbennun: Agreed, added to the [Frontend Fixes](https://github.com/spcl/dace/projects/10) project as a task.

dace/transformation/dataflow/prune_connectors.py
tbennun: It is correct, but slow for large SDFGs. In the next three PRs (Passes) it will become faster since it won't be applied on a pattern-matching basis.

dace/transformation/interstate/state_elimination.py
tbennun: Fixed

dace/symbolic.py
tbennun: Fixed

dace/transformation/subgraph/__init__.py
tbennun: Fixed

tbennun:
Good idea with the documentation. The API itself wasn't fixed at the beginning and was shifting around, hence it was not well documented. Now it seems stable enough to document and update the docs later. For now I applied the minor fixes.

dace/symbolic.py
tbennun: I have to revert this. Intel FPGA with OpenCL doesn't like namespaces :(

tbennun:
@alexnick83 I added some documentation. I'll leave making a **good** sample for a future PR.

alexnick83 (approved):

PR #962, opened 2022-03-02
Grid distributions

alexnick83:
Adds support for grid-based data/computation distributions.

- Basic implementation:
  - [x] Grid distributions, implemented with MPI Cartesian distributions
  - [x] Descriptors for subset of (distributed) data, implemented with MPI Sub-arrays (see [`MPI_create_subarray`](https://www.mpich.org/static/docs/v3.1/www3/MPI_Type_create_subarray.html))
  - [x] Re-destribution of data among different grid distributions

- Clean-up, tests, etc.:
  - [x] Grid distributions
  - [x] Sub-array descriptors
  - [x] Data re-distribution


dace/distr_types.py
```diff
@@ -0,0 +1,600 @@
...
+            raise ValueError('The dimensionality of the shape and correspondence list must match')
+        return True
+
+    def to_json(self):
+        attrs = serialize.all_properties_to_json(self)
+        retdict = {"type": type(self).__name__, "attributes": attrs}
+        return retdict
+
+    @classmethod
+    def from_json(cls, json_obj, context=None):
```
tbennun, line 264: looks like this function and the above one could benefit from the half-working `__new__` fix.

dace/libraries/mpi/nodes/dummy.py
```diff
@@ -0,0 +1,46 @@
...
+            node.name,
+            inputs={},
+            outputs={'__out'},
+            code='',
+            state_fields=node.fields)
+        return tasklet
+
+
+@dace.library.node
+class Dummy(dace.sdfg.nodes.LibraryNode):
```
tbennun, line 28: what's the use of this class? Should be documented here

dace/libraries/mpi/nodes/redistribute.py
```diff
@@ -0,0 +1,115 @@
...
+
+        code = f"""
+            int myrank;
+            MPI_Comm_rank(MPI_COMM_WORLD, &myrank);
+            MPI_Request* req = new MPI_Request[__state->{node._redistr}_sends];
+            MPI_Status* status = new MPI_Status[__state->{node._redistr}_sends];
+            MPI_Status recv_status;
+            if (__state->{array_a.pgrid}_valid) {{
+                for (auto __idx = 0; __idx < __state->{node._redistr}_sends; ++__idx) {{
+                    // printf("({redistr.array_a} -> {redistr.array_b}) I am rank %d and I send to %d\\n", myrank, __state->{node._redistr}_dst_ranks[__idx]);
```
tbennun, line 51: all these comments in the generated code are a bit cluttery (also appears in other expansions).

dace/sdfg/sdfg.py
```diff
@@ -298,6 +304,19 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
 
     debuginfo = DebugInfoProperty(allow_none=True)
 
+    _pgrids = Property(dtype=dict,
```
tbennun, line 307: why does this have an underscore here

dace/sdfg/sdfg.py
```diff
@@ -298,6 +304,19 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
...
+                       from_json=_arrays_from_json)
+    _subarrays = Property(dtype=dict,
+                          desc="Sub-array descriptors for this SDFG",
+                          to_json=_arrays_to_json,
+                          from_json=_arrays_from_json)
+    _rdistrarrays = Property(dtype=dict,
+                             desc="Sub-array redistribution descriptors for this SDFG",
+                             to_json=_arrays_to_json,
+                             from_json=_arrays_from_json)
+
```
tbennun, line 319: I am not a fan of spamming properties in SDFG. We have to find somewhere else.

dace/sdfg/sdfg.py
```diff
@@ -298,6 +304,19 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
 
     debuginfo = DebugInfoProperty(allow_none=True)
 
+    _pgrids = Property(dtype=dict,
+                       desc="Process-grid descriptors for this SDFG",
+                       to_json=_arrays_to_json,
+                       from_json=_arrays_from_json)
+    _subarrays = Property(dtype=dict,
```
tbennun, line 311: those should be DictProperty objects as the types are known!

dace/sdfg/sdfg.py
```diff
@@ -469,6 +496,21 @@ def arrays(self):
         """
         return self._arrays
 
+    @property
```
tbennun, line 499: A property of a Property? bound to be messy!

dace/sdfg/sdfg.py
```diff
@@ -1702,6 +1744,136 @@ def add_datadesc(self, name: str, datadesc: dt.Data, find_new_name=False) -> str
...
+            parent_grid = self._pgrids[parent_grid]
+
+        self._pgrids[grid_name] = ProcessGrid(grid_name, is_subgrid, shape, parent_grid, color, exact_grid, root)
+
+        self.append_init_code(self._pgrids[grid_name].init_code())
+        self.append_exit_code(self._pgrids[grid_name].exit_code())
+
+        return grid_name
+
+    def temp_subarray_name(self):
```
tbennun, line 1799: all of these methods have a replacement already: `find_new_name` https://github.com/spcl/dace/blob/master/dace/data.py#L99

tbennun: I just noticed that some merge conflict led to that function appearing there twice. could you also remove the older one of them in this PR?

tests/library/mpi/process_grids_test.py
```diff
@@ -0,0 +1,211 @@
...
+    size = commworld.Get_size()
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        func = CompiledSDFG(sdfg, ReloadableDLL(".dacecache/{n}/build/lib{n}.so".format(n=sdfg.name), sdfg.name))
```
tbennun, line 51: we have APIs for this, closest one to this is `load_precompiled_sdfg`: https://github.com/spcl/dace/blob/master/dace/sdfg/utils.py#L1166

tests/library/mpi/process_grids_test.py
```diff
@@ -0,0 +1,211 @@
...
```
tbennun, line 109: same

tests/library/mpi/process_grids_test.py
```diff
@@ -0,0 +1,211 @@
...
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = pgrid_bcast.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=pgrid_bcast.name))
+        func = CompiledSDFG(sdfg, ReloadableDLL(".dacecache/{n}/build/lib{n}.so".format(n=sdfg.name), sdfg.name))
```
tbennun, line 148: same

tests/library/mpi/process_grids_test.py
```diff
@@ -0,0 +1,211 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = pgrid_bcast.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=pgrid_bcast.name))
```
tbennun, line 189: same

tests/library/mpi/redistribute_test.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = matrix_2d_2d.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=matrix_2d_2d.name))
```
tbennun, line 54: same

tests/library/mpi/redistribute_test.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = matrix_2d_2d_2.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=matrix_2d_2d_2.name))
```
tbennun, line 115: same

tests/library/mpi/redistribute_test.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = matrix_2d_2d_3.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=matrix_2d_2d_3.name))
```
tbennun, line 185: same

tests/library/mpi/redistribute_test.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = vector_2d_2d.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=vector_2d_2d.name))
```
tbennun, line 246: maybe you should make a test util that only compiles on rank 0 and loads it on the rest, i.e., replacing lines 241-248 here.

tests/library/mpi/subarrays_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = block_scatter.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=block_scatter.name))
```
tbennun, line 34: .

tests/library/mpi/subarrays_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = block_scatter_bcast.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=block_scatter_bcast.name))
```
tbennun, line 77: .

tests/library/mpi/subarrays_test.py
```diff
@@ -0,0 +1,183 @@
...
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = block_gather.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
+        sdfg = dace.SDFG.from_file(".dacecache/{n}/program.sdfg".format(n=block_gather.name))
```
tbennun, line 120: .

tests/library/mpi/subarrays_test.py
```diff
@@ -0,0 +1,183 @@
...
+    even_size = (size // 2) * 2
+
+    if size < 2:
+        raise ValueError("Please run this test with at least two processes.")
+
+    if rank == 0:
+        sdfg = block_gather_reduce.to_sdfg()
+        func = sdfg.compile()
+    commworld.Barrier()
+    if rank > 0:
```
tbennun, line 162: ..phew

tbennun (changes requested):
Mostly the PR is good. There are two general comments about it:
1. with all this generated code, I'm wondering: could we just add some of that stuff to the runtime  and write it as C++ code?
2. I don't like adding new properties to SDFG that are not relevant to every SDFG. This is more like library properties. We need to discuss how we want them in and why are they relevant to an entire SDFG. (to be clear, I am also not a fan of `openmp_sections` living in SDFG, we just need a better option as this spams the SDFV view in VSCode)

another minor comment is that the PR should use the existing APIs as much as possible.

dace/distr_types.py
alexnick83: I followed the same class structure/methods as the descriptors in `dace/data`. Where can I find an example of this `__new__` fix?

dace/libraries/mpi/nodes/redistribute.py
alexnick83: Yes, it is, but they may be needed for quick debugging purposes the following weeks, so I left them there intentionally. They will be clean-up in a subsequent PR.

dace/sdfg/sdfg.py
alexnick83: It is supposed to be similar to `SDFG::_arrays`. There is also the corresponding `SDFG::pgrids` method.

dace/sdfg/sdfg.py
alexnick83: There are properties of the (DaCe) program though.

dace/sdfg/sdfg.py
alexnick83: Same as `SDFG::_arrays`, `SDFG::arrays`. If we consider that the array descriptor store does not follow the DaCe programming conventions, then it should be fixed ASAP.

dace/sdfg/sdfg.py
alexnick83: `temp_data_name` is still there and used though

alexnick83: If by duplicate you mean `dace::SDFG::_find_new_name` and `data::find_new_name`, then both are used.

alexnick83:
> Mostly the PR is good. There are two general comments about it:
> 
> 1. with all this generated code, I'm wondering: could we just add some of that stuff to the runtime  and write it as C++ code?
> 2. I don't like adding new properties to SDFG that are not relevant to every SDFG. This is more like library properties. We need to discuss how we want them in and why are they relevant to an entire SDFG. (to be clear, I am also not a fan of `openmp_sections` living in SDFG, we just need a better option as this spams the SDFV view in VSCode)
> 
> another minor comment is that the PR should use the existing APIs as much as possible.

Answering both 1 and 2.

There are 3 points of code generation:
- `dace::distr_types`: This is for DaCe Program/SDFG distributed descriptors, where we want code to run on init/deinit. This is code that should be reused and not be tied to a particular instance of a library node. Also, it is code that we do not want to have in the "critical path" of the runtime.
- Library nodes: Mainly the Redistribute node; outputs code in the SDFGState subgraph for actual MPI communication.
- Runtime `dace::comm`: Helper methods to be called by the generated code

It is probably possible to move some of the init/deinit code to the runtime (as helper methods). However, this code is heavily customized with (string) replacements. I am not sure it would be nice to have runtime methods where we pass 5 int and 10+ const char[] parameters.

dace/distr_types.py
tbennun: in the failing PR :D

dace/sdfg/sdfg.py
tbennun: `_arrays` is a broken property too. It only has an underscore to avoid it showing up in the viewer, but that's not true anymore and it shows up anyway.

tbennun (approved):

PR #963, opened 2022-03-02
Map auto-tuning

lukastruemper:


dace/optim/cutout_tuner.py
```diff
@@ -0,0 +1,73 @@
+import dace
+import numpy as np
+
+from abc import ABC, abstractmethod
+from typing import Dict
+
+from dace.optim.auto_tuner import AutoTuner
```
tbennun, line 7: same as above

dace/optim/map_permutation_tuner.py
```diff
@@ -0,0 +1,46 @@
+import dace
```
tbennun, line 1: missing copyright

samples/optimization/map_tuning.py
```diff
@@ -0,0 +1,37 @@
+import dace
```
tbennun, line 1: missing copyright

dace/optim/cutout_tuner.py
```diff
@@ -0,0 +1,73 @@
+import dace
```
tbennun, line 1: missing copyright

dace/optim/auto_tuner.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
```
tbennun, line 1: also missing copyright, see guidelines (and maybe add yourself to AUTHORS while you're at it)

tbennun (changes requested):

dace/optim/map_tiling_tuner.py
```diff
@@ -0,0 +1,97 @@
...
+            cutout = cutout_state(state, *subgraph_nodes)
+            cutout.instrument = dace.InstrumentationType.Timer
+
+            arguments = {}
+            for cstate in cutout.nodes():
+                for dnode in cstate.data_nodes():
+                    if cutout.arrays[dnode.data].transient:
+                        continue
+
+                    arguments[dnode.data] = np.copy(dreport.get_first_version(dnode.data))
```
tbennun, line 57: why copy the data? it's not going to change, is it?

dace/optim/map_tiling_tuner.py
```diff
@@ -0,0 +1,97 @@
...
+            results[None] = baseline
+
+            best_choice = None
+            best_runtime = baseline
+            cutout_json = cutout.to_json()
+            for point in self.space(parent_map):
+                # TODO: Hacky? Faster than deepcopy?
+                tiled_sdfg = SDFG.from_json(cutout_json)
+                # TODO: Inefficient
+                for xform in Optimizer(tiled_sdfg).get_pattern_matches(patterns=[MapTiling], options={"tile_sizes": point}):
```
tbennun, line 70: you should probably use `apply_to` on a particular map entry node, as received from cutouts (`parent_map`)?

dace/optim/map_tiling_tuner.py
lukastruemper: TypeError: Passing a numpy view (e.g., sub-array or "A.T") to DaCe programs is not allowed in order to retain analyzability. Please make a copy with "numpy.copy(...)". If you know what you are doing, you can override this error in the configuration by setting compiler.allow_view_arguments to True.

tbennun: which is why I enable view arguments in the other tuning example. maybe we should enable it here too

PR #965, opened 2022-03-04
Fix for issues with Tasklet-syntax and indirection

alexnick83:
Fixes the following issues:
- [x] Multi-dimensional indirection truncates the index and keeps only the first dimension.
- [x] Indirected arrays are not properly propagated tthrough all nested SDFG levels.


tbennun (approved):
LGTM if tests pass

PR #970, opened 2022-03-15
Enable allocating transients outside of nested SDFGs or kernels

tbennun:


dace/codegen/targets/cpp.py
```diff
@@ -209,22 +210,29 @@ def memlet_copy_to_absolute_strides(dispatcher, sdfg, memlet, src_node, dst_node
...
-        from dace.codegen.targets.cuda import CUDACodeGen  # Avoid import loop
-        if not CUDACodeGen._in_device_code:  # GPU kernels cannot access state
-            if not sdfg:
-                raise ValueError("Missing SDFG value")
-            return f'__state->__{sdfg.sdfg_id}_{name}'
+    if desc.storage != dtypes.StorageType.CPU_ThreadLocal:
+        if (desc.transient and desc.lifetime is dtypes.AllocationLifetime.Persistent):
+            from dace.codegen.targets.cuda import CUDACodeGen  # Avoid import loop
+            if not CUDACodeGen._in_device_code:  # GPU kernels cannot access state
+                return f'__state->__{sdfg.sdfg_id}_{name}'
```
alexnick83, line 229: Is it possible to arrive here with `sdfg is None`? I guess it raises an exception anyway (before and after code change), so maybe not so important.

dace/codegen/targets/cuda.py
```diff
@@ -426,25 +427,26 @@ def allocate_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, d
     def allocate_stream(self, sdfg, dfg, state_id, node, nodedesc, function_stream, declaration_stream,
                         allocation_stream):
         dataname = node.data
-        allocname = cpp.ptr(dataname, nodedesc, sdfg)
+        allocname = cpp.ptr(dataname, nodedesc, sdfg, self._frame)
         if nodedesc.storage == dtypes.StorageType.GPU_Global:
             fmtargs = {
-                'name': dataname,
+                'name': allocname,  # TODO: Handle persistent streams
```
alexnick83, line 433: So is `name` obsolete?

alexnick83 (approved):

dace/codegen/targets/cuda.py
tbennun: no, we need to bring it back later because of a `__state->` opener that we don't need in the declaration stream. However, I don't know how to take care of that right now. Hence the TODO

PR #976, opened 2022-04-08
Add support for Reference data descriptors

tbennun:
Typed/shaped pointers that can be set at runtime. Necessary for supporting several challenging programs, while skirting around optimizing _some parts_ of them.

As opposed to Views, there are no special graph rules (e.g. edge node) for references, as once they are set, they can be used as normal access nodes. Aliasing rules do apply though and transformations (e.g., redundant array) can be amended at a later PR. Future PRs may eliminate trivial References via passes/transformations too.

It is not accessible from frontends at this point on purpose, as those constructs are highly discouraged.

### DIFF
--- a/dace/codegen/targets/cpu.py
+++ b/dace/codegen/targets/cpu.py
@@ -198,6 +198,22 @@ class CPUCodeGen(TargetCodeGenerator):
             self._dispatcher.declared_arrays.add(aname, DefinedType.Pointer, ctypedef)
         allocation_stream.write(f'{aname} = {value};', sdfg, state_id, node)
 
+    def allocate_reference(self, sdfg: SDFG, dfg: SDFGState, state_id: int, node: nodes.AccessNode,
+                           global_stream: CodeIOStream, declaration_stream: CodeIOStream,
+                           allocation_stream: CodeIOStream):
+        name = node.data
+        nodedesc = node.desc(sdfg)
+        ptrname = cpp.ptr(name, nodedesc, sdfg, self._frame)
+
+        # Check if reference is already declared
+        declared = self._dispatcher.declared_arrays.has(ptrname)
+
+        if not declared:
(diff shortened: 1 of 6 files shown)


alexnick83 (commented):
Looks LGTM. I have a couple questions:
- Will transformations break programs using references? Since references are arrays, I am guessing that, yes, any transformations targeting arrays and access nodes will potentially cause issues. However, we may not care at this point since, as noted, only those who really know what they are doing should use references.
- References look like a very convenient way to implement multibuffering in the SDFG. Considering that references break analyzability, how do we plan stop users from abusing this feature?

tbennun:
> Looks LGTM. I have a couple questions:
> 
> * Will transformations break programs using references? Since references are arrays, I am guessing that, yes, any transformations targeting arrays and access nodes will potentially cause issues. However, we may not care at this point since, as noted, only those who really know what they are doing should use references.

Some transformations need to be adapted if they are removing arrays. What we can do is treat references as non-transient, which will block them from being removed. Thoughts?

> * References look like a very convenient way to implement multibuffering in the SDFG. Considering that references break analyzability, how do we plan stop users from abusing this feature?

We plan to stop them with lots of warnings on the frontend side, color them red in the renderer, and reference cleanup transformations (future PR).

alexnick83:
> Some transformations need to be adapted if they are removing arrays. What we can do is treat references as non-transient, which will block them from being removed. Thoughts?

That certainly will do the trick, although I am not sure we should. Maybe we should just update the transformations (perhaps in batches).

alexnick83 (approved):

tbennun:
> > Some transformations need to be adapted if they are removing arrays. What we can do is treat references as non-transient, which will block them from being removed. Thoughts?
> 
> That certainly will do the trick, although I am not sure we should. Maybe we should just update the transformations (perhaps in batches).

I propose we refactor this anyway to two functions that we will use in all transformations:
* can_remove_access_node
* remove_access_node_and_array

Or something similar. Next PR after Passes

PR #978, opened 2022-04-21
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.


tbennun:
@thomasrockhu-codecov Thanks for the PR! We disabled notifications as they were showing up a bit too early, with incorrect results until all the builds come in. even with the right number of reports before the comment it would show up too soon. Is there a way to time the comment and/or wait until no more builds arrive?

tbennun:
@thomasrockhu-codecov Yes, you understand correctly. It ended up being both the comment and the status checks. We would get notifications that the PRs "failed" due to that early report, and so we started to ignore them. We instead found it better to passively check the coverage % once in a while using the badge.

In any case, we'll give the informational checks a try, thanks!

Off topic, IMO the cleanest solution would be to somehow make an event hook that triggers codecov comments/status only after all expected status checks are complete. I wouldn't know how to get that from the available GitHub webhooks though (I saw seemingly relevant events like "check runs" / "workflows" / "statuses" but don't know if that would provide all the information).

tbennun (approved):

PR #979, opened 2022-04-22
Pass/Pipeline API

tbennun:


dace/data.py
```diff
@@ -402,20 +402,23 @@ class Array(Data):
...
-    shape = (4, 10)
-    strides = (12, 1)
-    start_offset = 3
-    total_size = 63   (= 3 + 12 * 5)
-    offset = (0, 0, 0)
-    ```
+    
+    .. code-block:: text
+
+        [xxx][          |xx]
```
alexnick83, line 408: I guess the change here adds some hint for sphinx documentation. However, it looks to me that line 408 is one space more to the right than it should (compared to line 409). Can you verify that it is properly aligned in the documentation output?

alexnick83 (approved):

dace/data.py
tbennun: Fixed now, thanks!

PR #982, opened 2022-04-26
Support callbacks with scalar return values

tbennun:


dace/codegen/targets/cpp.py
```diff
@@ -1188,10 +1188,10 @@ def visit_Name(self, node: ast.Name):
         except KeyError:
             defined_type = None
         if (self.allow_casts and isinstance(dtype, dtypes.pointer) and memlet.subset.num_elements() == 1):
-            return ast.Name(id="{}[0]".format(name), ctx=node.ctx)
+            return ast.parse(f"{name}[0]").body[0].value
```
alexnick83, line 1191: Just testing vscode PR functionality.

dace/frontend/python/newast.py
```diff
@@ -3973,11 +3974,19 @@ def parse_target(t: Union[ast.Name, ast.Subscript]):
         self._add_state('callback_%d' % node.lineno)
         self.last_state.set_default_lineinfo(self.current_lineinfo)
 
-        call_args = ', '.join(str(s) for s in allargs)
-        tasklet = self.last_state.add_tasklet(f'callback_{node.lineno}', {f'__in_{name}'
-                                                                          for name in args} | {'__istate'},
-                                              {f'__out_{name}'
-                                               for name in outargs} | {'__ostate'}, f'{funcname}({call_args})')
+        if callback_type.is_scalar_function() and len(callback_type.return_types) > 0:
```
alexnick83, line 3977: We should also consider cases where a callback has arguments and a return value. Perhaps in another PR.

alexnick83 (approved):
LGTM

dace/frontend/python/newast.py
tbennun: @alexnick83 testing tagging

AFAIU, `is_scalar_function` doesn't set limitation on the number of input arguments, just 0 or 1 scalar return values

Issue #980, opened 2022-04-22
Wrong result when running Program on FPGA Hardware


ImaGoodFella:
**Describe the bug** 
I have been trying to run my code on an Xilinx FPGA. When I run my code on the CPU or simulate it for the FPGA, the correct result is computed.  However, when I run my code on the FPGA hardware itself, the wrong result is computed. 

Here is the code, which produces an assertion error whenever I run it on the FPGA hardware. 
```python
import dace
import numpy as np

N = dace.symbol('N', dtype=dace.uint32)

daceType=dace.float64

@dace.program
def ComputeWAXPBY(a: daceType[1], x: daceType[N], b: daceType[1], y: daceType[N], res: daceType[N]):
    res[:] = a[0] * x + b[0] * y

ntype = np.float64

from numpy.random import default_rng
rng = default_rng(42)

N = 10000
x = rng.random((N,), ntype)
y = rng.random((N,), ntype)
a = rng.random((1,), ntype)
b = rng.random((1,), ntype)

res = np.zeros(N)

sdfg = ComputeWAXPBY.to_sdfg()
from dace.transformation.interstate import FPGATransformSDFG
sdfg.apply_transformations(FPGATransformSDFG)

sdfg(x=x, y=y, a=a, b=b, res=res, N=N)

print(abs(res - (a[0] * x + b[0] * y)).max())
assert (abs(res - (a[0] * x + b[0] * y)).max() < 10e-12)

```

**To Reproduce**
run `DACE_compiler_xilinx_mode=hardware python ComputeWAXPBY.py`

**Additional Context**
I am running this on a Xilinx FPGA with Vitis 2021.2. 

Thanks for the help.

PR #983, opened 2022-04-30
Xilinx decoupled array interfaces

TizianoDeMatteis:
In Xilinx, when the same container is accessed for reading and writing, the corresponding memory interface is decoupled (i.e., two pointers are code-generated, one for reading, one for writing).

While this can be sometimes needed to achieve II=1, in the general case it may hide dependencies to the compiler, resulting in designs that do not work in hardware (while they work in simulation)

For this reason, we disable array interface decoupling by default. It can be enabled by the user through dace.config

Some of the tests are affected by this: for example `vec_sum`, `matmul` and others can no longer achieve II=1 if decoupling is disabled.

For completeness, we check both configurations for a representative set of tests

Solves #980


dace/codegen/targets/xilinx.py
```diff
@@ -535,15 +594,27 @@ def generate_host_function_body(self, sdfg: dace.SDFG, state: dace.SDFGState, ke
         kernel_args = []
```
tbennun, line 594: add some type hint please

tbennun (changes requested):
LGTM apart from one minor comment

dace/codegen/targets/xilinx.py
```diff
@@ -591,7 +591,9 @@ def generate_host_function_body(self, sdfg: dace.SDFG, state: dace.SDFGState, ke
         :param instrumentation_stream: Code for profiling kernel execution time.
         '''
 
-        kernel_args = []
+        # Keep track of kernel arguments as (arg, interface_id) pair
+        kernel_args: list[Tuple[str, int]] = []
```
tbennun, line 595: you need to use typing.List. Before Python 3.9 this wasn't allowed

tbennun (changes requested):

tbennun (approved):
LGTM

PR #986, opened 2022-05-03
Support for numpy.arange

alexnick83:
Adds support for numpy.arange. The 'like` keyword is currently not supported.

### DIFF
--- a/dace/frontend/python/replacements.py
+++ b/dace/frontend/python/replacements.py
@@ -502,6 +502,58 @@ def _numpy_rot90(pv: 'ProgramVisitor', sdfg: SDFG, state: SDFGState, arr: str, k
     return arr_copy
 
 
+@oprepo.replaces('numpy.arange')
+@oprepo.replaces('dace.arange')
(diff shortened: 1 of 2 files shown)


tbennun:
@lukastruemper are you reviewing this?

lukastruemper (approved):
Very nice, thank you! I've never seen this "like" arg in practice, think we can merge for now

tbennun:
@alexnick83 Why wouldn't `add_transient_like` work in this case?

tbennun:
> @alexnick83 Why wouldn't `add_transient_like` work in this case?

@alexnick83 just read the documentation and I guess the `like` part could be used for custom descriptors as pre-padding and custom strides.

alexnick83:
> @alexnick83 Why wouldn't `add_transient_like` work in this case?

@tbennun The `like` keyword doesn't seem to do what it does in methods such as `numpy.empty_like` (where it isn't a keyword), at least in some testing I did. The documentation is also lacking. I think a more important issue is whether we should support symbolic bounds/steps, which can be quite tricky. Theoretically, the way I wrote the replacement should support symbolic values, but we should write our own rules about how this is supposed to work and appropriate tests.

PR #987, opened 2022-05-05
Container compatibility for config files

tbennun:
Respect reversed priority order when trying to save a file: home directory first, then cwd, then operate with in-memory config using the default settings.


TizianoDeMatteis (approved):
LGTM

PR #991, opened 2022-05-10
Various fixes

tbennun:


dace/sdfg/sdfg.py
```diff
@@ -298,7 +297,7 @@ class SDFG(OrderedDiGraph[SDFGState, InterstateEdge]):
     init_code = DictProperty(str, CodeBlock, desc="Code generated in the `__dace_init` function.")
     exit_code = DictProperty(str, CodeBlock, desc="Code generated in the `__dace_exit` function.")
 
-    orig_sdfg = SDFGReferenceProperty(allow_none=True)
+    orig_sdfg = OptionalSDFGReferenceProperty(allow_none=True)
```
gronerl, line 300: Is the `allow_none` still meaningful here? Can it ever be False given that the fallback behavior is to put None when deserializing.

tbennun: Optional doesn't necessarily mean that you cannot set it to None. It's optional in deserialization (but then necessarily becomes None). Ambiguous, yes.

PR #993, opened 2022-05-10
Transformations as SDFG Passes

tbennun:
- [x] Make Transformations into Pass subclasses
- [x] Refactor apply/apply_repeated to Pass pipelines
- [x] Modernize Transformation API to allow user-defined constructors (paving the way towards easy-to-use transformation dataclasses)

### DIFF
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -1082,7 +1082,8 @@ def fuse_states(sdfg: SDFG, permissive: bool = False, progress: bool = None) -> 
                 if u in skip_nodes or v in skip_nodes:
                     continue
                 candidate = {StateFusion.first_state: sd.node_id(u), StateFusion.second_state: sd.node_id(v)}
-                sf = StateFusion(sd, id, -1, candidate, 0, override=True)
+                sf = StateFusion()
+                sf.setup_match(sd, id, -1, candidate, 0, override=True)
                 if sf.can_be_applied(sd, 0, sd, permissive=permissive):
                     sf.apply(sd, sd)
                     applied += 1
@@ -1147,7 +1148,8 @@ def inline_sdfgs(sdfg: SDFG, permissive: bool = False, progress: bool = None, mu
             candidate = {
                 InlineMultistateSDFG.nested_sdfg: node_id,
             }
-            inliner = InlineMultistateSDFG(sd, id, state_id, candidate, 0, override=True)
+            inliner = InlineMultistateSDFG()
+            inliner.setup_match(sd, id, state_id, candidate, 0, override=True)
             if inliner.can_be_applied(state, 0, sd, permissive=permissive):
                 inliner.apply(state, sd)
                 counter += 1
@@ -1158,7 +1160,8 @@ def inline_sdfgs(sdfg: SDFG, permissive: bool = False, progress: bool = None, mu
         candidate = {
             InlineSDFG.nested_sdfg: node_id,
         }
-        inliner = InlineSDFG(sd, id, state_id, candidate, 0, override=True)
+        inliner = InlineSDFG()
+        inliner.setup_match(sd, id, state_id, candidate, 0, override=True)
         if inliner.can_be_applied(state, 0, sd, permissive=permissive):
             inliner.apply(state, sd)
             counter += 1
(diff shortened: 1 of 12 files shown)


dace/optimization/on_the_fly_map_fusion_tuner.py
```diff
@@ -93,7 +93,8 @@ def evaluate(self, config, cutout, measurements: int, **kwargs) -> float:
         maps_ = list(map(candidate.start_state.node, map_ids))
         subgraph = helpers.subgraph_from_maps(sdfg=candidate, graph=candidate.start_state, map_entries=maps_)
 
-        map_fusion = sg.SubgraphOTFFusion(subgraph, candidate.sdfg_id, candidate.node_id(candidate.start_state))
+        map_fusion = sg.SubgraphOTFFusion()
+        map_fusion.setup_match(subgraph, candidate.sdfg_id, candidate.node_id(candidate.start_state))
```
alexnick83, line 97: Are these warnings from codecov correct? If you cannot see them, it basically says "Added lines #L96 - L97 were not covered by tests." Since there seem quite a few of those in this file, should we ask @lukastruemper to write some suitable tests?

dace/sdfg/utils.py
```diff
@@ -1082,7 +1082,8 @@ def fuse_states(sdfg: SDFG, permissive: bool = False, progress: bool = None) ->
                 if u in skip_nodes or v in skip_nodes:
                     continue
                 candidate = {StateFusion.first_state: sd.node_id(u), StateFusion.second_state: sd.node_id(v)}
-                sf = StateFusion(sd, id, -1, candidate, 0, override=True)
+                sf = StateFusion()
+                sf.setup_match(sd, id, -1, candidate, 0, override=True)
```
alexnick83, line 1086: Is it still possible to initialize a transformation object with the Transformation's init method or use of `setup_match` is obligatory?

dace/transformation/dataflow/redundant_array.py
```diff
@@ -275,7 +282,10 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
                 for a in accesses:
                     subsets_intersect = False
                     for e in graph.out_edges(a):
-                        subset, _ = _validate_subsets(e, sdfg.arrays, src_name=a.data)
+                        try:
+                            subset, _ = _validate_subsets(e, sdfg.arrays, src_name=a.data)
+                        except (NotImplementedError, ValueError):
```
alexnick83, line 287: I am not sure if we actually want to raise an exception for any of those errors, i.e., there may be an actual bug in the code, rather than just being unable to infer the subsets. It might be worth to at least print something in debug mode, so someone can investigate if there are too many messages printed.

dace/transformation/interstate/state_fusion.py
```diff
@@ -443,11 +443,13 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
                                         return False
                                 found = outnode
 
-        from dace.codegen.targets.fpga import is_fpga_kernel  # avoid circular import
+        from dace.codegen.targets.fpga import \
```
alexnick83, line 446: Why did yapf do this??

alexnick83 (approved):
A few questions, but nothing blocking.

dace/sdfg/utils.py
tbennun: Yes, the use of `setup_match()` is only necessary before calling `apply`

dace/transformation/interstate/state_fusion.py
tbennun: :(

dace/optimization/on_the_fly_map_fusion_tuner.py
lukastruemper: Yes, will do! Re-writing parts of the tuner code currently.

I am actually thinking about even re-writing the fusion transformations. To me, it seems like we have multiple fusion transformations which are not really disjoint.

dace/transformation/dataflow/redundant_array.py
tbennun: Fixed now @alexnick83

PR #1000, opened 2022-05-23
Support callbacks with keyword arguments

tbennun:
- [x] Escape string arguments such that they generate valid code
- [x] Support kwargs, variable kwargs, and duplicate calls to same function

### DIFF
--- a/dace/frontend/python/astutils.py
+++ b/dace/frontend/python/astutils.py
@@ -663,3 +663,11 @@ def create_constant(value: Any, node: Optional[ast.AST] = None) -> ast.AST:
         newnode = ast.copy_location(newnode, node)
 
     return newnode
+
+
+def escape_string(value: str):
+    """ Converts special Python characters in strings back to their parsable version (e.g., newline to ``\n``) """
+    if sys.version_info >= (3, 0):
+        return value.encode("unicode_escape").decode("utf-8")
+    # Python 2.x
+    return value.encode('string_escape')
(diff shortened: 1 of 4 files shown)


dace/frontend/python/astutils.py
```diff
@@ -663,3 +663,11 @@ def create_constant(value: Any, node: Optional[ast.AST] = None) -> ast.AST:
...
 
     return newnode
+
+
+def escape_string(value: str):
+    """ Converts special Python characters in strings back to their parsable version (e.g., newline to ``\n``) """
+    if sys.version_info >= (3, 0):
+        return value.encode("unicode_escape").decode("utf-8")
+    # Python 2.x
+    return value.encode('string_escape')
```
alexnick83, line 673: Isn't this a bit pointless? I think that DaCe is completely incompatible with Python 2.

dace/frontend/python/preprocessing.py
```diff
@@ -300,6 +300,33 @@ def has_replacement(callobj: Callable, parent_object: Optional[Any] = None, node
...
+    Used for creating callbacks from C to Python with keyword arguments.
+    """
+    if not node.keywords:
+        return func
+    keywords = [kw.arg for kw in node.keywords]
+    poscount = len(node.args)
+
+    # Using two levels of functions to ensure keywords are stored with the callback
+    if poscount == 0:
+        def make_cb(keywords, _):
```
alexnick83, line 315: Why do we need two different definitions? Some issue with `(*)args`?

alexnick83 (approved):
Just some questions, nothing blocking

dace/frontend/python/astutils.py
tbennun: True

PR #1003, opened 2022-05-24
SubgraphFusion: Removing intermediate nodes when possible.

alexnick83:
This PR attempts to address the issue of non-local intermediate access nodes appearing inside (non-nested) Map bodies after applying SubgraphFusion. Consider the following problematic issue:
Screenshot from 2022-05-24 11-46-20

The intermediate access node is problematic because Array `p` is also used outside of the Map's scope. For example, when trying to nest the Map's body in another transformations, the SDFG breaks as seen below:
Screenshot from 2022-05-24 11-48-26

There are now multiple `p` access nodes; some of them are not even properly connected.

This PR addresses the issue by removing these nodes and reconnecting the adjacent edges through the MapEntry and MapExit to outer access nodes:
image


tbennun (approved):
LGTM

PR #1006, opened 2022-05-27
pr for propagation

acalotoiu:
In propagation, in annotate loop ranges, a check was added to stop the process if no increment_edge is found.


acalotoiu:
This is no longer necessary as the feature has already been fixed in master

Issue #1005, opened 2022-05-26
StateAssignElimination and StateFusion incorrect constant propagation with symbols


leporia:
**Describe the bug**
`StateAssignElimination` and `StateFusion` together do some constant propagation on symbols that result in incorrect (out-of-bound) values. 

**To Reproduce**
```
import dace

N = dace.symbol('N')
index = dace.symbol('index')
start_ptr = dace.symbol('start_ptr')

sdfg = dace.SDFG("test")
sdfg.add_array('A', shape=[N], dtype=dace.float32, transient=True)
sdfg.add_array('out_val', shape=[1], dtype=dace.int32, transient=True)

state_init1 = sdfg.add_state()
state_init2 = sdfg.add_state()
state_init3 = sdfg.add_state()
state_init4 = sdfg.add_state()
state_compute = sdfg.add_state()

sdfg.add_edge(state_init1, state_init2, \
        dace.InterstateEdge(assignments={"start_prt": "0", "N": \
        "2"}))
sdfg.add_edge(state_init2, state_init3, \
        dace.InterstateEdge(assignments={"start_ptr": "start_ptr+1"}))
sdfg.add_edge(state_init3, state_init4, \
        dace.InterstateEdge(assignments={"index": "1+start_ptr"}))
sdfg.add_edge(state_init4, state_compute, \
        dace.InterstateEdge(assignments={"start_ptr": "start_ptr+1"}))

tasklet1 = state_compute.add_tasklet(name="init1", inputs=[], outputs=["out"], \
        code="out = 1;", language=dace.Language.CPP)
tasklet2 = state_compute.add_tasklet(name="init2", inputs=["in"], \
        outputs=["out"], code="out = in;", language=dace.Language.CPP)

arr_A = state_compute.add_access("A")
memlet = dace.Memlet(expr="A", subset="index")
state_compute.add_memlet_path(tasklet1, arr_A, src_conn="out", memlet=memlet)

memlet = dace.Memlet(expr="A", subset="start_ptr")
state_compute.add_memlet_path(arr_A, tasklet2, dst_conn="in", memlet=memlet)

ret_val = state_compute.add_write("out_val")
memlet = dace.Memlet(expr="out_val", subset="0")
state_compute.add_memlet_path(tasklet2, ret_val, src_conn="out", memlet=memlet)

sdfg.save("untransformed.sdfg")

sdfg.simplify()

sdfg.save("transformed.sdfg")
```
[...]

PR #1008, opened 2022-05-30
Fix for issue #1005

alexnick83:
Fixes issue #1005 

StateFusion must not happen if the assigned symbols in the states' common interstate edge are free symbols in any of the first state's input edges.

### DIFF
--- a/dace/transformation/interstate/state_fusion.py
+++ b/dace/transformation/interstate/state_fusion.py
@@ -208,7 +208,7 @@ class StateFusion(transformation.MultiStateTransformation, transformation.Simpli
                 if e.data.assignments.keys() & symbols_used:
                     return False
                 # Also fail in the inverse; symbols assigned on the second edge are free symbols on the first edge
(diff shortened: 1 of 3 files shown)


dace/transformation/interstate/state_fusion.py
```diff
@@ -207,6 +207,9 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
             for e in in_edges:
                 if e.data.assignments.keys() & symbols_used:
                     return False
+                # Also fail in the inverse; symbols assigned on the second edge are free symbols on the first edge
+                if out_edges[0].data.assignments.keys() & set(e.data.free_symbols):
```
tbennun, line 211: you could use `new_assignments` here instead for performance

tbennun (changes requested):
Very minor comment, otherwise LGTM

dace/transformation/interstate/state_fusion.py
alexnick83: Done

tbennun (approved):

PR #1009, opened 2022-05-30
GPUTransformSDFG scalar-related fixes

alexnick83:
This PR attempts to fix GPUTransformSDFG issues related to scalars:
- Transient scalars that are copied to the GPU and then used in host-based control flow.
- Scalar Tasklets, i.e., Tasklets working on scalar values/subsets outside of a Map scope, becoming GPU kernels.

### DIFF
--- a/dace/transformation/interstate/gpu_transform_sdfg.py
+++ b/dace/transformation/interstate/gpu_transform_sdfg.py
@@ -192,6 +192,32 @@ class GPUTransformSDFG(transformation.MultiStateTransformation):
                                                                                    dst_array.desc(sdfg)))
 
         #######################################################
+        # Step 3: Collect free tasklets and check for scalars that have to be moved to the GPU
+
+        gpu_storage = [dtypes.StorageType.GPU_Global, dtypes.StorageType.GPU_Shared, dtypes.StorageType.CPU_Pinned]
+
+        gpu_scalars = {}
+        for node, state in sdfg.all_nodes_recursive():
+            if isinstance(node, nodes.Tasklet):
+                if (state.entry_node(node) is None
+                        and not scope.is_devicelevel_gpu(state.parent, state, node, with_gpu_default=True)):
+                    scalar_output = True
+                    scalars = set()
+                    for e in state.out_edges(node):
+                        last_edge = state.memlet_path(e)[-1]
+                        if isinstance(last_edge.dst, nodes.AccessNode):
+                            desc = sdfg.arrays[last_edge.dst.data]
+                            if isinstance(desc, data.Scalar):
+                                scalars.add(last_edge.dst.data)
+                                continue
+                            if desc.storage not in gpu_storage and last_edge.data.num_elements() == 1:
+                                continue
+                            scalar_output = False
+                    if not scalar_output:
+                        global_code_nodes[state].append(node)
+                        gpu_scalars.update({k: None for k in scalars})
+
+        #######################################################
         # Step 4: Modify transient data storage
 
         const_syms = xfh.constant_symbols(sdfg)
@@ -207,10 +233,22 @@ class GPUTransformSDFG(transformation.MultiStateTransformation):
                     if any(isinstance(state.memlet_path(e)[-1].dst, nodes.EntryNode) for e in state.out_edges(node)):
                         continue
 
-                    gpu_storage = [
-                        dtypes.StorageType.GPU_Global, dtypes.StorageType.GPU_Shared, dtypes.StorageType.CPU_Pinned
-                    ]
+                    # gpu_storage = [
+                    #     dtypes.StorageType.GPU_Global, dtypes.StorageType.GPU_Shared, dtypes.StorageType.CPU_Pinned
+                    # ]
                     if sdict[node] is None and nodedesc.storage not in gpu_storage:
+
+                        if isinstance(nodedesc, data.Scalar) and not node.data in gpu_scalars:
+                            continue
+                            # written_by_map = False
+                            # for e in state.in_edges(node):
+                            #     src = state.memlet_path(e)[0].src
+                            #     if isinstance(src, nodes.AccessNode) and sdict[src]:
+                            #         written_by_map = True
+                            #         break
+                            # if not written_by_map:
+                            #     continue
+                        
                         # NOTE: the cloned arrays match too but it's the same
                         # storage so we don't care
                         nodedesc.storage = dtypes.StorageType.GPU_Global
@@ -247,12 +285,19 @@ class GPUTransformSDFG(transformation.MultiStateTransformation):
         #######################################################
         # Step 6: Wrap free tasklets and nested SDFGs with a GPU map
 
-        # Collect free tasklets
-        for node, state in sdfg.all_nodes_recursive():
-            if isinstance(node, nodes.Tasklet):
-                if (state.entry_node(node) is None
-                        and not scope.is_devicelevel_gpu(state.parent, state, node, with_gpu_default=True)):
-                    global_code_nodes[state].append(node)
+        # # Collect free tasklets
+        # for node, state in sdfg.all_nodes_recursive():
+        #     if isinstance(node, nodes.Tasklet):
+        #         if (state.entry_node(node) is None
+        #                 and not scope.is_devicelevel_gpu(state.parent, state, node, with_gpu_default=True)):
+        #             writing_to_scalar = False
+        #             for e in state.out_edges(node):
+        #                 dst = state.memlet_path(e)[-1].dst
+        #                 if isinstance(dst, nodes.AccessNode) and isinstance(sdfg.arrays[dst.data], data.Scalar):
+        #                     writing_to_scalar = True
+        #                     break
+        #             if not writing_to_scalar:
+        #                 global_code_nodes[state].append(node)
 
         for state, gcodes in global_code_nodes.items():
             for gcode in gcodes:
@@ -286,14 +331,18 @@ class GPUTransformSDFG(transformation.MultiStateTransformation):
         #######################################################
         # Step 7: Introduce copy-out if data used in outgoing interstate edges
 
+        cloned_data = set(cloned_arrays.keys()).union(gpu_scalars.keys())
+
         for state in list(sdfg.nodes()):
             arrays_used = set()
             for e in sdfg.out_edges(state):
                 # Used arrays = intersection between symbols and cloned arrays
-                arrays_used.update(set(e.data.free_symbols) & set(cloned_arrays.keys()))
+                # arrays_used.update(set(e.data.free_symbols) & set(cloned_arrays.keys()))
+                arrays_used.update(set(e.data.free_symbols) & cloned_data)
 
             # Create a state and copy out used arrays
             if len(arrays_used) > 0:
+
                 co_state = sdfg.add_state(state.label + '_icopyout')
 
                 # Reconnect outgoing edges to after interim copyout state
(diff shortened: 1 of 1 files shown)


dace/codegen/targets/cpu.py
```diff
@@ -231,9 +235,11 @@ def declare_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, de
         if self._dispatcher.declared_arrays.has(ptrname):
             return
 
-        # If this is a view/reference, do not declare it here
-        if isinstance(nodedesc, (data.View, data.Reference)):
-            return
+        # NOTE: We actually want to declare Views here, but we may not want to declare References (to be clarified).
+        # Keeping the original code as a comment to ease debugging.
+        # # If this is a view/reference, do not declare it here
+        # if isinstance(nodedesc, (data.View, data.Reference)):
```
tbennun, line 241: please explain this

dace/codegen/targets/cpu.py
```diff
@@ -277,16 +283,26 @@ def allocate_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, d
         # Check if array is already declared
         declared = self._dispatcher.declared_arrays.has(alloc_name)
 
+        define_var = self._dispatcher.defined_vars.add
+        # NOTE: The code below fixes symbol-related issues with transient data originally defined in a NestedSDFG scope
+        # but promoted to be persistent. These data must have their free symbols replaced with the corresponding
+        # top-level SDFG symbols.
+        if nodedesc.lifetime == dtypes.AllocationLifetime.Persistent:
+            define_var = self._dispatcher.defined_vars.add_global
+            if sdfg.parent and any(str(s) in sdfg.parent_nsdfg_node.symbol_mapping for s in nodedesc.free_symbols):
```
tbennun, line 292: what if it's two layers down and in the grandparent SDFG it is not defined as a free symbol (but e.g., set in an interstate edge)

tbennun: what I mean is that the while loop below should cover that and fail gracefully

dace/codegen/targets/cuda.py
```diff
@@ -422,6 +427,18 @@ def allocate_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, d
...
                                                         allocation_stream)
 
+        # NOTE: The code below fixes symbol-related issues with transient data originally defined in a NestedSDFG scope
+        # but promoted to be persistent. These data must have their free symbols replaced with the corresponding
+        # top-level SDFG symbols.
+        if nodedesc.lifetime == dtypes.AllocationLifetime.Persistent:
+            if sdfg.parent and any(str(s) in sdfg.parent_nsdfg_node.symbol_mapping for s in nodedesc.free_symbols):
+                nodedesc = copy.deepcopy(nodedesc)
+                csdfg = sdfg
+                while csdfg.parent_sdfg:
```
tbennun, line 437: same as above comment

dace/codegen/targets/fpga.py
```diff
@@ -1051,6 +1051,18 @@ def declare_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, de
...
                        allocation_stream):
 
+        # NOTE: The code below fixes symbol-related issues with transient data originally defined in a NestedSDFG scope
+        # but promoted to be persistent. These data must have their free symbols replaced with the corresponding
+        # top-level SDFG symbols.
+        if nodedesc.lifetime == dtypes.AllocationLifetime.Persistent:
+            if sdfg.parent and any(str(s) in sdfg.parent_nsdfg_node.symbol_mapping for s in nodedesc.free_symbols):
+                nodedesc = copy.deepcopy(nodedesc)
+                csdfg = sdfg
+                while csdfg.parent_sdfg:
```
tbennun, line 1061: same (maybe it's better to refactor this method out)

dace/transformation/auto/auto_optimize.py
```diff
@@ -387,6 +397,11 @@ def set_fast_implementations(sdfg: SDFG, device: dtypes.DeviceType, blocklist: L
     # general nodes
     for node, _ in sdfg.all_nodes_recursive():
         if isinstance(node, nodes.LibraryNode):
+            # NOTE: LibraryNodes with sequential schedule on GPU must be expanded to CUDA kernel-compatible code.
+            # NOTE: Pure implementations are a safe choice for now but this should be revisited in the future.
+            if device == dtypes.DeviceType.GPU and node.schedule == dtypes.ScheduleType.Sequential:
+                node.implementation = "pure"
```
tbennun, line 403: for some reason it doesn't seem like the "right" fix. Maybe default `implementation` could depend on the device context it's executing in.

dace/transformation/subgraph/subgraph_fusion.py
```diff
@@ -547,12 +547,12 @@ def adjust_arrays_nsdfg(self, sdfg: dace.sdfg.SDFG, nsdfg: nodes.NestedSDFG, nam
...
-            nsdfg.data(nname).strides = tuple(strides)
-            nsdfg.data(nname).total_size = total_size
+            nsdfg.data(nname)._strides = tuple(strides)
+            nsdfg.data(nname)._total_size = total_size
 
         else:
-            nsdfg.data(nname).strides = sdfg.data(name).strides
-            nsdfg.data(nname).total_size = sdfg.data(name).total_size
+            nsdfg.data(nname)._strides = sdfg.data(name).strides
+            nsdfg.data(nname)._total_size = sdfg.data(name).total_size
```
tbennun, line 555: why?

dace/transformation/subgraph/subgraph_fusion.py
```diff
@@ -1123,47 +1123,61 @@ def change_data(transient_array, shape, strides, total_size, offset, lifetime, s
...
+                        in_subset = ie.data.dst_subset
+
+                # Create transient data corresponding to the union of the incoming subsets.
+                desc = sdfg.arrays[node.data]
+                name, new_desc = sdfg.add_temp_transient(in_subset.bounding_box_size(), desc.dtype, desc.storage)
+                new_node = graph.add_access(name)
+
+                # Reconnect incoming edges through the transient data.
+                for ie in graph.in_edges(node):
+                    mem = Memlet(f"{name}[{ie.data.dst_subset.offset_new(in_subset, True)}] -> {ie.data.src_subset}")
```
tbennun, line 1145: please use the keyword argument constructor

dace/transformation/subgraph/subgraph_fusion.py
```diff
@@ -1123,47 +1123,61 @@ def change_data(transient_array, shape, strides, total_size, offset, lifetime, s
...
+                    # Update memlet paths.
+                    for e in graph.memlet_path(new_edge):
+                        if e.data.data == node.data:
+                            e.data.data = name
+                            e.data.dst_subset.offset(in_subset, True)
+
+                # Reconnect outgoing edges through the transient data.
+                for oe in graph.out_edges(node):
+                    if in_subset.covers(oe.data.src_subset):
+                        mem = Memlet(
```
tbennun, line 1157: please use the keyword argument constructor

dace/transformation/transformation.py
```diff
@@ -181,7 +182,10 @@ def setup_match(self,
             expr = self.expressions()[expr_index]
             for value in subgraph.values():
                 if not isinstance(value, int):
-                    raise TypeError('All values of ' 'subgraph' ' dictionary must be ' 'instances of int.')
+                    raise TypeError('All values of '
```
tbennun, line 185: what in the name of yapf happened here?

tests/npbench/compute_test.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2019-2022 ETH Zurich and the DaCe authors. All rights reserved.
+# Original application code: NPBench - https://github.com/spcl/npbench
+import dace.dtypes
```
tbennun, line 3: does this even deserve a test of its own?

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -192,27 +192,196 @@ def apply(self, _, sdfg: sd.SDFG):
                                                                                    dst_array.desc(sdfg)))
 
         #######################################################
-        # Step 4: Modify transient data storage
+        # Step 4: Change all top-level maps and library nodes to GPU schedule
+
+        gpu_nodes = set()
+        for state in sdfg.nodes():
+            sdict = state.scope_dict()
```
tbennun, line 199: probably better to use scope_children and iterate over `schild[None]`

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -192,27 +192,196 @@ def apply(self, _, sdfg: sd.SDFG):
...
+                    if desc.storage not in gpu_storage and last_edge.data.num_elements() == 1:
+                        sset, ssout = _recursive_in_check(last_edge.src, state, gpu_scalars)
+                        scalset = scalset.union(sset)
+                        scalout = scalout and ssout
+                        continue
+                    scalout = False
+            return scalset, scalout
+
+        def _is_devicelevel_gpu(sdfg, state, node):
+            """ Recursive version of `is_devicelevel_gpu` method. """
```
tbennun, line 292: malformed docstring, see other comment

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -192,27 +192,196 @@ def apply(self, _, sdfg: sd.SDFG):
...
+                    if isinstance(dst, nodes.AccessNode):
+                        desc = sdfg.arrays[dst.data]
+                        desc.storage = dtypes.StorageType.GPU_Global
+
+        #######################################################
+        # Step 5: Collect free tasklets and check for scalars that have to be moved to the GPU
+
+        gpu_storage = [dtypes.StorageType.GPU_Global, dtypes.StorageType.GPU_Shared, dtypes.StorageType.CPU_Pinned]
+
+        def _recursive_out_check(node, state, gpu_scalars):
```
tbennun, line 237: why not move all three functions out of this method? also why not reuse the recursive version of is_devicelevel_gpu from the other file?

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -192,27 +192,196 @@ def apply(self, _, sdfg: sd.SDFG):
...
+                    csdfg = state.parent
+                    # If the tasklet is not adjacent only to scalars or it is in a GPU scope.
+                    # The latter includes NestedSDFGs that have a GPU-Device schedule but are not in a GPU kernel.
+                    if (not scalar_output or (csdfg.parent is not None
+                                              and csdfg.parent_nsdfg_node.schedule == dtypes.ScheduleType.GPU_Default)):
+                        global_code_nodes[state].append(node)
+                        gpu_scalars.update({k: None for k in scalars})
+
+        # NOTE: We execute the above algorithm a second time to catch any potential new GPU-scalars.
+        # TODO: Should we run this recursively until there are no new GPU-scalars?
```
tbennun, line 324: can we simply run this once, but reversed? Since new gpu scalars only affect tasklets if they are output, we only need it to run once, or am I wrong?

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -192,27 +192,196 @@ def apply(self, _, sdfg: sd.SDFG):
...
+                                              and csdfg.parent_nsdfg_node.schedule == dtypes.ScheduleType.GPU_Default)):
+                        global_code_nodes[state].append(node)
+                        gpu_scalars.update({k: None for k in scalars})
+
+        #######################################################
+        # Step 6: Modify transient data storage
 
         const_syms = xfh.constant_symbols(sdfg)
 
+        def _codenode_condition(node):
```
tbennun, line 347: same (factor out)

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -192,27 +192,196 @@ def apply(self, _, sdfg: sd.SDFG):
...
+                                    break
+                            if not used_in_gpu_scope:
+                                for e in state.out_edges(node):
+                                    if _codenode_condition(state.memlet_path(e)[-1].dst):
+                                        used_in_gpu_scope = True
+                                        break
+                            if not used_in_gpu_scope:
+                                continue
+                            for e in state.all_edges(node):
+                                for node in (e.src, e.dst):
```
tbennun, line 378: shouldn't this be using memlet_path[0].src and [-1].dst?

tbennun (changes requested):
some comments. let's discuss

dace/codegen/targets/cpu.py
alexnick83: We delete this.

dace/transformation/interstate/gpu_transform_sdfg.py
alexnick83: Data can have access nodes in multiple states. Therefore, if some data is moved to GPU memory in one state, then we need to rerun the check for all other states that is used in and have been already checked. There is probably a better way to handle this and this is why I have left {NOTE, TODO}.

dace/transformation/interstate/gpu_transform_sdfg.py
alexnick83: Since, `sdict[node] is None`, the memlet path should always consist of a single edge, right? Which means that the use of `memlet_path` above is uneeded ... I need to check again in the next PR.

tbennun: True.

dace/transformation/interstate/gpu_transform_sdfg.py
tbennun: Agreed.

PR #1010, opened 2022-05-31
Adding OMP attributes to Maps

alexnick83:
This PR adds direct support to Maps for OpenMP pragma attributes, such as:
- num_threads
- schedule {static, dynamic, guided}
- chunk-size

### DIFF
--- a/tests/openmp_test.py
+++ b/tests/openmp_test.py
@@ -4,13 +4,12 @@ import numpy as np
 from dace import dtypes, nodes
 
 N = dace.symbol("N")
-J = dace.symbol("J")
 
 
 @dace.program
 def arrayop(inp: dace.float32[N], out: dace.float32[N]):
     for i in dace.map[0:N]:
-        out = 2 * inp
+        out[i] = 2 * inp[i]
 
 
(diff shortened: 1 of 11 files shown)


tests/openmp_test.py
```diff
@@ -0,0 +1,35 @@
...
+import numpy as np
+from dace import dtypes, nodes
+
+N = dace.symbol("N")
+J = dace.symbol("J")
+
+
+@dace.program
+def arrayop(inp: dace.float32[N], out: dace.float32[N]):
+    for i in dace.map[0:N]:
```
tbennun, line 12: isn't the map redundant since `out[:] = 2 * inp` would create one?

tests/openmp_test.py
```diff
@@ -0,0 +1,35 @@
...
+        out = 2 * inp
```
tbennun, line 13:
```suggestion
        out[:] = 2 * inp
```

tbennun (changes requested):
I don't like the idea behind this PR. I left some comments but I don't think we should merge this

PR #1013, opened 2022-06-01
SDFV: Add web serving capability

orausch:
When `filename` argument of SDFV is an int, start a `http.server` to serve the generated HTML and the dependencies.

This is really useful when interactively debugging sdfgs on a remote machine; you can just do `sdfg.view(8000)` (e.g. from within PDB), local-forward that port via SSH and then view the sdfg.

### DIFF
--- a/dace/cli/sdfv.py
+++ b/dace/cli/sdfv.py
@@ -1,5 +1,5 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
-""" SDFG visualizer that uses Flask, HTML5, and Javascript. """
+""" SDFG visualizer that uses Jinja, HTML5, and Javascript. """
 
 import json
 import tempfile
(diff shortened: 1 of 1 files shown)


dace/cli/sdfv.py
```diff
@@ -6,50 +6,78 @@
...
         os.close(fd)
 
+    if isinstance(filename, int):
+        # link in the web resources
+        os.symlink(os.path.join(basepath, 'webclient'), os.path.join(dirname, 'webclient'))
+
+        # start the web server
+        import http.server
+        handler = partialclass(http.server.SimpleHTTPRequestHandler, directory=dirname)
+        httpd = http.server.HTTPServer(('localhost', 8000), handler)
```
tbennun, line 69: Don't you want to use the given port here?

tbennun (changes requested):
One comment and a question: how do you stop serving the webpage? Otherwise LGTM

tbennun (approved):

PR #1017, opened 2022-06-10
Support literal lists, tuples, sets, and dictionaries in callbacks

tbennun:
through argument flattening and marshalling

### DIFF
--- a/dace/frontend/python/preprocessing.py
+++ b/dace/frontend/python/preprocessing.py
@@ -10,7 +10,7 @@ import sympy
 import sys
 import warnings
 
-from typing import Any, Callable, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, OrderedDict, Set, Tuple, Union
 import dace
 from dace import data, dtypes, subsets, symbolic, sdfg as sd
 from dace.config import Config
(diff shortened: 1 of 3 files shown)


dace/frontend/python/preprocessing.py
```diff
@@ -300,36 +309,120 @@ def has_replacement(callobj: Callable, parent_object: Optional[Any] = None, node
...
+                        continue
+                    if i in instructions:
+                        unflatten, skip = instructions[i]
+                        unflattened.append(unflatten(all_args[i:i + skip]))
+                        skip -= 1
+                    else:
+                        unflattened.append(arg)
+
+                args = unflattened[:poscount]
+                kwargs = {kw: arg for kw, arg in zip(keywords, unflattened[poscount:])}
```
gronerl, line 397: I think the posargs all are before the kwargs in `all_args` so the keys of instructions are all larger than `poscount`. Couldn't this logic be simplified to: 
```suggestion
                args = all_args[:poscount]
                unflattened = []
                for i, (unflatten, skip) in sorted(instructions.items()):
                    unflattened.append(unflatten(all_args[i:i + skip]))
                kwargs = {kw: arg for kw, arg in zip(keywords, unflattened)}
```
Or maybe even shorter with more list comprehension.

gronerl: The current version with the skips and the counters looks like a weird `while`/C-style `for` emulation.

tbennun: ok, I like the loop over instructions a lot more! The unflattened/poscount is still an issue, because even positional arguments are flattened (think `print([a, b])`). I'll revise.

dace/frontend/python/preprocessing.py
```diff
@@ -300,36 +309,120 @@ def has_replacement(callobj: Callable, parent_object: Optional[Any] = None, node
...
                 return func(*args, **kwargs)
-            return cb_func
-    return make_cb(keywords, poscount)
+
+            return cb_func        
+    else:
+        # Optimization: In the keyword argument case, make the callback itself faster based on existence of pos. arguments
```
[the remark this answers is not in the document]
tbennun: Agree with comment.
For the question: it generates a different callback function based on the arguments, so that it doesn't need to subscript anything.

gronerl (approved):

PR #1018, opened 2022-06-14
Fixes to MoveLoopIntoMap

alexnick83:
MoveLoopIntoMap should not apply when:
- The matched loop has a multi-state body.
- There are nodes outside the matched Map that have the loop iteration variable in their free symbols.


phschaad (approved):
LGTM

PR #1020, opened 2022-06-15
Various fixes to parsing and codegen

tbennun:
- [x] is_array should fail for types
- [x] Make pystr_to_symbolic faster by caching
- [x] Fix persistent memory codegen error


acalotoiu (approved):
Looks good to me

PR #1021, opened 2022-06-16
Simplify passes (2): Dead Code Elimination

tbennun:
- [x] Dead state (control flow) elimination
- [x] Dead code (dataflow) elimination
    - [x] Add `has_side_effects` to tasklets and library nodes (in order to identify which dataflow can be removed)

### DIFF
--- a/dace/sdfg/graph.py
+++ b/dace/sdfg/graph.py
@@ -664,10 +664,13 @@ class OrderedDiGraph(Graph[NodeT, EdgeT], Generic[NodeT, EdgeT]):
         return self._nx.add_edge(src, dst, data=data)
 
     def remove_node(self, node: NodeT):
-        for edge in itertools.chain(self.in_edges(node), self.out_edges(node)):
-            self.remove_edge(edge)
-        del self._nodes[node]
-        self._nx.remove_node(node)
+        try:
+            for edge in itertools.chain(self.in_edges(node), self.out_edges(node)):
+                self.remove_edge(edge)
+            del self._nodes[node]
+            self._nx.remove_node(node)
+        except KeyError:
+            pass
 
     def remove_edge(self, edge: Edge[EdgeT]):
         src = edge.src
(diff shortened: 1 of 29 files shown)


dace/sdfg/graph.py
```diff
@@ -664,10 +664,13 @@ def add_edge(self, src: NodeT, dst: NodeT, data: EdgeT = None):
...
-            self.remove_edge(edge)
-        del self._nodes[node]
-        self._nx.remove_node(node)
+        try:
+            for edge in itertools.chain(self.in_edges(node), self.out_edges(node)):
+                self.remove_edge(edge)
+            del self._nodes[node]
+            self._nx.remove_node(node)
+        except KeyError:
+            pass
```
phschaad, line 673: Ignored deliberately or missed?

dace/sdfg/nodes.py
```diff
@@ -377,6 +386,24 @@ def validate(self, sdfg, state):
...
+        if self.side_effects is not None:
+            return self.side_effects
+
+        # If side effect property is not defined, find calls within tasklet
+        if self.code.language == dace.dtypes.Language.Python and self.code.code:
+            for stmt in self.code.code:
+                for n in ast.walk(stmt):
+                    if isinstance(n, ast.Call):
+                        cname = rname(n.func)
+                        if cname in sdfg.symbols or cname in sdfg.arrays:
```
phschaad, line 403: If the function call name is one of the symbols or data containers it has side effects? Might need a doc comment, or am I reading it wrong?

phschaad (approved):
LGTM, only 2 small questions / observations.

dace/sdfg/graph.py
tbennun: Removing a node that is already not in the graph now won't complain _and_ run in decent performance (otherwise need to look it up every time).

tbennun: It matches the behavior of networkx iirc

dace/sdfg/nodes.py
tbennun: the check basically asks "is this function a callback" (they may be `Scalar(dace.callback)` or symbol typed `dace.callback`. Will comment.

PR #738, opened 2021-06-04
Add code generator for Snitch

noah95:
This adds a new code generator back-end for the [Snitch System](https://github.com/pulp-platform/snitch). The new code generator adds:

- Emission of Snitch-compatible C-code
- Array allocation/de-allocation
- Array copy using the DMA engine
- Parallelism with OpenMP
- SSR data mover inference

Furthermore, there is a simple example in samples/snitch/axpy.py that shows how we can make use of SSR and DMA extensions.

@tbennun is familiar with the work.

PR #1024, opened 2022-06-18
Snitch codegen

am-ivanov:
Updated https://github.com/spcl/dace/pull/738


acalotoiu (approved):
LGTM, however, significant parts of snitch.py are not covered by tests so there might be additional issues down the line

PR #1025, opened 2022-06-19
Optional array support

tbennun:
Analyze which arrays can receive a value of None, which cannot, add checks, and optimize state machine accordingly.
- [x] Add `optional` property to Array
- [x] Add argument check for non-optional arrays with a None value
- [x] Support `Optional[T]` type hint
- [x] OptionalArrayInference pass:
    - [x] Infer which arrays are always read/written and thus not optional
    - [x] Propagate `optional` value to descriptors in nested SDFGs
    - [x] Evaluate `x is [not] None` conditions on inter-state edges based on property


acalotoiu (approved):
Looks good to me. For the future, maybe separate code beautification from content PRs.

PR #1040, opened 2022-07-01
Support customizing GPU block dimension per map

tbennun:


lukastruemper (approved):
Yes, nice addition.

However, I don't fully understand how the grid size for a >4D-map is currently computed, but can look it up by myself as it is unrelated to the PR

tbennun:
> Yes, nice addition.
> 
> However, I don't fully understand how the grid size for a >4D-map is currently computed, but can look it up by myself as it is unrelated to the PR

In the same function, it is computed as follows:
* Less than 3 dimensions are rounded up to three dimensions with size=1 for extra dimensions
* Anything above 3 dimensions is linearized (flattened) in kernel dimensions and then de-linearized within the map if the index is necessary

PR #1042, opened 2022-07-04
Simplify Passes (4): Array Elimination

tbennun:
Implement array elimination pass (covers redundant copies/views, merge source/sink arrays).


phschaad (approved):
LGTM

PR #1071, opened 2022-07-27
Support for `__array_interface__`

gronerl:
This PR implements support for objects implementing the `__array_interface__` similar to the already supported `__cuda_array_interface__`

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -33,9 +33,26 @@ def create_datadescriptor(obj, no_custom_desc=False):
         return obj.__descriptor__()
     elif not no_custom_desc and hasattr(obj, 'descriptor'):
         return obj.descriptor
-    elif isinstance(obj, (list, tuple, numpy.ndarray)):
-        if isinstance(obj, (list, tuple)):  # Lists and tuples are cast to numpy
-            obj = numpy.array(obj)
+    # elif isinstance(obj, (list, tuple, numpy.ndarray)):
+    elif dtypes.is_array(obj):
+        if dtypes.is_gpu_array(obj):
+            interface = obj.__cuda_array_interface__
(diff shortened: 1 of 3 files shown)


dace/data.py
```diff
@@ -33,9 +33,26 @@ def create_datadescriptor(obj, no_custom_desc=False):
...
+    # elif isinstance(obj, (list, tuple, numpy.ndarray)):
+    elif dtypes.is_array(obj):
+        if dtypes.is_gpu_array(obj):
+            interface = obj.__cuda_array_interface__
+            storage = dtypes.StorageType.GPU_Global
+        else:
+            interface = obj.__array_interface__
+            storage = dtypes.StorageType.CPU_Heap
+
+        dtype = dtypes.typeclass(numpy.dtype(interface['typestr']).type)
```
tbennun, line 45: supporting structs requires the following code (adapted from below):
```suggestion
        if hasattr(obj, 'dtype') and obj.dtype.fields is not None:  # Struct
            dtype = dtypes.struct('unnamed', **{k: dtypes.typeclass(v[0].type) for k, v in obj.dtype.fields.items()})
        else:   
            if numpy.dtype(interface['typestr']).type is numpy.void:  # Struct from __array_interface__
               if 'descr' in interface:
                    dtype = dtypes.struct('unnamed', **{k: dtypes.typeclass(numpy.dtype(v).type) for k, v in interface['descr']})
                else:
                    raise TypeError(f'Cannot infer data type of array interface object "{interface}"')
            else:
                dtype = dtypes.typeclass(numpy.dtype(interface['typestr']).type)
```

tbennun (changes requested):
Needs some changes

tbennun (approved):

PR #1073, opened 2022-07-28
Add proper symbolic equality checks for broadcasting

orausch:
Previously, broadcasting with symbolic shapes would not do symbolic equality checking when checking broadcasting. For example, the following arrays would not broadcast together:
```
A: shape [N,]
B: shape [N * (j + 1) - N * j]
```

I'm pretty sure that there are many other cases where this is being done incorrectly, but this helps for my use case for now.


alexnick83 (approved):

orausch:
BTW, I don't have merge rights

PR #1075, opened 2022-07-31
Set storage type for torch tensors when creating data descriptors

orausch:
Previously the storage of torch tensors was ignored. Also fixes a regression introduced in #1071.

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -43,10 +43,12 @@ def create_datadescriptor(obj, no_custom_desc=False):
 
         if hasattr(obj, 'dtype') and obj.dtype.fields is not None:  # Struct
             dtype = dtypes.struct('unnamed', **{k: dtypes.typeclass(v[0].type) for k, v in obj.dtype.fields.items()})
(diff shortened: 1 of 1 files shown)


dace/data.py
```diff
@@ -33,7 +33,7 @@ def create_datadescriptor(obj, no_custom_desc=False):
         return obj.__descriptor__()
     elif not no_custom_desc and hasattr(obj, 'descriptor'):
         return obj.descriptor
-    elif dtypes.is_array(obj):
+    elif dtypes.is_array(obj) and (hasattr(obj, '__array_interface__') or hasattr(obj, '__cuda_array_interface__')):
```
tbennun, line 36: What regression does this cause?

orausch: For torch tensors ``is_array`` is true, but ``__array_interface__`` doesn't exist. Thus this branch gets chosen but fails later when `__array_interface__`` is accessed.

tbennun (approved):

PR #1076, opened 2022-08-01
Support constants in SDFG.data, Propagate constants into NestedSDFGs.

orausch:
Previously, the code generator would not be aware when an input to a NestedSDFG is constant, causing compiler errors related to ``const``.

### DIFF
--- a/dace/codegen/targets/framecode.py
+++ b/dace/codegen/targets/framecode.py
@@ -64,6 +64,7 @@ class DaCeCodeGenerator(object):
 
                 # found a new nested sdfg: resolve symbols and constants
                 result = nsdfg.free_symbols.union(nsdfg.constants_prop.keys())
+                result |= self._symbols_and_constants[nsdfg._parent_sdfg.sdfg_id]
 
                 # check for constant inputs
                 for edge in state.in_edges(nested):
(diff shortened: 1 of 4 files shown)


dace/codegen/targets/framecode.py
```diff
@@ -52,13 +52,30 @@ def __init__(self, sdfg: SDFG):
         fsyms = self.free_symbols(sdfg)
         self.arglist = sdfg.arglist(scalars_only=False, free_symbols=fsyms)
 
+        # resolve all symbols and constants
+        # first handle root
+        self._symbols_and_constants[sdfg.sdfg_id] = sdfg.free_symbols.union(sdfg.constants.keys())
```
tbennun, line 57: Should be constants_prop

dace/codegen/targets/framecode.py
```diff
@@ -52,13 +52,30 @@ def __init__(self, sdfg: SDFG):
...
+        # then recurse
+        for nested, state in sdfg.all_nodes_recursive():
+            if isinstance(nested, nodes.NestedSDFG):
+                state: SDFGState
+
+                nsdfg = nested.sdfg
+
+                # found a new nested sdfg: resolve symbols and constants
+                result = nsdfg.free_symbols.union(nsdfg.constants.keys())
```
tbennun, line 66: Same

tbennun (changes requested):
Tiny performance comment

tbennun (approved):

PR #1077, opened 2022-08-03
OTFMapFusion re-implemented

lukastruemper:
So far, OTFMapFusion was only appliable in cases where the map parameters of both maps did match. I generalized the implementation by solving the equation systems of the memlets, i.e., 

second map reads tmp[i, j]
first map writes tmp[k, l]

we know: i -> k, j -> l

I am also renaming the map parameters temporarily in the equations so that the two maps can have the same parameters but in permuted order:

second map reads tmp[i, j]
first map writes tmp[j, i]
 
j -> i, i -> j

Update:

- I added support for tree-like cases
- Fixed a frequently occurring problem where the in_connector name was already taken

### DIFF
--- a/dace/transformation/dataflow/otf_map_fusion.py
+++ b/dace/transformation/dataflow/otf_map_fusion.py
@@ -237,6 +237,21 @@ class OTFMapFusion(transformation.SingleStateTransformation):
 
     @staticmethod
     def solve(first_params, write_accesses, second_params, read_accesses):
+        """ Infers the memory access for the write memlet given the
+            location/parameters of the read access.
+
+            Example:
+            - Write memlet: A[i + 1, j]
+            - Read memlet: A[k, l]
+            - Infer: k -> i - 1, l - > j
+
+            :param first_params: parameters of the first_map_entry.
+            :param write_accesses: subset ranges of the write memlet.
+            :param second_params: parameters of the second map_entry.
+            :param read_accesses: subset ranges of the read memlet.
+            :return: mapping of parameters.
+        """
+
         # Make sure that parameters of first_map_entry
         # are different symbols than parameters of
         # second_map_entry
--- a/dace/transformation/dataflow/otf_map_fusion.py
+++ b/dace/transformation/dataflow/otf_map_fusion.py
@@ -10,6 +10,7 @@ from dace.sdfg.state import SDFGState
 from dace.sdfg import nodes as nds
 from dace.memlet import Memlet
 from dace.sdfg import utils as sdutil
+from dace.subsets import Range
 from dace.transformation import transformation
 from dace import data as dt
 from dace import symbolic
@@ -182,14 +183,18 @@ class OTFMapFusion(transformation.SingleStateTransformation):
                         ranges = []
                         for i, access in enumerate(memlet.subset.ranges):
                             b, e, s = access
-                            b = sympy.sympify(b, locals=sdfg.constants)
-                            e = sympy.sympify(e, locals=sdfg.constants)
+                            b = sympy.sympify(b)
+                            e = sympy.sympify(e)
+                            s = sympy.sympify(s)
+
                             for param, sub in param_subs[i].items():
                                 b = b.subs(param, sub)
                                 e = e.subs(param, sub)
+                                s = s.subs(param, sub)
 
-                            ranges.append((b, e, s))
-                        memlet.subset.ranges = ranges
+                            ranges.append((str(b), str(e), str(s)))
+
+                        memlet.subset = Range(ranges)
 
                         in_connector = edge.src_conn.replace("OUT", "IN")
                         if in_connector in connector_mapping:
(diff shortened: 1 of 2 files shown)


dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -175,12 +236,64 @@ def map_memlet(memlet):
...
-        """Compute subset offset of `offset_memlet` relative to `base_memlet`."""
-        def offset(base_range, offset_range):
-            b0, e0, s0 = base_range
-            b1, e1, s1 = offset_range
-            assert e1 - e0 == b1 - b0 and s0 == s1
-            return int(e1 - e0)
-
-        return tuple(offset(b, o) for b, o in zip(base_memlet.subset.ranges, offset_memlet.subset.ranges))
+    def solve(first_params, write_accesses, second_params, read_accesses):
+        # Make sure that parameters of first_map_entry
```
phschaad, line 240: yapf

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -175,12 +236,64 @@ def map_memlet(memlet):
...
-    def _memlet_offsets(base_memlet, offset_memlet):
-        """Compute subset offset of `offset_memlet` relative to `base_memlet`."""
-        def offset(base_range, offset_range):
-            b0, e0, s0 = base_range
-            b1, e1, s1 = offset_range
-            assert e1 - e0 == b1 - b0 and s0 == s1
-            return int(e1 - e0)
-
-        return tuple(offset(b, o) for b, o in zip(base_memlet.subset.ranges, offset_memlet.subset.ranges))
+    def solve(first_params, write_accesses, second_params, read_accesses):
```
phschaad, line 239: This method could use a doc comment maybe

phschaad (changes requested):
LGTM overall, minor comments

dace/transformation/dataflow/otf_map_fusion.py
lukastruemper: No yapf for me, version mismatch?

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -175,12 +262,87 @@ def map_memlet(memlet):
...
+        """ Infers the memory access for the write memlet given the
+            location/parameters of the read access.
+
+            Example:
+            - Write memlet: A[i + 1, j]
+            - Read memlet: A[k, l]
+            - Infer: k -> i - 1, l - > j
```
tbennun, line 272: improper format, sphinx will complain. Instead, put the expressions in double-backticks:
```
``A[i + 1, j]``
```

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -175,12 +262,87 @@ def map_memlet(memlet):
...
-        """Compute subset offset of `offset_memlet` relative to `base_memlet`."""
-        def offset(base_range, offset_range):
-            b0, e0, s0 = base_range
-            b1, e1, s1 = offset_range
-            assert e1 - e0 == b1 - b0 and s0 == s1
-            return int(e1 - e0)
-
-        return tuple(offset(b, o) for b, o in zip(base_memlet.subset.ranges, offset_memlet.subset.ranges))
+    def solve(first_params, write_accesses, second_params, read_accesses):
+        """ Infers the memory access for the write memlet given the
```
tbennun, line 266: missing newline before docstring starts

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -175,12 +262,87 @@ def map_memlet(memlet):
...
+            :param second_params: parameters of the second map_entry.
+            :param read_accesses: subset ranges of the read memlet.
+            :return: mapping of parameters.
+        """
+
+        # Make sure that parameters of first_map_entry are different symbols than parameters of second_map_entry.
+        first_params_subs = {}
+        first_params_subs_ = {}
+        for i, param in enumerate(first_params):
+            s = sympy.symbols(f'f_{i}')
```
tbennun, line 285: symbolic.symbol(...)

tbennun (changes requested):
Looks ok in general but there are some misuses of dace symbolics. We cannot use sympy directly as-is. See comments.

tbennun (approved):
OK from me, but the random connector thing seems odd, especially since we have an API for this (`next_connector`): https://github.com/spcl/dace/blob/master/dace/sdfg/nodes.py#L177

phschaad (approved):
LGTM in general, only one thing:
For renaming the connectors I would use this: https://github.com/spcl/dace/blob/4645a6a34ababb48d0786eca9f26604e2f8073f1/dace/data.py#L128

lukastruemper:
> OK from me, but the random connector thing seems odd, especially since we have an API for this (`next_connector`): https://github.com/spcl/dace/blob/master/dace/sdfg/nodes.py#L177

Ah, this is good! I will fix and push later

PR #1078, opened 2022-08-03
Save only modified config entries

tbennun:


dace/config_schema.yml
```diff
@@ -631,101 +631,6 @@ required:
                         default_Darwin: ''
                         default_Windows: ''
 
-
-    execution:
```
phschaad, line 635: These were all deprecated?

phschaad (approved):
LGTM

dace/config_schema.yml
tbennun: Yes, DIODEv1 related

PR #1080, opened 2022-08-05
Replacements look up base classes

tbennun:


dace/frontend/common/op_repository.py
```diff
@@ -1,10 +1,30 @@
...
+             superclasses.
+    """
+    if isinstance(class_or_name, str):
+        return [class_or_name]
+
+    classes = [class_or_name.__name__]
+    for base in class_or_name.__bases__:
+        classes.extend(_get_all_bases(base))
+
+    return deduplicate(classes)
```
orausch, line 25: This should retain class inheritance order. For example, `View` has `Array` as a base, and the repository should prefer implementations for `View` above implementations for `Array`.

orausch: Ah, I see now that deduplicate handles this correctly, never mind!

orausch (approved):

PR #1081, opened 2022-08-06
Explicit support for string literals in Python frontend

### LINKED ISSUE #1072: String literal evaluated as variable in callback parameters
When passing a string literal in a callback argument which value is  _the exact name_ of another variable, it gets swapped with the variable `__repr__` instead of being treated as a string literal.

Code:
```
import dace
import numpy as np

def force_callback(fn):
    return fn

@force_callback
def a_callback(an_arr: np.ndarray, a_name: str):
    print(
        a_name
    )  # < This should print "an_arr" but will print the __repr__ of an_arr instead

@dace.program
def the_program(an_arr):
    an_arr[:] = 4
    a_callback(an_arr, "an_arr") # < any other string literal works fine

an_arr = np.ones((10))
the_program(an_arr)
```

Expecting: `an_arr` to be printed, but getting `[4. 4. 4. 4. 4. 4. 4. 4. 4. 4.]`

tbennun:
Before this PR, there was effectively no difference between a string and an array name.

Fixes #1072

### DIFF
--- a/dace/codegen/cppunparse.py
+++ b/dace/codegen/cppunparse.py
@@ -527,12 +527,16 @@ class CPPUnparser:
         self.leave()
 
     def _write_constant(self, value):
-        result = repr(value)
-        if isinstance(value, (float, complex)):
-            # Substitute overflowing decimal literal for AST infinities.
-            self.write(result.replace("inf", INFSTR))
+        # Special case for byte literal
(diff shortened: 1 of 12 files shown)


alexnick83 (changes requested):
Looks good. I would like to see examples with bytes and how this will interact with CRC16.

alexnick83 (approved):

PR #1085, opened 2022-08-11
Frontend fixes

alexnick83:
The PR addresses two issues:

The frontend has a feature where the write-subsets of augmented assignments appearing inside Maps are checked for dependency on the Map parameters. According to the results of this check, the frontend will avoid using WCR and instead generate a subgraph where the write-data appear explicitly as read-data too. However, this feature will not work correctly when a Map has multiple writes to the same dimension of a data container, with index expressions that create write-conflicts (see `tests.python_frontend.augassign_wcr_test.augassign_wcr4`). We cannot rectify the issue within the bounds of the frontend's current design, which is not to look ahead for the following statements in the code. Instead of removing the feature altogether, which may have undesired performance consequences for users (there is currently no `WCRtoAugAssign` transformation), a configuration variable `avoid_wcr` with default value `False` is added to enable/disable the feature.

When the frontend generates loops that contain an if-condition deciding between executing a break or continue statement, an extraneous source state is generated, which may lead to broken SDFGs after applying StateFusion. The issue is fixed by detecting these extraneous states and deleting them at the end of the frontend's loop-generation methods `visit_For` and `visit_While.` Note that the `visit_For` changes are ported from the `simplify-fixes` branch.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -3193,7 +3193,7 @@ class ProgramVisitor(ExtNodeVisitor):
                     independent = True
                     waccess = inverse_dict_lookup(self.accesses, (new_name, new_rng))
                     if self.map_symbols and waccess:
-                        if Config.get_bool('frontend', 'avoid_wcr'):
+                        if not Config.get_bool('frontend', 'avoid_wcr'):
                             independent = False
                         else:
                             for s in self.map_symbols:
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -3189,14 +3189,17 @@ class ProgramVisitor(ExtNodeVisitor):
             # Strict independent access check for augmented assignments
             if op:
                 independent = False
-                if Config.get_bool('frontend', 'avoid_wcr') and not _subset_is_local_symbol_dependent(new_rng, self):
+                if not _subset_is_local_symbol_dependent(new_rng, self):
                     independent = True
                     waccess = inverse_dict_lookup(self.accesses, (new_name, new_rng))
                     if self.map_symbols and waccess:
-                        for s in self.map_symbols:
-                            if s not in waccess[1].free_symbols:
-                                independent = False
-                                break
+                        if Config.get_bool('frontend', 'avoid_wcr'):
+                            independent = False
+                        else:
+                            for s in self.map_symbols:
+                                if s not in waccess[1].free_symbols:
+                                    independent = False
+                                    break
 
             # Handle output indirection
             output_indirection = None
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -3158,7 +3158,7 @@ class ProgramVisitor(ExtNodeVisitor):
             # Strict independent access check for augmented assignments
             if op:
                 independent = False
-                if not _subset_is_local_symbol_dependent(new_rng, self):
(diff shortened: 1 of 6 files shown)


tests/python_frontend/loops_test.py
```diff
@@ -457,6 +457,37 @@ def while_else(A: dace.float64[2]):
...
+        if cond > 0:
+            break
+        else:
+            i += 1
+            continue
+
+
+def test_branch_in_while():
+    sdfg = branch_in_while.to_sdfg(simplify=False)
+    sdfg.view()
```
tbennun, line 487:
```suggestion
```

dace/frontend/python/newast.py
```diff
@@ -2209,6 +2219,12 @@ def visit_For(self, node: ast.For):
                     self.sdfg.remove_edge(e)
                 self.sdfg.add_edge(next_state, loop_end, dace.InterstateEdge())
             self.loop_idx -= 1
+
+            for state in body_states:
+                if not nx.has_path(self.sdfg.nx, loop_guard, state):
+                    for e in self.sdfg.all_edges(state):
+                        self.sdfg.remove_edge(e)
```
tbennun, line 2226: no need to remove the edges, removing the node would do that

dace/frontend/python/newast.py
```diff
@@ -2302,6 +2318,15 @@ def visit_While(self, node: ast.While):
 
             # The state that all "break" edges go to
             loop_end = self._add_state(f'postwhile_{node.lineno}')
+        
+        body_states = set()
+        to_visit = [first_loop_state]
+        while to_visit:
```
tbennun, line 2324: why not use our existing traversal methods like dfs_conditional?

dace/frontend/python/newast.py
```diff
@@ -2319,6 +2344,12 @@ def visit_While(self, node: ast.While):
             self.sdfg.add_edge(next_state, loop_end, dace.InterstateEdge())
         self.loop_idx -= 1
 
+        for state in body_states:
+            if not nx.has_path(self.sdfg.nx, end_guard, state):
+                for e in self.sdfg.all_edges(state):
+                    self.sdfg.remove_edge(e)
```
tbennun, line 2350: no need to remove the edges, removing the node would do that

dace/frontend/python/newast.py
```diff
@@ -2194,6 +2195,15 @@ def visit_For(self, node: ast.For):
                 # The state that all "break" edges go to
                 loop_end = self._add_state(f'postloop_{node.lineno}')
 
+            body_states = set()
+            to_visit = [first_loop_state]
```
tbennun, line 2199: why not use our existing traversal methods like dfs_conditional?

tbennun (changes requested):
Ideas look good, just needs some reuse of existing infra

dace/frontend/python/newast.py
alexnick83: Apart from being, in general, difficult to locate these methods in DaCe, the problem is that it is not always clear what they do or what their constraints are. For example, according to `dfs_conditional` documentation, the input `G` DiGraph is assumed acyclic. After looking at the code, it seems likely that the documentation is wrong, but who knows? If anything, I would have made a new method that explicitly finds the multi-state body of loops because we do that in many transformations already without using these traversal methods

tbennun (approved):

PR #1092, opened 2022-08-18
Reduce node auto gpu expansion

hodelcl:
This PR includes two things:
1. function `get_reduction_schedule`
2. auto expansion of the reduce node

This PR adds the auto expansion to the set of available reduce node expansions. The auto expansion is aimed at GPUs. Its goal is to compute an optimal mapping of the reduction input to GPU threads and expands the reduce node according to the computed schedule. 

The schedule gets computed via the newly added function `get_reduction_schedule`, which takes as an input the dimension of the input tensor and a list of the axes to reduce and returns a Python dictionary `schedule`, which specifies the number of blocks and threads per block to use, etc. At the highest level, the schedule depends on whether contiguous data or non-contiguous data has to be reduced.

The actual expansion called `ExpandReduceGPUAuto` then calls `get_reduction_schedule` and creates the actual SDFG expanding the reduce node according to the schedule returned by `get_reduction_schedule`.

If the auto expansion cannot handle a specific input, it simply falls back to the pure expansion `ExpandReducePure`.


orausch:
Exciting! Can't wait to try this

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1075,6 +1077,403 @@ def expansion(node: 'Reduce', state: SDFGState, sdfg: SDFG, partial_width=16):
...
+
+        if schedule['error']:
+            # return pure expansion if error
+
+            ########################################################
+            ###    This gives segfault if we reduce all axes     ###
+            ###(GPU map with 1 thread block doesn't seem to work)###
+            ########################################################
+
+            pure_sdfg = ExpandReducePure.expansion(node, state, sdfg)
```
orausch, line 1132: It would be better to add this check to `can_be_applied`

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1075,6 +1077,403 @@ def expansion(node: 'Reduce', state: SDFGState, sdfg: SDFG, partial_width=16):
...
+        # input_dims = len(insubset)
+        # output_dims = len(outsubset)
+        input_data = sdfg.arrays[inedge.data.data]
+        output_data = sdfg.arrays[outedge.data.data]
+
+        in_type = input_data.dtype
+
+        if input_data.storage != dtypes.StorageType.GPU_Global:
+            # data doesnt reside on GPU --> return pure expansion
+            return ExpandReducePure.expansion(node, state, sdfg)
```
orausch, line 1110: It would be better to add this check to `can_be_applied`

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1075,6 +1077,403 @@ def expansion(node: 'Reduce', state: SDFGState, sdfg: SDFG, partial_width=16):
...
+
+        if (schedule['contiguous_dim']):
+            # we are reducing the contiguous dimension
+
+            vectorize = schedule['vectorize']
+
+            ome, omx = nstate.add_map('block_grid',
+                                      {'_o%d' % i: '0:%s' % symstr(sz)
+                                       for i, sz in enumerate(schedule['grid'])},
+                                      schedule=dtypes.ScheduleType.GPU_Device)
```
orausch, line 1171: You should probably check within `can_be_applied` if we already in a GPU kernel or not. Reductions can appear within other kernels after subgraph fusions

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1075,6 +1077,403 @@ def expansion(node: 'Reduce', state: SDFGState, sdfg: SDFG, partial_width=16):
...
+
+            outm = dace.Memlet.simple('_out',
```
orausch, line 1173: `dace.Memlet.simple` is deprecated, you can just use `dace.Memlet`

dace/libraries/standard/reduction_planner.py
```diff
@@ -0,0 +1,209 @@
...
+    # apply vectorization or not
+    use_vectorization = True
+    # apply mini-warps optimitation or not
+    use_mini_warps = True
+
+    # define the warp size
+    warp_size = 32
+    wide_load_size = 16  #bytes
+
+    schedule = {
```
orausch, line 26: You could maybe add a data class to help with readability a bit:
```python
from dace.frontend.python.replacements import Size
@dataclasses.dataclass
class ReductionSchedule:
  grid: List[Size]
  ...
```

orausch (commented):
Just a few comments

dace/libraries/standard/nodes/reduce.py
hodelcl: Done

dace/libraries/standard/reduction_planner.py
hodelcl: Done

dace/libraries/standard/reduction_planner.py
```diff
@@ -0,0 +1,219 @@
...
+                schedule.mini_warps = True
+                schedule.num_mini_warps = warp_size // schedule.block[0]
+        if use_vectorization and not schedule.mini_warps:
+            #check if we can use vectorization
+            if (shape[contiguous_dimension] % num_loaded_elements == 0) == True:
+                schedule.vectorize = True
+
+        if schedule.grid == []:
+            # TODO: solve this issue
+            warnings.warn('Falling back to pure expansion, since invalid schedule.')
```
tbennun, line 215: same two comments as above

dace/libraries/standard/reduction_planner.py
```diff
@@ -0,0 +1,219 @@
...
+            return schedule
+
+        # grid is all the dims expect the one to reduce and the last
+        schedule.grid = []
+        for i, s in enumerate(shape):
+            if i != contiguous_dimension and i != axes[0]:
+                schedule.grid.append(s)
+
+        if use_mini_warps:
+            #check if we can use mini_warps
```
tbennun, line 203:
```suggestion
            # Check if we can use mini_warps
```

tbennun (changes requested):
Overall looks good. There are some minor comments about how things are implemented. The planner could be more Pythonic. The most important part is to change the defaults and only use autogpu in auto_optimize.

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1089,12 +1471,14 @@ class Reduce(dace.sdfg.nodes.LibraryNode):
...
         'CUDA (block)': ExpandReduceCUDABlock,
         'CUDA (block allreduce)': ExpandReduceCUDABlockAll,
-        'FPGAPartialReduction': ExpandReduceFPGAPartialReduction
+        'FPGAPartialReduction': ExpandReduceFPGAPartialReduction,
+        'GPUAuto': ExpandReduceGPUAuto
         # 'CUDA (warp)': ExpandReduceCUDAWarp,
         # 'CUDA (warp allreduce)': ExpandReduceCUDAWarpAll
     }
 
-    default_implementation = 'pure'
```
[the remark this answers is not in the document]
hodelcl: Okay, I understand this logically, but I am not sure how the code would look like. How do I check if we have `auto_optimize`? And if we have `auto_optimize` for `DeviceType.GPU`, what do I do then? Set `default_implementation = 'GPUAuto'`?

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1075,6 +1077,386 @@ def expansion(node: 'Reduce', state: SDFGState, sdfg: SDFG, partial_width=16):
...
+            else:
+                input_subset.append('_o%d' % octr)
+                octr += 1
+
+        if (schedule.contiguous_dim):
+            # we are reducing the contiguous dimension
+
+            vectorize = schedule.vectorize
+
+            ome, omx = nstate.add_map('grid', {f'_o{i}': f'0:{symstr(sz)}'
```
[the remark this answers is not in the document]
hodelcl: Just to be sure: Instead of `f'0:{symstr(sz)}'` I do `subsets.Range([(0, sz-1, 1)])` ?

dace/libraries/standard/nodes/reduce.py
```diff
@@ -1075,6 +1077,386 @@ def expansion(node: 'Reduce', state: SDFGState, sdfg: SDFG, partial_width=16):
...
+        outsubset = dcpy(outedge.data.subset)
+        osqdim = outsubset.squeeze()
+        # input_dims = len(insubset)
+        # output_dims = len(outsubset)
+        raw_input_data = sdfg.arrays[inedge.data.data]
+        raw_output_data = sdfg.arrays[outedge.data.data]
+
+        in_type = raw_input_data.dtype
+
+        if raw_input_data.storage != dtypes.StorageType.GPU_Global:
```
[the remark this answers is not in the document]
hodelcl: Could you elaborate on that?

As far as I understand, we need to check if our whole reduce node is inside a bigger GPU kernel already, right?
And why is this bad? Because we cannot have a GPU map inside another GPU map in the SDFG?

And why does Pure not have the same issues?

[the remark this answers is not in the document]
orausch: > Because we cannot have a GPU map inside another GPU map in the SDFG?

yes

> And why does Pure not have the same issues?

Pure expands to maps with `Default` schedule, these get inferred correctly later.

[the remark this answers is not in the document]
Biirait: Ah yes, thank you. Makes sense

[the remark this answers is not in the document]
hodelcl: Ah yes, makes sense. Thanks

tbennun (approved):

PR #1093, opened 2022-08-19
Additional local storage tests

lukastruemper:
When calling to_json on LocalStorage transformations, I saw that it did not add the chosen array to the "array" field. I first thought it was a bug and wrote some tests to check but then realized this is by intention when no "array" was provided. Anyway, I think it's good to add the additional tests I wrote


lukastruemper:
I think, I just found one bug

tbennun (approved):

PR #1094, opened 2022-08-23
OTFMapFusion: Support for WCR, nested maps & connector renaming fixed

lukastruemper:
This PR extends and fixes OTFMapFusion

- Fusion of maps with WCR in second map
- Fusion of maps with NestedSDFGs
- Renaming of symbols to avoid symbol collision of map parameters
- Renaming of symbols in NestedSDFG to avoid collision with map parameters

### DIFF
--- a/dace/transformation/dataflow/otf_map_fusion.py
+++ b/dace/transformation/dataflow/otf_map_fusion.py
@@ -140,23 +140,15 @@ class OTFMapFusion(transformation.SingleStateTransformation):
         # c. Add edges for in access nodes of first map to second map
         connector_mapping = {}
         for edge in graph.in_edges(first_map_entry):
-            old_in_connector = edge.dst_conn
-            if not edge.data is None and not edge.data.data is None:
-                suffix = edge.data.data
-            else:
-                suffix = "OTF"
-
-            new_in_connector = "IN_" + suffix
+            new_in_connector = self.second_map_entry.next_connector(edge.dst_conn[3:])
+            new_in_connector = "IN_" + new_in_connector
             if not self.second_map_entry.add_in_connector(new_in_connector):
-                for n in range(2, 65536):
-                    new_in_connector = "IN_" + suffix + "_" + str(n)
-                    if self.second_map_entry.add_in_connector(new_in_connector):
-                        break
+                raise ValueError("Failed to add new in connector")
 
             memlet = copy.deepcopy(edge.data)
             graph.add_edge(edge.src, edge.src_conn, self.second_map_entry, new_in_connector, memlet)
 
-            connector_mapping[old_in_connector] = new_in_connector
+            connector_mapping[edge.dst_conn] = new_in_connector
 
         # Phase 2: Memlet matching
         # Collect memlet-array map
--- a/dace/transformation/dataflow/otf_map_fusion.py
+++ b/dace/transformation/dataflow/otf_map_fusion.py
@@ -13,7 +13,10 @@ from dace.sdfg import utils as sdutil
 from dace.subsets import Range
 from dace.transformation import transformation
 from dace import data as dt
+from dace import dtypes
 from dace import symbolic
+
+from dace.frontend.operations import detect_reduction_type
 
 
 class OTFMapFusion(transformation.SingleStateTransformation):
@@ -24,10 +27,6 @@ class OTFMapFusion(transformation.SingleStateTransformation):
     first_map_exit = transformation.PatternNode(nds.ExitNode)
     array = transformation.PatternNode(nds.AccessNode)
     second_map_entry = transformation.PatternNode(nds.EntryNode)
-
-    @staticmethod
-    def annotates_memlets():
-        return False
 
     @classmethod
     def expressions(cls):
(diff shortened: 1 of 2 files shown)


dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -50,44 +49,93 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
+                    return False
 
-            write_memlets[array] = memlet
+                length = symbolic.evaluate(e - b, symbols=sdfg.constants)
+                if length > 1:
+                    return False
 
-        # Memlets access must be inferable
-        first_map_entry = graph.entry_node(self.first_map_exit)
+            # WCR must have identity zero. Otherwise, we must specifically initialize the temporary data
```
tbennun, line 74: that is also the case for zero identity, but you use setzero. Please note in the code that you use `setzero` below.

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -50,44 +49,93 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
-                return False
+            # Only fuse scalars
+            for access in memlet.subset:
+                b, e, s = access
+                if s > 1:
+                    return False
+
+                length = symbolic.evaluate(e - b, symbols=sdfg.constants)
+                if length > 1:
+                    return False
```
tbennun, line 110: same as above.

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -96,102 +144,185 @@ def apply(self, graph: SDFGState, sdfg: SDFG):
...
             else:
-                raise ValueError("Failed to connect")
+                suffix = "OTF"
+
+            new_in_connector = "IN_" + suffix
+            if not self.second_map_entry.add_in_connector(new_in_connector):
+                for n in range(2, 65536):
+                    new_in_connector = "IN_" + suffix + "_" + str(n)
+                    if self.second_map_entry.add_in_connector(new_in_connector):
+                        break
```
tbennun, line 166: I still don't understand why you do not use `next_connector`

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -96,102 +144,185 @@ def apply(self, graph: SDFGState, sdfg: SDFG):
...
+                    node.map.params = inner_map_params
+
+                    # Substitue range definitions by symbols for second map
+                    ranges = []
+                    for access in node.map.range:
+                        b, e, s = access
+
+                        b = symbolic.pystr_to_symbolic(b)
+                        e = symbolic.pystr_to_symbolic(e)
+                        s = symbolic.pystr_to_symbolic(s)
```
tbennun, line 260: they should already be symbolic, no?

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -96,102 +144,185 @@ def apply(self, graph: SDFGState, sdfg: SDFG):
...
+
+                        b = b.subs(param_mapping)
+                        e = e.subs(param_mapping)
+                        s = s.subs(param_mapping)
+
+                        ranges.append((str(b), str(e), str(s)))
```
tbennun, line 266: oh no, please do not use strings if you do not have to

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -96,102 +144,185 @@ def apply(self, graph: SDFGState, sdfg: SDFG):
...
+                            b, e, s = access
+                            b = symbolic.pystr_to_symbolic(b)
+                            e = symbolic.pystr_to_symbolic(e)
+                            s = symbolic.pystr_to_symbolic(s)
+
+                            b = b.subs(scope_subs)
+                            e = e.subs(scope_subs)
+                            s = s.subs(scope_subs)
+
+                            ranges.append((str(b), str(e), str(s)))
```
tbennun, line 286: same

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -96,102 +144,185 @@ def apply(self, graph: SDFGState, sdfg: SDFG):
...
+                                                   self.second_map_entry.map.params, in_accesses)
+
+                # Copy actual content
+                first_map_inner_nodes = self._copy_first_map_contents(sdfg, graph, first_map_entry)
+
+                # Prepare: Symbol collision
+                # Second map defines symbols (params) and becomes the new outer scope of the copied nodes.
+                # Thus, newly introduced symbols in this subgraph must be renamed first to avoid conflicts.
+                # Note symbols of first map must be re-named according to the param mapping.
+                for node in first_map_inner_nodes:
```
tbennun, line 232: this won't work for multi-level nested maps, right?

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -96,102 +144,185 @@ def apply(self, graph: SDFGState, sdfg: SDFG):
...
 
-                            for param, sub in param_subs[i].items():
-                                b = b.subs(param, sub)
-                                e = e.subs(param, sub)
-                                s = s.subs(param, sub)
+                            b = b.subs(param_mapping)
+                            e = e.subs(param_mapping)
+                            s = s.subs(param_mapping)
 
                             ranges.append((str(b), str(e), str(s)))
```
tbennun, line 327: same

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -322,27 +454,40 @@ def solve(first_params, write_accesses, second_params, read_accesses):
             # Step is constant
             assert (s0 - s1) == 0
 
+            b_eq = sympy.Eq(b0, b1)
+            e_eq = sympy.Eq(e0, e1)
```
tbennun, line 458: used outside the `if`?

tbennun (changes requested):
I'm a bit worried about the symbol replacement

dace/transformation/dataflow/otf_map_fusion.py
lukastruemper: Yeah, super annoying. The solution is awkward, but calling "while add_in_connector == False: new_connector = next_connector(new_connector)" was causing problems. I couldn't debug them, but can try another time. I know that it occurs in hdiff when fusing all maps

dace/transformation/dataflow/otf_map_fusion.py
lukastruemper: No, basically everyhting nested, also no states, nested SDFGs etc. Here I would need your help to generalize it. There must be a helper function which could replace symbols in subgraphs?

lukastruemper:
> I'm a bit worried about the symbol replacement

I addressed the simple problems now, but as we discussed, the symbol replacement needs to be handled better. Maybe you could show me how to make a nice (and more general) solution?

dace/transformation/dataflow/otf_map_fusion.py
lukastruemper: I will look into this separately

dace/transformation/dataflow/otf_map_fusion.py
tbennun: `SubgraphView.replace` is your friend

dace/sdfg/replace.py
```diff
@@ -60,7 +60,10 @@ def replace_dict(subgraph: 'dace.sdfg.state.StateGraphView',
 
     # Replace in node properties
     for node in subgraph.nodes():
-        replace_properties_dict(node, repl, symrepl)
+        if isinstance(node, dace.nodes.NestedSDFG):
+            node.sdfg.replace_dict(repl, symrepl)
```
lukastruemper, line 64: @tbennun This causes many other tests to fail, is this a bad idea?

dace/sdfg/replace.py
```diff
@@ -148,6 +151,14 @@ def replace_properties_dict(node: Any,
                     propval[symname] = symbolic.pystr_to_symbolic(str(sym_mapping)).subs(symrepl)
                 except AttributeError:  # If the symbolified value has no subs
                     pass
+        elif isinstance(propclass, properties.ListProperty) and pname == 'params':
```
lukastruemper, line 154: @tbennun This seems to be a bug: When renaming a parameter of a map, the parameters gets renamed correctly in the scope of the map (memlets, etc) but doesn't get renamed in the map definition.

If I add these lines, many of our tests start to break

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -50,44 +49,98 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
+                return False
+            first_memlets[memlet.data] = memlet
+
+        second_memlets = {}
         for edge in graph.out_edges(self.second_map_entry):
-            read_memlet = edge.data
-            if read_memlet.data not in write_memlets:
+            memlet = edge.data
+
+            if not memlet.data in first_memlets:
```
tbennun, line 93: same as above

tbennun (changes requested):
Only minor comments remain, otherwise LGTM!

dace/transformation/dataflow/otf_map_fusion.py
```diff
@@ -50,44 +49,98 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
+                red_type = detect_reduction_type(memlet.wcr)
+                if red_type == dtypes.ReductionType.Custom:
+                    return False
 
-        # Memlets access must be inferable
-        first_map_entry = graph.entry_node(self.first_map_exit)
+                dtype = sdfg.arrays[memlet.data].dtype
+                identity = dtypes.reduction_identity(dtype, red_type)
+                if identity != 0:
+                    return False
```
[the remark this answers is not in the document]
lukastruemper: Yes, otherwise we cannot use the setzero on the temporary buffer for initialization.

lukastruemper:
> Only minor comments remain, otherwise LGTM!

Pushed the fixes

dace/transformation/dataflow/otf_map_fusion.py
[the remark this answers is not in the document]
lukastruemper: In an OTF v2, we could insert a NestedSDFG which initializes the temporary buffer first

tbennun (approved):
LGTM!

PR #1097, opened 2022-08-27
Fix multidimensional transpose

tbennun:
Including 1D "transposition", which does not do anything


alexnick83 (approved):

acalotoiu (approved):
Looks good to me.

PR #1099, opened 2022-08-29
Add comparison methods to StringLiteral

### LINKED ISSUE #1100: StringLiteral introduces API incompatibility
**Describe the bug**
Between versions 0.13.3 and 0.14 DaCe seems to have changed what it stores the the SDFG in a backward incompatible way. Where before it stored a `str`, it now stores a `StringLiteral`. This is fine, but the latter does not implement all the methods that the builtin `str` supports.

GT4Py has automatically pulled in dace v0.14, and it expects to be able to use `<` on string inputs it gets from DaCe. I will make a change there for now to work around the issue (calling `str(...)` on the `StringLiteral` to get back that value), but longer term `StringLiteral` should support these.

jdahm:
`StringLiteral` is missing comparison methods to make it API compatible with `str`.
This adds the necessary methods.

Resolves #1100.


jdahm:
Issue: #1100

tbennun (approved):
LGTM, thanks!

tbennun:
@jdahm post-merge thought: what if the other object is also a StringLiteral?

jdahm:
Good thought! Yes this is the case I meant to cover. What I merged is a "bug". I'll add a commit and open another PR.

PR #1106, opened 2022-09-06
LIKWID Instrumentation: GPU Support

lukastruemper:


samples/instrumentation/GPU_BASE.txt
```diff
@@ -0,0 +1,18 @@
+SHORT GPU Targets
```
tbennun, line 1: what are those two files? can they be removed?

tbennun (changes requested):
Please remove the two extra files, otherwise ok!

tbennun (approved):

PR #1112, opened 2022-09-20
Ssa

alexander-hammett:
Adds SSA conversion to the Python frontend that handles redefined and conditionally defined variables. The resulting phi nodes are implemented via creating references and assigning to them.


tbennun:
@alexander-hammett is this still in progress? many tests are failing and there is no config option to disable it

PR #1114, opened 2022-09-28
Transformation Fixes Round 2

alexnick83:
Assortment of fixes in transformations and passes:
- [x] Constant propagation when the start-state is a scope's guard instead of a dedicated initialization state.


tbennun (approved):
LGTM

PR #1119, opened 2022-10-07
PyObject Callbacks

tbennun:
Callbacks can return arbitrary pyobjects, which are only reusable in other callbacks.
Also includes bugfixes for callbacks and return values, as well as tests.


acalotoiu (approved):
LGTM. I appreciate adding tests to ensure the new features are covered from the start!

PR #1123, opened 2022-10-17
Generic slice view removal transformation

tbennun:
Supports squeezing and expansion ("unsqueezing"), as well as arbitrary subsets.


dace/sdfg/propagation.py
```diff
@@ -1202,21 +1202,50 @@ def _propagate_node(dfg_state, node):
...
+
+    # Otherwise, find other data container
+    result = copy.deepcopy(e.data)
+    if dst:
+        node = mpath[-1].dst
+    else:
+        node = mpath[0].src
+
+    # Fix memlet fields
+    result.data = node.data
```
alexnick83, line 1242: I am a bit confused about the semantics here. Is it implied that if we reach this part of the code, then the memlet is described (erroneously?) in the exact opposite way compared to the desired "aligned" memlet?

dace/transformation/auto/auto_optimize.py
```diff
@@ -334,7 +334,7 @@ def find_fast_library(device: dtypes.DeviceType) -> List[str]:
         if openblas.OpenBLAS.is_installed():
             result.append('OpenBLAS')
 
-        return result + ['OpenMP', 'pure']
+        return result + ['pure']
```
alexnick83, line 337: My understanding is that we have an issue with the OpenMP ReduceNode expansion. However, this change will basically disable in the auto-optimizer all OpenMP expansions. We probably don't have many OpenMP expansions now, but perhaps we should amend this in the future.

dace/codegen/targets/cpp.py
```diff
@@ -28,6 +28,9 @@
 from dace.sdfg import SDFG, is_devicelevel_gpu, SDFGState
 from dace.codegen.targets import fpga
 
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
tbennun: this is a standard `typing.TYPE_CHECKING` condition that is used by vscode and pycharm (among others incl. mypy) to use for string imports. This is the recommended Python way :tm:

alexnick83 (approved):

dace/sdfg/propagation.py
tbennun: I think we should change memlets accordingly

dace/transformation/auto/auto_optimize.py
tbennun: true. right now the only OpenMP implementation is this reduce node, and openmp is not well supported across compilers :( we had several complaints about this failing.

PR #1124, opened 2022-10-18
Passes Interface Improvements

phschaad:
- [x] Provide an interface to query all available passes
- [x] Group passes and pipelines into categories
- [x] Remove deprecated functions and transformations that were surpassed by passes
- [x] Make passes and pipelines use properties rather than (just) dataclasses
- [x] Rename the Pipeline map scope to `PipelineScope` to avoid namespace clashes with the new pass pipelines

This serves as a prerequisite for https://github.com/spcl/dace-vscode/pull/203


dace/transformation/pass_pipeline.py
```diff
@@ -272,12 +322,18 @@ class Pipeline(Pass):
...
 
     """
-    passes: List[Pass]
+
+    _passes = properties.ListProperty(element_type=Pass,
+                                      default=[],
+                                      desc='List of passes that this pipeline contains')
+    _pass_names = properties.SetProperty(element_type=str,
+                                         default=set(),
+                                         desc='List of passe names that this pipeline contains')
```
tbennun, line 342: why leading underscore??

tbennun, line 342:
```suggestion
    _passes = properties.ListProperty(element_type=Pass,
                                      default=[],
                                      desc='List of passes that this pipeline contains')
    _pass_names = properties.SetProperty(element_type=str,
                                         default=set(),
                                         desc='List of pass names that this pipeline contains')
```

dace/transformation/pass_pipeline.py
```diff
@@ -272,12 +322,18 @@ class Pipeline(Pass):
...
 
     def __init__(self, passes: List[Pass]):
-        self.passes = []
-        self.pass_names = set(type(p).__name__ for p in passes)
-        self.passes.extend(passes)
+        self._passes = []
+        self._pass_names = set(type(p).__name__ for p in passes)
```
tbennun, line 346: why make this a property? it is derived from passes

dace/transformation/passes/analysis.py
```diff
@@ -2,16 +2,19 @@
...
 import networkx as nx
 
 
+@properties.make_properties
 class StateReachability(ppl.Pass):
     """
     Evaluates state reachability (which other states can be executed after each state).
     """
 
+    CATEGORY: str = 'Analysis'
```
tbennun, line 16: I hope you don't show these in the UI...

dace/transformation/passes/constant_propagation.py
```diff
@@ -17,14 +17,17 @@ class _UnknownValue:
...
     Propagates constants and symbols that were assigned to one value forward through the SDFG, reducing
     the number of overall symbols.
     """
 
-    recursive: bool = True
-    progress: Optional[bool] = None
+    CATEGORY: str = 'Simplification'
+
+    recursive = properties.Property(dtype=bool, default=True, desc='Propagagte recursively through nested SDFGs')
+    progress = properties.Property(dtype=bool, default=None, allow_none=True, optional=True, desc='Show progress')
```
tbennun, line 30: what's optional=True?

dace/transformation/passes/fusion_inline.py
```diff
@@ -6,18 +6,26 @@
...
-    permissive: bool = False  #: If True, ignores some race conditions checks
-    progress: Optional[bool] = None  #: Whether to print progress, or None for default (print after 5 seconds)
+
+    CATEGORY: str = 'Simplification'
+
+    permissive = properties.Property(dtype=bool, default=False, desc='If True, ignores some race conditions checks.')
+    progress = properties.Property(dtype=bool,
+                                   default=None,
+                                   allow_none=True,
+                                   optional=True,
```
tbennun, line 27: same

dace/transformation/passes/fusion_inline.py
```diff
@@ -41,13 +49,21 @@ def report(self, pass_retval: int) -> str:
...
-    progress: Optional[bool] = None  #: Whether to print progress, or None for default (print after 5 seconds)
-    multistate: bool = True  #: If True, include multi-state inlining
+
+    CATEGORY: str = 'Simplification'
+
+    permissive = properties.Property(dtype=bool, default=False, desc='If True, ignores some checks on inlining.')
+    progress = properties.Property(dtype=bool,
+                                   default=None,
+                                   allow_none=True,
+                                   optional=True,
```
tbennun, line 64: same

dace/transformation/passes/util.py
```diff
@@ -0,0 +1,25 @@
...
+    Returns all available passes and pass pipelines as a set by recursing over Pass subclasses. 
+    :param all_passes: Include all passes, e.g., including PatternTransformation and other base passes.
+    """
+    full_pass_set = Pass.subclasses_recursive()
+
+    if not all_passes:
+        reduced_pass_set = set()
+        for p in full_pass_set:
+            if not issubclass(p, (PatternTransformation, SubgraphTransformation)) and not p in (
+                TransformationBase, VisitorPass, StatePass, Pipeline, FixedPointPipeline, ScopePass
```
tbennun, line 20: maybe just check the category to not be "hidden" (or "helper" as you call it here)

tbennun (changes requested):

dace/transformation/passes/analysis.py
phschaad: No they're excluded :-)

dace/transformation/pass_pipeline.py
phschaad: This way they're automatically hidden from the UI. I see that they're not in that sense 'private', however, so we can change it to no underscore and instead explicitly hide them in the UI, like the `CATEGORY`.

dace/transformation/pass_pipeline.py
phschaad: If it does not have a leading underscore it will have to be a property, otherwise the `properties` system complains.

dace/transformation/passes/util.py
phschaad: Good point.

dace/transformation/pass_pipeline.py
tbennun: I think the `(debug)` category is already hidden

dace/transformation/pass_pipeline.py
tbennun: not the same issue. Why is _pass_names a property? Should not be.

dace/transformation/passes/constant_propagation.py
tbennun: `allow_none=True` is what Optional equates to

tbennun (approved):
LGTM if tests pass

PR #1126, opened 2022-10-22
Documentation v2

### LINKED ISSUE #1011: Getting started tutorial needs updating
**Compiled SDFG does not require keyword args**

In the [Getting Started tutorial](https://nbviewer.org/github/spcl/dace/blob/master/tutorials/getting_started.ipynb), just before code cell 11, the notebook states:

"A compiled SDFG, however, has to be invoked like an SDFG, with keyword arguments only:"

This does not seem to be correct; I can run the line fine with `A` as a positional argument.

**Final SDFG does not match latest dace output**

In the [Explicit Dataflow example](https://nbviewer.org/github/spcl/dace/blob/master/tutorials/getting_started.ipynb#Explicit-Dataflow) the outputted graph does not match that produced using Dace 0.13.1 and Python 3.10.4:

[sse_sigma.txt](https://github.com/spcl/dace/files/8805690/sse_sigma.txt)

tbennun:
Second installment of documentation entries.

- [x] Update getting started tutorial with easier CompiledSDFG syntax (Fixes Issue #1011)
- [x] Audit entries for correctness w.r.t. code
    - [x]  Python frontend features / support
- [x] Python frontend preprocessing
- [x] Optimizing SDFGs with the SDFG API
- [x] Schedule Tree representation
- [x] Extending DaCe
    - [x] Working with symbolic expressions
    - [x] New library nodes
    - [x] New frontend
    - [x] DSLs and SDFGConvertible
    - [x] New Backends
    - [x] Writing your own instrumentation provider


acalotoiu (approved):
This is really nice - and should be required reading before starting to develop as it should hopefully help preempt a lot of design issues later on

doc/frontend/pysupport.rst
pratyai, line 1: I think small examples attached with the features would make things more immediately accessible. But if it's already illustrated in some other sibling page, ignore this comment.

doc/optimization/guidelines.rst
```diff
@@ -0,0 +1,84 @@
+.. _opt_guidelines:
+
+Optimization Guidelines
+=======================
+
+This page collects the high-level principles that underlie effective DaCe
+optimization. The other pages in :ref:`optimization` show *how* to apply
+specific transformations; this one is about *what* to aim for.
+
+Minimize data movement
```
pratyai, line 10: Do these also fall under "Profile, do not guess" rule? How well the current toolset help with these guidelines?

doc/sdfg/schedule_tree.rst
pratyai, line 1: This is nice :)

doc/sdfg/schedule_tree.rst
```diff
@@ -0,0 +1,177 @@
...
+  or :ref:`passes <pass_pipeline>`, most of which are written against the SDFG IR.
+* You need to **generate code**: code generation operates on SDFGs (see
+  :ref:`codegen`). To generate code from a schedule tree, convert it back to an SDFG
+  first.
+
+
+Converting Between SDFG and Schedule Tree
+-----------------------------------------
+
+Every SDFG can be converted to a schedule tree, and a schedule tree can be converted
```
pratyai, line 97: how can a general graph that's not a tree be converted to a tree? Are we duplicating the downstream states? What if there is a cycle?

doc/extensions/symbolic.rst
pratyai, line 1: I (personally) am not convinced that keeping a distinction between scalars and symbol at the "user-level" is the right call. A (possibly expensive, but correct) analysis should be able to tell when a scalar can be used symbolically or not.

tbennun: This is not user level. This is under the developers’ guide under extensions

doc/sdfg/schedule_tree.rst
tbennun: The tree has goto nodes and loop scopes for general state machines

doc/optimization/guidelines.rst
tbennun: I would rather not replicate the text between both sections.

doc/frontend/pysupport.rst
tbennun: That would be very lengthy and most of it is covered by the Python official documentation

PR #1136, opened 2022-11-03
Populating other-subset in edge.try_initialize

alexnick83:
This PR attempts to address issues that occur in InlineSDFG (unsqueeze memlet) due to missing other-subsets, especially when those subsets should be used for unsqueezing the memlets.

### DIFF
--- a/dace/memlet.py
+++ b/dace/memlet.py
@@ -13,6 +13,7 @@ from dace.frontend.operations import detect_reduction_type
 from dace.frontend.python.astutils import unparse
 from dace.properties import (Property, make_properties, DataProperty, SubsetProperty, SymbolicProperty,
                              DebugInfoProperty, LambdaProperty)
+from dace.sdfg import nodes
 
 if TYPE_CHECKING:
     import dace.sdfg.graph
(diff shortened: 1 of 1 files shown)


dace/memlet.py
```diff
@@ -390,6 +391,15 @@ def try_initialize(self, sdfg: 'dace.sdfg.SDFG', state: 'dace.sdfg.SDFGState',
         if (self.data is not None and self.subset is None):
             self.subset = subsets.Range.from_array(sdfg.arrays[self.data])
 
+        other_data = None
+        if self._is_data_src and isinstance(path[-1].dst, nodes.AccessNode):
+            other_data = path[-1].dst.data
+        elif isinstance(path[0].src, nodes.AccessNode):
```
tbennun, line 397: maybe you need `not self._is_data_src` here? @alexnick83

alexnick83: No, I don't think so. Here, I am trying to find the "other data," i.e., the path's destination if the memlet is a data source or the path's source if the memlet is not a data source. There is also the check that the source/destination is an access node so that we avoid trying to extract data ranges out of code nodes. Obviously, this causes a lot of issues, and it must be investigated further.

tbennun: what I mean is that the elif would still enter even if is_data_src is true here (but the destination is a code node), so it's just a boolean inversion common issue with the branch.

alexnick83: 🦌, thanks for pointing out

PR #1144, opened 2022-11-08
Fixed problem with Intel FPGA and long long

sarahtr:


dace/codegen/CMakeLists.txt
```diff
@@ -26,7 +26,8 @@ set(DACE_XILINX_ENABLE_DEBUGGING OFF CACHE STRING "Inject debugging cores to ker
 set(DACE_INTELFPGA_MODE "simulation" CACHE STRING "Type of compilation/execution [emulator/simulator/hardare].")
 set(DACE_INTELFPGA_HOST_FLAGS "" CACHE STRING "Extra flags to host compiler.")
 set(DACE_INTELFPGA_KERNEL_FLAGS "" CACHE STRING "Extra flags to kernel compiler.")
-set(DACE_INTELFPGA_TARGET_BOARD "a10gx" CACHE STRING "Target FPGA board.")
+#set(DACE_INTELFPGA_TARGET_BOARD "a10gx" CACHE STRING "Target FPGA board.")
+set(DACE_INTELFPGA_TARGET_BOARD "p520_max_sg280l" CACHE STRING "Target FPGA board.")
```
TizianoDeMatteis, line 30: Please revert this change.

We have the other BSP (p520) in our development server, but, by default, intel compiler is shipped with the BSP for the Arria 10 FPGA (a10gx). This is why the Unit test failed.

You can edit your .dace.conf to use the p520_max_sg280l BSP locally

dace/codegen/targets/fpga.py
```diff
@@ -2002,6 +2002,13 @@ def _generate_MapEntry(self, sdfg, dfg, state_id, node, function_stream, callsit
                         # This could occur, for example, if they are defined in terms of other symbols, which
                         # is the case in a tiled map
                         pass
+                    
+                    # To enforce opencl type long instead of c type long long for intel fpga
+                    v = dace.config.Config.get("compiler", "fpga", "vendor")
+                    if v.casefold() == 'intel_fpga'.casefold():
+                        if loop_var_type.__contains__("long long"):
```
TizianoDeMatteis, line 2009: Better to use `substring in string` for the check.
You can also apply the replace directly, it will not change anything if it does not found the substring

dace/codegen/targets/intel_fpga.py
```diff
@@ -562,6 +562,11 @@ def generate_module(self, sdfg, state, kernel_name, module_name, subgraph, param
             if isinstance(p, dace.data.View):
                 continue
             arg = self.make_kernel_argument(p, pname, is_output, True)
+            
+            #change c type long long to opencl type long
+            if arg.__contains__("long long"):
```
TizianoDeMatteis, line 567: same comment as before.

Also note that arg can be none (see the if two lines below)

dace/codegen/targets/intel_fpga.py
```diff
@@ -753,7 +758,11 @@ def generate_nsdfg_arguments(self, sdfg, dfg, state, node):
                 continue
             desc = sdfg.arrays[in_memlet.data]
             ptrname = cpp.ptr(in_memlet.data, desc, sdfg, self._frame)
-            defined_type, defined_ctype = self._dispatcher.defined_vars.get(ptrname, 1)
+            defined_type, defined_ctype = self._dispatcher.defined_vars.get(ptrname, 1) #Ctype -> dacetype
+
+            #change c type long long to opencl type long
```
TizianoDeMatteis, line 763: Same comment as before

dace/codegen/targets/intel_fpga.py
```diff
@@ -753,7 +758,11 @@ def generate_nsdfg_arguments(self, sdfg, dfg, state, node):
                 continue
             desc = sdfg.arrays[in_memlet.data]
             ptrname = cpp.ptr(in_memlet.data, desc, sdfg, self._frame)
-            defined_type, defined_ctype = self._dispatcher.defined_vars.get(ptrname, 1)
+            defined_type, defined_ctype = self._dispatcher.defined_vars.get(ptrname, 1) #Ctype -> dacetype
```
TizianoDeMatteis, line 761: I didn't get the comment, could you explain it?

dace/config_schema.yml
```diff
@@ -557,7 +557,8 @@ required:
 
                     board:
                         type: str
-                        default: a10gx
+                        #default: a10gx
+                        default: p520_max_sg280l
```
TizianoDeMatteis, line 561: Revert this change

tests/npbench/polybench/simple_add_test.py
```diff
@@ -0,0 +1,100 @@
+## copied jacobi 1d and simplified the kernel
```
TizianoDeMatteis, line 1: I would remove change a bit this test:
- you can move it under the `tests/fpga` folder, and give a name more related to the thing that wants to test (something like "long_long_opencl_test.py")
- I would add a brief comment on top explaining what's the purpose of the test
- I would test only FPGA, and in particular intel_fpga. So no CPU/GPU test.  To run only for Intel FPGA you can mark the test with `@intel_fpga_test()` rather than `@fpga_test`
- Cleanup the code, removing CPU/GPU related code
- no need to have all the sizes, just use and pick one
- remove unused imports

TizianoDeMatteis (changes requested):
Overall looks good, but there are things to fix.

dace/codegen/targets/intel_fpga.py
sarahtr: This was a debugging comment I forgott to remove. I used it to remember where to change the type.

tests/fpga/long_long_opencl_test.py
```diff
@@ -69,17 +59,8 @@ def run_simple_add(device_type: dace.dtypes.DeviceType):
...
-
-
-@pytest.mark.gpu
-def test_gpu():
-    run_simple_add(dace.dtypes.DeviceType.GPU)
-
-
-
-@fpga_test(assert_ii_1=False)
+#@fpga_test(assert_ii_1=False)
```
TizianoDeMatteis, line 62: remove

tests/fpga/long_long_opencl_test.py
```diff
@@ -1,19 +1,13 @@
...
+from dace.fpga_testing import intel_fpga_test
 from dace.transformation.interstate import FPGATransformSDFG, InlineSDFG
-from dace.transformation.dataflow import StreamingMemory, StreamingComposition
-from dace.transformation.auto.auto_optimize import auto_optimize
-from dace.config import set_temporary
 
-# Dataset sizes
-# TSTEPS, N
-sizes = {"mini": (30), "small": (120), "medium": (400), "large": (2000), "extra-large": (4000)}
+#N
```
TizianoDeMatteis, line 10: Remove

tests/fpga/long_long_opencl_test.py
```diff
@@ -1,19 +1,13 @@
-## copied jacobi 1d and simplified the kernel
+## simple test case to check that for the intel fpga the openCL type long gets used instead of C type long long
```
TizianoDeMatteis, line 1: Starts with capital letter (and use multi line comments with `'''` if it does not fit in a line with 120 chars)

tests/fpga/long_long_opencl_test.py
```diff
@@ -39,17 +33,13 @@ def run_simple_add(device_type: dace.dtypes.DeviceType):
...
     B_ref = np.copy(B)
 
-    if device_type in {dace.dtypes.DeviceType.CPU, dace.dtypes.DeviceType.GPU}:
-        # Parse the SDFG and apply auto-opt
-        sdfg = simple_add_kernel.to_sdfg()
-        sdfg = auto_optimize(sdfg, device_type)
-        sdfg(A, B, N=N)
-    elif device_type == dace.dtypes.DeviceType.FPGA:
+
+    if device_type == dace.dtypes.DeviceType.FPGA:
```
TizianoDeMatteis, line 42: You can also remove this check on the Device Type, and the use of `device_type` at all. This will be only executed for Intel FPGA anyway

TizianoDeMatteis (changes requested):
Just some minor comments left.

On top of that, please use `yapf` to format your code. The DaCe yapf configuration can be found here: https://github.com/spcl/dace/blob/master/.style.yapf. We normally use version `0.31.0`

You can either configure VScode to yapf automatically when you save (in that case you have to configure the style), or do it from a shell with `yapf -i path/to/file`

tests/fpga/long_long_opencl_test.py
```diff
@@ -0,0 +1,73 @@
...
+    # Parse SDFG and apply FPGA friendly optimization
+    sdfg = simple_add_kernel.to_sdfg(simplify=True)
+    applied = sdfg.apply_transformations([FPGATransformSDFG])
+    assert applied == 1
+
+    # Use FPGA Expansion for lib nodes, and expand them to enable further optimizations
+    from dace.libraries.blas import Dot
+    Dot.default_implementation = "FPGA_PartialSums"
+    sdfg.expand_library_nodes()
+    sdfg.apply_transformations_repeated([InlineSDFG], print_report=True)
```
TizianoDeMatteis, line 49: This may be uncessary. Please remove if there is no library node in this test.

tests/npbench/polybench/jacobi_1d_test.py
```diff
@@ -85,7 +85,7 @@ def test_gpu():
     run_jacobi_1d(dace.dtypes.DeviceType.GPU)
 
 
-@pytest.mark.skip(reason="Intel FPGA missing support for long long")
+#@pytest.mark.skip(reason="Intel FPGA missing support for long long")
```
TizianoDeMatteis, line 88: Just delete it

TizianoDeMatteis (changes requested):
Please check the other npbench tests that were not working for IntelFPGA because of the `long long` issue, if now they work with your fix

TizianoDeMatteis (approved):

PR #1158, opened 2022-11-24
GEMM: CSR expansion support

lukastruemper:


tbennun:
Probably not the best way to go imo

lukastruemper:
> Probably not the best way to go imo. Maybe convert the matmul node to a csrmm node first?

Yeah, could be safer to use if you can convert to a CSR node, but I would need help on that

alexnick83:
> > Probably not the best way to go imo. Maybe convert the matmul node to a csrmm node first?
> 
> Yeah, could be safer to use if you can convert to a CSR node, but I would need help on that

There is a prototype CSRMM node here: https://github.com/spcl/dace/blob/soap-merge-master/dace/libraries/sparse/nodes/csrmm.py
I don't have an ETA for merging into master though.

PR #1177, opened 2022-12-08
auto_optimize now properly chooses GPUAuto expansion for reduce nodes

hodelcl:
This PR adds very little code to `auto_optimize.py`. The added code ensures that the `GPUAutoExpansion` gets used for reduce nodes, when `auto_optimize` is used to optimize the produced SDFG.

Before, it would always choose `CUDA (device)` even though the `GPUAuto` expansion is higher in the `implementation_prio`.

### DIFF
--- a/dace/libraries/standard/reduction_planner.py
+++ b/dace/libraries/standard/reduction_planner.py
@@ -138,8 +138,6 @@ def get_reduction_schedule(in_array: Array,
     :return: ReductionSchedule object that descibes the GPU schedule used to perform the reduction
     """
 
-    # use_mini_warps = False
-
     @dataclasses.dataclass
     class ReductionSchedule:
         grid: List[Size]  #: dimension of the grid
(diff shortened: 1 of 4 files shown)


dace/libraries/standard/reduction_planner.py
```diff
@@ -99,7 +100,7 @@ def simplify_input(shape, strides, axes):
         r = rem[1]
         s = shape[rem[0]]
         for i in range(len(out_strides)):
-            if out_strides[i] > r:
+            if (out_strides[i] > r) == True:
                 out_strides[i] //= s
```
tbennun, line 104: are you sure that this is ok to do? if it's symbolic what would happen?

tbennun:
@hodelcl any update?

hodelcl:
I think I spotted the bug, but I am unsure about how to resolve it. It is in this line where we set the size of the grid:

https://github.com/spcl/dace/blob/262b8c1b2c3f447bda33ebaac7b6b65055e1726d/dace/libraries/standard/reduction_planner.py#L269

 If `shape[contiguous_dimension]` is an integer, this works fine, even if it is for example 4. Then this will generate `dim3(1, 1, 1)` for the grid size in the CUDA code. However, if `shape[contiguous_dimension]` is a symbol (e.g. `M`), this will generate this grid size: `dim3(int_ceil((M / 32), 1), 1, 1)`. And if `M` is again 4, then this will be `dim3(0, 1, 1)`, which does not work.

tbennun:
> I think I spotted the bug, but I am unsure about how to resolve it. It is in this line where we set the size of the grid:
> 
> https://github.com/spcl/dace/blob/262b8c1b2c3f447bda33ebaac7b6b65055e1726d/dace/libraries/standard/reduction_planner.py#L269
> 
> If `shape[contiguous_dimension]` is an integer, this works fine, even if it is for example 4. Then this will generate `dim3(1, 1, 1)` for the grid size in the CUDA code. However, if `shape[contiguous_dimension]` is a symbol (e.g. `M`), this will generate this grid size: `dim3(int_ceil((M / 32), 1), 1, 1)`. And if `M` is again 4, then this will be `dim3(0, 1, 1)`, which does not work.

Clearly then it should use ‘symbolic.int_ceil(…, 32)’ in the Python code. If it’s numeric it will evaluate and if not it will at least stay 1

hodelcl:
Thanks for the suggestion, it works now!

hodelcl:
@tbennun The PR is now ready for a review.

tbennun (approved):

PR #1179, opened 2022-12-10
fix to use data.find_new_name in _find_new_name

acalotoiu:
Replaces _find_new_name logic with data.find_new_name implementation.

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -27,6 +27,7 @@ from dace.sdfg.scope import ScopeTree
 from dace.sdfg.replace import replace, replace_properties, replace_properties_dict
 from dace.sdfg.validation import (InvalidSDFGError, validate_sdfg)
 from dace.config import Config
+from dace.data import find_new_name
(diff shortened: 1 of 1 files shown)


dace/sdfg/sdfg.py
```diff
@@ -27,6 +27,7 @@
 from dace.sdfg.replace import replace, replace_properties, replace_properties_dict
 from dace.sdfg.validation import (InvalidSDFGError, validate_sdfg)
 from dace.config import Config
+from dace.data import find_new_name
```
tbennun, line 30: do not import functions or classes from other modules (we are making this a requirement for future code and have yet to clean the past code). instead, use dt.find_new_name (if you look at line 25 it is already imported)

tbennun (changes requested):

dace/sdfg/sdfg.py
acalotoiu: Done.

tbennun (approved):

PR #1181, opened 2022-12-17
BatchedMatMul: MKL gemm_batch support

lukastruemper:


dace/libraries/blas/nodes/batched_matmul.py
```diff
@@ -84,6 +84,75 @@ class ExpandBatchedMatMulMKL(ExpandTransformation):
...
+        MKL_INT k_array[group_count] = {{ {K} }};
+        char transa[group_count] = {{ '{ta}' }};
+        char transb[group_count] = {{ '{tb}' }};
+        {dtype} alpha_array[group_count] = {{ {alpha} }};
+        {dtype} beta_array[group_count] = {{ {beta} }};
+        MKL_INT lda_array[group_count] = {{ {lda} }};
+        MKL_INT ldb_array[group_count] = {{ {ldb} }};
+        MKL_INT ldc_array[group_count] = {{ {ldc} }};
+
+        const {dtype}** A = new const {dtype}*[{BATCH}];
```
tbennun, line 132: why malloc in the middle of a potentially performance-critical code??

tbennun: btw that's the reason why we didn't implement it thus far.

tbennun (changes requested):

tbennun:
you can try adding the allocation to the init code (if it depends on input symbols or is constant) or the stack (if the batch size is known and small)

lukastruemper:
> you can try adding the allocation to the init code (if it depends on input symbols or is constant) or the stack (if the batch size is known and small)

I don't know how to move this part to the init part. For BERT with batch size 6-8, the old implementation is much slower than this version

tbennun:
I'll approve for now, but this should probably use some inflatable workspace.

PR #1193, opened 2023-02-17
AMD-related fixes and rocBLAS GEMM

tbennun:
- [x] Auto-detect whether existing GPU is HIP or NVIDIA
- [x] Fix HIP compilation issues and better detect GPU arch
- [x] Add rocBLAS implementation of GEMM
- [x] Allow external libraries to change streams for interoperability

### DIFF
--- a/dace/codegen/common.py
+++ b/dace/codegen/common.py
@@ -1,12 +1,15 @@
 # Copyright 2019-2022 ETH Zurich and the DaCe authors. All rights reserved.
 import ast
 from copy import deepcopy
-from dace import data, dtypes, sdfg as sd, symbolic
+import ctypes.util
+from dace import config, data, dtypes, sdfg as sd, symbolic
 from dace.sdfg import SDFG
 from dace.properties import CodeBlock
 from dace.codegen import cppunparse
 from functools import lru_cache
 from io import StringIO
+import os
+import subprocess
 from typing import List, Optional, Set, Union
 import warnings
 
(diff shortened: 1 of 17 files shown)


dace/codegen/common.py
```diff
@@ -94,3 +97,50 @@ def unparse_interstate_edge(code_ast: Union[ast.AST, str], sdfg: SDFG, symbols=N
...
+
+@lru_cache()
+def get_gpu_backend() -> str:
+    """
+    Returns the currently-selected GPU backend. If automatic,
+    will perform a series of checks to see if an NVIDIA device exists,
+    then if an AMD device exists, or fail.
+    Otherwise, chooses the configured backend in ``compiler.cuda.backend``.
+    """
+    backend: str = config.Config.get('compiler', 'cuda', 'backend')
```
alexnick83, line 110: Does it make sense to keep calling this `cuda` instead of `gpu`? I suppose that the CUDA code is "hipified," and that warrants keeping the name, but maybe it should be changed in the future.

dace/codegen/instrumentation/data/data_dump.py
```diff
@@ -1,204 +1,205 @@
...
+        if desc.storage == dtypes.StorageType.GPU_Global:
+            self._setup_gpu_runtime(sdfg, global_stream)
+            preamble, postamble, ptrname = self._generate_copy_to_device(node, desc, ptrname)
+
+        # Write code
+        inner_stream.write(preamble, sdfg, state_id, node_id)
+        inner_stream.write(
+            f'__state->serializer->restore({ptrname}, {cpp.sym2cpp(desc.total_size - desc.start_offset)}, '
+            f'"{node.data}", "{uuid}");\n', sdfg, state_id, node_id)
+        inner_stream.write(postamble, sdfg, state_id, node_id)
```
alexnick83, line 205: Did anything change in this file apart from adding `from dace.codegen import common` in line 8?

dace/codegen/targets/cpp.py
```diff
@@ -1350,6 +1348,7 @@ def synchronize_streams(sdfg, dfg, state_id, node, scope_exit, callsite_stream,
         # If we are the ones to remove the last terminator, release memory
         terminators.remove(scope_exit)
         if len(terminators) == 0:
+            backend = common.get_gpu_backend()
```
alexnick83, line 1351: Why isn't this (and line 1374) just replacing line 1331? Just curious; do we expect to somehow handle streams on different cards (and architectures) in the future?

dace/codegen/tools/get_hip_arch.cpp
```diff
@@ -17,9 +16,16 @@ int main(int argc, char **argv) {
...
-    ss << prop.major * 10 << prop.minor;
-    architectures.insert(ss.str());
+
+    // Find architecture, ignoring anything that succeeds a colon
+    void *colon = memchr(prop.gcnArchName, ':', 256);
+    if (colon)
+      *(char *)colon = '\0';
+
+    // Trim starting "gfx"
+    std::string str = prop.gcnArchName + 3;
```
alexnick83, line 26: These things are inevitable, but maybe check if the string starts with "gfx"?

alexnick83 (approved):

dace/codegen/instrumentation/data/data_dump.py
tbennun: Yes, it gets the backend with the new function

dace/codegen/common.py
tbennun: We should, it’s a bigger refactor PR though

dace/codegen/tools/get_hip_arch.cpp
tbennun: You’re right. I’ll keep the gfx here and remove its addition in the cmake file

dace/codegen/targets/cpp.py
tbennun: Since getting the architecture is now a more involved process that can fail and raise exceptions, this change only triggers it if a GPU is actually used

dace/codegen/tools/get_hip_arch.cpp
tbennun: Fixed

PR #1196, opened 2023-02-18
Fix integer division in Python frontend

tbennun:
Two serious symbolic issues were tracked to the following behavior:
* `symstr` printed divisions as the C++ versions, even when the code is intended to be read as Python
* The floor division operator was read by `pystr_to_symbolic` as `floor(a/b)`

This PR resolves this issue and adds an appropriate test.


dace/data.py
```diff
@@ -615,7 +616,8 @@ def validate(self):
             raise TypeError('Strides must be the same size as shape')
 
         if any(not isinstance(s, (int, symbolic.SymExpr, symbolic.symbol, symbolic.sympy.Basic)) for s in self.strides):
-            raise TypeError('Strides must be a list or tuple of integer ' 'values or symbols')
+            raise TypeError('Strides must be a list or tuple of integer '
```
alexnick83, line 619: Doesn't this fit in a single line?

dace/frontend/python/replacements.py
```diff
@@ -1904,7 +1909,9 @@ def _scalar_sym_binop(visitor: ProgramVisitor, sdfg: SDFG, state: SDFGState, lef
     ">=": sp.GreaterThan,
     "<=": sp.LessThan,
     ">": sp.StrictGreaterThan,
-    "<": sp.StrictLessThan
+    "<": sp.StrictLessThan,
+    # Boolean ops
```
alexnick83, line 1913: Duplicate comment from line 1902?

dace/libraries/blas/nodes/dot.py
```diff
@@ -544,9 +543,11 @@ def validate(self, sdfg, state):
                 desc_res = sdfg.arrays[e.data.data]
 
         if desc_x.dtype != desc_y.dtype:
-            raise TypeError("Data types of input operands must be equal: " f"{desc_x.dtype}, {desc_y.dtype}")
+            raise TypeError("Data types of input operands must be equal: "
```
alexnick83, line 546: Doesn't fit in the same line?

dace/libraries/blas/nodes/dot.py
```diff
@@ -544,9 +543,11 @@ def validate(self, sdfg, state):
...
+                            f"{desc_x.dtype}, {desc_y.dtype}")
         if desc_x.dtype.base_type != desc_res.dtype.base_type:
-            raise TypeError("Data types of input and output must be equal: " f"{desc_x.dtype}, {desc_res.dtype}")
+            raise TypeError("Data types of input and output must be equal: "
```
alexnick83, line 549: Same

dace/libraries/blas/nodes/ger.py
```diff
@@ -288,7 +287,8 @@ def validate(self, sdfg, state):
                 desc_y = sdfg.arrays[memlet.data]
 
         if size_a is None or size_x is None:
-            raise ValueError("Expected at least two inputs to Ger " "(matrix A and vector x)")
+            raise ValueError("Expected at least two inputs to Ger "
```
alexnick83, line 290: Same

dace/sdfg/propagation.py
```diff
@@ -970,7 +979,8 @@ def propagate_memlets_nested_sdfg(parent_sdfg, parent_state, nsdfg_node):
 
                     if inside_memlet.wcr is not None:
                         if (memlet.wcr is not None and memlet.wcr != inside_memlet.wcr):
-                            warnings.warn('Memlet appears with more than one' ' type of write-conflict resolution.')
+                            warnings.warn('Memlet appears with more than one'
```
alexnick83, line 982: Same line?

dace/sdfg/propagation.py
```diff
@@ -1013,7 +1023,8 @@ def propagate_memlets_nested_sdfg(parent_sdfg, parent_state, nsdfg_node):
                     # union of the ranges to merge the subsets.
                     if memlet.subset is not None:
                         if memlet.subset.dims() != subset.dims():
-                            raise ValueError('Cannot merge subset ranges ' 'of unequal dimension!')
+                            raise ValueError('Cannot merge subset ranges '
```
alexnick83, line 1026: Same

dace/sdfg/propagation.py
```diff
@@ -1314,7 +1325,8 @@ def propagate_memlet(dfg_state,
 
     if arr is None:
         if memlet.data not in sdfg.arrays:
-            raise KeyError('Data descriptor (Array, Stream) "%s" not defined ' 'in SDFG.' % memlet.data)
+            raise KeyError('Data descriptor (Array, Stream) "%s" not defined '
```
alexnick83, line 1328: Same

dace/symbolic.py
```diff
@@ -971,7 +971,8 @@ class BitwiseOpConverter(ast.NodeTransformer):
         ast.BitXor: 'BitwiseXor',
         ast.Invert: 'BitwiseNot',
         ast.LShift: 'LeftShift',
-        ast.RShift: 'RightShift'
+        ast.RShift: 'RightShift',
+        ast.FloorDiv: 'int_floor',
```
alexnick83, line 975: This may be a major one, so please check this thoroughly. Python floor division rounds the remainder towards negative infinity. We have implemented this in the runtime, so you may want to call an appropriate method. From `dace/runtime/include/dace/math.h`:
```cpp
// Computes integer floor, rounding the remainder towards negative infinity.
// https://stackoverflow.com/a/39304947
template <typename T, std::enable_if_t<std::is_integral<T>::value && std::is_signed<T>::value>* = nullptr>
static DACE_CONSTEXPR DACE_HDFI T int_floor_ni(const T& numerator, const T& denominator) {
    auto divresult = std::div(numerator, denominator);
    T corr = (divresult.rem != 0 && ((divresult.rem < 0) != (denominator < 0)));
    return (T)divresult.quot - corr;
}
template <typename T, std::enable_if_t<std::is_integral<T>::value && std::is_unsigned<T>::value>* = nullptr>
static DACE_CONSTEXPR DACE_HDFI T int_floor_ni(const T& numerator, const T& denominator) {
    T quotient = numerator / denominator;
    T remainder = numerator % denominator;
    T corr = (remainder != 0 && ((remainder < 0) != (denominator < 0)));
    return quotient - corr;
}

// Computes Python floor division
template<typename T, std::enable_if_t<std::is_integral<T>::value>* = nullptr>
static DACE_CONSTEXPR DACE_HDFI T py_floor(const T& numerator, const T& denominator) {
    return int_floor_ni(numerator, denominator);
}
template<typename T, std::enable_if_t<!std::is_integral<T>::value && std::is_floating_point<T>::value>* = nullptr>
static DACE_CONSTEXPR DACE_HDFI T py_floor(const T& numerator, const T& denominator) {
    return (T)std::floor(numerator / denominator);
}
template<typename T>
static DACE_CONSTEXPR DACE_HDFI std::complex<T> py_floor(const std::complex<T>& numerator, const std::complex<T>& denominator) {
    std::complex<T> quotient = numerator / denominator;
    quotient.real(std::floor(quotient.real()));
    quotient.imag(0);
    return quotient;
}
```

alexnick83 (changes requested):
Please check that you actually get the correct result for negative numbers. You may want to use `py_floor` or `int_floor_ni.`

dace/symbolic.py
tbennun: This only applies to memlets here. There is an easy fix: if we are dealing with memlets, we will generate truncation, otherwise, int_floor will generate a call to py_floor. Does that work?

alexnick83: I see. By memlets, do you mean subsets generated due to slicing expressions, e.g., `A[i//5] = 3`? This should be fine because it would only be incorrect if combined with negative indices. However, symbolic negative indices are not supported anyway, while constant negative indices will be evaluated with the CPython interpreter while parsing.

alexnick83: I would like to support symbolic negative indices in the future, so could you leave a NOTE somewhere in the code to warn about this behavior?

acalotoiu: This would be useful towards handling Fortran codes as well. We are currently leaning towards normalizing to 0, but this does make the SDFG version of code harder to identify with the original - and the more array indexing changes, the harder debugging becomes.

alexnick83 (approved):

PR #1198, opened 2023-02-20
Provide passes to detect shadowed scalar reads and fission them into separte scalars

phschaad:


acalotoiu (approved):
Looks good to me! I appreciate you added tests from the start!

PR #1200, opened 2023-02-26
LoopToMap detects shared transients

alexnick83:
When applying LoopToMap to a for-loop whose body is nested to complete the transformation, any transient data used exclusively inside the body become transient data of the nested SDFG, i.e., thread-local variables. Although this is technically always correct, it could be more efficient in the case of Arrays that can be partitioned to disjoint subsets per loop iteration. This PR adds to LoopToMap functionality to detect and keep such data shared among all Map iterations.

### DIFF
--- a/dace/transformation/interstate/loop_to_map.py
+++ b/dace/transformation/interstate/loop_to_map.py
@@ -8,7 +8,7 @@ import sympy as sp
 import networkx as nx
 from typing import Dict, List, Optional, Set, Tuple
 
-from dace import dtypes, memlet, nodes, registry, sdfg as sd, symbolic, subsets
+from dace import data as dt, dtypes, memlet, nodes, registry, sdfg as sd, symbolic, subsets
 from dace.properties import Property, make_properties, CodeBlock
 from dace.sdfg import graph as gr, nodes
 from dace.sdfg import SDFG, SDFGState, InterstateEdge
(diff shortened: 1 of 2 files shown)


dace/transformation/interstate/loop_to_map.py
```diff
@@ -291,6 +291,48 @@ def test_read_memlet(self, sdfg: SDFG, state: SDFGState, edge: gr.MultiConnector
...
+        :param name: The name of array.
+        :param itervar: The for-loop iteration variable.
+        :param sdfg: The SDFG containing the states that comprise the body of the for-loop.
+        :param states: A list of states that comprise the body of the for-loop.
+        :return: True if the array is thread-local, otherwise False.
+        """
+
+        desc = sdfg.arrays[name]
+        if not isinstance(desc, dt.Array):
+            # Scalars are always thread-local.
```
tbennun, line 312: That will not work for streams, please change condition

tbennun (changes requested):
One tiny change

dace/transformation/interstate/loop_to_map.py
alexnick83: Do you think that it is safe to apply this for Streams?

tbennun (approved):

PR #1201, opened 2023-03-01
Cutout Improvements

phschaad:
This PR separates out parts of the changes proposed in #1079 and adds additional improvements.

# Multistate Cutouts

Allow cutouts from SDFG where the cutout consists of one or more entire states that form a coherent, self-contained SDFG. In practice, that means that there is a single distinct entry state to the cutout's state machine. This mode of constructing cutouts will _always_ create a copy of states and does not contain a reference-cutout mode like in-state cutouts, since states hold a back reference to the SDFG they're contained in and that reference must point to the new cutout SDFG for the cutout to validate.

# "Alibi Nodes" in Cutouts

If a cutout inside of a scope region borders one of the scope nodes (exit or entry), the cutout attempts to cross the scope via each memlet path to connect the corresponding data containers. In the example image below, a cutout of the tasklet `prog_39` would result in the cutout procedure attempting to add the map exit and the access node to `C` to the cutout. Having only the map exit in the cutout, but not the entry trivially leads to a fault. Including the entry node would require including the entire content of the scope. Both these solutions are not what the cutout should intuitively contain.

image

To address this, this PR adds "Alibi Nodes" to cutouts that cross scope borders. What this means, is that in the example above, the access across the map exit border is detected and instead of inserting both the exit node and the outside access node to the cutout, the outgoing memlet from `prog_39` is analyzed, and a new data container with a shape corresponding to the accessed subset is inserted into the cutout SDFG. The memlet is then attached to that new alibi data container instead. This further solves the problem that the memlet crossing the scope border may have symbols in the index term, which can only be defined by the map itself.

# Determine Cutout Input/Output Configuration

The input configuration is any data container read inside the cutout that may contain data from _before_ the cutout is executed in the original SDFG. Analogously, the output configuration consists of any data container that is written to inside the cutout, which may be read in the original SDFG _after_ the cutout was executed.

Previously, *any* transient data container used in the cutout, which was defined outside the cutout, was made global to capture possible inputs and side effects of the cutout. Now, only data containers in the input or output configurations are made global to make the necessary argument set for the cutout smaller and hide away data containers where the lifespan is limited to the scope of the cutout itself.

# Misc

Remove deprecated functions to obtain inputs and outputs from SDFGs. They are semantically invalid. They return any sink nodes in the SDFG as outputs, and any source nodes as inputs. Transient sinks or sources, however, are not considered inputs and outputs. The functions are used nowhere, and their functionality is (correctly) covered by `sdfg.arglist`.


tbennun (approved):
LGTM

phschaad:
Thanks!

PR #1204, opened 2023-03-02
Warn reparse_sdfg or recompile keyword arguments

somu5307:
A warning message will be sent to user if reparse_sdfg or recompile keyword arguments are passed.


tbennun (changes requested):
Thank you for your contribution. This is not exactly the right change (it would only cover the case of `program.to_sdfg(reparse_sdfg=False)`, which would be a tad odd). The goal is to cover `@dace.program(reparse_sdfg=False)` followed by `prog.to_sdfg()`.

You can do that by using the fields in the `DaCeProgram` object (look at `self.recreate_sdfg` for example)

somu5307:
Thank you for your response i will look into it.

tbennun:
Closed due to #1223

PR #1207, opened 2023-03-03
Fix for LoopToMap when applied on multi-nested loops

alexnick83:
This PR fixes an issue where valid parallel loops are not converted to Maps when their bodies consist of other loops.


acalotoiu (approved):
Looks good to me - is this a significant performance penalty though?
Should we mark this a potential performance challenge for future analysis/improvement?

alexnick83:
> Looks good to me - is this a significant performance penalty though? Should we mark this a potential performance challenge for future analysis/improvement?

Tbh, it shouldn't. I just added a DFS call to find the correct states. I have sort of merged it already to the Fortran branch, so if you didn't notice any significant slowdown there, it should be ok.

PR #1208, opened 2023-03-06
Cutout-Assisted Debugging

phschaad:
This PR contains varios improvements to how cutouts are handle, together with a number of extensions to the capabilities around data instrumentation. The common goal is to enable a workflow where transformations or codegen issues can be localized and more easily diagnosed / debugged through the help of minimal test cases. The concrete list of changes consists of:

- [x] Making cutouts an SDFG subclass to allow them to carry additional data such as their input/output configurations and what SDFG they came from.
- [x] Removing hashes from the serialization test. They are not useful for testing the serialization framework and cause failures even just if the set of hashed properties does not match exactly with the set of serialized properties. This may be the case, for example, if an SDFG subclass (such as cutouts) is serialized to an SDFG.
- [x] Allowing cutouts to be constructed from transformations (i.e., their set of affected nodes) based on pattern matching or subgraphs.
- [x] Allowing symbols to be dumped and restored through data instrumentation on SDFG states.
- [x] Making `InstrumentedDataReport`s able to read dumped symbol values.
- [x] Capturing the cutout's inputs/outputs through a dry run of the full SDFG, including symbol values.


dace/runtime/include/dace/serialization.h
```diff
@@ -89,6 +89,58 @@ class DataSerializer {
...
+        }
+
+        // Write contents to file
+        ss << "/" << filename << "_" << version;
+        std::ofstream ofs(ss.str(), std::ios::out);
+        ofs << symbol_value;
+    }
+
+    template <typename T>
+    T restore_symbol(const std::string &symbol_name, const std::string &filename, T symbol_oldval) {
```
tbennun, line 122: why do you need the old value?

tests/sdfg/cutout_test.py
```diff
@@ -248,16 +248,16 @@ def test_multistate_cutout_complex_expand():
...
+    ct: SDFGCutout = SDFGCutout.multistate_cutout(s4, s5, s6, s7)
+    state_names = [s.name for s in ct.states()]
+    assert len(state_names) == 7
+    assert ('s1' in state_names)
+    assert ('s2' in state_names)
+    assert ('s3' in state_names)
+    assert ('s4' in state_names)
+    assert ('s5' in state_names)
+    assert ('s6' in state_names)
+    assert ('s7' in state_names)
```
tbennun, line 262: no test for symbol instrumemtation?

dace/sdfg/analysis/cutout.py
```diff
@@ -404,11 +607,11 @@ def _extend_subgraph_with_access_nodes(state: SDFGState, subgraph: StateSubgraph
...
 
 
-def determine_cutout_reachability(
+def _determine_cutout_reachability(
         ct: SDFG,
         sdfg: SDFG,
-        in_translation: TranslationDictT,
-        out_translation: TranslationDictT,
+        in_translation: dict,
+        out_translation: dict,
```
tbennun, line 614: type hints need to be Python 3.7 compatible (`Dict[K, V]`)

dace/sdfg/analysis/cutout.py
```diff
@@ -442,8 +645,8 @@ def determine_cutout_reachability(
     return (inverse_cutout_reach, cutout_reach)
 
 
-def cutout_determine_input_config(ct: SDFG, inverse_cutout_reach: Set[SDFGState], in_translation: TranslationDictT,
-                                  out_translation: TranslationDictT) -> Set[str]:
+def _cutout_determine_input_config(ct: SDFG, inverse_cutout_reach: Set[SDFGState], in_translation: dict,
```
tbennun, line 648: same

dace/sdfg/analysis/cutout.py
```diff
@@ -505,8 +708,8 @@ def cutout_determine_input_config(ct: SDFG, inverse_cutout_reach: Set[SDFGState]
     return input_configuration
 
 
-def cutout_determine_output_configuration(ct: SDFG, cutout_reach: Set[SDFGState], in_translation: TranslationDictT,
-                                          out_translation: TranslationDictT) -> Set[str]:
+def _cutout_determine_output_configuration(ct: SDFG, cutout_reach: Set[SDFGState], in_translation: dict,
```
tbennun, line 711: same

tbennun (changes requested):
Looks good overall. Some questions and minor changes.

tests/sdfg/cutout_test.py
phschaad: added

dace/runtime/include/dace/serialization.h
phschaad: We don't, forgot to remove, thanks

tbennun (approved):
Looks great, thank you!

PR #1212, opened 2023-03-10
Fix-intermediate-nodes

alexnick83:
Ths PR introduces a new transformation that tries to remove intermediate write from Map subgraphs. This first version targets the pattern `WriteAccess -> MapExit,` where the edge has an empty Memlet.


acalotoiu (approved):
Looks good to me.

PR #1214, opened 2023-03-12
Restricted Symbol SSA and ScalarFission fixes

phschaad:
1. Allow symbols to be transformed into a restricted SSA form (i.e. without phi-nodes)
2. Fix a bug in ScalarFission with which fissioning happened too aggressively in the case of branching scopes with subscopes in one of the branches. This has lead to subscopes being fissioned into a separate data container, while they may in fact be read again in the parent scope.


acalotoiu (approved):
Looks good to me. You can merge once that tests pass :)

PR #1216, opened 2023-03-13
Map-fission-single-data-multi-connectors

alexnick83:
When fissioning a Map with a single nested SDFG in its body, there is a bug when reconnecting the AccessNodes around the Map to the NestedSDFG node. If an AccessNode has outgoing edges to more than one NestedSDFG connector, then only one will be reconnected. This PR addresses this issue.


acalotoiu (approved):
Looks good to me. 
Should we keep the old code as comments in the fixed code?
I would suggest removing it altogether.

alexnick83:
> Looks good to me. Should we keep the old code as comments in the fixed code? I would suggest removing it altogether.

Oh, I have commented-out tests. I will fix this first.

PR #1217, opened 2023-03-15
Doc: Installation: Add OpenBLAS symlink tip

nrwahl2:
On some systems with OpenBLAS installed, libblas.so and liblapacke.so may need to be created manually as symlinks to libopenblas.so.


tbennun (approved):
LGTM, thanks!

PR #1221, opened 2023-03-16
Fix-nested-sdfg-deepcopy

alexnick83:
This PR attempts to solve issues that arise when (deep) copying NestedSDFG nodes and nested SDFGs. When deep-copying an SDFG object, the "parent" attributes should now be set to None. Deep-copying a NestedSDFG node should now set the nested SDFGs `parent_nsdfg_node` attribute to the copy. Adding a NestedSDFG node to an SDFGState with the `add_node` method should properly update the nested SDFGs "parent" attributes.


tbennun (approved):
LGTM (how have we never seen this before). Any test you want to add?

PR #1222, opened 2023-03-19
Scalar Write Shadow Analysis Fix

phschaad:
Scalar write shadow analysis was trying to be too conservative in the context of loops and consequently coarsened the scope of writes incorrectly, which caused problems for Scalar Fission.


acalotoiu (approved):
Looks good! It also fixes the original issue we reported

Issue #1150, opened 2022-11-11
`DeadDataFlowElimination` removes connector, which removes the type hint of a tasklet variable


gronerl:
The following example
```
sdfg = dace.SDFG("test")
state = sdfg.add_state()
sdfg.add_array("out", dtype=dace.float64, shape=(10,))
sdfg.add_array("cond", dtype=dace.bool, shape=(10,))
sdfg.add_array("tmp", dtype=dace.float64, shape=(10,), transient=True)
tasklet, *_ = state.add_mapped_tasklet(
    code="""if _cond:
    _tmp = 3.0
else:
    _tmp = 7.0
_out = _tmp
""",
    inputs={"_cond": dace.Memlet(subset="k", data="cond")},
    outputs={
        "_out": dace.Memlet(subset="k", data="out"),
        "_tmp": dace.Memlet(subset="k", data="tmp"),
    },
    map_ranges={"k": "0:10"},
    name="test_tasklet",
    external_edges=True,
)
sdfg.simplify()
sdfg.compile()

```
fails to compile after simplify. The cause is that DeadDataflowElimination removes a connector from a Tasklet which leaves a variable without typehint, resulting in this generated code:
```
                  ///////////////////
                  // Tasklet code (test_tasklet)
                  if (_cond) {
                      auto _var = 3.0;
                  }
                  else {
                      auto _var = 7.0;
                  }
                  _out = _var;
                  ///////////////////
```

BhuvanAde:
Hey @tbennun, can I take up this issue?

tbennun:
@BhuvanAde of course!

BhuvanAde:
Thanks!

PR #1226, opened 2023-03-21
Fix DeadDataFlowElimination removing neccessary typehint

Luo-Yihang:
### Motivation
This is pr tries to fix the issue #1150, which suggests `DeadDataFlowElimination` may unintentionally remove the required type hint of a tasklet variable, leading to the failure of sdfg compilation.

### Modification
Added the type hint expression to the beginning of the tasklet.

### May need confirm
- [ ] Will it be better to add the type hint only if the variable is found in the tasklet code?
- [ ] Is it correct that the type hint is hardcoded to `dace.float64`? If not, what is the appropriate way to obtain the variable type?


tbennun:
@Luo-Yihang the type hint should not be hardcoded to float64. You can get the type of the connector by first querying it (connectors are dictionaries), and if it's None you need to infer it from the memlet (which data container is connected to it).

It would be nice to not do it if the connector name is not found in the tasklet (see `ASTFindReplace`)

tbennun:
Superseded by #1499

PR #1229, opened 2023-03-23
Standardize jocobi_1d scpripts as other distributed samples

Luo-Yihang:
### Motivation
The sample scripts under `samples/distributed` have output logs (like time spent, validation correctness) expcept for the `jacobi_1d.py`. It would be beneficial for us to add time logs and validation output for this script as well.

### Modification
Added time logs and validation results in `jacobi_1d.py`.

### Outcome
I have tested the result, a sample output running under 2 processes would be (the format is just like other scripts result):
```
Median is 2ms
=======Validation=======
OK
```


alexnick83 (approved):
LGTM

PR #1230, opened 2023-03-23
Fix atomic operation detection for exactly-overlapping ranges

tbennun:


acalotoiu (approved):
Looks good to me! Thanks for adding the tests!

PR #1231, opened 2023-03-23
Fix-gpu-transform-copy-out

alexnick83:
This PR solves an issue with GPUTransform where copying an array from GPU global memory back to the host overwrites values that weren't updated in the kernel.

### DIFF
--- a/dace/transformation/interstate/gpu_transform_sdfg.py
+++ b/dace/transformation/interstate/gpu_transform_sdfg.py
@@ -8,6 +8,7 @@ from dace.transformation import transformation, helpers as xfh
 from dace.properties import Property, make_properties
 from collections import defaultdict
 from copy import deepcopy as dc
+from sympy import floor
 from typing import Dict
 
 gpu_storage = [dtypes.StorageType.GPU_Global, dtypes.StorageType.GPU_Shared, dtypes.StorageType.CPU_Pinned]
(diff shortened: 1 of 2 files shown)


dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -214,6 +217,26 @@ def apply(self, _, sdfg: sd.SDFG):
             name = sdfg.add_datadesc('gpu_' + onodename, newdesc, find_new_name=True)
             cloned_arrays[onodename] = name
 
+            # The following ensures that when writing to a subset of an array, we don't overwrite the rest of the array
+            # when copying back to the host. This is done by adding the array to the `inputs_nodes,` while will copy
```
tbennun, line 221:
```suggestion
            # when copying back to the host. This is done by adding the array to the `input_nodes,` which will copy
```

dace/transformation/interstate/gpu_transform_sdfg.py
```diff
@@ -214,6 +217,26 @@ def apply(self, _, sdfg: sd.SDFG):
...
+            # the entire array to the GPU.
+            if (onodename, onode) not in input_nodes:
+                found_full_write = False
+                full_subset = sbs.Range.from_array(onode)
+                try:
+                    for state in sdfg.nodes():
+                        for node in state.nodes():
+                            if (isinstance(node, nodes.AccessNode) and node.data == onodename):
+                                for e in state.in_edges(node):
+                                    if e.data.get_dst_subset(e, state) == full_subset:
```
tbennun, line 231: would this work for dynamic memlets (i.e., writing to one index but the subset is the full array)

tbennun (changes requested):

dace/transformation/interstate/gpu_transform_sdfg.py
alexnick83: No, it will not work. Are you suggesting that we fail this check if the memlet is dynamic? Good idea.

tbennun (approved):
Nice, LGTM

PR #1232, opened 2023-03-23
ConditionalOptionalArrayResolver pass in preprocess

FlorianDeconinck:
Introduce a new pass in preprocessing that aims to reduce the well define case of optional hinted arrays. E.g.

```python
def program(array: Optional[]):
  if array is None:
    ...
  else:
    ...
    
program(None, ...)
 ```

See #1209 for full background

The PR adds a new pass and fix the `optional` flag on sdfg.arrays.
Test `test_optional_array_inference_via_parse` in `tests/python_frontend/optional_array_test.py`

### DIFF
--- a/dace/frontend/python/preprocessing.py
+++ b/dace/frontend/python/preprocessing.py
@@ -1260,10 +1260,16 @@ class LoopUnroller(ast.NodeTransformer):
 
 class CallTreeResolver(ast.NodeVisitor):
 
-    def __init__(self, closure: SDFGClosure, globals: Dict[str, Any]) -> None:
+    def __init__(
+        self,
+        closure: SDFGClosure,
+        globals: Dict[str, Any],
+        caller_argtypes: Dict[str, data.Data],
+    ) -> None:
         self.closure = closure
         self.seen_calls: Set[str] = set()
         self.globals = globals
+        self.caller_argtypes = caller_argtypes
 
     def _eval_args(self, node: ast.Call) -> Dict[str, Any]:
         res = {}
@@ -1288,13 +1294,19 @@ class CallTreeResolver(ast.NodeVisitor):
 
         return res
 
-    def _get_given_args(self, node: ast.Call, function: 'DaceProgram') -> Set[str]:
+    def _get_given_args(
+        self,
+        node: ast.Call,
+        function: 'DaceProgram',
+        caller_argtypes: Dict[str, data.Data],
+    ) -> Set[str]:
         """ Returns a set of names of the given arguments from the positional and keyword arguments """
         from dace.frontend.python.parser import DaceProgram  # Avoid import loop
 
         posargs = node.args
         kwargs = [kwarg.arg for kwarg in node.keywords]
         result = set()
+        argtypes = {}
 
         if not isinstance(function, DaceProgram):
             # Make set of parameters from __sdfg_signature__
@@ -1325,6 +1337,7 @@ class CallTreeResolver(ast.NodeVisitor):
             # END OF VARIABLE-LENGTH ARGUMENTS
             elif sig_kind is inspect.Parameter.POSITIONAL_ONLY:
                 if arg_ind < nargs:
+                    argtypes[aname] = caller_argtypes[posargs[arg_ind].id]
                     result.add(aname)
                     arg_ind += 1
             elif sig_kind is inspect.Parameter.POSITIONAL_OR_KEYWORD:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: dace/frontend/python/preprocessing.py [hunk 1]
```diff
@@ -195,6 +196,57 @@ def visit_IfExp(self, node: ast.IfExp) -> Any:
...
+    def _resolve_array_optional(self, node: ast.Compare, is_none: bool):
+        # Is lhs or rhs None?
+        if isinstance(node.left, ast.Constant) and node.left is None:
+            cmp_to_test = node.comparators[0]
+        elif isinstance(node.comparators[0], ast.Constant) and node.comparators[0].value is None:
+            cmp_to_test = node.left
+        else:
+            return self.generic_visit(node)
+
+        # Is the non-None cmp and array? Check it's optional status
```
alexnick83, line 224:
```suggestion
        # Is the non-None cmp an array? Check it's optional status
```

alexnick83 (changes requested):
It looks quite good to me. However, I think you should make a test with the original code you posted to showcase the issue to make sure that there are no issues with "redefinitions" of views. Also, you should add your name to the `AUTHORS` file.

FlorianDeconinck:
Fixing test it seem my original logic is flawed. Type hinting an array as `Optional` doesn't tell us anything more than "we should check at runtime". The way parsing is setup we skip looking at the actual given argument if any hinting is setup (e.g. `argtypes` build by `_get_type_annotations`, especially line `577`)

So best we can do when we have a type hint is "maybe this will go away, we have to wait for call time to know". When no type hint is given, then we can infer because we are looking at given args, but isn't that more a side effect? Having the system behaves differently when type hinted or not seems dubious at best.

alexnick83:
> Fixing test it seem my original logic is flawed. Type hinting an array as `Optional` doesn't tell us anything more than "we should check at runtime". The way parsing is setup we skip looking at the actual given argument if any hinting is setup (e.g. `argtypes` build by `_get_type_annotations`, especially line `577`)
> 
> So best we can do when we have a type hint is "maybe this will go away, we have to wait for call time to know". When no type hint is given, then we can infer because we are looking at given args, but isn't that more a side effect? Having the system behaves differently when type hinted or not seems dubious at best.
> 
> I can write the `optional` inference by actually resolving the given argument but that means making the SDFG dependent on the way it's called. E.g.
> 
> ```python
> @dace.program
> def meh( A: Optional[dace.Float64[2]]):
>  if A is None:
>   ...
>  
>  meh(A) <- one SDFG
>  meh(None) <- another SDFG
> ```
> 
> Isn't that problematic in the wider usage of DaCe?

I think that this is fine. In fact, @tbennun isn't the purpose of the closure to make separate SDFGs only when the arguments change in a significant way?

FlorianDeconinck:
Seems fair, I will write the code that go down that path and we can reason on it.

tbennun:
Just make sure that if type hints were given, it’s AOT compilable

FlorianDeconinck:
TL;DR: to make sure AOT based on type hints only is supported, no inference can be done, including in the `OptionalArrayInference` which might be a bug.

Going deeper on this, clearly the AOT & JIT clashes.

My understanding is as follows.

Given
```python

@dace.program
def program(hinted_a: Optional[dace.float64[2]], b):
   if hinted_a is None:
      ...
   if b is None:
      ...
```

This should be compilable with
1. program.to_sdfg(A, B)
2.  program.to_sdfg()

When down in parsing, I can't reason with `given_args` because of 2. Therefore I can only read the `hint` for the optional flag. This means that `Optional=True` is not an information we can infer on, since at call time we can have either an array or `None` in the case of 2 without triggering recompilation.

Happy to dive in this with you guys whenever you have time

FlorianDeconinck:
The only option left is to restrict this pass to nested calls made with clear `None` type e.g.

```python

def nested(A. B):
   ..

@dace.program
def program(A, B):
   nested(A, None) <- we can infer here when doing program.to_sdfg()
```

dace/frontend/python/preprocessing.py
tbennun: also, "its"

alexnick83:
> The only option left is to restrict this pass to nested calls made with clear `None` type e.g.
> 
> ```python
> def nested(A. B):
>    ..
> 
> @dace.program
> def program(A, B):
>    nested(A, None) <- we can infer here when doing program.to_sdfg()
> ```
> 
> but we don't have the information in preprocess if this is the top-level or a nested call

The underlying issue is that we just don't support the redefinition of arrays and views (the latter is the issue in your case). An experimental SDFG feature supports redefinition, and I am considering using it to support redefining views **as an experiment**. However, I don't have an ETA because I will probably need to spend a day on that, and it may be hard to commit before the SC deadline. I will update you.

PR #1236, opened 2023-03-27
Make-intermediate-map-writes-external

alexnick83:
This PR added a helper method for redirecting a Map's internal write to an output Access node. It further amens LoopToMap to make use of this method.


acalotoiu (approved):
LGTM

acalotoiu (changes requested):
Could you please add an additional test to ensure the new scenario is covered?

alexnick83:
> Could you please add an additional test to ensure the new scenario is covered?

The changes in the last 6 commits are already covered by NPBench tests. Do we want more specific ones?

PR #1242, opened 2023-04-18
Inline preprocessor command

tbennun:
This PR adds the `dace.inline` function to the Python frontend. The preprocessor will call the internal expression at compile-time and replace the value with the results.


alexnick83 (approved):
LGTM

Issue #1220, opened 2023-03-16
float32 detection on pow


FlorianDeconinck:
Consider the following code
``` python
import dace
import numpy as np


@dace.program
def f32_pow_failure(array):
    array = array**3.3


A = np.ones(shape=(10), dtype=np.float32)
f32_pow_failure(A)
```

This will fail on compile because `dace::pow(float*, double)` is not a valid templated version. Of course, we want `3.3` to be a float32
I could write `array = array**np.float32(3.3)` but if the code float precision is meant to be parametrized it brings other problem (type inference) and unneeded burden on the developer to type out the literal in code.

Could we have inference of the literal floating precision looking at the variable on lhs of the `**` ?

PS: this pattern was found many times in atmospheric science code

nrwahl2:
FWIW it's correct after replacement. The tasklet's Python code is `__out = __in1 ** dace.float32((3.3))`. But the generated C++ (abridged) is
```
float __in1 = array[__i0];
float __out;
__out = dace::math::pow(__in1, 3.3); 
```

which loses the casting of 3.3 to 32-bit float

alexnick83:
> FWIW it's correct after replacement. The tasklet's Python code is `__out = __in1 ** dace.float32((3.3))`. But the generated C++ (abridged) is
> 
> ```
> float __in1 = array[__i0];
> float __out;
> __out = dace::math::pow(__in1, 3.3); 
> ```
> 
> which loses the casting of 3.3 to 32-bit float. Specifically it looks like we lose it when `DaCeKeywordRemover.visit_BinOp` evaluates the right-hand side of the operation, transforming it from a `Call` to a `Constant`.
> 
> The compile failure starts with commit [60a97f8](https://github.com/spcl/dace/commit/60a97f8a1c8586c3622433060a760fe5d0648c02). Not currently sure exactly how that commit broke us, and whether it's somehow problematic in itself or unmasked a different problem. I suspect it just unmasked the evaluation behavior above.

Correct; it looks like it was introduced inadvertently while trying to optimize code generation for integer powers. We are fixing it.

FlorianDeconinck:
Cheers!

PR #1243, opened 2023-04-19
Fix for #1220

alexnick83:
This PR addresses the issue of implicit casting in (floating point) pow operations being stripped from the generated code due to an optimization concerning integer pow operations.


tbennun (approved):
LGTM!

Issue #1164, opened 2022-11-30
Python `with` statement code generation identifier is not unique enough


FlorianDeconinck:
**Describe the bug**
The `with` statement generates in C code a pair of `__with_XXX___enter`, `__with_XXX___exit` statements, with XXX the line number in the original source. It's also how those symbols are earmarked in the SDFG. Unfortunately, this is can cause nasty clashes when:
- `with` statement from two different file are at the same line
- code change _outside_ the DaCe handled code path end up changing the line of a with statement that is considered by DaCe, which invalidates running the `.so` (bad symbol) when technically nothing changed in the code DaCe should care about.

**To Reproduce**

It would be a two file reproducer with `with` statement sharing a file number.

**Expected behavior**

The python frontend handling with statement properly is a very good feature and shouldn't be discarded. A more robust sanitization of the with statement should be found.

Proposal: `with util.timer.clock("mainloop")` to be sanitized as `__with_util_timer_clock_mainloop_X___enter` with X a global counter on with statements to keep ordering consistent.

FlorianDeconinck:
Here is a better reproducer. It probably wasn't reproduce-able because you have to _load an existing SDFG_ previously compiled

Below the key is to add code that _isn't in the DaCe parsing path_ but that do not change anything (comments). If you shift the line count, the SDFG can't reload

```
import dace
import numpy as np
import contextlib


class Ctx:
    def __init__(self) -> None:
        self.did_start = False
        self.should_pass = False

    @contextlib.contextmanager
    def mgr(self, name: str):
        self.start(name)
        yield
        self.stop()

    def start(self, name: str):
        if name == "pass":
            self.did_start = True

    def stop(self):
        if self.did_start:
            self.should_pass = True


ctx = Ctx()


@dace.program
def a_program(A):
    with ctx.mgr("pass"):
        A[:] = 0


np_arr = np.random.random((10))
sdfg = a_program.to_sdfg(np_arr)
# a_program.load_precompiled_sdfg(".dacecache/a_program", np_arr)
a_program(np_arr)


# Next add comment line 5
# Comment line 36
# Uncomment line 37
# Add a new line (with a few comments for example) _before_ line 29
# >>> KeyError: 'Missing program argument "__with_31___enter__"'
```

FlorianDeconinck:
Sorry for the delay in checking.

My tests all pass, so it looks solid. I will update and run the model a little bit later - closing for now. Cheers!

PR #1253, opened 2023-05-09
Investigating Issue #1164

alexnick83:


tests/python_frontend/context_manager_test.py
```diff
@@ -36,5 +36,56 @@ def prog(A: dace.float64[20]):
...
+    sdfg = ctxmgr_name_clashing_0.to_sdfg()
+    
+    for i, f in enumerate([ctxmgr_name_clashing_0, ctxmgr_name_clashing_1]):
+
+        if i > 0:
+            f.load_precompiled_sdfg(sdfg.build_folder)
+
+        a_count = 0
+        b_count = 0
+        for _ in range(100):
```
tbennun, line 80: This test looks overly complicated. Any reason?

tests/python_frontend/context_manager_test.py
```diff
@@ -36,5 +36,56 @@ def prog(A: dace.float64[20]):
...
+        else:
+            prog_b()
+        return i
+    
+    sdfg = ctxmgr_name_clashing_0.to_sdfg()
+    
+    for i, f in enumerate([ctxmgr_name_clashing_0, ctxmgr_name_clashing_1]):
+
+        if i > 0:
+            f.load_precompiled_sdfg(sdfg.build_folder)
```
tbennun, line 76: should we not call `sdfg.compile()` first?

tbennun (commented):
Some questions but otherwise LGTM

dace/dtypes.py
```diff
@@ -599,6 +599,7 @@ def from_json(json_obj, context=None):
 
         try:
             typeclass = json_to_typeclass(json_obj['ctype'], context)
+            return typeclass()
```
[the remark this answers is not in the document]
alexnick83: This is part of a greater fix. I identified the issue by having a callback to the `print` method, which is registered on the SDFG as a `dtypes.pyobject` Scalar. When reading the SDFG from file, it becomes `dtypes.opaque`. Calling the program with loaded SDFG subsequently fails, because the opaque type is missing methods needed by ctypes. The return statement is needed because `opaque(pyobject)` is opaque, not pyobject. You can comment out the line and run the new test to see the exact ctypes error.

tests/python_frontend/context_manager_test.py
alexnick83: Yes and no. I was testing various modifications but the test combines all the reported issues:
- context managers with potentially the same name (when auto simplification is on)
- reloading from file and calling through another program 

If you are asking about the `count`s, it is just to confirm that both context managers are executed despite having the same name. It could be simplified of course.

tests/python_frontend/context_manager_test.py
alexnick83: It is compiled for `i = 0`.

tbennun (approved):

PR #1256, opened 2023-05-17
SymPy 1.12

alexnick83:
Fixes for compatibility with SymPy 1.12.

The main issue concerns symbolic substitution. In SymPy 1.12, the new values are sympified with `strict=True,` which allows the sympification of only scalar values (no `None` values, `Callable`s. `Iterable`s, and strings). This PR addressed the issue in the following manner:
- `None` values, `Callable`s, and `Iterable`s are filtered out from the replacement dictionaries since it is unlikely that their corresponding old values need to be replaced.
- Strings are first converted to `sympy.symbol`s since they should be symbol names.

### DIFF
--- a/tests/inlining_test.py
+++ b/tests/inlining_test.py
@@ -4,6 +4,7 @@ from dace.transformation.interstate import InlineSDFG, StateFusion
 from dace.libraries import blas
 from dace.library import change_default
 import numpy as np
+import os
 import pytest
 
 W = dace.symbol('W')
@@ -263,8 +264,6 @@ def test_inline_unsqueeze3():
             assert (np.array_equal(B[:, i], np.zeros((5, ), np.int32)))
 
 
-# NOTE: Issue with serialization
-@pytest.mark.skip
 def test_inline_unsqueeze4():
 
     @dace.program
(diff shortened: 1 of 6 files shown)


tests/inlining_test.py
```diff
@@ -281,7 +280,10 @@ def inline_unsqueeze(A: dace.int32[2, 5], B: dace.int32[5, 3]):
 
     A = np.arange(10, dtype=np.int32).reshape(2, 5).copy()
     B = np.zeros((5, 3), np.int32)
+    last_value = os.environ.get('DACE_testing_serialization', '0')
+    os.environ['DACE_testing_serialization'] = '0'
```
tbennun, line 284: use `with dace.config.set_temporary`

tests/npbench/polybench/correlation_test.py
```diff
@@ -71,7 +72,10 @@ def run_correlation(device_type: dace.dtypes.DeviceType):
         # Parse the SDFG and apply autopot
         sdfg = correlation_kernel.to_sdfg()
         sdfg = auto_optimize(sdfg, device_type)
+        last_value = os.environ.get('DACE_testing_serialization', '0')
```
tbennun, line 75: same

tbennun (changes requested):

tbennun (approved):

PR #1259, opened 2023-05-22
Fix compile errors

edopao:
- Some import paths were not compatible with python3.6
- The g++ compiler on MacOS installed through Homebrew seems to rely on MacOS SDK built with clang, for which __unused is a reserved word (an attribute), therefore variables cannot be named __unused.


acalotoiu (commented):
Looks good to me - but I would like to also hear from @tbennun.

tbennun (approved):

edopao:
@tbennun @acalotoiu Thank you for the review. I am not an authorized user for this project, so merging is blocked.

PR #1262, opened 2023-05-24
Determine schedule type based on surrounding storage types

### LINKED ISSUE #1185: Python frontend should use storage type of array to determine map schedule
@alexnick83 
In the following example, the storage types of the output arrays are well-inferred:

```python
@dace.program
def add(a: dace.float32[10, 10] @ dace.StorageType.GPU_Global, 
        b: dace.float32[10, 10] @ dace.StorageType.GPU_Global):
    return a + b @ b

add.to_sdfg().view()
```

However, the map generated by `a + b` would have the `Default` schedule, which translates to a CPU map, which subsequently crashes.

Additionally, the following code should probably fail gracefully, due to the output storage / computation schedule not being able to be inferred:
```python
@dace.program
def add(a: dace.float32[10, 10] @ dace.StorageType.GPU_Global, 
        b: dace.float32[10, 10]):
    return a + b
```
but it doesn't fail.

tbennun:
This is a lighter version of #1241 without the entire SDFG device revamp. This PR fixes #1185 by inferring schedule types of maps based on the surrounding storage types (skipping scalar values). As opposed to #1241, this PR does not remove the `GPU_Default` schedule.

Comment for changelog @phschaad:

The schedule type of a scope (e.g., a Map) is now also determined by the surrounding storage. If the surrounding storage is ambiguous, dace will fail with a nice exception. This means that codes such as the one below:
```python
@dace.program
def add(a: dace.float32[10, 10] @ dace.StorageType.GPU_Global, 
        b: dace.float32[10, 10] @ dace.StorageType.GPU_Global):
    return a + b @ b
```
will now automatically run the `+` and `@` operators on the GPU.

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -218,6 +218,8 @@ class CompiledSDFG(object):
             
             :return: the ctypes.Structure representation of the state struct.
         """
+        if not self._libhandle:
+            raise ValueError('Library was not initialized')
 
         return ctypes.cast(self._libhandle, ctypes.POINTER(self._try_parse_state_struct())).contents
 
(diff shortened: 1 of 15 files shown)


phschaad (approved):
This looks good to me.

alexnick83 (approved):

PR #1265, opened 2023-05-26
Feature/log10

oelbert:
This pull request adds log10 as a dace function, similar to log and log2, and adds tests to math_test and ufunc_test.

### DIFF
--- a/dace/frontend/python/replacements.py
+++ b/dace/frontend/python/replacements.py
@@ -692,6 +692,14 @@ def _log(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, input: str):
     return _simple_call(sdfg, state, input, 'log')
 
 
+@oprepo.replaces('log10')
+@oprepo.replaces('dace.log10')
+@oprepo.replaces('numpy.log10')
+@oprepo.replaces('math.log10')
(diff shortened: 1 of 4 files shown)


tbennun (approved):

tbennun:
LGTM, thanks!

dace/frontend/python/replacements.py
```diff
@@ -692,6 +692,14 @@ def _log(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, input: str):
     return _simple_call(sdfg, state, input, 'log')
 
 
+@oprepo.replaces('log10')
+@oprepo.replaces('dace.log10')
+@oprepo.replaces('numpy.log10')
```
alexnick83, line 697: AFAIK the NumPy log10 ufunc is already supported through the Python frontend's ufunc interface (see `dace/frontend/python/replacements.py`). Could you confirm this and remove the above line so it doesn't interfere with calls that, e.g., use ufunc keyword arguments (any supported by DaCe)?

alexnick83 (changes requested):

dace/frontend/python/replacements.py
oelbert: I can remove this line but I haven't been able to confirm that `numpy.log10` works

alexnick83 (approved):

PR #1268, opened 2023-06-06
State Fusion Extension with happens before dependency edge

acalotoiu:
This is a version of StateFusion that can fuse some instances of WAR and WAR conflicts by using an empty memlet as a happens-before dependency edge. This allows other not connected subgraphs within the states fused to be optimized.


dace/transformation/interstate/state_fusion_with_happens_before.py
```diff
@@ -0,0 +1,589 @@
...
+
+
+def top_level_nodes(state: SDFGState):
+    return state.scope_children()[None]
+
+
+class StateFusionExtended(transformation.MultiStateTransformation):
+    """ Implements the state-fusion transformation extended to fuse states with RAW and WAW dependencies.
+        An empty memlet is used to represent a dependency between two subgraphs with RAW and WAW dependencies.
+        The merge is made by identifying the source in the first state and the sink in the second state, and linking the bottom of the appropriate source subgraph in the first state with the top of the appropriate sink subgraph in the second state.
```
phschaad, line 37: yapf

dace/transformation/interstate/state_fusion_with_happens_before.py
```diff
@@ -0,0 +1,589 @@
...
+            x for x in reversed(list(nx.topological_sort(first_state._nx)))
+            if isinstance(x, nodes.AccessNode) and sdict[x] is None
+        ]
+        for node in second_state.nodes():
+            if isinstance(node, nodes.NestedSDFG):
+                # update parent information
+                node.sdfg.parent = first_state
+            #we only want to add the node once!    
+            try:
+                first_state.add_node(node)
```
phschaad, line 523: Why is the try / except necessary here, when is a node being added that already exists?

dace/transformation/interstate/state_fusion_with_happens_before.py
```diff
@@ -0,0 +1,589 @@
...
+            # the first state is an initial state (in which case the new initial
+            # state would be ambiguous).
+            first_in_edges = graph.in_edges(first_state)
+            second_in_edges = graph.in_edges(second_state)
+            if ((not second_state.is_empty() or not first_state.is_empty() or len(first_in_edges) == 0)
+                    and len(second_in_edges) != 1):
+                return False
+
+            # Get connected components.
+            first_cc = [cc_nodes for cc_nodes in nx.weakly_connected_components(first_state._nx)]
```
phschaad, line 261: The can_be_applied is quite computationally expensive for large graphs, how does this hold up in our pipeline for a decently sized application? I see it having a noticeable impact for graphs with larger states. Do you have any measurements / data?

phschaad (changes requested):

dace/transformation/interstate/state_fusion_with_happens_before.py
acalotoiu: It's just as expensive as the baseline state fusion (and it's currently not part of the pipeline) - so unless state fusion itself is too slow, this should be ok.

dace/transformation/interstate/state_fusion_with_happens_before.py
acalotoiu: You are correct - it is not. I was not able to create a case where this is needed. Likely another bug that was fixed, or SC tiredness issue

phschaad (approved):

PR #1283, opened 2023-06-23
GPU User Experience Improvements

tbennun:
- [x] Post-SDFG error checking and informative guidance
- [ ] More information in syncdebug prints (e.g., actual grid/block sizes)
- [ ] Warnings/errors on conflicting behavior with gpu_block_size, configuration, and intenal maps
- [ ] Validation error upon reading from a non-accessible host memory (e.g. GPU global) outside an accelerator context
- [ ] Argument errors
- [ ] `__launch_bounds__` support in map nodes: `-1` disables, `0` (default) sets if block size is constant, any other number sets explicitly

### DIFF
--- a/dace/codegen/codegen.py
+++ b/dace/codegen/codegen.py
@@ -31,7 +31,7 @@ def generate_headers(sdfg: SDFG, frame: framecode.DaCeCodeGenerator) -> str:
     exit_params = (sdfg.name, sdfg.name)
     proto += 'typedef void * %sHandle_t;\n' % sdfg.name
     proto += 'extern "C" %sHandle_t __dace_init_%s(%s);\n' % init_params
-    proto += 'extern "C" void __dace_exit_%s(%sHandle_t handle);\n' % exit_params
+    proto += 'extern "C" int __dace_exit_%s(%sHandle_t handle);\n' % exit_params
     proto += 'extern "C" void __program_%s(%sHandle_t handle%s);\n' % params
     return proto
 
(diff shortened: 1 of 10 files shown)


mcopik:
I think it might be worth considering a case where we fail at the first CUDA error, not just check at the end - this will make debugging much easier and faster. We could make it an optional feature if there's any concern that an additional check of a return value from CUDA calls would add an unnecessary performance tax.

Happy to help with this feature by improving & merging my solution for this.

tbennun:
We don’t want the overhead by default, that’s what syncdebug is for. It does what you proposed and more (and even more after this PR).

tbennun:
@mcopik I followed your suggestion in the latest commit

dace/transformation/passes/fusion_inline.py
```diff
@@ -82,3 +83,43 @@ def apply_pass(self, sdfg: SDFG, _: Dict[str, Any]) -> Optional[int]:
...
+
+
+@dataclass(unsafe_hash=True)
+@properties.make_properties
+class FixNestedSDFGReferences(ppl.Pass):
+    """
+    Fixes nested SDFG references to parent state/SDFG/node
+    """
+
+    CATEGORY: str = 'Simplification'
```
phschaad, line 95: It may make sense to introduce a category called 'Housekeeping' or something simliar for passes like this. I assume it will not be the last one that doesn't 'simplify' anything, just cleans stuff up.

dace/runtime/include/dace/cuda/stream.cuh
```diff
@@ -28,252 +28,236 @@
...
+      return i % m_capacity_mask;
+  }
+
+  __device__ __forceinline__ void push(const T &item) {
+    uint32_t allocation = atomicAggInc(m_pending);
+    m_data[get_addr(allocation)] = item;
+  }
+
+  /*
+  __device__ __forceinline__ void push(T *items, int count)
```
phschaad, line 145: Was here before the PR, but what's this doing here?

phschaad (approved):
LGTM, minor comments to consider.

PR #1287, opened 2023-06-27
Fix writes to nested definitions

alexnick83:
Fixes a bug that renders subscripted accesses to data defined inside a nested scope (e.g., a Map) to full-range accesses to the same data.


acalotoiu (approved):
LGTM. Thanks for adding a new test to test the functionality!

PR #1288, opened 2023-06-28
Added mpi4py send/recv support and alltoall library node

Com1t:
## Summary
Added mpi4py send/recv support and alltoall library node

## Description
Added replacements and unit tests for mpi4py.MPI.COMM_WORLD.Send and mpi4py.MPI.COMM_WORLD.Recv.
Implemented the library node and unit test for the alltoall collective.

### DIFF
--- a/dace/frontend/common/distr.py
+++ b/dace/frontend/common/distr.py
@@ -256,7 +256,7 @@ def _Reduce(pv: 'ProgramVisitor',
 
 @oprepo.replaces('mpi4py.MPI.COMM_WORLD.Alltoall')
 @oprepo.replaces('dace.comm.Alltoall')
-def _allreduce(pv: 'ProgramVisitor',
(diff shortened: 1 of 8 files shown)


alexnick83 (changes requested):
Looks good. Let's also add to this PR a replacement for AlltoAll.

tbennun:
@alexnick83 wouldn’t we rather use replaces_method of the mpi4py communicator object instead?

alexnick83:
> @alexnick83 wouldn’t we rather use replaces_method of the mpi4py communicator object instead?

That would make sense only if the communicator Python object existed. We should prefer creating the communicators in C directly. Using a preexisting communicator could be interesting and it may be possible; mpi4py returns the Fortran ID of the communicators. However, I have not done any tests to see if passing this integer to C and calling `MPI_Comm_f2c` returns a working handle. Technically it should work.

PR #1291, opened 2023-06-30
GPUTransform fixes and refactoring

alexnick83:


acalotoiu (approved):
LGTM

alexnick83:
> LGTM

I was thinking that more work could be done here but since it passes the tests we could merge it and leave the refactoring for another PR. What do you think?

acalotoiu:
> > LGTM
> 
> I was thinking that more work could be done here but since it passes the tests we could merge it and leave the refactoring for another PR. What do you think?

agreed

PR #1294, opened 2023-07-04
Externally-managed memory lifetime

tbennun:
This PR is important for interoperability with other frameworks, whose memory management is external. It provides an allocation lifetime and generates two functions to get required sizes and set pointers externally.


dace/codegen/targets/framecode.py
acalotoiu, line 1: This requires more explanation IMO.

acalotoiu (approved):
LGTM, only one small instance where more explanation is necessary. Can be merged directly after adding it.

PR #1297, opened 2023-07-05
Refactoring Frontend's Variable-Access Visitors

alexnick83:
This PR refactors the frontend's visitors responsible for access to variables/data (`visit_Name`, `visit_Subscript`, etc.). The intent is to unify handling of read and write accesses in the assignment visitors to simplify supporting Structures.


acalotoiu (approved):
Looks good to me

PR #1300, opened 2023-07-07
Fix issue where multidimensional arrays with a singleton dimension sometimes fail

computablee:
Encountered a bug where multidimensional arrays with a singleton dimension can result in a failing SDFG. @tbennun guided me on the bug fix, and has said he'll write the unit tests. Leaving this as a draft until appropriate unit tests can be written.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -482,6 +482,7 @@ def add_indirection_subgraph(sdfg: SDFG,
         code = "lookup = {arr}[{index}]"
 
     newsubset = [r[0] if isinstance(r, tuple) else r for r in newsubset]
+    newsubset = [s for s, shp in zip(newsubset, array.shape) if shp != 1]
     if ind_entry:  # Amend newsubset when a range is indirected
         for i, idx in enumerate(nonsqz_dims):
             newsubset[idx] = '__i%d' % i


alexnick83:
@tbennun Isn't this a code generation error? If the data being indirected has n dimensions, then the lookup index should also have n indices. The issue seems to be in `DaCeKeywordRemover._subscript_exr`:
```python
            # Pointer to a single element can use all strides
            is_scalar = not isinstance(dtype, dtypes.pointer)
            if is_scalar or data._prod(subset_size) != 1:
                strides = [
                    s for i, s in enumerate(strides) if i not in indexdims and not (s == 1 and subset_size[i] == dimlen)
                ]
```
Shouldn't the scalar detection be `if is_scalar or data._prod(subset_size_ == 1` (equality)? I may have misunderstood what this code is doing and it is squeezing the strides on purpose but we need to somehow document this. Still, expecting squeezed data in the code generator looks like a dangerous design.

tbennun:
@alexnick83 unfortunately not. We thought so too at first and then found out that a memlet with [0, …] is already fed into the tasklet, which squeezes that dimension. So the tasklet mismatches the incoming memlet already in the frontend

alexnick83:
> @alexnick83 unfortunately not. We thought so too at first and then found out that a memlet with [0, …] is already fed into the tasklet, which squeezes that dimension. So the tasklet mismatches the incoming memlet already in the frontend

Then the test is not showcasing the problem (and showing a code generator problem instead). Array `a` has shape 1, 2, 3; so the first index will be 0 anyway. I tried the test in master and the access is not squeezed, all dimensions are in the memlet.

tbennun:
Superseded by #1302

PR #1307, opened 2023-07-10
Fixes Intemediate Node Removal in SubgraphFusion

alexnick83:
The PR fixes removal of intermediate nodes to non-local data in SubgraphFusion, taking into account dependencies among those nodes.


acalotoiu (approved):
Looks good to me.

PR #1309, opened 2023-07-11
Cherry-picks tensor-related libraries/replacements

alexnick83:
Ports TensorTranspose and TensorDot to the master branch to continue developing tensor-related libraries.


tbennun (changes requested):
This all looks good except for one thing: I think tensor transpose should live somewhere else (standard/blas/linalg) rather than a ttranspose library of its own

tbennun (approved):
Nice

PR #1312, opened 2023-07-12
Fix MapFusion with NestedSDFGs

alexnick83:
This PR:
- Adds a validation check for edges/Memlet paths and NestedSDFGs; if a NestedSDFG is src/dst of a Memlet path, then the dst/src must be an AccessNode.
- Amends MapFusion to create new intermediate data, even when the access is Scalar, when connecting one or more NestedSDFGs.


phschaad (approved):
LGTM, thanks!

PR #1316, opened 2023-07-13
Gemm, Gemv: Support symbolic expressions

lukastruemper:


dace/libraries/blas/nodes/gemm.py
```diff
@@ -1010,15 +1010,30 @@ def validate(self, sdfg, state):
             if dst_conn == '_a':
                 subset = dc(memlet.subset)
                 subset.squeeze()
-                size0 = subset.size()
+                size0 = []
```
alexnick83, line 1013: How about we add to `dace.symbolic.evaluate` a `throw_error` flag with default value `True.` When set to `False,` it returns the symbolic expression instead of raising `TypeError.` Then you could rewrite this (and everything below) as follows:
```python
size0 = dace.symbolic.evaluate(subset, sdfg.constants, throw_error=False)
```
I will bring it up in the next meeting.

alexnick83 (commented):
It looks good, but is it really solving the underlying issue? It may be better to change validation to throw warnings instead of errors if the sizes don't match (but potentially could, depending on what values symbols take). I will bring it up in the next meeting because I would like everyone's opinion.

I will not approve this yet, but ping me if it is blocking you.

dace/libraries/blas/nodes/gemm.py
phschaad: Using `dace.symbolic.simplify` with a symbols map sounds like an elegant solution for this. The `throw_error` flag may be a bit confusing imo.

phschaad (approved):
Also generally LGTM. Other point as already discussed between the two of you - I agree.

tbennun:
I don't think this is the right solution. Instead, we should warn on those indeterminate cases rather than error out.
In SymPy:
* `M == M - 3` returns False
* `M == M` returns True
* `M == N` returns `M == N` aka indeterminate

if we check `(size1 == size2) == False` (note the `== False` rather than `is False` because SymPy is SymPy), that would be preferable.

alexnick83:
@lukastruemper @tbennun @tbennun 

I made #1321 as an alternative. If you agree, we can merge it, and Lukas can continue adding tests and amending the validation methods.

phschaad:
> @lukastruemper @tbennun @phschaad
> 
> I made #1321 as an alternative. If you agree, we can merge it, and Lukas can continue adding tests and amending the validation methods.

Very nice, I'll be glad to review too once ready

PR #1321, opened 2023-07-15
Relaxed Shape/Size Equality Checks

alexnick83:
This PR introduces a new equality-check method, mainly aimed at dimension/mode length checks. The new method returns `True` if the two arguments are equal, `False` if they are unequal, and `None` if the result cannot be determined. The PR also introduces example usages in matrix-multiplication-related functionality, utilizing the method to print warnings instead of throwing errors when matrix multiplication could be valid depending on the values symbols take.


phschaad (approved):
LGTM, thanks

PR #1323, opened 2023-07-18
Proposal for resolving benign validation errors when auto-optimizing for non-CPU devices

alexnick83:
A standard workflow for optimizing programs to run on non-CPU devices involves instrumenting smaller kernels assuming that the inputs and output (non-transient data) are already on the device, i.e., no copies from the host to the device and back are needed. This is usually achieved by changing the non-transient data's storage to, e.g., GPU or FPGA global memory and then applying the appropriate device-specific transformation. However, when auto-optimizing, many other transformations are executed before GPU/FPGATransform. During this process, the SDFG is most likely invalid because the non-transient data on the device may appear to be accessed in the host's context, resulting in validation errors. Such errors are benign since the device-specific transformations will appropriately update schedules and storages, and they will introduce any needed copies.

This PR proposes a quick way to address the above issue. It introduces a new configuration parameter, `check_accessibility`, under the `experimental` category, having by default the value `True`. The fix consists of two components:
- The `dace.sdfg.validation._accessible` helper method checks the parameter's value and immediately returns `True` if the parameter is set to `False`.
- The auto-optimizer sets the parameter to `False`, if the selected device is other than CPU. It restores the original value of the parameter (as set by the user) just before applying GPU or FPGATransform.

### DIFF
--- a/dace/config_schema.yml
+++ b/dace/config_schema.yml
@@ -898,6 +898,13 @@ required:
                 title: Undefined Symbol Check
                 description: >
                     Check for undefined symbols in memlets during SDFG validation.
+            
+            check_accessibility:
+                type: bool
+                default: True
+                title: Accessibility Check
+                description: >
+                    Check if data containers can be accessed in the SDFG contexts with corresponding AccessNodes.
 
     #############################################
     # Features for unit testing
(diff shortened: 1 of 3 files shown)


tbennun:
@alexnick83 shall we close this?

BenWeber42:
Hi, thanks a lot for the excellent description. It sounds like this PR tends to be more of a work-around than a proper fix. Is this issue still a priority?

alexnick83:
I think this is superseded by PR #1405 but the correct workflow should be documented.

PR #1325, opened 2023-07-20
state.read_and_write_sets() ignore reads directly after writes

Sajohn-CH:
`read_and_write_sets()` filters out reads which are written directly before to the same AccessNode. Will only work if the read and write memlet go to the same AccessNode

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -298,11 +298,10 @@ class StateGraphView(object):
 
         # Get scopes
         for node, scopenodes in sdc.items():
-            scope_exit_nodes = [v for v in scopenodes if isinstance(v, nd.ExitNode)]
             if node is None:
                 exit_node = None
             else:
-                exit_node = next(iter(scope_exit_nodes))
+                exit_node = next(v for v in scopenodes if isinstance(v, nd.ExitNode))
             scope = ScopeTree(node, exit_node)
             result[node] = scope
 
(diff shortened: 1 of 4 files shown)


Sajohn-CH:
I also had to change `RefineNestedAccess` to check if a memlet is in the read and write set before refining it, as it expects a connector which is only there it it is in the read_and_write_sets

dace/sdfg/state.py
```diff
@@ -298,10 +298,11 @@ def scope_tree(self) -> 'dace.sdfg.scope.ScopeTree':
 
         # Get scopes
         for node, scopenodes in sdc.items():
+            scope_exit_nodes = [v for v in scopenodes if isinstance(v, nd.ExitNode)]
```
alexnick83, line 301: Does this make any meaningful difference, or is it for debugging? If it is only for debugging, I would suggest, after completing the PR, switching it to how it was before because it is somewhat slower.

dace/sdfg/state.py
```diff
@@ -502,13 +503,22 @@ def _read_and_write_sets(self) -> Tuple[Dict[AnyStr, List[Subset]], Dict[AnyStr,
             # is read is not counted in the read set
             for n in utils.dfs_topological_sort(sg, sources=sg.source_nodes()):
                 if isinstance(n, nd.AccessNode):
-                    for e in sg.in_edges(n):
+                    in_edges = sg.in_edges(n)
+                    out_edges = sg.out_edges(n)
+                    # Filter out memlets which go out but the same data is written to the AccessNode by another memlet
+                    for out_edge in out_edges:
+                        for in_edge in in_edges:
+                            if in_edge.data.data == out_edge.data.data and \
```
alexnick83, line 511: I suggest using parentheses rather than invertest slash, for example:
```python
if (in_edge.data.data == out_edge.data.data and
        in_edge.data.dst_subset.covers(out_edge.data.src_subset)):
    out_edges.remove(out_edge)
```

alexnick83 (changes requested):
It looks pretty good; only minor comments. Also, add your name in the [`AUTHORS` file](https://github.com/Sajohn-CH/dace/blob/bb00fea35013b66c153d89c5ba31ede70c0120d2/AUTHORS#L39)

dace/sdfg/state.py
Sajohn-CH: Was for debugging and switched it back

alexnick83 (approved):

PR #1330, opened 2023-07-25
Add `CPU_Persistent` map schedule (OpenMP parallel regions)

tbennun:
This PR adds the ability to generate per-thread code without a for loop, using `#pragma omp parallel`.

It inherits all the OpenMP properties that the parallel for has (e.g., schedule, number of threads), and internal OpenMP loops detect that they are nested in a parallel region.

It also allows the user to obtain the current thread ID (i.e., the map parameter) and the total number of runtime threads via the reserved `__omp_num_threads` symbol.


acalotoiu (approved):
LGTM

PR #1333, opened 2023-07-26
Validate NestedSDFG inout connectors

alexnick83:
This PR adds an extra validation check for NestedSDFGs: connectors that are both input and output must connect to the same arrays in the parent SDFG.

EDIT: The PR also addresses an issue where SubgraphFusion leads to SDFGs that fail the above validation check.


acalotoiu (changes requested):
LGTM, this is a very important fix! One small request: please add explanatory comments to the functions added to utils, as they might be useful beyond the scope of this PR.

acalotoiu (approved):
Thank you!

PR #1336, opened 2023-07-27
Fix some underlying issues with tensor core sample

computablee:
After messing with the tensor core example some, myself and @tbennun noticed some issues with the sample. This PR fixes a few issues.

1. The storage types (e.g., `Accumulator`, `TensorCore_*`) have been added to `defined_vars`. In some cases not demonstrated in `hgemm`, the codegen generates improper passes to functions. By setting the type to `DefinedType.Stream`, we ensure that the WMMA fragments are always passed as an object reference.
2. The frontend is greatly redone and simplified, changing the use of `@replaces` to simple `dace.tasklet`s. As with 1, certain situations not exposed with `hgemm` showed that memlets are not always properly generated with the `frag_fill` and `wmma` functions. Changing their implementation to `dace.tasklet` ensures that memlets are now properly generated in all use cases, along with simplifying their implementation to a large degree.

### DIFF
--- a/samples/codegen/tensor_cores.py
+++ b/samples/codegen/tensor_cores.py
@@ -27,6 +27,7 @@ from dace.transformation.interstate import GPUTransformSDFG
 from dace.sdfg.graph import MultiConnectorEdge
 from dace.sdfg.state import StateSubgraphView
 from dace.codegen.prettycode import CodeIOStream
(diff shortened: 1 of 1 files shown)


tbennun (approved):
LGTM, I would recommend to change the name `Stream` in DefinedTypes to `Object` or something that would make more sense.

PR #1338, opened 2023-07-30
Remove unused global data descriptor shapes from arguments

tbennun:
This PR allows symbols that exist only in non-transient arrays to generate code successfully. This avoids the case of unused program arguments and enables compilation of more programs, while retaining symbolic analysis.

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -243,15 +243,26 @@ class Data:
         """Returns a string for a C++ function signature (e.g., `int *A`). """
         raise NotImplementedError
 
-    @property
-    def free_symbols(self) -> Set[symbolic.SymbolicType]:
-        """ Returns a set of undefined symbols in this data descriptor. """
+    def used_symbols(self, all_symbols: bool) -> Set[symbolic.SymbolicType]:
+        """
+        Returns a set of symbols that are used by this data descriptor.
+
+        :param all_symbols: Include not-strictly-free symbols that are used by this data descriptor,
+                            e.g., shape and size of a global array.
+        :return: A set of symbols that are used by this data descriptor. NOTE: The results are symbolic
+                 rather than a set of strings.
+        """
         result = set()
-        if self.transient:
+        if self.transient or all_symbols:
             for s in self.shape:
                 if isinstance(s, sp.Basic):
                     result |= set(s.free_symbols)
         return result
+
+    @property
+    def free_symbols(self) -> Set[symbolic.SymbolicType]:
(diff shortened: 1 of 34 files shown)


dace/data.py
```diff
@@ -689,20 +701,23 @@ def as_arg(self, with_types=True, for_call=False, name=None):
...
         for s in self.strides:
             if isinstance(s, sp.Expr):
                 result |= set(s.free_symbols)
-        if isinstance(self.total_size, sp.Expr):
-            result |= set(self.total_size.free_symbols)
         for o in self.offset:
             if isinstance(o, sp.Expr):
                 result |= set(o.free_symbols)
-
+        if self.transient or all_symbols:
```
alexnick83, line 712: Why is the transient check only here? What about strides and offset just above?

alexnick83 (commented):
Some questions on the selected implementation.

dace/sdfg/sdfg.py
```diff
@@ -1290,14 +1291,22 @@ def free_symbols(self) -> Set[str]:
...
 
-        # Start with the set of SDFG free symbols
-        free_syms |= set(self.symbols.keys())
-
-        # Exclude data descriptor names and constants
+        # Exclude data descriptor names, constants, and shapes of global data descriptors
+        not_strictly_necessary_global_symbols = set()
         for name, desc in self.arrays.items():
             defined_syms.add(name)
+            if not desc.transient:
```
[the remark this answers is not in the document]
tbennun: You’re right and I started to rewrite it as such.

dace/data.py
tbennun: Strides and offset are always necessary if you have any memlet (a[i*stride] will be generated in the code).

dace/memlet.py
```diff
@@ -176,15 +177,16 @@ def to_json(self):
     @staticmethod
     def from_json(json_obj, context=None):
         ret = Memlet()
-        dace.serialize.set_properties_from_json(ret,
-                                                json_obj,
-                                                context=context,
-                                                ignore_properties={'src_subset', 'dst_subset', 'num_accesses', 'is_data_src'})
-        
+        dace.serialize.set_properties_from_json(
```
alexnick83, line 180: Unrelated to the PR, but why is YAPF constantly switching the formatting of such lines?

alexnick83 (approved):
It appears the ssa-symbol tests didn't need to be amended after all. This makes more sense since this PR should not change the result of `free_symbols`.

PR #1341, opened 2023-08-02
Validation improvements for Scalars.

### LINKED ISSUE #1334: Validation does not catch transient=False on scalar result in nsdfg
**Describe the bug**
The storage of result data in a nested-SDFG should be non transient. I am referring to the array or scalar node with the same name as an output connector of the nested-SDFG. Validation is able to catch when `transient=False` is set on array storage, but not when it is set on scalar storage. Validation will silently pass, and simplify will produce a wrong SDFG (empty state in nested-SDFG).

**To Reproduce**
I applied some modification to the example code samples/sdfg_api/nested_states.py
```
diff --git a/samples/sdfg_api/nested_states.py b/samples/sdfg_api/nested_states.py
index b489b94e1..25c845fe3 100644
--- a/samples/sdfg_api/nested_states.py
+++ b/samples/sdfg_api/nested_states.py
@@ -27,9 +27,9 @@ sub_sdfg = dace.SDFG('nested_sub')
 
 # Declare arrays for nested SDFG (the names can differ from the top-level SDFG)
 # In this case, we read only one element out of the full arrays
-sub_sdfg.add_array('sA', [1], dace.float32)
-sub_sdfg.add_transient('sB', [1], dace.float32)
-sub_sdfg.add_array('sC', [1], dace.float32)
+sub_sdfg.add_scalar('sA', dace.float32)
+sub_sdfg.add_scalar('sB', dace.float32, transient=True)
+sub_sdfg.add_scalar('sC', dace.float32, transient=True)
 
 # Create nested states
 state0 = sub_sdfg.add_state('subs0')
@@ -57,6 +57,8 @@ state.add_memlet_path(nsdfg, mx, Aout, memlet=dace.Memlet(data='A', subset='k'),
 
 # Validate correctness of SDFG
 sdfg.validate()
+sdfg.simplify()
+sdfg.view()
 
 ######################################
 if __name__ == '__main__':
```

**Expected behavior**
Validation should catch the wrong transient flag, to avoid the error in simplify pass.

I have checked the source code and I found this TODO in the method `NestedSDFG::validate()`
```
        for dname, desc in self.sdfg.arrays.items():
            # TODO(later): Disallow scalars without access nodes (so that this
            #              check passes for them too).
            if isinstance(desc, data.Scalar):
                continue
            if not desc.transient and dname not in connectors:
                raise NameError('Data descriptor "%s" not found in nested SDFG connectors' % dname)
            if dname in connectors and desc.transient:
                raise NameError('"%s" is a connector but its corresponding array is transient' % dname)
```
The problem is that allowing `transient=True` on scalar result will break simplify (see below).

**Screenshots**
The result of simplify will be an empty state inside the nested-SDFG:
[nested_main.pdf](https://github.com/spcl/dace/files/12183144/nested_main.pdf)

alexnick83:
- [x] Removes validation skip for Scalars in nested SDFGs.

Expected to fix #1334

### DIFF
--- a/dace/sdfg/nodes.py
+++ b/dace/sdfg/nodes.py
@@ -636,10 +636,6 @@ class NestedSDFG(CodeNode):
                     f'Connector "{conn}" was given but is not a registered data descriptor in the nested SDFG. '
                     'Example: parameter passed to a function without a matching array within it.')
         for dname, desc in self.sdfg.arrays.items():
-            # TODO(later): Disallow scalars without access nodes (so that this
-            #              check passes for them too).
-            # if isinstance(desc, data.Scalar):
-            #     continue
             if not desc.transient and dname not in connectors:
                 raise NameError('Data descriptor "%s" not found in nested SDFG connectors' % dname)
             if dname in connectors and desc.transient:
--- a/dace/sdfg/nodes.py
+++ b/dace/sdfg/nodes.py
@@ -628,8 +628,8 @@ class NestedSDFG(CodeNode):
         for dname, desc in self.sdfg.arrays.items():
             # TODO(later): Disallow scalars without access nodes (so that this
             #              check passes for them too).
-            if isinstance(desc, data.Scalar):
-                continue
+            # if isinstance(desc, data.Scalar):
+            #     continue
             if not desc.transient and dname not in connectors:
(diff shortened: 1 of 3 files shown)


tbennun (approved):
Good

PR #1342, opened 2023-08-03
Fortran fix memlet indices

Sajohn-CH:
Fixes error with memlet indices. Parser made implicit assumption about position of index a loop iterations over. Also added a testcase.


Sajohn-CH:
@acalotoiu

acalotoiu (approved):
LGTM. Thank you!

PR #1343, opened 2023-08-03
Add Object to defined types in code generation and some documentation

tbennun:


acalotoiu (approved):
LGTM. Still not a fan of having references, and we might need to flag/discuss how to write transformations in their presence.

tbennun:
Not the same kind of references :) these are c++ by-reference parameters. Before that we could only do scalars by value (which would copy the object around) or arrays by pointer (in which case we would dereference that object which would fail compilation)

PR #1347, opened 2023-08-04
Added mpi4py communicator split support

Com1t:
## Summary
Added mpi4py communicator split support by creating a new library node.

## Description
Added comm split library node, test, and replacement in frontend.
In this library node, the pseudo-writing edge can be used to ensure the order of mpi nodes.


BenWeber42:
I take it this is also a result of GSoC'23? What's the status here?
Looks like this would be a nice addition to DaCe?

Issue #1351, opened 2023-08-18
More than one HBM bank with automatic SDFG generation


gabrielrodcanal:
Hi there,

**Describe the bug**
I am building an SDFG from a kernel ported to Python like this:
```
sdfg = kernel.to_sdfg()
sdfg.apply_transformations(FPGATransformSDFG)
sdfg(u=u,v=v)
```

By default, this connects every kernel port to HBM[0]. Is there any way to modify this? I want to run this kernel efficiently on big sizes, and for that I need to be able to map the kernel ports to memory banks other than HBM[0].

**To Reproduce**
1. Save the following kernel as `issue_kernel.py`
```
import dace
import numpy as np
from dace.transformation.interstate import FPGATransformSDFG

DIM_X = 256
DIM_Y = 256
DIM_Z = 128


@dace.program
def issue_kernel(u:dace.float64[DIM_X,DIM_Y,DIM_Z],v:dace.float64[DIM_X,DIM_Y,DIM_Z]):
    nx = DIM_X
    ny = DIM_Y
    nz = DIM_Z

    for i in range(nx):
        for j in range(ny):
            for k in range(nz):
                u[i,j,k] = 10.0
                v[i,j,k] = 20.0



if __name__ == "__main__":
    u = np.array([[[0 for k in range(DIM_X)] for j in range(DIM_Y)] for i in range(DIM_Z)], dtype=np.double)
    v = np.array([[[0 for k in range(DIM_X)] for j in range(DIM_Y)] for i in range(DIM_Z)], dtype=np.double)

    sdfg = issue_kernel.to_sdfg()
    sdfg.apply_transformations(FPGATransformSDFG)
    sdfg(u=u,v=v)
```
2. Run `python3 issue_kernel.py`.

The `link.cfg` file under `.dacecache/issue_kernel/src/xilinx/device/` doesn't connect the kernel ports to memory banks. Hence, every port is connected to HBM[0]. The host code is generated accordingly, so although we can get a bitstream properly connected modifying this file manually, the run will fail with error `memidx == -1`.

How can we generate this connectivity without altering the kernel code? I've seen this is usually done by building the SDFG manually: https://github.com/spcl/dace/blob/master/samples/fpga/gemm_systolic_vectorized.py. I would rather not do this.

Best,
Gabriel

TizianoDeMatteis:
Hi @gabrielrodcanal,

[...]

PR #1352, opened 2023-08-21
Have memory type as argument for fpga auto interleave

TizianoDeMatteis:
Allow the user to specify the targeted type of memory when performing automatic buffer interleaving for FPGA.

Addresses issues raised in #1351


tbennun (approved):
LGTM

PR #1353, opened 2023-08-21
Trivial map elimination init

Sajohn-CH:
The EliminateTrivialMap transformation fails to properly connect the content of the map if it is a write only map and the map to be removed is nested inside another map. This is as the current solution uses memlet_path which does not work on the empty memlets going away from the map entry.

This adds a testcase and solution which seems to work for my specific needs but are not polished and might not work in all cases.


dace/transformation/dataflow/trivial_map_elimination.py
```diff
@@ -48,12 +49,16 @@ def apply(self, graph, sdfg):
...
         if len(remaining_ranges) == 0:
             # Redirect map entry's out edges
+            write_only_map = False
             for edge in graph.out_edges(map_entry):
                 path = graph.memlet_path(edge)
                 index = path.index(edge)
 
-                # Add an edge directly from the previous source connector to the destination
-                graph.add_edge(path[index - 1].src, path[index - 1].src_conn, edge.dst, edge.dst_conn, edge.data)
+                if len(path) > 1:
```
alexnick83, line 57: Technically, you could also use [`Memlet.is_empty`](https://github.com/spcl/dace/blob/f99bbcabdf993d5c969094a8c803bf6037e835b5/dace/memlet.py#L218), which indicates that the edge does not carry/read data.

dace/transformation/dataflow/trivial_map_elimination.py
```diff
@@ -48,12 +49,16 @@ def apply(self, graph, sdfg):
...
+                    # Add an edge directly from the previous source connector to the destination
+                    graph.add_edge(path[index - 1].src, path[index - 1].src_conn, edge.dst, edge.dst_conn, edge.data)
+                else:
+                    write_only_map = True
```
alexnick83, line 61: It may be unlikely, but one could construct on purpose a MapScope with at least two input edges, one empty and one non-empty. I don't think that this violates semantics. In such a case, `write_only_map` would be `True,` although the Map would not really be write-only. Perhaps the boolean logic can be rewritten so that the default value is `True` and is switched to `False` when a non-empty edge is detected. I think this would work for all cases.

alexnick83 (changes requested):
LGTM, but maybe check the case I described where the MapEntry has two output edges, one empty and one non-empty.

dace/transformation/dataflow/trivial_map_elimination.py
Sajohn-CH: I changed it as suggested and added a testcase in [this commit](cdba4e45ce95fcaf2c515f00e191096432a8cd31). Though I think there should be an argument if this is wanted. As my code before that commit leads to the situation where the empty edge is added again. One could argue that this is to be expected as it was there before and it is not the job the this transformation to remove these edges. In that case one would probably also have to consider the case if there are any empty memlets as this code assumes there is only one. Which sparks the general question what should happen with them.

acalotoiu: I propose we leave it at the version it is now - assuming it covers current cases.

alexnick83 (approved):

PR #1362, opened 2023-09-04
Bugfix: Duplicate memlet error with add_memlet_path

lukastruemper:


tbennun:
I think it's the API user's responsibility that a unique memlet is passed into `add_memlet_path`. Who is calling it with the same object more than once?

tbennun:
To be clear, this would create slowdowns as for almost all the `add_memlet_path` calls, a new memlet is created for them.

PR #1363, opened 2023-09-04
Improvements to work depth analysis

hodelcl:
This PR adds various improvements to the work depth analysis for SDFGs. These improvements include:

- **Performance improvements**: Analysis now runs in minutes compared to hours (or days?) on big SDFGs. Biggest performance gain comes from optimising the Max expressions which occur from branches. Instead of `Max(x + a, x + b)`, we now formulate it as `x + Max(a, b)`. If `x` also contains a max expression, the performance gains are even bigger, since this generally allows for more Max expressions to be evaluated symbolically.
- **Assumptions**: It is now possible for the user to pass assumptions for the size of the symbols to the analysis. For example one can assume that `N>5` and hence the analysis evaluates `Max(N, 5)` to `N`. Sympy does not support such assumptions out of the box, hence this PR includes `assumptions.py`, which handles all kinds of equal / greater / lesser assumptions. The user can enter the assumptions directly into the VS code extension as implemented by [this other PR](https://github.com/spcl/dace-vscode/pull/234).
- **wcr edges**: Priorly, wcr edges were ignored by the analysis, which meant that reductions had work and depth of 0. Now, wcr edges are included in the analysis and a wcr edge with _N_ writes to the same memory location adds _N_ to the depth (and work respectively).
- Some more minor improvements.


tests/sdfg/work_depth_tests.py
```diff
@@ -1,14 +1,18 @@
...
+from dace.sdfg.work_depth_analysis.assumptions import ContradictingAssumptions
 import sympy as sp
 
 from dace.transformation.interstate import NestSDFG
 from dace.transformation.dataflow import MapExpansion
 
+from pytest import raises
+
 # TODO: add tests for library nodes (e.g. reduce, matMul)
+# TODO: add tests for average parallelism
```
phschaad, line 15: Note: please keep those in mind, would be great to have some tests for these in the next PR.

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -5,7 +5,7 @@
 import argparse
 from collections import deque
 from dace.sdfg import nodes as nd, propagation, InterstateEdge
-from dace import SDFG, SDFGState, dtypes
+from dace import SDFG, SDFGState, dtypes, int64
```
phschaad, line 8: `dtypes` is already imported, so using `dtypes.int64` instead is preferred if possible.

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -267,11 +250,10 @@ def tasklet_work(tasklet_node, state):
 
 
 def tasklet_depth(tasklet_node, state):
-    # TODO: how to get depth of CPP tasklets?
-    # For now we use depth == work:
     if tasklet_node.code.language == dtypes.Language.CPP:
+        # For now we simply take depth == work for CPP tasklets.
```
phschaad, line 254:
```suggestion
        # Depth == work for CPP tasklets.
```

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -282,19 +264,37 @@ def tasklet_depth(tasklet_node, state):
 
 
 def get_tasklet_work(node, state):
-    return tasklet_work(node, state), -1
+    return sp.sympify(tasklet_work(node, state)), sp.sympify(-1)
```
phschaad, line 267: There's a ton of these `sympify` calls all over the place in this file. Are these really necessary? Can Sympy not handle it if it has to deal with, say, an integer `0` or `-1`, or a string?

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -304,6 +304,11 @@ def sdfg_work_depth(sdfg: SDFG, w_d_map: Dict[str, Tuple[sp.Expr, sp.Expr]], ana
     :param w_d_map: Dictionary which will save the result.
     :param analyze_tasklet: Function used to analyze tasklet nodes.
     :param symbols: A dictionary mapping local nested SDFG symbols to global symbols.
+    :param detailed_analysis: If True, detailed analysis gets used. For each branch, we keep track of its condition
```
phschaad, line 307: Due to the performance impact, I would consider making this parameter default to `False` in the call signature. Thoughts @acalotoiu ?

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -430,7 +495,14 @@ def scope_work_depth(state: SDFGState,
     this can be done in linear time by traversing the graph in topological order.
 
     :param state: The state in which the scope to analyze is contained.
-    :param sym_map: A dictionary mapping symbols to their values.
+    :param w_d_map: Dictionary saving the final result for each SDFG element.
+    :param analyze_tasklet: Function used to analyze tasklets. Either analyzes just work, work and depth or average parallelism.
+    :param symbols: A dictionary mapping local nested SDFG symbols to global symbols.
+    :param detailed_analysis: If True, detailed analysis gets used. For each branch, we keep track of its condition
+    and work depth values for both branches. If False, the worst-case branch is taken. Discouraged to use on bigger SDFGs,
+    as computation time sky-rockets, since expression can became HUGE (depending on number of branches etc.).
```
phschaad, line 503: Same as above.

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -534,64 +631,129 @@ def scope_work_depth(state: SDFGState,
...
     """
     Analyze the work and depth of a state.
 
     :param state: The state to analyze.
     :param w_d_map: The result will be saved to this map.
     :param analyze_tasklet: Function used to analyze tasklet nodes.
     :param symbols: A dictionary mapping local nested SDFG symbols to global symbols.
+    :param detailed_analysis: If True, detailed analysis gets used. For each branch, we keep track of its condition
+    and work depth values for both branches. If False, the worst-case branch is taken. Discouraged to use on bigger SDFGs,
+    as computation time sky-rockets, since expression can became HUGE (depending on number of branches etc.).
```
phschaad, line 683: Same.

dace/sdfg/work_depth_analysis/work_depth.py
```diff
@@ -534,64 +631,129 @@ def scope_work_depth(state: SDFGState,
...
     :note: SDFGs should have split interstate edges. This means there should be no interstate edges containing both a
         condition and an assignment.
     :param sdfg: The SDFG to analyze.
     :param w_d_map: Dictionary of SDFG elements to (work, depth) tuples. Result will be saved in here.
-    :param analyze_tasklet: The function used to analyze tasklet nodes. Analyzes either just work, work and depth or average parallelism.
+    :param analyze_tasklet: Function used to analyze tasklet nodes. Analyzes either just work, work and depth or average parallelism.
+    :param assumptions: List of strings. Each string corresponds to one assumption for some symbol, e.g. 'N>5'.
+    :param detailed_analysis: If True, detailed analysis gets used. For each branch, we keep track of its condition
+    and work depth values for both branches. If False, the worst-case branch is taken. Discouraged to use on bigger SDFGs,
+    as computation time sky-rockets, since expression can became HUGE (depending on number of branches etc.).
```
phschaad, line 706: Same

dace/sdfg/work_depth_analysis/assumptions.py
```diff
@@ -0,0 +1,285 @@
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
+
+import sympy as sp
+from typing import Tuple, Dict
```
phschaad, line 4:
```suggestion
from typing import Dict
```

phschaad (changes requested):
Looks very good overall, small comments and questions only.

dace/sdfg/work_depth_analysis/work_depth.py
hodelcl: Sounds good to me!

hodelcl:
Accidentally pushed different work to this branch, sorry! Reverted it again.

phschaad (approved):
LGTM, thanks for addressing the comments!

PR #1366, opened 2023-09-06
Structure Support to NestedSDFGs and Python Frontend

alexnick83:
Adds basic support for nested data (Structures) to the Python frontend. It also resolves issues with the use of Structures in nested SDFG scopes (mostly code generation).

NOTE: This PR handles only CPU code generation and fixes issues with libraries and transformations discovered during testing. It doesn't handle GPU/FPGA code generation, which will be studied in subsequent PRs.


dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -17,6 +17,7 @@
 from dace.frontend.python import astutils
 from dace.sdfg import nodes, propagation, utils
 from dace.sdfg.graph import MultiConnectorEdge, SubgraphView
+from dace.sdfg.replace import replace_properties_dict
```
tbennun, line 20: no need

dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -628,7 +629,7 @@ def _modify_access_to_access(self,
                                                                   matching_edge.data,
                                                                   use_dst_subset=True)
                             new_memlet = in_memlet
-                            new_memlet.other_subset = out_memlet.dst_subset
+                            new_memlet.other_subset = out_memlet.subset
```
tbennun, line 632: Is that not less safe? Which case does this pertain to?

tests/npbench/weather_stencils/vadv_test.py
```diff
@@ -211,7 +211,9 @@ def run_vadv(device_type: dace.dtypes.DeviceType):
     return sdfg
 
 
-def test_cpu():
+def test_cpu(monkeypatch):
+    # NOTE: Serialization fails because of "k - k" expression simplified to "0"
+    monkeypatch.setenv("DACE_testing_serialization", 0)
```
tbennun, line 216: Did we not disable serialization in tests altogether?

tbennun (changes requested):
Overall looks good

dace/sdfg/replace.py
```diff
@@ -21,6 +21,12 @@ def _internal_replace(sym, symrepl):
 
     # Filter out only relevant replacements
     fsyms = set(map(str, sym.free_symbols))
+    # TODO/NOTE: Could we return the generated strings below as free symbols from Attr instead or ther will be issues?
```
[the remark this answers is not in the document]
alexnick83: If we have a symbolic expression `a.i.b`, then we want to consider the tokens `a`, `a.i`, and `a.i.b`. With the current design, it should not be possible to replace `i` with `j`. Instead, the replacement should be `a.i` with `a.j` or `c.j`, i.e., we need the full name.

The note is about investigating if the code (below) generating the tokens could be refactored to call free-symbols on `Attr` symbolic functions or issues will arise.

dace/sdfg/sdfg.py
```diff
@@ -1217,25 +1224,65 @@ def remove_node(self, node: SDFGState):
         return super().remove_node(node)
 
     def states(self):
-        """ Returns the states in this SDFG, recursing into state scope blocks. """
-        return list(self.all_states())
+        """ Alias that returns the nodes (states) in this SDFG. """
+        return self.nodes()
```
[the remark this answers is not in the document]
alexnick83: Yes, there were problems with the auto-merging in some files. I didn't notice this.

dace/transformation/interstate/sdfg_nesting.py
alexnick83: The problem arose in a (new) test, where it turned out that `dst_subset` was None. For outgoing edges (line 658 on the same file), we use `subset` (not `src_subset`), and I believe the original code above is just wrong.

tests/npbench/weather_stencils/vadv_test.py
alexnick83: It looks like we didn't make the PR yet.

dace/codegen/targets/cpu.py
```diff
@@ -620,16 +646,17 @@ def _emit_copy(
             #############################################
             # Corner cases
 
+            # NOTE: This looks obsolete but keeping it commented out in case tests fail.
```
phschaad, line 649: Looks like they pass

dace/codegen/targets/cpu.py
```diff
@@ -295,23 +303,40 @@ def declare_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, de
...
-        alloc_name = cpp.ptr(name, nodedesc, sdfg, self._frame)
+                       allocation_stream, allocate_nested_data: bool = True):
+        alloc_name = cpp.ptr(node.data, nodedesc, sdfg, self._frame)
         name = alloc_name
 
-        if nodedesc.transient is False:
+        tokens = node.data.split('.')
+        top_desc = sdfg.arrays[tokens[0]]
+        # NOTE: Assuming here that all Structure members share transient/storage/lifetime properties.
+        # TODO: Study what is needed in the DaCe stack to ensure this assumption is correct.
```
phschaad, line 313: Like with the other comments, if we cannot guarantee in the stack that this assumption is an invariant across our pipeline, we may want to verify and warn

phschaad (approved):
LGTM overall, see comments though - curious about thoughts

dace/libraries/standard/nodes/transpose.py
```diff
@@ -149,18 +151,21 @@ def expansion(node, state, sdfg):
...
+            alpha = f"{cast}dace::blas::BlasConstants::Get().Complex64Pone()"
         elif dtype == dace.complex128:
             func = "zomatcopy"
-            alpha = "dace::blas::BlasConstants::Get().Complex128Pone()"
+            cast = "(double*)"
+            alpha = f"{cast}dace::blas::BlasConstants::Get().Complex128Pone()"
         else:
             raise ValueError("Unsupported type for OpenBLAS omatcopy extension: " + str(dtype))
-        _, _, (m, n) = _get_transpose_input(node, state, sdfg)
+        # TODO: Add stride support
```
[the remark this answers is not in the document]
alexnick83: Yes, but it should go in a separate PR. I just noticed the issue due to a new test using transposition. Otherwise, it is unrelated tot his PR.

dace/sdfg/infer_types.py
```diff
@@ -381,6 +382,8 @@ def _get_storage_from_parent(data_name: str, sdfg: SDFG) -> dtypes.StorageType:
     parent_sdfg = parent_state.parent
 
     # Find data descriptor in parent SDFG
+    # NOTE: Assuming that all members of a Structure have the same storage type.
+    data_name = data_name.split('.')[0]
```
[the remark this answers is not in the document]
alexnick83: To clarify, this is a developer note and doesn't need to propagate to the user. Currently, the structure creation routines do not ensure common storage type, lifetime, etc. I am suggesting that this should be the case and, for the time being, this is enforced by ignoring the storage type, lifetime, etc., of the nested data. I wouldn't put a warning here; rather make it clear in the documentation (to be written when the nested data design starts to finalize).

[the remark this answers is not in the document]
alexnick83: Also, this may change. I discussed it a bit with Tal and didn't see any legitimate case where part of the structure should have a different storage type or lifetime than the rest, but maybe we have missed something.

dace/codegen/targets/cpu.py
```diff
@@ -620,16 +646,17 @@ def _emit_copy(
...
-                return
+            # if (isinstance(memlet.subset, subsets.Indices) and memlet.wcr is None
+            #         and self._dispatcher.defined_vars.get(vconn)[0] == DefinedType.Scalar):
+            #     stream.write(
+            #         "%s = %s;" % (vconn, self.memlet_ctor(sdfg, memlet, dst_nodedesc.dtype, False)),
+            #         sdfg,
+            #         state_id,
+            #         [src_node, dst_node],
+            #     )
+            #     return
```
tbennun, line 659:
```suggestion
```

dace/codegen/targets/framecode.py
```diff
@@ -671,7 +677,7 @@ def determine_allocation_lifetime(self, top_sdfg: SDFG):
                     for node in state.nodes():
                         if not isinstance(node, nodes.AccessNode):
                             continue
-                        if node.data != name:
+                        if node.data.split('.')[0] != name:
```
tbennun, line 680: use root_data?

dace/sdfg/sdfg.py
```diff
@@ -1220,22 +1227,62 @@ def states(self):
...
+                    yield from node.sdfg.all_sdfgs_recursive()
+
+    def all_edges_recursive(self):
+        """ Iterate over all edges in this SDFG, including state edges,
+            inter-state edges, and recursively edges within nested SDFGs,
+            returning tuples on the form (edge, parent), where the parent is
+            either the SDFG (for states) or a DFG (nodes). """
+        for e in self.edges():
+            yield e, self
+        for node in self.nodes():
```
tbennun, line 1254: Another merge issue with Phil's recent changes

tbennun (changes requested):
Looks good, one merge fix and minor changes

tbennun:
@alexnick83 there are two consistently failing tests in this PR (GPU).

alexnick83:
> @alexnick83 there are two consistently failing tests in this PR (GPU).

I don't see something in the PR that should directly cause these issues, but maybe the changes assisted in bringing to light pre-existing bugs.

Specifically, the `adi` test fails because `__tmp62` and `__tmp63` are not passed as arguments to the GPU kernels. The interesting thing here is that both of them have as **original** allocation scope the GPU kernel itself. However, because they are placed in GPU global memory instead of registers, the allocation scope changes to the corresponding State. I will try to fix the argument passing (some other PR may have broken it), but we should look into fixing the storage types.

PR #1367, opened 2023-09-08
Add offset normalization to Fortran frontend

mcopik:
This PR contributes a Fortran AST pass to normalize all array offsets. In addition, it fixes a few additional problems:
* Adds a pass to assign parent scope to each AST node.
* Adds a pass accumulating all variable declarations in a given scope.
* Fixes handling of Fortran array assignment when offsets are present.

@acalotoiu I'm going to do a few more tests covering more complex function invocations and 3D arrays. Please look when you can to verify this is the intended behavior.


dace/frontend/fortran/ast_transforms.py
acalotoiu, line 1: I think there can be other parents: modules.

acalotoiu (changes requested):
The parent of a node can also be a module no just a function, subroutine or program. The full complexity of Fortan in general is even broader, but modules we require support for directly.

mcopik:
> The parent of a node can also be a module no just a function, subroutine or program. The full complexity of Fortan in general is even broader, but modules we require support for directly.

@acalotoiu Fixed in commit `70c33dd913376cd87b46887da51da2b5d939f10f`, I also added a test verifying we catch the module as a parent node.

acalotoiu (approved):
LGTM now. Thanks!

PR #1369, opened 2023-09-14
Support attributes in symbolic expressions

tbennun:
I do not know how to make tests for that.

### DIFF
--- a/dace/codegen/targets/cpu.py
+++ b/dace/codegen/targets/cpu.py
@@ -299,9 +299,6 @@ class CPUCodeGen(TargetCodeGenerator):
         name = node.data
         alloc_name = cpp.ptr(name, nodedesc, sdfg, self._frame)
         name = alloc_name
-        # # NOTE: `expr` may only be a name or a sequence of names and dots. The latter indicates nested data and
-        # # NOTE: structures. Since structures are implemented as pointers, we replace dots with arrows.
-        # alloc_name = alloc_name.replace('.', '->')
 
         if nodedesc.transient is False:
             return
@@ -1184,9 +1181,6 @@ class CPUCodeGen(TargetCodeGenerator):
         if not types:
             types = self._dispatcher.defined_vars.get(ptr, is_global=True)
         var_type, ctypedef = types
-        # # NOTE: `expr` may only be a name or a sequence of names and dots. The latter indicates nested data and
-        # # NOTE: structures. Since structures are implemented as pointers, we replace dots with arrows.
-        # ptr = ptr.replace('.', '->')
 
         if fpga.is_fpga_array(desc):
             decouple_array_interfaces = Config.get_bool("compiler", "xilinx", "decouple_array_interfaces")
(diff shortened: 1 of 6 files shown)


alexnick83 (approved):
We will need updates for multi-nested Structures, but let's do it step by step.

dace/codegen/targets/cpu.py
```diff
@@ -299,9 +299,9 @@ def allocate_array(self, sdfg, dfg, state_id, node, nodedesc, function_stream, d
         name = node.data
         alloc_name = cpp.ptr(name, nodedesc, sdfg, self._frame)
         name = alloc_name
-        # NOTE: `expr` may only be a name or a sequence of names and dots. The latter indicates nested data and
-        # NOTE: structures. Since structures are implemented as pointers, we replace dots with arrows.
-        alloc_name = alloc_name.replace('.', '->')
+        # # NOTE: `expr` may only be a name or a sequence of names and dots. The latter indicates nested data and
+        # # NOTE: structures. Since structures are implemented as pointers, we replace dots with arrows.
+        # alloc_name = alloc_name.replace('.', '->')
```
tbennun, line 304:
```suggestion
```
@alexnick83

dace/codegen/targets/cpu.py
```diff
@@ -1184,9 +1184,9 @@ def memlet_definition(self,
         if not types:
             types = self._dispatcher.defined_vars.get(ptr, is_global=True)
         var_type, ctypedef = types
-        # NOTE: `expr` may only be a name or a sequence of names and dots. The latter indicates nested data and
-        # NOTE: structures. Since structures are implemented as pointers, we replace dots with arrows.
-        ptr = ptr.replace('.', '->')
+        # # NOTE: `expr` may only be a name or a sequence of names and dots. The latter indicates nested data and
+        # # NOTE: structures. Since structures are implemented as pointers, we replace dots with arrows.
+        # ptr = ptr.replace('.', '->')
```
tbennun, line 1189:
```suggestion
```

dace/codegen/tools/type_inference.py
```diff
@@ -405,6 +405,9 @@ def _infer_dtype(t: Union[ast.Name, ast.Attribute]):
 
 def _Attribute(t, symbols, inferred_symbols):
     inferred_type = _dispatch(t.value, symbols, inferred_symbols)
+    if (isinstance(inferred_type, dtypes.pointer) and isinstance(inferred_type.base_type, dtypes.struct) and
```
tbennun, line 408: @alexnick83 what about plain structs (i.e., not pointers)

alexnick83: Since we have decided that Structures will always resolve to a pointer to a struct, I think it is better to not handle that case (now), since it might be something unrelated to nested data and cause further confusion.

phschaad: Note: This assumption was valid for the time being, but I am now arguing to change that assumption. For cases where the type of an expression is being evaluated, where the `t.value` type of an Attribute expression is not merely a structure, but an array of structures (indexed through a Slice), the returned `inferred_type` is no longer a pointer but directly a plain struct.

Example: `my_array_of_structs[i].my_attribute`

Commit https://github.com/spcl/dace/commit/e2561ed5bda44808b910940093728de808213508 implements a handle for this additional case.

PR #1370, opened 2023-09-20
Speed up StateReachability pass for large state machines

tbennun:
The current algorithm in networkx is suboptimal. With this change, I noticed a 55x speedup in computing state reachability on realistic programs.

### DIFF
--- a/dace/transformation/passes/analysis.py
+++ b/dace/transformation/passes/analysis.py
@@ -42,15 +42,60 @@ class StateReachability(ppl.Pass):
             # The implementation below is faster
             # tc: nx.DiGraph = nx.transitive_closure(sdfg.nx)
 
-            for n, v in nx.all_pairs_shortest_path_length(sdfg.nx):
-                result[n] = set(t for t, l in v.items() if l > 0)
-                # Add self-edges
-                if n in sdfg.successors(n):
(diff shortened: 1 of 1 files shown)


tbennun:
@acalotoiu @mcopik tests fail because of an unrelated issue (has to do with fparser)

mcopik:
@tbennun Indeed, it looks like an incorrect dependency version. It's really strange because it should have popped up earlier. I will look into this.

mcopik:
@tbennun @acalotoiu It seems the module structure changed in `fparser` between `0.1.1` and `0.1.2`. In the latter, they introduced a sub-module for `Fortran2008`, which should work with this import structure. However, our requirements specify exactly the version `0.1.2` - I'm not sure why an older version would be installed.

mcopik:
@acalotoiu @tbennun Mystery solved - they released a new version two days ago, drastically changing the package structure. While `requirements.txt` specifies an exact match on `0.1.2`, our `setup.py` has a different dependency: `fparser >= 0.1.2`.

I think it's better to move forward with the newer version - I will propose a PR to resolve this issue.

acalotoiu (approved):
LGTM. Thank you! It is a great performance improvement :).

PR #1373, opened 2023-09-22
Added mpi4py RMA support 

Com1t:
## Summary
Added mpi4py RMA library nodes and replacements.

## Description
Implemented RMA transmission functions, such as put/get/accumulate.
And fence for active synchronization, lock/unlock/flush for passive synchronization.
Besides MPI functions, put/get/accumulate require an synchronization before and after it, thus a logical corrctness checker is introduced in `distr.py`.


alexnick83:
It is from GSoC'23. It needs some adjustments on the frontend based on internal discussions before we merge.

PR #1386, opened 2023-10-06
Python 3.12

alexnick83:


pytest.ini
```diff
@@ -14,6 +14,7 @@ markers =
     scalapack: Test requires ScaLAPACK (Intel MKL and OpenMPI). (select with '-m scalapack')
     datainstrument: Test uses data instrumentation (select with '-m datainstrument')
     hptt: Test requires the HPTT library (select with '-m "hptt')
+    py312: Test requires Python 3.12 or later (select with '-m "py312"')
```
tbennun, line 17: do we need this anymore, since the test is commented out?

tbennun (changes requested):

pytest.ini
alexnick83: Currently, it does nothing. It is still unclear to me why both approaches I tried (if/else with `sys.version_info` and the mark) failed to work. It looks like pytest parses (?) the code in any case and crashes because the new type alias statement is invalid (older version) Python code. I would say that, ultimately, we want this mark, but if it fails to work it is useless.

alexnick83: Removed.

tbennun (approved):

PR #1387, opened 2023-10-09
K caching changing strides

Sajohn-CH:
Adds two features used in my thesis
- Change the strides in arrays in a given SDFG
- Adapt subgraph fusion to better be able to perform K-caching


dace/sdfg/state.py
```diff
@@ -674,9 +675,16 @@ def arglist(self, defined_syms=None, shared_transients=None) -> Dict[str, dt.Dat
 
         # Add scalar arguments from free symbols
         defined_syms = defined_syms or self.defined_symbols()
+        # Fix for dynamic map inputs appearing in free symbols
+        from dace.sdfg.utils import dynamic_map_inputs
+        dynamic_inputs = set()
+        for node in self.nodes():
+            if isinstance(node, nd.MapEntry):
+                dynamic_inputs.update([e.dst_conn for e in dynamic_map_inputs(self.graph, node)])
+        free_symbols = self.free_symbols - dynamic_inputs
```
Sajohn-CH, line 684: I don't remember adding something in here but without these changes I get an undefined symbol error. Same goes for the changes below. I didn't test of all of them or only one is required.

dace/config_schema.yml
```diff
@@ -813,6 +813,12 @@ required:
         default: false
         title: Debug printing
         description: Enable verbose printouts.
+    
+    debugpass:
+        type: bool
+        default: false
+        title: Debug transformation pass
+        description: Enable saving SDFG before applying a (sub-)pass.
```
Sajohn-CH, line 821: I don't remember adding this but if I remove it I get a problem as `simplify.py` seems to use it.

dace/transformation/passes/scalar_to_symbol.py
```diff
@@ -227,7 +227,7 @@ def find_promotable_scalars(sdfg: sd.SDFG, transients_only: bool = True, integer
     interstate_symbols = set()
     for edge in sdfg.edges():
         interstate_symbols |= edge.data.free_symbols
-    for candidate in (candidates - interstate_symbols):
```
Sajohn-CH, line 230: Without this change the generated cloudsc code fails, I did not investigate it further and according to git blame the change does not stem from me.

dace/transformation/interstate/loop_to_map.py
```diff
@@ -557,12 +586,18 @@ def apply(self, _, sdfg: sd.SDFG):
...
             # can_be_applied makes sure there are no sympy functions in each of
             # the loop expresions
             access_node = body.add_read(rd)
-            body.add_memlet_path(access_node, entry, dst_conn=rd, memlet=memlet.Memlet(rd))
+            # Find new name
+            new_rd = sdfg._find_new_name(rd)
+            repl_dict[rd] = new_rd
+            body.add_memlet_path(access_node, entry, dst_conn=new_rd, memlet=memlet.Memlet(rd))
+        if repl_dict:
+            map.range.replace(repl_dict)
```
Sajohn-CH, line 599: If I don't add these changes, which don't stem from my, cloudsc fails

dace/transformation/interstate/loop_unroll.py
```diff
@@ -8,10 +8,14 @@
 from dace.properties import Property, make_properties
```
Sajohn-CH, line 8: There are some changes in this file which seem to be required but I didn't try out if it needs all of them or not. If I reset that file to the state of master I get an undefined symbol error

dace/config_schema.yml
Sajohn-CH: The usages where in other changes which I could undo

BenWeber42:
What's the status of this PR?
This thesis is advised by @acalotoiu ?
It adds 500k LoC. I assume it isn't meant to be merged as is? :sweat_smile: 
What are the valuable contributions for DaCe? I guess it's in the `dace` and `tests` folders? (dace doesn't need the `cloudsc_thesis` folder?)

BenWeber42:
ok, thanks a lot for the explanation. I guess the intention is that this PR will be closed eventually (after the thesis & its main contributions to DaCe have been wrapped up).

PR #1392, opened 2023-10-13
Add tensor storage format abstraction

JanKleine:
Format abstraction is based on [https://doi.org/10.1145/3276493].

It adds a new `Tensor` data descriptor as subclass of `Structure`, which
automatically has the necessary fields for the relevant indices.

The following index types are currently supported:
- Dense 
- Compressed (stores coordinates in segmented array akin to CSR)
- Singelton (stores one coordinate for every parent coordinate; used in COO)
- Range (computes coordinates based in current and parent dimension sizes + offset from parent)
- Offset (computes coordinates based on offset from parent coordinate)

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -576,7 +576,7 @@ class TensorIndex(ABC):
 
     @property
     @abstractmethod
-    def unique(sefl) -> bool:
+    def unique(self) -> bool:
         """
         True if coordinate in the level are unique, False otw.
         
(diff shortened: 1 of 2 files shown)


tbennun:
@JanKleine Since there is a lot of class infrastructure and code being added in this PR, I'd consider adding it to a new file (maybe dace.data as a folder).

Minor notes: there are some occurrences of `sefl` instead of `self` and `Singelton` should be spelled `Singleton`

JanKleine:
> Since there is a lot of class infrastructure and code being added in this PR, I'd consider adding it to a new file (maybe dace.data as a folder).

I tried to do that but circular dependencies seem to be a pain. Philipp also didn't know how to fix those easily. Do you (@tbennun) know if there is an easy way to do it? Otherwise I would postpone this to a different PR.

phschaad (approved):
LGTM, thank you for including the good doc comments and tests!

PR #1399, opened 2023-10-17
replace |& which is not widely supported

tim0s:
Some versions of bash (which is the only thing the shebang asks for) do not support |&, on the other hand 2>&1 does the same thing and is widely supported.


tbennun:
FYI test_all.sh is deprecated afaik

tim0s:
> FYI test_all.sh is deprecated afaik

Then it should be deleted? What is the up-to-date way to run all tests? Maybe we can add this into the installation guide? Also some of the tests seem dated, i.e., they generate warnings about deprecated functions.

tbennun (approved):

PR #1402, opened 2023-10-18
Change strides move assignment outside if

Sajohn-CH:
Adds functions/transformations from my thesis to:
- Change strides of existing data arrays
- Move constants assignment outside of if/else construct

Adds fixes for:
- Codegen getting struct depening on some scope situation
- Map expansion using hardcoded schedule instead of taking existing onej


acalotoiu (approved):
LGTM

acalotoiu:
> LGTM

However, first ensure that no tests are failing!

dace/transformation/dataflow/map_expansion.py
```diff
@@ -47,7 +47,7 @@ def apply(self, graph: dace.SDFGState, sdfg: dace.SDFG):
         new_maps = [
             nodes.Map(current_map.label + '_' + str(param), [param],
                       subsets.Range([param_range]),
-                      schedule=dtypes.ScheduleType.Sequential)
+                      schedule=current_map.schedule)
```
tbennun, line 50: will not work for GPU maps.

Sajohn-CH: Could you elaborate why do you think this would not work in GPU maps? The reason why I put this in is that I had maps expanded which were not sequential and suddenly one of the maps was sequential, which goes against my understanding of what MapExpansion should do. Though when I disable this change some (maybe all, didn't check all) of the failing tests pass again.

Sajohn-CH:
Regarding the failing tests and the change in map expansion. One of the failing tests seems to assume that map expansion sets the expanded map to sequential schedule, as it [sets some back to default](https://github.com/spcl/dace/blob/master/tests/transformations/otf_map_fusion_test.py#L633-L634), if I instead set the innermost map to sequential the test now passes. So I see two possibilities here, either my understanding of map expansion is incorrect or some tests depend on the incorrect behaviour and fixing this would potentially be a new issue.

Issue #1383, opened 2023-10-06
failed to run some FPGA samples


remifan:
**Describe the bug**
I can run ` dace/samples/fpga/rtl/fladd.py` without issues but failed in running other rtl examples. Here is the error log:

```
python dace/samples/fpga/rtl/pipeline.py
                                                                      
Traceback (most recent call last):                                                                                             
  File "/home/<user>/Workspace/dace/pipeline.py", line 164, in <module>                                                          
    sdfg(A=a, B=b, N=N)                                                                                                        
  File "/home/<user>/Applications/miniconda3/envs/dace/lib/python3.11/site-packages/dace/sdfg/sdfg.py", line 2342, in __call__   
    binaryobj = sdfg.compile()                                                                                                 
                ^^^^^^^^^^^^^^                                                                                                 
  File "/home/<user>/Applications/miniconda3/envs/dace/lib/python3.11/site-packages/dace/sdfg/sdfg.py", line 2254, in compile    
    program_objects = codegen.generate_code(sdfg, validate=validate)                                                           
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                           
  File "/home/<user>/Applications/miniconda3/envs/dace/lib/python3.11/site-packages/dace/codegen/codegen.py", line 231, in genera
te_code                                                                                                                        
    (global_code, frame_code, used_targets, used_environments) = frame.generate_code(sdfg, None)                               
                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                               
  File "/home/<user>/Applications/miniconda3/envs/dace/lib/python3.11/site-packages/dace/codegen/targets/framecode.py", line 831,
 in generate_code                                                                                                              
    states_generated = self.generate_states(sdfg, global_stream, callsite_stream)                                              
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                              
  File "/home/<user>/Applications/miniconda3/envs/dace/lib/python3.11/site-packages/dace/codegen/targets/framecode.py", line 408,
 in generate_states                                                                                                            
    callsite_stream.write(cft.as_cpp(self, sdfg.symbols), sdfg)                                                                
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                       
[...]

PR #1403, opened 2023-10-18
RTL codegen "line" error

carljohnsen:
The issue in #1383 uncovered flaws in the RTL samples: 
1. Unsupported RTL configurations not failing gracefully. 
  The specific error arose when an RTL tasklet tried to access a stream during simulation mode, which is currently not supported in simulation mode, as the produced Verilator code does not mimick actual behaviour, resulting in a deadlock. Current "fix" is to emit this shortcoming as a comment in the resulting code. It should not always fail (hence the emission to a comment), since the RTL code generator is partially used by the Xilinx code generator, leading to said faulty code to be omitted anyways; at least for the samples/tests that reach this point.
2. The samples (at least not all of them) did not set the required DaCe configuration entries. 
  Some of the samples did correctly use `with dace.config.set_temporary()`, forcing them to run as they're inteded to be run. This has been changed, so all RTL samples run in their respective modes. Furthermore, some of the samples did not have this requirement mentioned in their comments, which has also been fixed.

### DIFF
--- a/dace/codegen/targets/rtl.py
+++ b/dace/codegen/targets/rtl.py
@@ -102,6 +102,10 @@ class RTLCodeGen(target.TargetCodeGenerator):
                 elif isinstance(arr, data.Scalar):
                     line: str = "{} {} = {};".format(dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn,
                                                      edge.src.data)
+                elif isinstance(arr, data.Stream):
+                    # TODO Streams are currently unsupported, as the proper behaviour has to be implemented to avoid deadlocking.
+                    line: str = "// Unsupported read from ({}) variable '{}' from stream '{}'".format(
+                        dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn, edge.src_conn)
         elif isinstance(edge.src, nodes.MapEntry) and isinstance(edge.dst, nodes.Tasklet):
             rtl_name = self.unique_name(edge.dst, sdfg.nodes()[state_id], sdfg)
             self.n_unrolled[rtl_name] = symbolic.evaluate(edge.src.map.range[0][1] + 1, sdfg.constants)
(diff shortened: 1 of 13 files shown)


samples/fpga/rtl/rtl_multi_tasklet.py
```diff
@@ -1,11 +1,10 @@
-# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
-"""
-    Two sequential RTL tasklets connected through a memlet.
-"""
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
+#
+# Two sequential RTL tasklets connected through a memlet.
+#
+# It is intended for running simulation xilinx targets.
```
tbennun, line 5: The previous docstring version is the standard way of defining a file.

tbennun (changes requested):
Thanks for the fix! Two minor comments.

dace/codegen/targets/rtl.py
```diff
@@ -102,6 +102,10 @@ def copy_memory(self, sdfg: sdfg.SDFG, dfg: state.StateSubgraphView, state_id: i
                 elif isinstance(arr, data.Scalar):
                     line: str = "{} {} = {};".format(dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn,
                                                      edge.src.data)
+                elif isinstance(arr, data.Stream):
+                    # TODO Streams are currently unsupported, as the proper behaviour has to be implemented to avoid deadlocking.
+                    line: str = "// Unsupported read from ({}) variable '{}' from stream '{}'".format(
+                        dst_node.in_connectors[edge.dst_conn].ctype, edge.dst_conn, edge.src_conn)
```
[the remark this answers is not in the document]
carljohnsen: An error no, as the RTL backend is used by the Xilinx backend, which may hit this case without using the faulty code. I've further elaborated this in the comment. I've added the information as a warning as well.

samples/fpga/rtl/rtl_multi_tasklet.py
carljohnsen: Ah, right, I see! I've reverted them "back" to this format! :)

tbennun (approved):

PR #1405, opened 2023-10-20
Option for utilizing GPU global memory

alexnick83:
A standard workflow when transforming GPUs for GPU execution involves changing the storage of non-transient data to GPU global memory so that they are read/written directly from/to the device. Due to changes to SDFG validation, this workflow frequently fails because changing the storage (temporarily) invalidates reads and writes in InterstateEdges. This PR addresses the issue in the case of auto-optimization by adding an option to change the storage just before applying GPUTransformation.


phschaad (approved):
LGTM

PR #1406, opened 2023-10-24
Fix array indirection to memlet subset promotion

BenWeber42:
The current solution is rather hacky. I want to run the tests first to see the impacts of this change.
Additionally, there is no test yet, because validation doesn't catch the erroneous SDFG yet.

Overall, it's not clear currently how to solve the issue and the PR might change as we progress...


dace/transformation/passes/scalar_to_symbol.py
```diff
@@ -322,6 +322,7 @@ def __init__(self, in_edges: Dict[str, mm.Memlet], out_edges: Dict[str, mm.Memle
 
     def visit_Subscript(self, node: ast.Subscript) -> Any:
         # Convert subscript to symbol name
+        node = self.generic_visit(node)
```
BenWeber42, line 325: Post-order traversal _just_ fixes this issue, but is mostly a work-around. `_range_is_promotable` should really be changed here. But that's difficult without having to do larger changes.

tests/double_index_bug_test.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
+from sympy import core as sympy_core
+
+import dace
+from dace import subsets
+
+
+def test_double_index_bug():
```
BenWeber42, line 8: Maybe I should put this test somewhere else? :thinking:

tbennun: Probably in tests/passes/scalar_to_symbol_promotion_test.py?

tests/double_index_bug_test.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
```
tbennun, line 1: year

dace/transformation/passes/scalar_to_symbol.py
tbennun: Sounds like that's the right approach then. I'm not sure it's a workaround to be honest.

tbennun (approved):
Looks good, see comment about test.

tbennun (approved):

PR #1408, opened 2023-10-25
Fix an inappropriate test expression to remove a logical short circuit

munahaf:
In file: snitch.py, the comparison of Collection length creates a logical short circuit. 
The length of a collection is always greater than or equal to zero. So testing that a length is less than zero is always false.

I suggested that the Collection length comparison should be removed since it is meaningless. 

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.


tbennun:
Thank you for posting this PR! Could you explain how this code poses a security risk?

munahaf:
This issue is not related with security. The tool we used found this logical bug; it is easy to fix, so we reported it.

I will update the PR as suggested.

PR #1410, opened 2023-10-26
Symbol specialization in `auto_optimizer()` never took effect.

philip-paul-mueller:
The `dict` that was storing all the symbols, `known_symbols`, was emptied just after its creation. The efect was that the specialization never took effect.

### DIFF
--- a/dace/transformation/auto/auto_optimize.py
+++ b/dace/transformation/auto/auto_optimize.py
@@ -647,7 +647,6 @@ def auto_optimize(sdfg: SDFG,
     if symbols:
         # Specialize for all known symbols
         known_symbols = {s: v for (s, v) in symbols.items() if s in sdfg.free_symbols}
-        known_symbols = {}
         for (s, v) in symbols.items():
             if s in sdfg.free_symbols:
                 if isinstance(v, (int, float)):
--- a/dace/transformation/auto/auto_optimize.py
+++ b/dace/transformation/auto/auto_optimize.py
@@ -646,7 +646,7 @@ def auto_optimize(sdfg: SDFG,
 
     if symbols:
         # Specialize for all known symbols
-        known_symbols = {s: v for (s, v) in symbols.items() if s in sdfg.free_symbols}
+        known_symbols = {}
         for (s, v) in symbols.items():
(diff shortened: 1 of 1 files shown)


phschaad:
I'm a bit confused, the change (now) seems to be the exact inverse of what it originally was. While I see that this line should be removed, shouldn't this have no effect on behavior whatsoever since it is effectively a dead (or succeeded) line of code?

philip-paul-mueller:
I think the difference is because of the following loop (see file `dace/transformation/auto/auto_optimize.py`, line 650ff).
There on line 653 it is checked if the value `v` is a `sympy.Integer` and if so it tries to transform it into a Python `int`, in case this fails the symbol is ignored  (I am not sure why though), and will thus not enter `known_symbols`.
Therefore `known_symbols` will only contain Python `float` and `int`.

However, if `known_symbols` is first directly constructed, this transformation is not done, and the "offending" (in the sense of having type `sympy.Integer` which can not be transformed to `int`) symbols will remain inside the `dict`.
While the loop will replace all `sympy.Integer`s that can be transformed to Python `int` with a corresponding Python `int` it will not remove the `sympy.Integer` for which this transformation fails and this probably causes the issues we are seeing.

BenWeber42:
I agree. Basically, if you only look at these two lines, it would seem that the reset should be removed (as initially proposed):
https://github.com/spcl/dace/blob/66913220ea600492db59cf8e536271b36c1554bd/dace/transformation/auto/auto_optimize.py#L649-L650

However, looking at the larger context, it becomes clear that the dict comprehension of the first initialization is a strict super set of the later filtering code:
https://github.com/spcl/dace/blob/66913220ea600492db59cf8e536271b36c1554bd/dace/transformation/auto/auto_optimize.py#L649-L659

And this is in line with the doc-string:
https://github.com/spcl/dace/blob/66913220ea600492db59cf8e536271b36c1554bd/dace/transformation/auto/auto_optimize.py#L559

So it seems like it's best to initialize empty and remove the redundant dict comprehension.

BenWeber42 (approved):

phschaad (approved):

PR #1412, opened 2023-10-30
Implement Subsetlist and covers_precise

matteonu:
This PR adds new functionality to the `Subset` class

- `Subsetlist` is a new Subset that acts as a wrapper of existing subset types. It is essentially a list of Subset objects. The union of two Subsetlists is just the list concatenation.
- `covers_precise()` implements a set cover operation on Subsets of type Range and Indices. If it returns true the first input subset definitely covers the second input subset. Therefore it also takes steps into account.

### DIFF
--- a/dace/cli/sdfv.py
+++ b/dace/cli/sdfv.py
@@ -36,7 +36,11 @@ def view(sdfg: dace.SDFG, filename: Optional[Union[str, int]] = None):
     """
     # If vscode is open, try to open it inside vscode
     if filename is None:
-        if 'VSCODE_IPC_HOOK_CLI' in os.environ or 'VSCODE_GIT_IPC_HANDLE' in os.environ:
+        if (
+            'VSCODE_IPC_HOOK' in os.environ
+            or 'VSCODE_IPC_HOOK_CLI' in os.environ
+            or 'VSCODE_GIT_IPC_HANDLE' in os.environ
+        ):
             filename = tempfile.mktemp(suffix='.sdfg')
             sdfg.save(filename)
             os.system(f'code {filename}')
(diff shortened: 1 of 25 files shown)


tbennun:
Thank you for the contribution. Before reviewing the PR, I have a minor naming comment: perhaps UnionSubset is a better name?

matteonu:
> Thank you for the contribution. Before reviewing the PR, I have a minor naming comment: perhaps UnionSubset is a better name?

I renamed it `SubsetUnion` now instead of `UnionSubset` (I switched the two words in my head somehow). I hope this works too. If not, I can change it again of course.

matteonu:
Seems like `subset_intersects_test.py` fails in the build pipeline but I cannot reproduce it locally for some reason.

tbennun:
It looks like DaCe is still installing SymPy 1.9, and that might be the reason: https://github.com/spcl/dace/actions/runs/6706005794/job/18223680540?pr=1412#step:4:521

 Maybe try changing setup.py to reflect that (`sympy >= 1.9`)

acalotoiu (changes requested):
Thank you for the contribution! Please make it clear what all the tests are testing.

matteonu:
I think it is not possible to run the tests on python3.7 with sympy 1.12. According to the official sympy [website](https://docs.sympy.org/latest/install.html) version 1.12 only supports python3.8 and above. I am not sure what else could cause the failing test. I tried to reproduce it by running the commands of the failing test from the `[.github/workflows/general-ci.yml](https://github.com/spcl/dace/blob/3c10cb126ee68b2558f636a73aedf71f4863a125/.github/workflows/general-ci.yml)` locally but I pass the failing test.

matteonu:
Now the test should work.

matteonu:
> Now the test should work.
It seems like I forgot to set the sympy version back to 1.9 in the requirements.txt. I fixed it now.

acalotoiu (approved):
LGTM now! Thank you!

Issue #1396, opened 2023-10-16
Can not create a DaCe program `double` in Python


philip-paul-mueller:
It might be a bit of a silly issue, but I made a DaCe program with the name `double` as in:
```python
@dace.program
def double():
	...
```
As `double()` is a valid name for a Python function it is not in C++. Everything works as expected except calling `compile()`. The code generator generate a state structure for the program with the name `double_t`, however, by chance it also creates the typedef `double_t` to alias the C++ type `double`.
Interestingly, well at least for me, if this program is used inside another dace program the error vanishes.

There was already a similar issue in the (#1164) in the past. I think the best way to tackle this issue is to update the name mangling process.
However, I am not sure if the issue is worth the effort.

philip-paul-mueller:
I did some more investigation and realized that I made a small mistake, DaCe is not creating the `typedef` that causes the error it is the C standard. It seems that a `C99` conforming library has to [define `double_t` and `float_t`](https://en.cppreference.com/w/c/numeric/math/float_t), C++ also does this but in the `std` namespace.
Since DaCe creates a state struct (correct me if I am wrong, but it seems to be always empty) with the name of the program, the compiler complains.

PR #1413, opened 2023-10-31
Better mangeling of the state struct in the code generator

philip-paul-mueller:
This PR addresses issue  #1396, by modifying the way how the name of the state struct is created.

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -239,6 +239,7 @@ class CompiledSDFG(object):
         return ctypes.cast(self._libhandle, ctypes.POINTER(self._try_parse_state_struct())).contents
 
     def _try_parse_state_struct(self) -> Optional[Type[ctypes.Structure]]:
+        from dace.codegen.targets.cpp import mangle_dace_state_struct_name  # Avoid import cycle
         # the path of the main sdfg file containing the state struct
         main_src_path = os.path.join(os.path.dirname(os.path.dirname(self._lib._library_filename)), "src", "cpu",
                                      self._sdfg.name + ".cpp")
@@ -247,7 +248,7 @@ class CompiledSDFG(object):
         code_flat = code.replace("\n", " ")
 
         # try to find the first struct definition that matches the name we are looking for in the sdfg file
-        match = re.search(f"struct {self._sdfg.name}_t {{(.*?)}};", code_flat)
+        match = re.search(f"struct {mangle_dace_state_struct_name(self._sdfg)} {{(.*?)}};", code_flat)
         if match is None or len(match.groups()) != 1:
             return None
 
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -239,7 +239,6 @@ class CompiledSDFG(object):
         return ctypes.cast(self._libhandle, ctypes.POINTER(self._try_parse_state_struct())).contents
 
     def _try_parse_state_struct(self) -> Optional[Type[ctypes.Structure]]:
-        from dace.codegen.targets.cpp import mangle_dace_state_struct_name  # Avoid import cycle
         # the path of the main sdfg file containing the state struct
         main_src_path = os.path.join(os.path.dirname(os.path.dirname(self._lib._library_filename)), "src", "cpu",
                                      self._sdfg.name + ".cpp")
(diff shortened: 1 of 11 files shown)


philip-paul-mueller:
According to the log `tests/python_frontend/indirections_test.py::test_indirection_scalar_range_nsdfg` failed due to a segmentation fault. However, I run it on my machine without an error. This could indicate some deeper issue. But I am not fully sure.

phschaad:
> According to the log `tests/python_frontend/indirections_test.py::test_indirection_scalar_range_nsdfg` failed due to a segmentation fault. However, I run it on my machine without an error. This could indicate some deeper issue. But I am not fully sure.
> 
> ```python
> pytest tests/python_frontend/indirections_test.py::test_indirection_scalar_range_nsdfg
> ```

This has become a common issue in the past few days. It fails on the same test configuration every time, if it fails. For now just restarting the workflow works, but we have to look into this issue further.

BenWeber42:
> > According to the log `tests/python_frontend/indirections_test.py::test_indirection_scalar_range_nsdfg` failed due to a segmentation fault. However, I run it on my machine without an error. This could indicate some deeper issue. But I am not fully sure.
> > ```python
> > pytest tests/python_frontend/indirections_test.py::test_indirection_scalar_range_nsdfg
> > ```
> 
> This has become a common issue in the past few days. It fails on the same test configuration every time, if it fails. For now just restarting the failing workflow works, but we have to look into this issue further.

Thanks for the context. I am tracking this here #1421. I initiated a re-run for the tests.

philip-paul-mueller:
@BenWeber42 Interestingly  [rebooting](https://www.youtube.com/watch?v=nn2FB1P_Mn8) even works on non Windows machines.

tbennun:
I am not sure we should make this change. Several codes depend on the existing naming convention, and the benefit is only for specific names that we should maybe disallow in validation.

philip-paul-mueller:
Having this function has in my view multiple advantages and I do not find any disadvantages:
- There is _one single place_ where the name is composed, so if you have to change the name for whatever reason, there is one place to do that. (It will only get worse if time passes.)
- It might be a convention, but having a function indicates this directly.
- It immediately makes clear what you do.

Furthermore, while this might not be much of an argument, I learned that: "For everything you do more than once write a function" and we compose this name multiple times.

tbennun:
> Having this function has in my view multiple advantages and I do not find any disadvantages:
> 
> * There is _one single place_ where the name is composed, so if you have to change the name for whatever reason, there is one place to do that. (It will only get worse if time passes.)
> * It might be a convention, but having a function indicates this directly.
> * It immediately makes clear what you do.
> 
> Furthermore, while this might not be much of an argument, I learned that: "For everything you do more than once write a function" and we compose this name multiple times.

I wasn't referring to the function, that part is good. I was referring to the part where we make the code less readable by changing the struct's name. One could add a dace config such as `codegen.state_struct_suffix` which is `_t` by default.

philip-paul-mueller:
Oh, you were referring to the _generated_ C++ code. But I really like your idea with the configuration.

philip-paul-mueller:
I have implemented your idea, but the default value of the configuration now defaults to `_state_t` instead of `_t` which.

By the way would it make sense to move the keys for the code generator to their own hierarchy?

BenWeber42:
This seems very much like an ad-hoc solution for one instance of a much more general problem. The general problem is that Python identifiers and C++ identifiers (as generated by code-gen) are not the same. A common solution for such a problem is to have a name-mangling system. As far as I know, we currently don't have a proper name mangling system in DaCe (based on `git grep -E "mangle"`).

I think we should aim towards a general solution to the problem, because DaCe needs this anyway if we want DaCe to be robust. If we start with just an ad-hoc solution, we risk growing a name mangling system that is not suitable to solve the general problem.

I think we should take a step back and first consider a general name mangling system for DaCe. Once we have a better view of this, we can think of introducing this step by step. In particular, how should this one bug be fixed, such that the solution facilitates a path towards a more general name mangling system.

Regarding the configuration option, my big question is whether our name mangling should be configurable. We could also consider name mangling to be something internal, that shouldn't be configurable. Especially, because it's much easier to later add configuration options than to remove/change them.

philip-paul-mueller:
@BenWeber42 
I agree, we should make the code generators more robust and for this a general name mangeling scheme should be created. But as far as I can tell this is a rather big task.

tbennun (approved):
LGTM

PR #1419, opened 2023-11-01
Fix for VS Code debug console: view opens sdfg in VS Code and not in browser

kotsaloscv:
Test system Mac OS

This small modification makes sure that in both modes (interactive & debug) the `sdfg.view()` will open the graph in the VS Code env (if present) and not in the available browser.

The `VSCODE_IPC_HOOK_CLI` env var is present in an interactive session from the VS Code termninal, but it is not present in the debug console. In the debugging mode, it is the `VSCODE_IPC_HOOK` env var which is present.

### DIFF
--- a/dace/cli/sdfv.py
+++ b/dace/cli/sdfv.py
@@ -36,7 +36,7 @@ def view(sdfg: dace.SDFG, filename: Optional[Union[str, int]] = None):
     """
     # If vscode is open, try to open it inside vscode
     if filename is None:
-        if 'VSCODE_IPC_HOOK_CLI' in os.environ or 'VSCODE_GIT_IPC_HANDLE' in os.environ:
+        if 'VSCODE_IPC_HOOK' in os.environ or 'VSCODE_GIT_IPC_HANDLE' in os.environ:
             filename = tempfile.mktemp(suffix='.sdfg')
             sdfg.save(filename)
             os.system(f'code {filename}')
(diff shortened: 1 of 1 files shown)


phschaad:
Thank you for the PR - weird, this seems to work fine on Windows, but in that case Mac OS VSCode seems to handle the environment variables differently. Is `VSCODE_IPC_HOOK` present in both environments then, terminal and debug console?

dace/cli/sdfv.py
```diff
@@ -36,7 +36,7 @@ def view(sdfg: dace.SDFG, filename: Optional[Union[str, int]] = None):
     """
     # If vscode is open, try to open it inside vscode
     if filename is None:
-        if 'VSCODE_IPC_HOOK_CLI' in os.environ or 'VSCODE_GIT_IPC_HANDLE' in os.environ:
+        if 'VSCODE_IPC_HOOK' in os.environ or 'VSCODE_GIT_IPC_HANDLE' in os.environ:
```
phschaad, line 39: I just checked - in my environment (Ubuntu20), VSCode does not set the `VSCODE_IPC_HOOK` variable in the terminal, nor the debug console, weirdly enough. In that case I would request that we test for both `VSCODE_IPC_HOOK` and `VSCODE_IPC_HOOK_CLI`.

phschaad (changes requested):

phschaad (approved):
Thanks!

Issue #1389, opened 2023-10-12
Numpy `fill` accepts only numbers


TizianoDeMatteis:
**Describe the bug**

Current replacement for `numpy.fill` checks that the value is a Number (or boolean). This prevents using a variable.

**To Reproduce**
```Python
import dace
import numpy as np

N = dace.symbol("N", dace.int32)


@dace.program
def fill_dace(A: dace.float64[N, N], value: dace.float64):
    A.fill(value)


N = 8
value = 0.3
input = np.ndarray((N, N)).astype(np.float64)
input_ref = np.copy(input)

fill_dace(input, value)
input_ref.fill(value)
assert np.allclose(input, input_ref)
```

1. Run the code above
2.  It will return a `DaceSyntaxError`

**Expected behavior**
Allow also variables (or parameters) to be passed to fill

philip-paul-mueller:
I will take a look at it.

philip-paul-mueller:
I think I have found a way to do this.
Currently `fill()` is delegated to  `_elementwise()`, which only accepts values.
There must be some kind of switch to handle different cases.

PR #1420, opened 2023-11-01
Numpy fill accepts also variables 

philip-paul-mueller:
This PR is for addressing issue [#1389](https://github.com/spcl/dace/issues/1389).


philip-paul-mueller:
While this PR allows for code such as:
```python
import dace
import numpy as np

N = dace.symbol("N", dace.int32)

@dace.program
def fill_dace(A: dace.float64[N, N], value: dace.float64):
    A.fill(value)
```

it does currently not allows for code such as:
```python
@dace.program
def fill_dace_not_working(A: dace.float64[N, N], value: dace.float64):
    A.fill(value + 1)
```

What is funny though is that the code:
```python
@dace.program
def fill_dace_does_also_not_work(A: dace.float64[N, N], value: dace.float64):
    value2 = value + 1
    A.fill(value2)
```
will also not work but:
```python
@dace.program
def fill_dace_will_work_but_why(A: dace.float64[N, N], value: dace.float64):
    value2 = value + 1
    A.fill(value2)
    return value2
```
will work, for some strange reason and I have no idea why?

philip-paul-mueller:
**Current State**:         
To summarize the current state:
- `A.fill(value)`: Where value is an _argument_ -> Works
- `A.fill(Expr)`: Where `Expr` is an expression, such as `value + 1` -> Does not work (probably need some big changes)
- `c = value + 2; A.fill(c)`: Where `value` is an argument -> Does not work (This is probably a bug in the Python frontend, see above)

philip-paul-mueller:
Commit [a29ae4a](https://github.com/spcl/dace/pull/1420/commits/a29ae4a6b102fd7cebbca0858a5541c356697f20) should now solve everything of the above list.

dace/frontend/python/astutils.py
```diff
@@ -480,25 +481,27 @@ class ExtNodeVisitor(ast.NodeVisitor):
...
+                            visitor_name = "visit_TopLevel" + type(value).__name__
+                            if hasattr(self, visitor_name):
+                                visitor = getattr(self, visitor_name)
+                                return visitor(value)
                             else:
-                                self.visit(value)
+                                return self.visit(value)
                         else:
-                            self.visit(value)
+                            return self.visit(value)
```
BenWeber42, line 504: I don't think we should return here, otherwise we cause an early exit of the loop and the remaining elements of the `old_value` list won't be visited.

This is an `ast.NodeVisitor`, so modifications to the AST are not intended. Not sure what `ast.NodeVisitor` does when it goes through a list. I assume it won't `return` either.

philip-paul-mueller: You are right it is inconsistent. I do not know why I did it, I think the final return irritated my.
What is more interesting is, that _why_ all tests have passed?

BenWeber42: Yes, I was surprised by this too. Unfortunately, this is dead-code testing wise:
https://app.codecov.io/gh/spcl/dace/blob/master/dace%2Ffrontend%2Fpython%2Fastutils.py#L492

The whole `ExtNodeVisitor` and `ExtNodeTransformer` classes seem a bit strange to me. I don't see how a simple `visit_Expr` basically accomplishes the same with the normal `ast.NodeVisitor` and `ast.NodeTransformer` classes.
I suspect that in older Python versions things were different. I don't think it's worth spending more time into this.

dace/frontend/python/replacements.py
```diff
@@ -605,11 +605,10 @@ def _elementwise(pv: 'ProgramVisitor',
...
         state.add_mapped_tasklet(
             name="_elementwise_",
-            map_ranges={'__i%d' % i: '0:%s' % n
-                        for i, n in enumerate(inparr.shape)},
-            inputs={'__inp': Memlet.simple(in_array, ','.join(['__i%d' % i for i in range(len(inparr.shape))]))},
+            map_ranges={f'__i{dim}': f'0:{N}' for dim, N in enumerate(inparr.shape)},
+            inputs={'__inp': Memlet.simple(in_array, ','.join([f'__i{dim}' for dim in range(len(inparr.shape))]))},
             code=code,
-            outputs={'__out': Memlet.simple(out_array, ','.join(['__i%d' % i for i in range(len(inparr.shape))]))},
+            outputs={'__out': Memlet.simple(out_array, ','.join([f'__i{dim}' for dim in range(len(inparr.shape))]))},
```
BenWeber42, line 611: Here we just replace the old formatting mechanism (with the `%` _operator_) with formatting strings (`f"..."`)?

As far as I can tell, it doesn't really change the meaning of the code?

I think this is fine. I just want to make sure I didn't miss anything :smile:

dace/frontend/python/astutils.py
philip-paul-mueller: I agree.

dace/frontend/python/replacements.py
```diff
@@ -4232,10 +4231,38 @@ def _ndarray_copy(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, arr: str) ->
...
+    state.add_mapped_tasklet(
+        '_numpy_fill_',
+        map_ranges={f"__i{dim}": f"0:{s}" for dim, s in enumerate(shape)},
+        inputs=inputs,
+        code=f"__out = {body}",
+        outputs={'__out': dace.Memlet.simple(arr, ",".join([f"__i{dim}" for dim in range(len(shape))]))},
+        external_edges=True
+    )
+
+    return arr
```
BenWeber42, line 4265:
```suggestion
def _ndarray_fill(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, arr: str, value: Union[str, Number,
                                                                                           sp.Expr]) -> str:
    assert arr in sdfg.arrays

    if isinstance(value, sp.Expr):
        raise NotImplementedError(
            f"`{arr}.fill` is not implemented for symbolic expressions (`{value}`).")  # Look at `full`.

    if isinstance(value, (Number, np.bool_)):
        body = value
        inputs = {}
    elif isinstance(value, str) and value in sdfg.arrays:
        value_array = sdfg.arrays[value]
        if not isinstance(value_array, data.Scalar):
            raise mem_parser.DaceSyntaxError(
                pv, None, f"`{arr}.fill` requires a scalar argument, but `{type(value_array)}` was given.")
        body = '__inp'
        inputs = {'__inp': dace.Memlet(data=value, subset='0')}
    else:
        raise mem_parser.DaceSyntaxError(pv, None, f"Unsupported argument `{value}` for `{arr}.fill`.")

    shape = sdfg.arrays[arr].shape
    state.add_mapped_tasklet(
        '_numpy_fill_',
        map_ranges={
            f"__i{dim}": f"0:{s}"
            for dim, s in enumerate(shape)
        },
        inputs=inputs,
        code=f"__out = {body}",
        outputs={'__out': dace.Memlet.simple(arr, ",".join([f"__i{dim}" for dim in range(len(shape))]))},
        external_edges=True)

    return arr
```

I would suggest the following changes:

- `f"The type of fill value `{f}` not supported! It was '{type(value)}'"`, here `f` is undeclared
- Tried to change control-flow to be easier to understand
- Tried to improve error reporting
- Formatted with yapf

dace/frontend/python/replacements.py
philip-paul-mueller: Yes this is is the case, it was easier to read for me.

dace/frontend/python/replacements.py
```diff
@@ -4232,10 +4231,40 @@ def _ndarray_copy(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, arr: str) ->
...
+        inputs = {}
+    elif isinstance(value, str) and value in sdfg.arrays:
+        value_array = sdfg.arrays[value]
+        if not isinstance(value_array, data.Scalar):
+            raise mem_parser.DaceSyntaxError(
+                pv, None, f"`{arr}.fill` requires a scalar argument, but `{type(value_array)}` was given.")
+        body = '__inp'
+        inputs = {'__inp': dace.Memlet(data=value, subset='0')}
+    else:
+        raise mem_parser.DaceSyntaxError(pv, None, f"Unsupported argument `{value}` for `{arr}.fill`.")
```
BenWeber42, line 4253:
```suggestion
        raise NotImplementedError(
            f"{arr}.fill is not implemented for symbolic expressions ({value}).")  # Look at `full`.

    if isinstance(value, (Number, np.bool_)):
        body = value
        inputs = {}
    elif isinstance(value, str) and value in sdfg.arrays:
        value_array = sdfg.arrays[value]
        if not isinstance(value_array, data.Scalar):
            raise mem_parser.DaceSyntaxError(
                pv, None, f"{arr}.fill requires a scalar argument, but {type(value_array)} was given.")
        body = '__inp'
        inputs = {'__inp': dace.Memlet(data=value, subset='0')}
    else:
        raise mem_parser.DaceSyntaxError(pv, None, f"Unsupported argument '{value}' for {arr}.fill.")
```

BenWeber42: I removed the backticks (`), since existing error reporting doesn't seem to use them.

BenWeber42 (approved):

PR #1423, opened 2023-11-02
Rename misleading uses of API

tbennun:
`symstr` and `symbols_in_ast` have now been renamed.


BenWeber42:
Hey, thanks for the PR. I need a bit more time to consider this case. In particular, because I think we have the opportunity to further clean-up the API. It seems like some of these (`symbols_in_ast`/`names_in_ast`, `symlist`) are not often used. We also discussed that we should look into how often these methods are used. I couldn't do this yet. But if they are not used often, we can consider refactoring them more aggressively. In particular, we could move `symbols_in_ast`/`names_in_ast` to a different module, because the current name & implementation doesn't seem related to the sympy layer (seems misleading to have it in the symbolic module). Additionally, I get the impression that the `symlist` function is an _older API_, which we could consider removing altogether (the return type of dict also seem rather strange).

tbennun:
I agree with both points, names in ast sounds like something that should be in dace.frontend.python.astutils. symlist can be erased and most of its uses call for .values() anyway

BenWeber42:
If you like, I can take over ownership of the branch and work on the suggestions.

tbennun:
That would be great, thank you

tbennun:
Unfortunately, due to Structures and `Attr()` attributes, symlist has a specific use beyond finding names in ASTs and we cannot merge this PR as-is. It will have to be redone to be merged.

PR #1425, opened 2023-11-06
Implement writeset underapproximation

matteonu:
This PR adds a pass that under-approximates the write-sets of loops and maps based on the propagation pass. Currently it only supports unconditional writes and affine subscript expressions.


tbennun:
@matteonu if you are using the Pass API, you should probably put the file under dace/transformation/passes/...

acalotoiu (approved):
LGTM. Thank you!

PR #1427, opened 2023-11-07
Implement ArrayFission

matteonu:
This PR adds the ArrayFission pass, which is the array counterpart of the existing ScalarFission pass. The main functionality is outsourced in a separate pass (VariableFission), which will be used in a future PR for a new version of Scalar Fission.


acalotoiu (changes requested):
This seems to overlap with the write set approximation review that is being merged. Could you please remove those aspects that are already included there?

matteonu:
> This seems to overlap with the write set approximation review that is being merged. Could you please remove those aspects that are already included there?

You are right. I will fix that as soon as the other PR is closed.

Issue #1433, opened 2023-11-17
Missing symbols for data access on inter-state edge


edopao:
**Describe the bug**
After uplift to dace v0.15, one SDFG which was working before started to show compilation errors. The latest DaCe is moving a data access to an inter-state edge. For the data-access, the symbols that define array strides are needed for code generation. The SDFG was validated, before and after the simplify pass, but it did not compile for CPU. When skipping the simplify pass, the compilation did work. The problem has been narrowed down to the scalar-to-symbol promotion, which is moving a data access to an inter-state edge. Then, the method `_used_symbols_internal` needs to be update to account for data containers, including symbolic shape and strides.

**To Reproduce**
The branch [bug-missing-symbols](https://github.com/edopao/dace/tree/bug-missing-symbols) contains a unit test to reproduce the issue, and the proposed fix.

**Expected behavior**
Not clear if the introduction of data access on an inter-state edge was expected, in the first place.

PR #1434, opened 2023-11-17
Fix codegen with data access on inter-state edge

edopao:
After uplift to dace v0.15, one SDFG which was working before started to show compilation errors. The latest DaCe is moving a data access to an inter-state edge. For the data-access, the symbols that define array strides are needed for code generation. The SDFG was validated, before and after the simplify pass, but it did not compile for CPU. When skipping the simplify pass, the compilation did work. The problem has been narrowed down to the scalar-to-symbol promotion, which is moving a data access to an inter-state edge. Then, the method `_used_symbols_internal` needs to be update to account for data containers, including symbolic shape and strides.

This issue was reported in #1433. This PR contains a unit test to reproduce the issue and verify the proposed fix.


BenWeber42 (changes requested):
Hey, sorry for the delayed review. I briefly talked with Tal, and the changes to DaCe are very good. Thanks a lot for that!

I would like to make some suggestions to the test:

1. Move the test to this file: `tests/codegen/codegen_used_symbols_test.py`
2. Try to reduce the test-case. Particularly, build the SDFG directly instead of starting with a different SDFG and then applying transformations until you reach the SDFG that triggers the bug

Regarding (2), I am worried that if the transformations change, then that might nullify the test. We don't really want to test the transformations here (most of the `assert`s just check that the test should work correctly).

edopao:
@BenWeber42 Thank you for the review. Very good comment, I totally agree. I had focused my test more on the trigger of this issue rather than on its cause. I have pushed a commit where I apply the changes you proposed.

PR #1437, opened 2023-11-21
Added `nan` to the DaCe `math` namespace

philip-paul-mueller:
Before this was generating an error, because there was no object `nan` inside the `dace::math` namespace. This commit adds a `nan` object to the namespace, the implementation is based on `typeless_pi`.


dace/runtime/include/dace/nan.h
```diff
@@ -0,0 +1,112 @@
...
+    namespace math
+    {
+        //////////////////////////////////////////////////////
+        // Defines a typeless Pi
+        struct typeless_nan
+        {
+            operator int() const = delete;
+            operator float() const
+            {
+                return std::nanf("");
```
tbennun, line 18: `std::nan*` use `strto*`, which will be slow. Please use numeric_limits' NaN: https://en.cppreference.com/w/cpp/types/numeric_limits/quiet_NaN

tbennun (changes requested):

dace/runtime/include/dace/nan.h
philip-paul-mueller: Thanks for reminding me about that thing.

tbennun (approved):

PR #1438, opened 2023-11-21
FPGA tutorial

TizianoDeMatteis:
Basic FPGA tutorial.

The run jupyter notebook can be accessed here: [https://nbviewer.org/github/spcl/dace/blob/fpga_tutorial/tutorials/fpga.ipynb](https://nbviewer.org/github/spcl/dace/blob/fpga_tutorial/tutorials/fpga.ipynb)

**Note: ** currently there are glitches in the rendered notebook


tutorials/fpga.ipynb
tbennun, line 1: No inline comments:
* date -> dace
* explict -> explicit
* commented code blocks can just be code blocks with a comment in them
* optmize -> optimize
* "library nodes Library nodes" -> "library nodes. Library nodes are [high-level nodes](https://spcldace.readthedocs.io/en/latest/sdfg/ir.html#library-nodes) that can represent a wide variety of operators. In this case, the two matrix multiplications"...
* "expension this FPGA" -> "expansion. This FPGA"
* "expensions" -> "expansions" (apply everywhere)
* "need of off-chip" -> "need for off-chip"
* "multiple memory buffer being" -> "multiple memory buffers being"
* "we can distributed" -> "we can distribute"
* ", that" -> ", which"
* "of its program" -> "of their program"
* "applying DaCe auto-optimization heuristic" -> either "heuristics" or "the DaCe". The former is likely better
* "a set of simplification to" -> "a set of simplification passes to"
* "of the [...] ones" -> "of [...]."
* "can not currently being tuned by the user." - what does this mean?
* "Let's start again from the SDFG parsing" -> "Let's start again from parsing the SDFG"
* "explicitely" -> "explicitly"
* "execute FPGA program" -> "execute FPGA programs"
* "properly" -> ""
* "For example, to enable Hardware execution via command line (other" -> "Hardware execution can be enabled via the command line, or other"
* "as follow" -> "as follows"
* ", that" -> ", which"
* "bistream" -> "bitstream"
* "executed the program" -> "execute the program"

tbennun (changes requested):

tutorials/fpga.ipynb
TizianoDeMatteis: Thanks, these should be fixed now.

tbennun (approved):

PR #1337, opened 2023-07-28
CMakeLists.txt Improvements for CUDA

kylosus:
This PR bumps `cmake` version to `3.17` and replaces the deprecated `find_package(CUDA)` with [FindCUDAToolkit](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html), with a number of improvements to the compilation process:
- CUDA Include and Library directories are now handled automatically by `cmake`
- CUDA architecture handling is reworked: No more regex in `CMakeLists.txt` or manual `-gencode` string generation in python code.
- CUDA source files are now included directly in the targets: cmake handles proper compilation and linking of device code automatically.
- Similar modifications to `OpenMP` and `Threads` targets

PR #1441, opened 2023-11-22
Fix CUDA high-dimensional test

tbennun:
Related to #1337


BenWeber42 (approved):
Thanks!

Since this testcase only works with `uin64` (for which overflow is well-defined to wrap around), we could also improve the assertion to require full equality:

```python
    diff = np.linalg.norm(B_regression - B) / (N * M * K * L)
    print('Difference:', diff)
    assert diff <= 1e-5
```

I leave it up to you whether you want to include this in this PR. I can also implement this separately.

PR #1450, opened 2023-11-27
Fix array storage in TrivialMapElimination

edopao:
After removal of trivial maps, access nodes directly connected to a tasklet should not have GPU global storage. Set array storage to default. Added unittest for which validation failed in baseline.

### DIFF
--- a/dace/transformation/dataflow/trivial_map_elimination.py
+++ b/dace/transformation/dataflow/trivial_map_elimination.py
@@ -1,6 +1,7 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
 """ Contains classes that implement the trivial-map-elimination transformation. """
 
+from dace import dtypes
 from dace.sdfg import nodes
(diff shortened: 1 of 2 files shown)


tbennun:
This is an assumption I am not sure we should be making. When TrivialMapElimination is applied, the arrays should all be Default storage and the storage should be set later (Default would also set arrays to GPU if the map is scheduled to a GPU_Device).

edopao:
@tbennun Thank you Tal for this comment. I was probably applying the GPU storage too early, before optimize transformations:

1. sdfg.simplify()
2. autoopt.apply_gpu_storage(sdfg)
3. sdfg.apply_gpu_transformations()
4. sdfg = autoopt.auto_optimize(...)

and I changed it to:

1. sdfg.simplify()
2. sdfg = autoopt.auto_optimize(...)
3. sdfg.apply_gpu_transformations()

With the new sequence, I can run my sdfg on GPU without the change in this PR.

PR #1452, opened 2023-11-27
Do not serialize non-default fields by default

tbennun:
Added a configuration entry (enabled by default) that serializes only the modified fields in an SDFG. This leads to a reduction in size.

Merging this PR is contingent on updating the SDFG renderer to use the defaults/metadata for properties.


tbennun:
@phschaad do not merge until a release (not prerelease) of sdfv and the vscode plugin are both done.

PR #1457, opened 2023-11-29
`SDFG.arg_names` was not a member but a class variable.

philip-paul-mueller:
The member variable `arg_names` was not created in the constructor. Thus a class variable was created instead which was shared among all instances.


philip-paul-mueller:
According to my understanding the reason why it has not surfaced yet is due to `dace/frontend/python/parser.py:903`.

tbennun (approved):
LGTM, thanks!

PR #1459, opened 2023-11-29
Remove unused deps

### LINKED ISSUE #1458: Are `flask` & `requests` required dependencies?
They are both required installs as per the setup.py https://github.com/spcl/dace/blob/16c569b8bb71b0e3dfbe87206c975dc83427beca/setup.py#L76 but I can't see them used in the source?

**Is your feature request related to a problem? Please describe.**
Unneeded requirements can bloat environments

**Describe the solution you'd like**
Remove if they really are unneeded

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

jack-mcivor:
Fixes #1458

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ certifi==2023.7.22
 charset-normalizer==3.1.0
 click==8.1.3
 dill==0.3.6
-Flask==2.3.2
 fparser==0.1.3
 idna==3.4
 importlib-metadata==6.6.0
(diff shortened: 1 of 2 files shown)


tbennun:
It used to be a requirement, but not anymore. Thank you for finding this!

tbennun (approved):

jack-mcivor:
Ok great. Can we please run the tests again? I added jinja2, which is required here https://github.com/spcl/dace/blob/master/dace/cli/sdfv.py#L16

I also removed some transitive dependencies from requirements.txt which are required through flask only

PR #1461, opened 2023-11-29
Consider data descriptor offset when propagating memlets

lukastruemper:
This PR adds the data descriptor's offsets to the memlet propagation.


tbennun:
@acalotoiu I thought we were removing/deprecating offsets?

acalotoiu:
> @acalotoiu I thought we were removing/deprecating offsets?

Yes - but while we intend to do that, actually removing them is not an engineering effort planned for the short or medium term. I propose we allow the fix for the current state for now.

acalotoiu (approved):
LGTM

tbennun:
@lukastruemper do you know why this is failing and could you please fix that? Thanks!

PR #1462, opened 2023-11-30
Use State Fissioning to Generalize Transformations

lukastruemper:


dace/sdfg/utils.py
```diff
@@ -1295,6 +1295,10 @@ def inline_sdfgs(sdfg: SDFG, permissive: bool = False, progress: bool = None, mu
     from dace.transformation.interstate import InlineSDFG, InlineMultistateSDFG
 
     counter = 0
+
```
tbennun, line 1298: This will be slow. Please do not use apply_transformations_repeated in this pass as it is designed to be scalable w.r.t. large SDFGs.

dace/transformation/dataflow/prune_connectors.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
 from typing import Set, Tuple
 import re
+import networkx as nx
```
tbennun, line 4:
```suggestion
```

dace/transformation/helpers.py
```diff
@@ -687,6 +687,85 @@ def state_fission(sdfg: SDFG, subgraph: graph.SubgraphView, label: Optional[str]
...
+    # Bookkeeping
+    successors = set([node])
+    boundary_nodes = set()
+    orig_edges = set()
+
+    # Collect predecessors
+    if not isinstance(node, nodes.AccessNode):
+        for edge in state.in_edges(node):
+            for e in state.memlet_path(edge):
+                successors.add(e.src)
```
tbennun, line 704: This does not look like a successor. Perhaps what you want is to call it something like nodes_to_keep (for new state)

tests/inlining_test.py
```diff
@@ -372,6 +415,7 @@ def test(A: dace.float64[96, 32], B: dace.float64[42, 32]):
     # test_regression_reshape_unsqueeze()
     test_empty_memlets()
     test_multistate_inline()
```
tbennun, line 417: I'm more worried about a case where an access node is used around the nested SDFG. Please add a test case for the following:
```
  A[0]         B[0]
(A) -> [nsdfg] ----> (B)
    \             / B[1]
     -> [tasklet]   
    A[1]
```
i.e., same output access node, same input access node.

tbennun: According to the algorithm, it would do nothing. It would be important to see if `state_fission_after` will generate an empty state or some orphaned access nodes in this case.

tbennun (changes requested):
This is great, wanted to do that for a while. The reason why I didn't (esp. with inline multistate SDFG) is because of corner cases where the nested SDFG is not the only branch in the dataflow graph. Please add another test (or even more) to reflect on that behavior.
Otherwise looks good!

tests/inlining_test.py
```diff
@@ -127,15 +127,16 @@ def outerprog(A: dace.float64[20]):
...
 
     sdfg = outerprog.to_sdfg(simplify=True)
-    from dace.transformation.interstate import InlineMultistateSDFG
-    sdfg.apply_transformations(InlineMultistateSDFG)
-    assert sdfg.number_of_nodes() in (4, 5)
 
     A = np.random.rand(20)
     expected = np.copy(A)
-    outerprog.f(expected)
+    sdfg(expected)
```
tbennun, line 133: The test may fail in unexpected ways. `outerprog.f ` will run the original function and is what we expect should be the result of the function

tests/inlining_test.py
```diff
@@ -152,16 +153,103 @@ def outerprog(A: dace.float64[20]):
...
+    assert len(sdfg.states()) == 1
+    assert len([node for node in sdfg.start_state.data_nodes()]) == 3
+
+    A = np.random.rand(10)
+    B = np.random.rand(10)
+    C = np.random.rand(10)
+    expected_a = np.copy(A)
+    expected_b = np.copy(B)
+    expected_c = np.copy(C)
+    sdfg(expected_a, expected_b, expected_c)
```
tbennun, line 243: same

tests/inlining_test.py
```diff
@@ -152,16 +153,103 @@ def outerprog(A: dace.float64[20]):
...
+
+                b = 2 * a
+
+    sdfg = outerprog.to_sdfg(simplify=False)
+    sdfg.apply_transformations_repeated((StateFusion, InlineSDFG))
+    assert len(sdfg.states()) == 1
+
     A = np.random.rand(20)
-    expected = np.copy(A)
-    outerprog.f(expected)
```
tbennun, line 161: same

tests/inlining_test.py
```diff
@@ -152,16 +153,103 @@ def outerprog(A: dace.float64[20]):
             nested(A)
 
     sdfg = outerprog.to_sdfg(simplify=True)
+
+    A = np.random.rand(20)
+    expected = np.copy(A)
+    sdfg(expected)
```
tbennun, line 159: same

tbennun (changes requested):

tbennun (approved):

PR #1463, opened 2023-11-30
In-out connector's global source when connector becomes out-only at outer SDFG scopes.

alexnick83:
Adds utility-method support for the case of an in-out nested SDFG connector that is out-only at outer SDFG scopes.

### DIFF
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -1847,6 +1847,11 @@ def get_global_memlet_path_src(sdfg: SDFG, state: SDFGState, edge: MultiConnecto
         if len(pedges) > 0:
             pedge = pedges[0]
             return get_global_memlet_path_src(psdfg, pstate, pedge)
+        else:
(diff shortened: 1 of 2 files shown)


dace/sdfg/utils.py
```diff
@@ -1847,6 +1847,11 @@ def get_global_memlet_path_src(sdfg: SDFG, state: SDFGState, edge: MultiConnecto
         if len(pedges) > 0:
             pedge = pedges[0]
             return get_global_memlet_path_src(psdfg, pstate, pedge)
+        else:
```
tbennun, line 1850: Should we do this for dst too?

tbennun (approved):

dace/sdfg/utils.py
alexnick83: I don't think that the inverse is possible, i.e., if something is written in an inner scope, it should not be possible that it is only read in an outer scope.

PR #1442, opened 2023-11-22
Fix for CUDA codegen

edopao:
This PR addresses #1388: fix python codegen and `SharedToGlobal1D` template to generate correct code for write without reduction.

PR #1471, opened 2023-12-05
Fix for floordiv on GPU target

edopao:
This PR addresses a `floordiv` runtime error detected on GPU target: `std::ifloor` returns zero with integer argument (for example, the result of an `int` division).

This change was initially part of #1442. It has been moved to this PR to let #1442 focus on the CUDA-codegen issue.


tbennun (changes requested):
This looks good overall. I think there should be a test for correctness but otherwise approved.

edopao:
@tbennun Test added, I have also updated the return type of the `ifloor` template for integral argument type.

edopao:
Auto-format on save, I will remove all white-space changes. Sorry for that.

tbennun:
> Auto-format on save, I will remove all white-space changes. Sorry for that.

also, dace uses single-quote strings

edopao:
@tbennun PR updated with (working) test.

tbennun:
@edopao This makes the GPU tests 20 minutes (~60%) longer. I'd propose a more focused test regime.

tbennun (changes requested):
see comments

edopao:
@tbennun I agree, we should keep the duration of the CI jobs within a reasonable limit. I have updated the test.

PR #1479, opened 2023-12-13
Test propagation of strides for tasklet indirection (was working on v0.14.4)

edopao:
I was trying to bump the version of dace in gt4py dependencies from 0.14.4 to 0.15.1. Some tests in gt4py cartesian (the legacy gt4py version) start to fail, because of an exception in cpp code generation. I was able to reproduce the issue in the test case which is added by this PR.

The exception originates from:
```
self = <dace.codegen.targets.cpp.DaCeKeywordRemover object at 0x134060f70>
slicenode = <ast.Tuple object at 0x134060c10>, target = '_mat'

    def _subscript_expr(self, slicenode: ast.AST, target: str) -> symbolic.SymbolicType:
        ...
    
        if isinstance(visited_slice, ast.Tuple):
            # If slice is multi-dimensional and writes to array with more than 1 elements, then:
            # - Assume this is indirection (?)
            # - Soft-squeeze the slice (remove unit-modes) to match the treatment of the strides above.
            if target not in self.constants:
                desc = self.sdfg.arrays[dname]
                if isinstance(desc, data.Array) and data._prod(desc.shape) != 1:
                    elts = [e for i, e in enumerate(visited_slice.elts) if desc.shape[i] != 1]
            else:
                elts = visited_slice.elts
            if len(strides) != len(elts):
>               raise SyntaxError('Invalid number of dimensions in expression (expected %d, '
                                  'got %d)' % (len(strides), len(elts)))
E               SyntaxError: Invalid number of dimensions in expression (expected 2, got 1)
```

This PR proposes a fix, although I do not fully understand this code and there might be a better solution.

### DIFF
--- a/tests/subarray_test.py
+++ b/tests/subarray_test.py
@@ -1,13 +1,13 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
-import dace as dp
+import dace as dc
 import numpy as np
 
-W = dp.symbol('W')
+W = dc.symbol('W')
 
 
-@dp.program
+@dc.program
 def subarray(A, B):
-    @dp.map(_[0:W])
+    @dc.map(_[0:W])
     def subarrays(i):
         a << A[:, i, i, i]
         a2 << A(1)[i, i, i, :]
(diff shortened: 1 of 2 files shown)


tbennun:
@edopao @alexnick83 we had a discussion on this in one of our meetings. Instead of trying to "fill in the missing parts" of the memlet in case of a partial subscript (e.g., A[i] instead of A[i, 0]), in a tasklet we have to produce all the indices explicitly.

edopao:
> @edopao @alexnick83 we had a discussion on this in one of our meetings. Instead of trying to "fill in the missing parts" of the memlet in case of a partial subscript (e.g., A[i] instead of A[i, 0]), in a tasklet we have to produce all the indices explicitly.

Thanks, I interpret this comment as that the SDFG is not valid; and that, in this test case, gt4py should generate a tasklet with all explicit indices like this:
```
        tasklet = state.add_tasklet(
            'vec_to_mat', {'_vec'}, {'_mat'}, f"\
for i in range({N}):\
    _mat[0, i, i] = _vec[0, i] + dace.float64(1)\
        ")
```

Is it maybe even better to generate the arrays in the nested SDFGs without the extra dimension with size 1?
```
        sdfg.add_array(
            'vec_field', (N), dc.float64,
            strides=(vec_stride_d0,)
        )
        sdfg.add_array(
            'mat_field', (N, N), dc.float64,
            strides=(mat_stride_d0, mat_stride_d1)
        )
```

I will try this out. This SDFGs are generated from the legacy gt4py which I am not so familiar with.

tbennun:
That is correct.

edopao:
@tbennun I have tried your suggestion, setting all explicit indices in array access inside the tasklet. That worked for the case I previously showed in my new test case. In the latest commit, you can still refer to this case on the tasklet output memlet:
```
        sdfg.add_array(
            'mat_field', (1, N, N), dc.float64,
            strides=(mat_stride_W, mat_stride_d0, mat_stride_d1)
        )
        ...
        tasklet = state.add_tasklet(
            'vec_to_mat', {'_vec'}, {'_mat'}, f"\
for i in range({N}):\
    _mat[0, i, i] = _vec[0, i] + dace.float64(1)\
        ")
         state.add_edge(
            tasklet, '_mat',
            state.add_access('mat_field'), None,
            dc.Memlet.from_array('mat_field', sdfg.arrays['mat_field'])
        )      
```
However, I encountered a different case for which it was not enough to set all explicit indices. This case is reproduced on the tasklet input memlet:
```
        sdfg.add_array(
            'vec_field', (W, N), dc.float64,
            strides=(vec_stride_W, vec_stride_d0)
        )
        tasklet = state.add_tasklet(
            'vec_to_mat', {'_vec'}, {'_mat'}, f"\
for i in range({N}):\
    _mat[0, i, i] = _vec[0, i] + dace.float64(1)\
        ")
        state.add_edge(
            state.add_access('vec_field'), None,
            tasklet, '_vec',
            dc.Memlet.simple('vec_field', f"_w, 0:{N}")
        )
```
In this second case, the array slice is at index `_w`, not at `0`, so we need the stride for its corresponding dimension. This required some changes in the codegen, which are included in latest commit.

However, these changes caused `subarray_test.test` to fail. Therefore, I have also updated this test to use all explicit indices for array access inside tasklet.

tests/subarray_test.py
```diff
@@ -18,16 +18,112 @@ def subarrays(i):
...
+        dc.symbol(s) for s in ['vec_stride_W', 'vec_stride_d0']
+    )
+    mat_stride_W, mat_stride_d0, mat_stride_d1 = (
+        dc.symbol(s) for s in ['mat_stride_W', 'mat_stride_d0', 'mat_stride_d1']
+    )
+
+    def build_nsdfg():
+        sdfg = dc.SDFG('vec_to_mat')
+        sdfg.add_symbol('_w', dc.int32)
+        sdfg.add_array(
```
edopao, line 44: I have reverted all changes to code generation and legacy tests. I have found a pattern for array access inside the tasklet that works, although I don't understand it.

@tbennun Access to `_mat` requires all explicit indices, while access to `_vec` only works if I exclude the index for the `_w` slice.

edopao:
This PR should be an issue, opening #1486 to investigate it.

PR #1485, opened 2023-12-20
Reference-to-View pass and comprehensive reference test suite

tbennun:
Implements a reference-to-view pass (converting references to views if they are only set to one particular subset). Also improves the simplify pipeline in the presence of Reference data descriptors and adds multiple tests that use references.

### DIFF
--- a/dace/transformation/passes/reference_reduction.py
+++ b/dace/transformation/passes/reference_reduction.py
@@ -1,0 +1,247 @@
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
+from collections import defaultdict
+import copy
+from typing import Any, Dict, List, Optional, Set, Tuple
+
+from dace import SDFG, SDFGState, data, properties, Memlet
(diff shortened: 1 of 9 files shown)


dace/transformation/passes/reference_reduction.py
```diff
@@ -0,0 +1,247 @@
...
+            source = next(iter(reference_sources[cand]))
+            fsyms = source.subset.free_symbols
+            if not fsyms:
+                continue
+
+            for state, node in refsets[cand]:
+                # Check if any of the symbols is a scope symbol
+                entry = state.entry_node(node)
+                while entry is not None:
+                    if fsyms & entry.new_symbols(sdfg, state, {}):
```
phschaad, line 124: Lines 124-129 are not hit by the tests, if it's easy to add one that causes these edge cases then it may be good to add one. Otherwise nevermind.

phschaad (approved):
Nice addition, LGTM

PR #1487, opened 2023-12-20
Scalar to symbol promotion bugfix

tbennun:
When a scalar is used more than once in the same state, promotion incorrectly creates a data race and further promotes scalars incorrectly.


tbennun:
There were no bad SDFG codegens, only bad SDFGs to begin with.

PR #1492, opened 2023-12-29
CuPy fixes and special cases for HIP

tbennun:
CuPy with AMD HIP does not support `__cuda_array_interface__`. This PR adds special cases to support CuPy with HIP.
The PR also fixes the reduce node's GPUAuto schedule with respect to warp size.


acalotoiu (changes requested):
The PR is useful, and important and looks good to me. However, not all features added are tested. ~~Could you please add a test to ensure correctness?~~

acalotoiu (approved):
The PR is useful, and important and looks good to me.

PR #1498, opened 2024-01-07
More NumPy operation implementations

### LINKED ISSUE #1295: Incorrect parsing of parentheses in DaCe programs
It looks like parentheses don't get parsed correctly.
```
import dace
import numpy as np

N, F_in, F_out =  2, 3, 4

@dace.program
def fn(a, b, c):
  c[:] = (a.T @ b).T

a = np.random.rand(N, F_in)
b = np.random.rand(N, F_out)
c_expected = np.zeros((F_out, F_in))
c = np.zeros((F_out, F_in))

fn.f(a, b, c_expected)
fn(a, b, c)
print(c)
print(c_expected)
assert np.allclose(c, c_expected)
```
Throws an error:
```
DaceSyntaxError: Use of undefined variable "(a"
```

tbennun:
- [x] Concatenation and stacking (numpy.concatenate, numpy.stack, and their variants)
- [x] numpy.linspace
- [x] Fix nested attribute parsing (Fixes #1295)
- [x] numpy.clip
- [x] numpy.split and its variants
- [x] numpy.full variants (zeros, ones, etc.) with a single value for shape (`np.zeros(N)`)
- [x] NumPy-compatible numpy.arange dtype inference
- [x] `numpy.fft.{fft, ifft}`


dace/frontend/python/newast.py
```diff
@@ -4426,7 +4430,7 @@ def visit_Call(self, node: ast.Call, create_callbacks=False):
                 arg = self.scope_vars[modname]
             else:
                 # Fallback to (name, object)
-                arg = (modname, self.defined[modname])
+                arg = modname
```
alexnick83, line 4433: Is this change handled already in the code somehow, i.e., is `self.defined` queried on demand?

dace/frontend/python/newast.py
```diff
@@ -4673,30 +4683,39 @@ def visit_NameConstant(self, node: NameConstant):
         return self.visit_Constant(node)
 
     def visit_Attribute(self, node: ast.Attribute):
-        # If visiting an attribute, return attribute value if it's of an array or global
-        name = until(astutils.unparse(node), '.')
-        result = self._visitname(name, node)
+        result = self.visit(node.value)
+        if isinstance(result, (tuple, list, dict)):
+            if len(result) > 1:
```
alexnick83, line 4688: Is this the fix for attributes on expressions?

dace/frontend/python/replacements.py
```diff
@@ -527,30 +535,43 @@ def _arange(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, *args, **kwargs):
     else:
         start, stop, step = args
 
+    if isinstance(start, str):
```
alexnick83, line 538: The underlying issue here is data-dependent array shape?

dace/frontend/python/replacements.py
```diff
@@ -527,30 +535,43 @@ def _arange(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, *args, **kwargs):
...
+            shape = (stop - start,)
+        else:
+            shape = (symbolic.int_ceil(stop - start, step), )
     else:
         shape = (np.ceil((stop - start) / step), )
 
-    if not isinstance(shape[0], Number) and ('dtype' not in kwargs or kwargs['dtype'] == None):
-        raise NotImplementedError("The current implementation of numpy.arange requires that the output dtype is given "
-                                  "when at least one of (start, stop, step) is symbolic.")
+    # Infer dtype from input arguments
```
alexnick83, line 560: The previous code looks suspicious. There is probably a reason we didn't fall back to using `_result_type.` Probably `arange` only works with integers or something. I guess it is not important if the code is "correct" (numpy-wise), but it is good to keep in mind.

dace/libraries/standard/nodes/transpose.py
```diff
@@ -140,27 +146,37 @@ class ExpandTransposeOpenBLAS(ExpandTransformation):
...
             alpha = "1.0"
+            cast = ''
         elif dtype == dace.complex64:
             func = "comatcopy"
             alpha = "dace::blas::BlasConstants::Get().Complex64Pone()"
+            cast = '(float*)'
         elif dtype == dace.complex128:
             func = "zomatcopy"
             alpha = "dace::blas::BlasConstants::Get().Complex128Pone()"
+            cast = '(double*)'
```
alexnick83, line 171: I have the impression that this has been fixed in another PR, but maybe it is stuck in the queue. Just noting it in case merge issues appear later.

alexnick83 (approved):
Looks fine

dace/frontend/python/newast.py
tbennun: It’s a code path that we didn’t cover before in coverage. The rest of the code assumes a string return type rather than a tuple, and crashes somewhere else.

dace/frontend/python/newast.py
tbennun: Exactly

dace/frontend/python/replacements.py
tbennun: Yes, though generally solvable with a symbol assignment and another state, for now at least symbols and constants should be supported.

dace/frontend/python/replacements.py
tbennun: It may have been written before/without knowledge of _result_type

dace/libraries/standard/nodes/transpose.py
tbennun: It was fixed for some BLAS operators but not others. Also there is some tensor transpose PR but I think it’s GPU focused.

PR #1499, opened 2024-01-07
`DeadDataFlowElimination` will add type hint when removing a connector

luca-patrignani:
Issue #1150 that `DeadDataflowElimination` removes a connector from a Tasklet which leaves a variable without type hint.
This PR tries to fix this bug by adding a type hint expression for a variable which is used in the tasklet. It adds the type hint only if the variable is used inside the tasklet code (I checked using `ASTFindReplace`).
The PR also adds a test which is literaly the code presented in #1150 and asserts the presence of the type hint and checks if it compiles.

### May need confirmation
- [ ] Did I use `ASTFindReplace` correctly?
- [ ] If the type inference fails no type hint is added. Is it the right solution?
- [ ] Does the test even make sense? (I don't have much experience in unit testing).

This is my first PR for this project so be patient with me.


luca-patrignani:
@tbennun can you please approve the workflows?

dace/transformation/passes/dead_dataflow_elimination.py
```diff
@@ -137,16 +139,23 @@ def apply_pass(self, sdfg: SDFG, pipeline_results: Dict[str, Any]) -> Optional[D
...
-                                    # Add definition
                                     leaf.src.code.code = f'{ctype.as_arg(leaf.src_conn)};\n' + leaf.src.code.code
+                                elif leaf.src.code.language == dtypes.Language.Python:
+                                    if ctype is not None:
+                                        ast_find = astutils.ASTFindReplace({}, trigger_names={leaf.src_conn})
+                                        try:
+                                            for code in leaf.src.code.code:
+                                                ast_find.generic_visit(code)
+                                        except astutils.NameFound:
+                                            leaf.src.code.code = [ast.parse(f'{leaf.src_conn}: dace.{ctype.to_string()}\n')] + leaf.src.code.code
```
tbennun, line 158: This will likely not work well, the type of the AST node that ast.parse results in is ast.Module.

tbennun:
Thank you for the PR! Unfortunately our CI is currently undergoing fixing so failing tests may not mean that your PR broke them.

tbennun:
@luca-patrignani as for your questions:
1. It is a bit of a nonstandard use for ASTFindReplace, but should be OK. Please add comments annotating what the try/except block means so that future developers can easily see that you do not intend to replace anything.
2. If done correctly, type inference should always pass, and if not the code generator will use the `auto` keyword (which was done before this PR and not always correct).
3. The test looks good. If you're already compiling the SDFG, I would also run it with random numpy arrays and verify that it produces the correct output (`np.where` will produce the same answer in the Python world).

luca-patrignani:
@tbennun can you please approve the workflows once again?

luca-patrignani:
@tbennun when you have some free time can we continue this PR review?

tbennun:
@luca-patrignani please stop merging the master branch to this one. When merging the PR, we will squash and merge the updated version as necessary.

tbennun (approved):
LGTM, thank you again for contributing this fix!

PR #1504, opened 2024-01-13
Generalize StructArrays to ContainerArrays and refactor View class structure

tbennun:
This PR enables the use of an array data descriptor that contains a nested data descriptor (e.g., ContainerArray of Arrays). Its contents can then be viewed normally with View or StructureView.
With this, concepts such as jagged arrays are natively supported in DaCe (see test for example).
Also adds support for using ctypes pointers and arrays as arguments to SDFGs.

This PR also refactors the notion of views to a View interface, and provides views to arrays, structures, and container arrays. It also adds a syntactic-sugar/helper API to define a view of an existing data descriptor.


dace/codegen/compiled_sdfg.py
```diff
@@ -516,13 +510,15 @@ def _construct_args(self, kwargs) -> Tuple[Tuple[Any], Tuple[Any]]:
                     if atype.optional is False:  # If array cannot be None
                         raise TypeError(f'Passing a None value to a non-optional array in argument "{a}"')
                     # Otherwise, None values are passed as null pointers below
+                elif isinstance(arg, ctypes._Pointer):
```
alexnick83, line 513: Could we leave a note here about the use of an underscored type? I am mostly concerned about the developers changing or removing it in a future version.

alexnick83 (approved):
Looks fine

dace/codegen/compiled_sdfg.py
tbennun: Interestingly, it is part of the Python stable API: https://docs.python.org/3/library/ctypes.html#ctypes._Pointer

PR #1505, opened 2024-01-14
Generalize container views

tbennun:
This PR refactors the notion of views to a `View` interface, and provides views to arrays, structures, and container arrays. It also adds a syntactic-sugar/helper API to define a view of an existing data descriptor. Depends on merging #1504


alexnick83 (approved):
Looks fine

PR #1513, opened 2024-02-01
Dependency Edge Hotfix

Berke-Ates:
This PR adds an additional check for dependency edges during codegen for CPU and omits the copy operation generation accordingly.


phschaad (approved):

dace/codegen/targets/cpu.py
```diff
@@ -588,6 +588,9 @@ def _emit_copy(
         elif isinstance(src_node, nodes.CodeNode) and isinstance(dst_node, nodes.CodeNode):
             # Code->Code copy (not read nor write)
             raise RuntimeError("Copying between code nodes is only supported as part of the participating nodes")
+        elif uconn == None and vconn == None and memlet.data == None and dst_schedule is dtypes.ScheduleType.Sequential:
```
tbennun, line 591: all of the tests here are inverted. `is None` and `== <enum entry>` is what you should use.

Berke-Ates: Thanks! Fixed it

tbennun (approved):

PR #1518, opened 2024-02-13
Modernize HIP CMake commands, fix corner cases

tbennun:
Enables support for newer HIP versions with CMake, fix branch ordering when making data descriptors.


phschaad (approved):
Just out of curiosity, what was the issue with respect to the data construction ordering?

tbennun:
> Just out of curiosity, what was the issue with respect to the data construction ordering?

PyTorch tensors expose `__*_array_interface__` but fail to actually retrieve it.

PR #1526, opened 2024-02-19
Fixed an issue in the Memlet duplication verification.

philip-paul-mueller:
It was not considered that an empty memlet has `data` equal to `None`. However, empty Memlets should not be considered as duplicated data as they do not move data.

### DIFF
--- a/dace/sdfg/validation.py
+++ b/dace/sdfg/validation.py
@@ -589,7 +589,9 @@ def validate_state(state: 'dace.sdfg.SDFGState',
                 f'Duplicate memlet detected: "{e}". Please copy objects '
                 'rather than using multiple references to the same one', sdfg, state_id, eid)
         references.add(id(e))
-        if id(e.data) in references:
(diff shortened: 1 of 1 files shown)


dace/sdfg/validation.py
```diff
@@ -589,7 +589,9 @@ def validate_state(state: 'dace.sdfg.SDFGState',
                 f'Duplicate memlet detected: "{e}". Please copy objects '
                 'rather than using multiple references to the same one', sdfg, state_id, eid)
         references.add(id(e))
-        if id(e.data) in references:
+        if(e.data.is_empty()):
```
tbennun, line 592:
```suggestion
        if e.data.is_empty():
```
Please see our coding guidelines for style.

tbennun (changes requested):

dace/sdfg/validation.py
philip-paul-mueller: Sorry, I always forget that.

PR #1531, opened 2024-02-22
Updated and fixed the MapDimShuffle tranformation.

philip-paul-mueller:
The reordering of the dimension was quite strange, it is now much more cleaner. Furthermore, the `tile_sizes` parameter of the map range was ignored, i.e. not changed. Furthermore, the parameters are now correclty copied. Before it was just assigned, thus if later the parameter list of the map was again changed this effect would propagate to _all_ maps that where treated.


philip-paul-mueller:
@lukastruemper 
I added one of the tests you asked, however, the `None` case can not be tested, because such a transformation can not be created.
The test was there before (I just moved it from the `apply()` to the `can_be_applied()` function) so I keep it.

PR #1535, opened 2024-02-26
Fixed a bug in the map fusion transformation.

philip-paul-mueller:
Essentially the bug happens because the transformation considers any array with one elements as scalar, including arrays such as `(1, 1)`. This commit allows the scalar branch of the transformation to handle this kind of situation. Another solution would be to "promote" it to an array.

In addition this commit adds some test cases for these kind of events.

### DIFF
--- a/dace/transformation/dataflow/map_fusion.py
+++ b/dace/transformation/dataflow/map_fusion.py
@@ -475,7 +475,7 @@ class MapFusion(transformation.SingleStateTransformation):
                 for other_edge in out_edges:
                     if other_edge is not edge:
                         graph.remove_edge(other_edge)
-                        mem = Memlet(data=local_name, other_subset=other_edge.data.dst_subset)
+                        mem = Memlet(data=local_name, subset="0", other_subset=other_edge.data.dst_subset)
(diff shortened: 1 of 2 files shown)


BenWeber42:
The test seems to pass on `master`. I think we should have a test that triggers the bug on `master`...

I looked a bit deeper into the transformation. I also think that it's better for the scalar branch to handle this case, because inside the map it's better to use a scalar transient instead of a transient array where all dimensions have size 1.

BenWeber42:
We briefly discussed this:

- The _scalar branch_ ist the better place to fix this (instead of _array branch_)
- AlexNik confirmed that `"0"` is the correct subset for memlets from scalars

We are now looking into the `assert` (seems to highlight another bug) and whether the test triggers the bug (it triggers new code paths in the transformation, and, thus, increases code coverage, so is strictly an improvement).

philip-paul-mueller:
As far as I can tell there is still a bug in the array branch.

philip-paul-mueller:
Has been superseded by [PR#1629](https://github.com/spcl/dace/pull/1629)

Issue #424, opened 2020-10-22
(Feature) Warn on potential data races


tbennun:
For example, a map that has the same transient name multiple times and the memlet ranges overlap should trigger a warning.

Might be expensive to check so may require adding a config entry for this.

meshtag:
I am interested in working on this issue. Can someone provide a minimum example which produces an SDFG with memlet ranges overlap.

tbennun:
@meshtag sounds good! In order to create this situation, one would have to manually create an SDFG with the API, since our Python frontend will not let the two statements be in one state.
There are many examples of how we create two-state SDFGs that have races here:
https://github.com/spcl/dace/blob/master/tests/transformations/state_fusion_test.py

Please let us know if you have a specific example you want to create or if we can help with the SDFG API!

meshtag:
Apologies for the late reply.

I had a look at SDFG API and wanted to confirm whether the following SDFG is a valid example of memlet range overlap : 
```
import dace
import numpy as np

def memlet_range_overlap():
    sdfg = dace.SDFG('memlet_range_overlap')
    sdfg.add_array('A', [5], dace.int32)
    sdfg.add_array('B', [5], dace.int32)
    state = sdfg.add_state()
    r = state.add_read('A')
    t1 = state.add_tasklet('first_tasklet', {'a'}, {'b'}, 'b[0] = a[0] + 10')
    t2 = state.add_tasklet('second_tasklet', {'a'}, {'b'}, 'b[1] = a[1] - 20')
    w = state.add_write('B')
    state.add_edge(r, None, t1, 'a', dace.Memlet('A[0:2]'))
    state.add_edge(r, None, t2, 'a', dace.Memlet('A[0:2]'))
    state.add_edge(t1, 'b', w, None, dace.Memlet('B[0:2]'))
    state.add_edge(t2, 'b', w, None, dace.Memlet('B[2:4]'))

    A = np.ones([5], dtype=np.int32)
    B = np.zeros([5], dtype=np.int32)
    sdfg(A=A, B=B)
    print(A)
    print(B)

if __name__ == '__main__':
    memlet_range_overlap()

```

tbennun:
Looks like it is not. The writes to B will be to indices 0,1 in t1; and 2,3 in t2.
There are overlapping reads but that should not be a data race issue.

tbennun:
Merged in #1541, thanks!

PR #1541, opened 2024-03-04
Warn on potential data races

luca-patrignani:
This PR aims to raise a warning if two or more memlets are writing to the same access node and their ranges are overlapping. It also adds a new config entry to activate this check. 
This PR tries to close issue #424.

Needs confirmation:
- [ ] Are the new tests overcomplicated?
- [ ] Do we need some unit tests for `are_intersecting`?


luca-patrignani:
@tbennun  when you got some free time can you review this? Thanks

dace/subsets.py
```diff
@@ -813,6 +813,17 @@ def intersects(self, other: 'Range'):
...
         return True
 
+    @staticmethod
+    def are_intersecting(ranges: List['Range']) -> bool:
+        """ Check if any of the given ranges are intersecting
+            :param ranges: the ranges to be checked.
+        """
+        for i in range(len(ranges)):
+            for j in range(i+1, len(ranges)):
+                if ranges[i].intersects(ranges[j]):
```
tbennun, line 823: Subset.intersects has four potential return values: True (if definitely intersects), False (if definitely does not intersect), None (unsupported cases), or raising TypeError (if the result is indeterminate).
Additionally, Range will not cover all cases (e.g., Indices).

You may want to use [dace.subsets.intersects](https://github.com/spcl/dace/blob/master/dace/subsets.py#L1294), which has only three return values (True, False, None) and accepts all subset types. Also, you will need to address the None case (ignoring it). Lastly, I would move the loop itself to the validation routine and not add another API for this.

dace/viewer/webclient
tbennun, line 1: There should be no changes to the submodule. Please revert this.

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
+
+import warnings
+import dace
+import numpy as np
+import pytest
+
```
tbennun, line 7: These tests are OK, but you may want to add more tests for data races. Examples:

1.
```
(A) -> (B)
(C) -> (A)
```

2.
```
(A) -> map -> (B)
(A) -> map -> (B)
```
(i.e., with two different access nodes)

3. symbolic overlap (2*i, i)
4. constant memlet overlap (3:10, 6:12) and "almost-overlaps" (3:10, 10:20)

tbennun (changes requested):
Thank you for the contribution! Overall this is a good start, but it will not cover the majority of cases of data races that may occur in dace programs (see comments). Also, please use the `subsets.intersects` API.

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    N = 6
+    A = np.arange(N, dtype=np.int32)
+    B = np.zeros((N,), dtype=np.int32)
+    C = 20 * A
+
+    with pytest.warns(UserWarning):
+        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
+            sdfg(N=N, A=A, B=B, C=C)
```
tbennun, line 163: wouldn't you rather call `sdfg.validate()` here instead of running the program?

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    N = 6
+    A = np.arange(2*N, dtype=np.int32)
+    B = np.zeros((2*N,), dtype=np.int32)
+    C = 20 * A
+
+    with pytest.warns(UserWarning):
+        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
+            sdfg(N=N, A=A, B=B, C=C)
```
tbennun, line 202: same as above

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+        output_nodes={"B": B}
+    )
+
+    A = np.arange(20, dtype=np.int32)
+    B = np.zeros((20,), dtype=np.int32)
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", UserWarning)
+        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
+            sdfg(A=A, B=B)
```
tbennun, line 275: same

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+        output_nodes={"B": B}
+    )
+
+    N = 6
+    A = np.arange(N//2, dtype=np.int32)
+    B = np.zeros((N,), dtype=np.int32)
+    with warnings.catch_warnings():
+        warnings.simplefilter("error", UserWarning)
+        with dace.config.set_temporary("experimental.check_race_conditions", value=True):
+            sdfg(N=N, A=A, B=B)
```
tbennun, line 44: wouldn't you rather call sdfg.validate() here instead of running the program?

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+        input_nodes={"A": A},
+        output_nodes={"B": B}
+    )
+
+    N = 6
+    A = np.arange(N, dtype=np.int32)
+    B = np.zeros((N,), dtype=np.int32)
+    with pytest.warns(UserWarning):
+        with dace.config.set_temporary("experimental.check_race_conditions", value=True):
+            sdfg(N=N, A=A, B=B)
```
tbennun, line 81: wouldn't you rather call sdfg.validate() here instead of running the program?

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    N = 6
+    A = np.arange(N, dtype=np.int32)
+    B = np.zeros((N,), dtype=np.int32)
+    C = 20 * A
+
+    with pytest.warns(UserWarning):
+        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
+            sdfg(N=N, A=A, B=B, C=C)
```
tbennun, line 122: wouldn't you rather call sdfg.validate() here instead of running the program?

tests/warn_on_potential_data_race_test.py
```diff
@@ -0,0 +1,285 @@
...
+        input_nodes={"A": A},
+        output_nodes={"B": B}
+    )
+
+    A = np.arange(12, dtype=np.int32)
+    B = np.zeros((12,), dtype=np.int32)
+
+    with pytest.warns(UserWarning):
+        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
+            sdfg(A=A, B=B)
```
tbennun, line 238: same as above

dace/sdfg/validation.py
```diff
@@ -790,6 +792,33 @@ def validate_state(state: 'dace.sdfg.SDFGState',
...
+
+        for node_label in node_labels:
+            write_memlet_subsets = write_accesses[node_label]
+            read_memlet_subsets = read_accesses[node_label]
+            # check write-write data races
+            for i in range(len(write_memlet_subsets)):
+                for j in range(i+1, len(write_memlet_subsets)):
+                    intersects = subsets.intersects(write_memlet_subsets[i], write_memlet_subsets[j])
+                    if intersects is not None:
+                        if intersects:
```
tbennun, line 812: if you wish, you can replace that with `if intersects is True:`

dace/sdfg/validation.py
```diff
@@ -790,6 +792,33 @@ def validate_state(state: 'dace.sdfg.SDFGState',
...
+                            warnings.warn(f'Memlet range overlap while writing to "{node}" in state "{state.label}"')
+            # check read-write data races
+            for write in write_memlet_subsets:
+                for read in read_memlet_subsets:
+                    intersects = subsets.intersects(write, read)
+                    if intersects is not None:
```
tbennun, line 818: same as above

dace/sdfg/validation.py
```diff
@@ -790,6 +792,33 @@ def validate_state(state: 'dace.sdfg.SDFGState',
...
+    if Config.get_bool("experimental.check_race_conditions"):
+        node_labels = []
+        write_accesses = defaultdict(list)
+        read_accesses = defaultdict(list)
+        for node in state.data_nodes():
+            node_labels.append(node.label)
+            write_accesses[node.label].extend([e.data.dst_subset for e in state.in_edges(node)])
+            read_accesses[node.label].extend([e.data.src_subset for e in state.out_edges(node)])
+
+        for node_label in node_labels:
```
tbennun, line 804: A few notes on this test:
1. You should keep the actual access nodes in your dictionary. If a map that reads from `A[i]` also writes to `A[i]`, that would trigger a validation warning incorrectly. Instead you need to check if:
  * The node is the same (in that case, write-write conflicts may still happen)
  * If the nodes are not the same, whether there is a path from the first to the second (use `has_path` to check that). If there is a path, then there is no conflict (read/write or write/write). This is another reason for this check being expensive. There may be some analysis passes in `dace.transformation.passes` that give you the paths between dependent access nodes, but creating a dictionary would be an optimization. To be safe you can use `has_path` for now and leave a comment that this could be optimized.
2. Write-conflict resolution memlets (`memlet.wcr is not None`) are exempt from write-write conflicts (maybe not read-write)

Please also add an elementwise map test (A->mapentry-A[i]->tasklet-A[i]->mapexit->A) just to make sure.

tbennun (changes requested):
This is much improved! The data race check still needs one more change to be correct (see comment)

dace/sdfg/validation.py
luca-patrignani: I tried to follow your notes but I felt kind of lost, especially for point 2. Consider this two test cases:

```
def test1():
    sdfg = dace.SDFG('memlet_overlap_with_wcr')
    state = sdfg.add_state()
    sdfg.add_array("A", (20,), dace.int32)
    sdfg.add_array("B", (1,), dace.int32)
    A = state.add_read("A")
    B = state.add_write("B")

    state.add_mapped_tasklet(
        name="first_reduction",
        code="b = a",
        inputs={"a": dace.Memlet(data="A", subset="k")},
        outputs={"b": dace.Memlet(data="B", subset="0", wcr="lambda old, new: old + new")},
        map_ranges={"k": "0:20"},
        external_edges=True,
        input_nodes={"A": A},
        output_nodes={"B": B}
    )

    state.add_mapped_tasklet(
        name="second_reduction",
        code="b = a",
        inputs={"a": dace.Memlet(data="A", subset="k")},
        outputs={"b": dace.Memlet(data="B", subset="0", wcr="lambda old, new: old * new")},
        map_ranges={"k": "0:20"},
        external_edges=True,
        input_nodes={"A": A},
        output_nodes={"B": B}
    )

    with warnings.catch_warnings():
        warnings.simplefilter("error", UserWarning)
        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
            sdfg.validate()

def test2():
    sdfg = dace.SDFG('memlet_overlap_with_wcr')
    state = sdfg.add_state()
    sdfg.add_array("A", (20,), dace.int32)
    sdfg.add_array("B", (1,), dace.int32)
    A = state.add_read("A")
    B = state.add_write("B")

    state.add_mapped_tasklet(
        name="first_reduction",
        code="b = a",
        inputs={"a": dace.Memlet(data="A", subset="k")},
        outputs={"b": dace.Memlet(data="B", subset="0", wcr="lambda old, new: old + new")},
        map_ranges={"k": "0:20"},
        external_edges=True,
        input_nodes={"A": A},
        output_nodes={"B": B}
    )

    state.add_mapped_tasklet(
        name="second_reduction",
        code="b = a",
        inputs={"a": dace.Memlet(data="A", subset="k")},
        outputs={"b": dace.Memlet(data="B", subset="0")},
        map_ranges={"k": "0:20"},
        external_edges=True,
        input_nodes={"A": A},
        output_nodes={"B": B}
    )

    with warnings.catch_warnings():
        warnings.simplefilter("error", UserWarning)
        with dace.config.set_temporary('experimental', 'check_race_conditions', value=True):
            sdfg.validate()
```

Are they considered data races? At this moment my code raises no warnings.
I'm probably missing something important, sorry for this long PR review.

luca-patrignani:
@tbennun when I got some free time can you review the last changes? I left a reply on your last comment.

phschaad:
Pinging @tbennun

tbennun (approved):
Thank you for making the corrections, LGTM!

tbennun:
@luca-patrignani @phschaad if possible, can we revive this so we can merge it? Thanks!

phschaad:
The PR sis revived in https://github.com/spcl/dace/pull/1712 :)

PR #1543, opened 2024-03-07
Fixing some particular bugs in the redundant array transformations.

philip-paul-mueller:
The bug happened during the update of some memlets. Before the transformation updated the memlets unconditionally. However, this PR changes the behavior such that the update only happen if the memlet operates on the data that is removed.


dace/transformation/dataflow/redundant_array.py
```diff
@@ -558,8 +558,12 @@ def apply(self, graph, sdfg):
             wcr = e1.data.wcr
             wcr_nonatomic = e1.data.wcr_nonatomic
             for e3 in path:
+                if e3.data.data != in_array.data:
```
alexnick83, line 561: Something to check is whether `other_subset` exists. That still has to be updated.

PR #1544, opened 2024-03-08
Add dtype for numpy.uintp which is compatible with C uintptr_t

kotsaloscv:
Need this to pass C pointers to DaCe sdfg and reinterpret cast them inside a tasklet


alexnick83 (approved):
Can you add a small test to show example usage and ensure that proper code is generated?

tbennun (approved):

tests/uintptr_t.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
```
tbennun, line 1: Copyright year

tests/uintptr_t.py
```diff
@@ -0,0 +1,22 @@
...
+import dace
+import ctypes
+import numpy as np
+
+
+def test_uintptr_t():
```
tbennun, line 7: Test is irrelevant, please try using it in a dace program so that we see the full code generation pipeline working. See tests/numpy for some examples you can add the type to

tests/uintptr_t.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
+import dace
```
tbennun, line 2: File name does not follow our name convention

tbennun (changes requested):

tests/uintptr_t_test.py
```diff
@@ -0,0 +1,35 @@
...
+    size_of_np_uintp = np.uintp().itemsize
+    # Dace uintptr_t representation
+    size_of_dace_uintp = dace.uintp.bytes
+
+    assert size == size_of_np_uintp == size_of_dace_uintp
+
+    A = np.ndarray([N, M], dtype=np.uintp)
+    out = empty_like1(A)
+
+    assert (out.dtype == np.uintp) and (out.nbytes == M*N*size)
```
tbennun, line 31: this test will likely always succeed, because the types are managed outside of dace. Is there a way to fill the array with a value that you can verify?

alexnick83 (approved):
Looks good now. Thanks!

kotsaloscv:
@tbennun is it ok to merge?

tbennun:
No. The test is not portable (assumes g++ exists, assumes /tmp exists, assumes Linux) and too elaborate for what it should test. I’ll make a simple test myself and edit your branch if you don’t mind

kotsaloscv:
> No. The test is not portable (assumes g++ exists, assumes /tmp exists, assumes Linux) and too elaborate for what it should test. It also leaks 4 bytes. I’ll make a simple test myself and edit your branch if you don’t mind

more than welcome to suggest another test.

tbennun:
@kotsaloscv see new test

tbennun (approved):

PR #1546, opened 2024-03-11
Fixed edge consolidation

philip-paul-mueller:
The main issue was a typo were `src` was accessed instead of `dst`.

### DIFF
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -641,7 +641,7 @@ def remove_edge_and_dangling_path(state: SDFGState, edge: MultiConnectorEdge):
         else:
             if e.dst_conn:
                 e.dst.remove_in_connector(e.dst_conn)
-                e.src.remove_out_connector('OUT' + e.dst_conn[2:])
+                e.dst.remove_out_connector('OUT' + e.dst_conn[2:])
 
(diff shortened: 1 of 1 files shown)


dace/sdfg/utils.py
```diff
@@ -641,9 +644,10 @@ def remove_edge_and_dangling_path(state: SDFGState, edge: MultiConnectorEdge):
         else:
             if e.dst_conn:
                 e.dst.remove_in_connector(e.dst_conn)
-                e.src.remove_out_connector('OUT' + e.dst_conn[2:])
+                e.dst.remove_out_connector('OUT' + e.dst_conn[2:])
 
         # Continue traversing upwards
+        # TODO: What is about the children of the tree.
```
tbennun, line 650: What does this mean?

philip-paul-mueller: It was a question for you (not you personally but the DaCe Developer).
I was just wondering, why the children where not handled?

philip-paul-mueller: Okay this makes sense.

philip-paul-mueller: Updated.

tbennun (approved):

PR #1548, opened 2024-03-14
Fix Profiler + Minor improvements

JanKleine:
There is currently a bug in the profiler, that the first run recorded includes the runtime of all the warmup runs, making one ugly outlier. This is because `times[0]` is set before the warmups and is never properly reset.

This fixes that and makes the surrounding code a bit cleaner (e.g. using `np.diff` instead of list comprehension).

I also added some options that make the profiler more usable when it's called inside of other `tqdm` progress bars (allows setting `tqdm`'s `leave=False` and stop it from printing results).

List of changes:
- Fix: timing error
- Add: `tqdm_leave` parameter
- Add: `print_results` parameter
- Cleanup: Only print `Profiling...` if `tqdm` is not used
- Cleanup: use `warnings.warn()` instead of `print("WARNING: ...")`
- Cleanup: remove unused variable for return value (didn't seem to be used)
- Cleanup: make diff calculation more concise

### DIFF
--- a/dace/builtin_hooks.py
+++ b/dace/builtin_hooks.py
@@ -40,10 +40,10 @@ def profile(
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
-    :param tqdm_leave: Sets the leave parameter of the tqdm progress bar (usefull 
+    :param tqdm_leave: Sets the ``leave`` parameter of the ``tqdm`` progress bar (usefull
         for nested progress bars). Ignored if tqdm progress bar is not used.
-    :param print_results: Whether or not to print the average execution time after 
-        all repetitions (usefull to disable when using nested progress bars).
+    :param print_results: Whether or not to print the median execution time after
+        all repetitions.
     :note: Running functions multiple times may affect the results of the program.
     """
     from dace.frontend.operations import CompiledSDFGProfiler  # Avoid circular import
--- a/dace/builtin_hooks.py
+++ b/dace/builtin_hooks.py
@@ -40,6 +40,10 @@ def profile(
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
+    :param tqdm_leave: Sets the leave parameter of the tqdm progress bar (usefull 
+        for nested progress bars). Ignored if tqdm progress bar is not used.
+    :param print_results: Whether or not to print the average execution time after 
+        all repetitions (usefull to disable when using nested progress bars).
     :note: Running functions multiple times may affect the results of the program.
     """
     from dace.frontend.operations import CompiledSDFGProfiler  # Avoid circular import
--- a/dace/builtin_hooks.py
+++ b/dace/builtin_hooks.py
@@ -15,7 +15,12 @@ if TYPE_CHECKING:
 
 
 @contextmanager
-def profile(repetitions: int = 100, warmup: int = 0):
+def profile(
+    repetitions: int = 100,
+    warmup: int = 0,
+    tqdm_leave: bool = True,
+    print_results: bool = True,
+):
     """
     Context manager that enables profiling of each called DaCe program. If repetitions is greater than 1, the
     program is run multiple times and the average execution time is reported.
@@ -51,7 +56,7 @@ def profile(repetitions: int = 100, warmup: int = 0):
             yield hook
             return
 
-    profiler = CompiledSDFGProfiler(repetitions, warmup)
+    profiler = CompiledSDFGProfiler(repetitions, warmup, tqdm_leave, print_results)
 
     with on_compiled_sdfg_call(context_manager=profiler):
         yield profiler
--- a/dace/builtin_hooks.py
+++ b/dace/builtin_hooks.py
@@ -40,7 +40,7 @@ def profile(
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
-    :param tqdm_leave: Sets the ``leave`` parameter of the ``tqdm`` progress bar (usefull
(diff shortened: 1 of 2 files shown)


dace/builtin_hooks.py
```diff
@@ -35,6 +40,10 @@ def profile(repetitions: int = 100, warmup: int = 0):
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
+    :param tqdm_leave: Sets the leave parameter of the tqdm progress bar (usefull
```
tbennun, line 43:
```suggestion
    :param tqdm_leave: Sets the ``leave`` parameter of the ``tqdm`` progress bar (useful 
```

dace/builtin_hooks.py
```diff
@@ -35,6 +40,10 @@ def profile(repetitions: int = 100, warmup: int = 0):
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
+    :param tqdm_leave: Sets the leave parameter of the tqdm progress bar (usefull 
+        for nested progress bars). Ignored if tqdm progress bar is not used.
+    :param print_results: Whether or not to print the average execution time after
```
tbennun, line 45:
```suggestion
    :param print_results: Whether or not to print the median execution time after 
```

dace/builtin_hooks.py
```diff
@@ -35,6 +40,10 @@ def profile(repetitions: int = 100, warmup: int = 0):
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
+    :param tqdm_leave: Sets the leave parameter of the tqdm progress bar (usefull 
+        for nested progress bars). Ignored if tqdm progress bar is not used.
+    :param print_results: Whether or not to print the average execution time after 
+        all repetitions (usefull to disable when using nested progress bars).
```
tbennun, line 46:
```suggestion
        all repetitions.
```

dace/frontend/operations.py
```diff
@@ -47,34 +57,45 @@ def __init__(self, repetitions: int = 0, warmup: int = 0) -> None:
...
-                      'dependency tqdm...\n\tTo see a live progress bar please install '
-                      'tqdm (`pip install tqdm`)\n\tTo disable this feature (and '
-                      'this warning) set `profiling_status` to false in the dace '
-                      'config (~/.dace.conf).')
+                warnings.warn(
+                    "Cannot show profiling progress, missing optional dependency "
+                    "tqdm...\n\tTo see a live progress bar please install tqdm "
+                    "(`pip install tqdm`)\n\tTo disable this feature (and this "
+                    "warning) set `profiling_status` to false in the dace config "
+                    "(~/.dace.conf)."
```
tbennun, line 80: why are the quotes double? see contributing guidelines.

dace/frontend/operations.py
```diff
@@ -47,34 +57,45 @@ def __init__(self, repetitions: int = 0, warmup: int = 0) -> None:
...
+        # zeros to overwrite start time, followed by indices for each repetition
+        iterator = chain(repeat(0, self.warmup), range(1, self.repetitions + 1))
 
-        times = [start] * (self.repetitions + 1)
-        ret = None
-        print('\nProfiling...')
-
-        iterator = range(self.warmup + self.repetitions)
-        if Config.get_bool('profiling_status'):
+        if Config.get_bool("profiling_status"):
```
tbennun, line 63: double quotes, see contributing guidelines and apply everywhere

dace/frontend/operations.py
```diff
@@ -47,34 +57,45 @@ def __init__(self, repetitions: int = 0, warmup: int = 0) -> None:
...
             try:
                 from tqdm import tqdm
-                iterator = tqdm(iterator, desc="Profiling", file=sys.stdout)
+
+                iterator = tqdm(
+                    iterator,
+                    desc="Profiling",
```
tbennun, line 69: double quotes

dace/frontend/operations.py
```diff
@@ -105,7 +127,7 @@ def __call__(self, compiled_sdfg: 'CompiledSDFG', args: Tuple[Any, ...]):
         # Restore state after skipping contents
         compiled_sdfg.do_not_execute = old_dne
 
-        return ret
+        return None
```
tbennun, line 130: Why this change?

tbennun (changes requested):

dace/frontend/operations.py
phschaad: I don't think we have it in our contributing guidelines, do we? We just mention that we follow Google's style, and that just states to be consistent; which [we're not](https://github.com/spcl/dace/blob/master/dace/sdfg/sdfg.py#L183) 🙃

phschaad: Unless I'm missing it, we should probably add it

tbennun: oh dear, we should. thx!

dace/frontend/operations.py
JanKleine: > - Cleanup: remove unused variable for return value (didn't seem to be used)

The variable didn't seem serve any purpose. So basically: "while I'm already here, might as well..."

dace/builtin_hooks.py
```diff
@@ -40,10 +40,10 @@ def profile(
 
     :param repetitions: The number of times to run each DaCe program.
     :param warmup: Number of additional repetitions to run the program without measuring time.
-    :param tqdm_leave: Sets the leave parameter of the tqdm progress bar (usefull 
+    :param tqdm_leave: Sets the ``leave`` parameter of the ``tqdm`` progress bar (usefull
```
tbennun, line 43:
```suggestion
    :param tqdm_leave: Sets the ``leave`` parameter of the ``tqdm`` progress bar (useful
```

tbennun (changes requested):

tbennun (approved):

PR #1553, opened 2024-03-20
Fix bug in map_fusion transformation

edopao:
Four-lines bugfix and associated test case for map_fusion transformation.

Without this change, the test would fail in SDFG validation with error:
`dace.sdfg.validation.InvalidSDFGEdgeError: Memlet data does not match source or destination data nodes) (at state state, edge __s0_n1None_n3IN_T[0] (V:None -> numeric:_inp))`


edopao:
@alexnick83 I can give you some context here. The problem addressed in this PR is different from the problem addressed in #1535. The problem addressed here blocks a delivery on gt4py main branch (for reference https://github.com/GridTools/gt4py/pull/1502), which simplifies the icon4py SDFGs.

dace/transformation/dataflow/map_fusion.py
```diff
@@ -480,6 +480,10 @@ def fuse_nodes(self, sdfg, graph, edge, new_dst, new_dst_conn, other_edges=None)
             else:
                 local_node = edge.src
                 src_connector = edge.src_conn
+                if isinstance(edge.src, nodes.AccessNode):
```
alexnick83, line 483: From the PR description, I understand that this fixes a case where the memlet is left carrying an "obsolete" transient scalar from a (previous?) fusion. The `__s0_...` name is characteristic of that. This is indeed the correct way to fix the issue. However, I would place a few more safeguards (or, if they are already satisfied, put a note to explain how and why they are satisfied for future reference). Namely:
- Is this change really needed, i.e., does the memlet indeed match neither the source nor the destination target?
- Is the priority order programmed (src -> dst) always correct if a change is needed? Does the (other) subset need to be checked? What about `Memlet._is_data_src`? The change above is definitely correct if both the src and dst are scalars. However, they could be arrays (for example, think of the "strange" subgraphs created by SubgraphFusion; you may have already had issues with them before).

If this is blocking and urgent, my suggestion would be to narrow the application of the above fix as follows:
```python
test_data = []
test_data.append(edge.src.data if isinstance(edge.src, nodes.AccessNode) else None)
test_data.append(edge.dst.data if isinstance(edge.dst, nodes.AccessNode) else None)
current_data = edge.data.data
if current_data is not None and current_data not in test_data and list(edge.data.subset) == [0]:
    for new_data in test_data:
        if new_data is not None and isinstance(sdfg.arrays[new_data], dace.data.Scalar):
            edge.data.data = new_data
````
The above code is indicative and not tested (you don't need the None values in test data unless you want to know if src or dst was selected), but I think it resolves your issue while avoiding triggering in cases that have not been studied yet.

**Review by alexnick83 — CHANGES_REQUESTED** [review]
The fix looks correct for your use case, but I am unsure if it will always lead to correct results due to how it is applied. I explain in more detail in my comment. The fact that this bug exists in the first place shows that we just don't have enough tests for a lot of different cases regarding MapFusion. Therefore, I suggest that you limit the applicability of your fix until we can study related issues in more detail. Please find an example in my comment and let me know if it is something you can work with. If it is urgent, I can check again on short notice.

File: dace/transformation/dataflow/map_fusion.py (hunk 1 above)
Comment (edopao) [reply dace/transformation/dataflow/map_fusion.py:483]: Thank you for a quick and informative reply. I will look in detail and try your suggestion. I just wanted to quickly answer to the question mark about the origin of the transient scalar:
> from a (previous?) fusion

No, the transient scalar originates from the same fusion step, just a few lines above, at line 463:
```
            edge.data.data = local_name
            edge.data.subset = "0"
```

edopao:
Thank you @alexnick83 for reviewing this PR. I have limited the application of this fix to the known valid case, as you suggested.
However, I skipped this line:
`if current_data is not None and current_data not in test_data and list(edge.data.subset) == [0]:`
because it is always true: the `current_data` is set at line 463-464.

alexnick83 (approved):

Issue #1568, opened 2024-05-06
[Python frontend]: Python 3.11.7 + enumerate + subscript breakage


FlorianDeconinck:
**Describe the bug**
The interaction of python `3.11.7` and `enumerate` + `subscript` is breaking the parsing. The following code is a reduction of a code in NASA/NOAA dynamical core. This works with python `3.8.X`

**To Reproduce**

**With python-3.11.7:**

```python

import numpy as np
import dace

tracer_variables = ["vapor", "rain", "nope"]


@dace.program
def enumerate_parsing(
    A,
    tracers: dace.compiletime,  # Dict[str, np.float64]
):
    for i, q in enumerate(tracer_variables[0:2]):
        tracers[q][:] = A


sz = (3, 3)
A_field = np.zeros(sz)
all_qs = {
    "ice": np.ones(sz),
    "rain": np.ones(sz),
    "vapor": np.ones(sz),
}

dace.config.Config.set(
    "frontend",
    "unroll_threshold",
    value=False,
)

enumerate_parsing(A_field, all_qs)
```

Traceback:

```log
...
  File "/home/fgdeconi/work/git/ndsl/external/dace/dace/frontend/python/preprocessing.py", line 764, in visit_Subscript
    return self._visit_potential_constant(node.value.elts[gslice], True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fgdeconi/work/git/ndsl/external/dace/dace/frontend/python/preprocessing.py", line 735, in _visit_potential_constant
    return self.generic_visit(node)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fgdeconi/work/git/ndsl/external/dace/dace/frontend/python/preprocessing.py", line 486, in generic_visit
    return super().generic_visit(node)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fgdeconi/work/git/ndsl/external/dace/dace/frontend/python/astutils.py", line 453, in generic_visit
    for field, old_value in ast.iter_fields(node):
  File "/home/fgdeconi/.pyenv/versions/3.11.7/lib/python3.11/ast.py", line 260, in iter_fields
    for field in node._fields:
                 ^^^^^^^^^^^^
AttributeError: 'list' object has no attribute '_fields'
```

alexnick83:
First evaluation:

[...]

PR #1570, opened 2024-05-07
Fix Subscript literal evaluation for List

FlorianDeconinck:
Looking at: https://github.com/spcl/dace/issues/1568

The code was blindly calling down to a `_visit_potential_constant` which is written for single element rather collection of them. Unroll the list, like the `dict` is done in the `if` above.

### DIFF
--- a/dace/frontend/python/preprocessing.py
+++ b/dace/frontend/python/preprocessing.py
@@ -761,7 +761,7 @@ class GlobalResolver(astutils.ExtNodeTransformer, astutils.ASTHelperMixin):
                     if gkey == gslice:
                         return self._visit_potential_constant(v, True)
             elif isinstance(node.value, ast.List):  # List
-                if isinstance(node.value.elts[gslice], ast.List):
+                if isinstance(node.value.elts[gslice], List):
                     visited_list = astutils.copy_tree(node.value)
                     visited_list.elts.clear()
                     for v in node.value.elts[gslice]:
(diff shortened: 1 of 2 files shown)


dace/frontend/python/preprocessing.py
```diff
@@ -760,8 +760,19 @@ def visit_Subscript(self, node: ast.Subscript) -> Any:
                         continue
                     if gkey == gslice:
                         return self._visit_potential_constant(v, True)
-            else:  # List or Tuple
-                return self._visit_potential_constant(node.value.elts[gslice], True)
+            elif isinstance(node.value, ast.List):  # List
+                if isinstance(node.value.elts[gslice], List):
```
alexnick83, line 764: `isinstance(..., (ast.List, ast.Tuple))`

tests/python_frontend/unroll_test.py
```diff
@@ -169,6 +169,29 @@ def tounroll(A: dace.float64[3]):
     assert np.allclose(a, np.array([1, 2, 3]))
 
 
+def test_list_global_enumerate():
```
alexnick83, line 172: Add a second test:
```python
def test_tuple_global_enumerate():
    tracer_variables = ("vapor", "rain", "nope")
...
```

alexnick83 (changes requested):
Thank you! The fix is very good. Could you also add a test for tuples (see comments on code)? Currently, `ast.List` and `ast.Tuple` are very similar (both have `elts`), therefore, it makes sense to treat them both similarly. Furthermore, a test will be very useful if this changes in a future Python version.

alexnick83 (approved):
LGTM

PR #1582, opened 2024-06-02
Fixed an error regarding the `pow()` function of the DaCe runtime.

philip-paul-mueller:
Currently there was an inplementation for `T pow(T, T)`, where `T` is a template and two for the case where `T` is a signed or unsigned integer. This is a problem if we consider the call `pow(double, long long)` because the call is ambigious and results in a compiler error. As a solution I added the function `T pow(T, E)`, where both `T` and `E` are templates.

### DIFF
--- a/dace/runtime/include/dace/math.h
+++ b/dace/runtime/include/dace/math.h
@@ -482,7 +482,7 @@ namespace dace
         template<typename T, typename E>
         DACE_CONSTEXPR DACE_HDFI T pow(const T& a, const E& b)
         {
-            return (T)std::pow(a, (T)b);
+            return (T)std::pow(a, b);
         }
 
 #ifndef DACE_XILINX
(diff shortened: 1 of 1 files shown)


dace/runtime/include/dace/math.h
```diff
@@ -479,30 +479,29 @@ namespace dace
             return (T)std::pow(a, b);
         }
 
+        template<typename T, typename E>
+        DACE_CONSTEXPR DACE_HDFI T pow(const T& a, const E& b)
+        {
+            return (T)std::pow(a, (T)b);
```
BenWeber42, line 485: Is the cast really necessary? Since C++11, there is an overload for when the two arguments have different types (overload A): https://en.cppreference.com/w/cpp/numeric/math/pow

Shouldn't we just forward directly to that overload and let the standard library pick the implementation?

philip-paul-mueller: Good idea.

BenWeber42:
I'm closing this one, since it's superseded by https://github.com/spcl/dace/pull/1583

PR #1583, opened 2024-06-05
Made `dace::math::pow` forward to `std::pow` more generic

BenWeber42:
This overload solution will forward the argument types separately and use the same return type.


BenWeber42:
My attempt at combining https://github.com/spcl/dace/pull/1582 and https://github.com/spcl/dace/pull/1521.

Berke-Ates:
Thanks for the work! Looks good to me!

BenWeber42:
I want to separate the Xilinx Vitis HLS compiler flag change into its own PR: https://github.com/spcl/dace/pull/1585
If we can merge that PR, I will remove the change in this PR. Then it should be ready for review.

philip-paul-mueller:
LGTM

tests/python_frontend/power_operator_test.py
```diff
@@ -35,8 +37,35 @@ def test_pow_neg_exp():
...
+@pytest.mark.parametrize("a_type", in_types)
+@pytest.mark.parametrize("b_type", in_types)
+def test_pow_types(a_type, b_type):
+    @dace.program
+    def pow_types(A: a_type[1], B: b_type[1], R: dace.float64[1]):
+        with dace.tasklet(dace.Language.Python):
+            a << A[0]
+            b << B[0]
+            r >> R[0]
+            """r = a ** b"""
```
tbennun, line 51: why is this a string and not just `r = a ** b`?

tbennun (changes requested):

tests/python_frontend/power_operator_test.py
Berke-Ates: If we don't use the string, the generated CPP uses pointers directly instead of reading out the array. Perhaps an optimization bug that needs to be investigated. Alternatively we could add wrappers for pointer types.

image

BenWeber42: Oof, good catch. I quickly checked the SDFG (without string/as code) and it contains a tasklet with `r = (None ** None)`. I guess then somehow `__g_a` and `__g_b` get introduced.

I also tried `R[0] = A[0] ** B[0]` (without the explicit tasklet), but this leads to a tasklet with code `__out = (dace.float64(__tmp0) ** dace.float64(__tmp1))`. The casts completely defeat the purpose of this PR.

It seems like using the string notation is currently the only reliable way to properly test the C++ overloads.

My suggestion would be to separate these issues from this PR. I can open GH issues with tests for those. So that we can merge this PR, since it's unrelated to these issues.

BenWeber42: First issue: https://github.com/spcl/dace/issues/1588
I looked into the other one as well, but I don't really understand the issue yet. I saw some interesting behavior with different types...

BenWeber42: I changed it based on the feedback from https://github.com/spcl/dace/issues/1588

tbennun (approved):

Issue #1560, opened 2024-04-26
state.topological_sort is incorrect, while sdfg.utils.dfs_topological_sort is correct


ThrudPrimrose:
**Describe the bug**
When I call:
`dace.sdfg.utils.dfs_topological_sort(state)`
I have a correct topological sorting of nodes within a state but, calling:
`state.topological_sort()`
Results with an ordering that is not correct.

**To Reproduce**
I have a minimal program to reproduce the issue. In the example I reproduce, an access node, which has an edge to an map node, is listed after the map when sorted with `state.topological_sort`

**Expected behavior**
When I print the nodes and edges in the order received from the nodes and their outgoing edges are printed as follows:

```
...
dataA  |  [dataA:None  -(gpu_dataA[0:N] <- [0:N])->  gpu_dataA:None]
kernelAddBToAOnGPUNoWCR_42[i=0:N]  |  [kernelAddBToAOnGPUNoWCR_42[i=0:N]:OUT_dataA  -(gpu_dataA[0:N])->  gpu_dataA:None]
gpu_dataA  |  [gpu_dataA:None  -(gpu_dataA[0:N])->  kernelAddBToAOnGPUNoWCR_42[i=0:N]:IN_dataA]
gpu_dataA  |  [gpu_dataA:None  -(dataA[0:N] <- [0:N])->  dataA:None]
dataA  |  []
```

Expected would be: (the access node gpu_dataA that as outgoing edge to the kernel, needs to be before the kernel)
```
gpu_dataA  |  [gpu_dataA:None  -(gpu_dataA[0:N])->  kernelAddBToAOnGPUNoWCR_42[i=0:N]:IN_dataA]
gpu_dataB  |  [gpu_dataB:None  -(gpu_dataB[0:N])->  kernelAddBToAOnGPUNoWCR_42[i=0:N]:IN_dataB]
kernelAddBToAOnGPUNoWCR_42[i=0:N]  |  [kernelAddBToAOnGPUNoWCR_42[i=0:N]:OUT_dataB  -(gpu_dataB[i])->  _Mult_:__in2, kernelAddBToAOnGPUNoWCR_42[i=0:N]:OUT_dataA  -(gpu_dataA[i])->  _Add_:__in2]
_Mult_  |  [_Mult_:__out  -(__tmp10[0])->  __tmp10:None]
__tmp10  |  [__tmp10:None  -(__tmp10[0])->  _Add_:__in1]
[...]

Issue #1538, opened 2024-02-28
Error in Simplify pass (ConstantPropagation)


edopao:
The Simplify pass introduces an error in the attached SDFG: [sdfg1_before_simplify.json](https://github.com/spcl/dace/files/14434365/sdfg1_before_simplify.json)

The error is introduces by ConstantPropagagation subpass. See the SDFG before ([sdfg2_before_constant_propagation.json](https://github.com/spcl/dace/files/14434364/sdfg2_before_constant_propagation.json)) and after ([sdfg3_after_constant_propagation.json](https://github.com/spcl/dace/files/14434363/sdfg3_after_constant_propagation.json)) applying ConstantPropagagation.

edopao:
I should probably add that the attached SDFG is quite simple, don't be worry of opening it. The error is introduced in the tasklet that writes to the output array. What happens is that ConstantPropagation propagates the symbol assignment on the false-branch edge to the output write. The problem is that it ignores the symbol assignment on the true-branch.

PR #1589, opened 2024-06-12
Fix constant propagation failing due to invalid topological sort

phschaad:
Constant propagation fails for certain graph structures due to an issue with `dace.sdfg.graph.Graph.topological_sort`. This is related to #1560.


edopao:
Interesting, this also an old issue of mine #1538

tbennun (approved):

PR #1590, opened 2024-06-12
Rename misleading topological_sort to bfs_nodes

BenWeber42:
This is currently work in progress to see how we can best fix this misleading naming.

Fixes https://github.com/spcl/dace/issues/1560

Since https://github.com/spcl/dace/issues/1560 is still in flux, we have to make sure the PR stays in sync with what we are discussing in https://github.com/spcl/dace/issues/1560.

Additionally, at the call-sites of the _previous topological_sort_ (before renaming), there are various comments to use a topoligical sort. After the renaming, they become misleading, so we should probably fix/improve those comments.

### DIFF
--- a/dace/codegen/targets/cuda.py
+++ b/dace/codegen/targets/cuda.py
@@ -2096,7 +2096,7 @@ gpuError_t __err = {backend}LaunchKernel((void*){kname}, dim3({gdims}), dim3({bd
 
         # Get all non-sequential scopes from the same level
         all_scopes = [
-            node for node in parent_scope.topological_sort(scope_entry)
+            node for node in parent_scope.bfs_nodes(scope_entry)
             if isinstance(node, nodes.EntryNode) and node.map.schedule != dtypes.ScheduleType.Sequential
         ]
 
--- a/dace/codegen/targets/fpga.py
+++ b/dace/codegen/targets/fpga.py
@@ -1848,7 +1848,7 @@ std::cout << "FPGA program \\"{state.label}\\" executed in " << elapsed << " sec
         parent_scope = dfg.scope_subgraph(parent_scope_entry)
 
         # Get all scopes from the same level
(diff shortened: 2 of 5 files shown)


tbennun (approved):
Thank you

dace/codegen/targets/fpga.py
```diff
@@ -1848,7 +1848,7 @@ def get_next_scope_entries(self, sdfg, dfg, scope_entry):
         parent_scope = dfg.scope_subgraph(parent_scope_entry)
 
         # Get all scopes from the same level
-        all_scopes = [node for node in parent_scope.topological_sort() if isinstance(node, dace.sdfg.nodes.EntryNode)]
+        all_scopes = [node for node in parent_scope.bfs_nodes() if isinstance(node, dace.sdfg.nodes.EntryNode)]
```
tbennun, line 1851: why is this not `.nodes()`?

dace/codegen/targets/cuda.py
```diff
@@ -2096,7 +2096,7 @@ def get_next_scope_entries(self, dfg, scope_entry):
 
         # Get all non-sequential scopes from the same level
         all_scopes = [
```
tbennun, line 2098: I am not exactly sure why this does not use `scope_subgraph` instead.

dace/codegen/targets/fpga.py
BenWeber42: hmm, a bit of a guess, but maybe because of the next line:

```Python
return all_scopes[all_scopes.index(scope_entry) + 1:]
```

```.nodes()``` might return a different order.

dace/codegen/targets/cuda.py
BenWeber42: hmm, I'll try to look into this later. But for now I'd like to wrap up this PR. If we should change this, I think it should be in a separate PR.

But thanks for pointing it out. Looks like there's a potential to improve the code there.

PR #1591, opened 2024-06-12
Fix DeadDataflowElimination pass for read-write map arrays

affifboudaoud:


phschaad:
Is this still active? If so, is something blocking this from being reviewed / merged?

affifboudaoud:
This was better addressed by a different PR. I will run the test case again and close this PR.

PR #1598, opened 2024-06-17
Use file object directly in `temporary_config()`

philip-paul-mueller:
The context manager uses `NamedTemporaryFile` to store the current configuration, to later restore them. Instead of passing the file object directly to the save function, it just passes the file name, i.e. the save (and the load function) will open the file again, which is in itself not a problem. However, on the Github Windows image this leads to a permission error (using the created file object is fine).

This commit solves this by adding the `file` argument to `Config.save()` that allows to pass a file object directly to the function. The same change is applied to the load function of the config object.

### DIFF
--- a/dace/config.py
+++ b/dace/config.py
@@ -3,6 +3,7 @@ import contextlib
 import os
 import platform
 import tempfile
+import io
 from typing import Any, Dict
 import yaml
 import warnings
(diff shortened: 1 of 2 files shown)


dace/config.py
```diff
@@ -39,10 +40,11 @@ def temporary_config():
             Config.set("optimizer", "autooptimize", value=True)
             foo()
     """
-    with tempfile.NamedTemporaryFile() as fp:
-        Config.save(fp.name)
+    with tempfile.NamedTemporaryFile(mode='w+t') as fp:
```
tbennun, line 43: what does 't' mean in `mode`?

philip-paul-mueller: It means text mode, on UNIX `\n` is the newline character, but on Windows it is `\r\n` (https://en.wikipedia.org/wiki/Newline#Representation). In that mode `\r\n` is transformed into `\n` it is essentially a compatibility layer, see `help(open)`.
I used it because it preserves old behaviour, but it is probably not needed.

tbennun (approved):

Issue #1595, opened 2024-06-13
Memlet Error in RedundantArray (simplify)


philip-paul-mueller:
I found a bug in simplify, as far as I can tell it is located it is inside `RedundantArray`
The error is about a Memlet, that performs some reshaping.
A further issue that makes the whole thing complicated is, that the error depends on the processing order, the bug only appears in some cases, which makes it hard to debug.
I was able to find a minimal example that triggers the bug, but not always, so you have to run it multiple times.

```python
import dace

"""
Minimal example for bug in `RedundantArray`

Essentially the SDFG perfroms the following computation.
def foo(A: dace.float64[6, 6, 6]) -> dace.float64[36, 1, 6]:
    return A.reshape((36, 1, 6))

I located the error in the `RedundantArray` transformation, that screws up with the
Memlets. Furthermore, the input array is in FORTRAN order, but it should also work
with C order since it happens before code generation.

I M P O R T A N T
=================
The bug is not deterministic, it depends on the processing order!
It happens if node `a` is removed instead of Node `b`, never mind that that node could
also be removed. If you run the script and it does not fail, try again until it happens.
"""




sdfg = dace.SDFG("invalid_sdfg")

_, input_desc = sdfg.add_array(
        "input",
        shape=(6, 6, 6),
        transient=False,
        strides=(1, 6, 36),
        dtype=dace.float64,
)
_, a_desc = sdfg.add_array(
        "a",
        shape=(6, 6, 6),
        transient=True,
        strides=(36, 6, 1),
        dtype=dace.float64,
)
_, b_desc = sdfg.add_array(
        "b",
        shape=(36, 1, 6),
        transient=True,
        strides=(6, 6, 1),
        dtype=dace.float64,
)
_, output_desc = sdfg.add_array(
        "output",
        shape=(36, 1, 6),
        transient=False,
        strides=(6, 6, 1 ),
        dtype=dace.float64,
)

state = sdfg.add_state("state", is_start_block=True)
input_an = state.add_access("input")
a_an = state.add_access("a")
b_an = state.add_access("b")
output_an = state.add_access("output")

state.add_edge(
        input_an,
        None,
        a_an,
        None,
        dace.Memlet.from_array("input", input_desc),
)

state.add_edge(
        a_an,
        None,
        b_an,
        None,
        dace.Memlet.simple(
            "a",
            subset_str="0:6, 0:6, 0:6",
            other_subset_str="0:36, 0, 0:6",
        )
)

state.add_edge(
[...]

PR #1603, opened 2024-06-18
Fixed `RedundantArray`'s handling of "reshaping" Memlets

philip-paul-mueller:
This PR fixes an issue that was reported as [issue 1595](https://github.com/spcl/dace/issues/1595), that was traced back to `RedundantArray`.

The commit adds a deterministic test, unlike the one in the issue, that fails without the fix.
The underlying problem is, that the transformation does not correctly handle a Memlet that performs a reshaping.
This commit does not really solves the issue, instead it adds a special case for this particular case and then handles it correctly as I was unable to modify the code to handle it correctly.
It is not a nice solution, but it works.


alexnick83 (changes requested):
What happens if the first/input array has multiple incoming edges? It looks like the transformation will apply as before your fix and produce a similar error. You don't need to make the transformation work on such corner cases, but maybe it should not match those cases in the first place. How about refactoring the new login into a separate helper method and calling it also in `can_be_applied`?

philip-paul-mueller:
I like that idea.

alexnick83 (approved):

PR #1604, opened 2024-06-19
Renamed `graph.bfs_edges` to `edge_bfs`

BenWeber42:
`networkx` has similar methods `bfs_edges` and `edge_bfs` for their graph classes. There is a slight and documented difference between the two: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.traversal.edgebfs.edge_bfs.html#networkx.algorithms.traversal.edgebfs.edge_bfs

Unfortunately, our `graph.bfs_edges` behaves differently to `networkx`'s `bfs_edges`. In fact, it behaves exactly like `networkx`'s `edge_bfs`. To avoid future confusion (just happened to me), I propose renaming our `bfs_edges` to `edge_bfs` to be consistent with `networkx`.


tbennun (approved):
Sounds good

PR #1613, opened 2024-07-04
Adds support for NumPy version 2

alexnick83:


tests/numpy/unop_test.py
```diff
@@ -29,7 +29,10 @@ def test_not():
     B = np.zeros((5, 5), dtype=np.int64).astype(np.bool_)
     regression = np.logical_not(A)
     nottest(A, B)
-    assert np.alltrue(B == regression)
+    if np.lib.NumpyVersion(np.__version__) >= '2.0.0b1':
```
tbennun, line 32: should we change this to 2.0.0?

tbennun (approved):
Looks great!! Very clean and captures the exact numpy 2 differences I was thinking of. I have one minor comment on a beta version in a test if you want to correct it.

tests/numpy/unop_test.py
alexnick83: I don't think it makes a practical difference, but we should change it for consistency.

PR #1620, opened 2024-07-11
SymPy 1.13 fixes

BenWeber42:


dace/symbolic.py
```diff
@@ -5,12 +5,14 @@
 import pickle
 import re
 from typing import Any, Callable, Dict, Iterable, Optional, Set, Tuple, Union
-import warnings
 import numpy
 
 import sympy.abc
 import sympy.printing.str
 
+# FIXME(ben): do we need to add `packaging` as a dependency to DaCe?
```
alexnick83, line 13: The alternative, `pkg_resources`, is deprecated. I do not believe that the standalone `packaging` module is included by default in a barebones Python installation, although other modules may have it in their dependencies (it is in my conda environment for some reason). I would add it to the dependencies to finish this PR.

alexnick83: Pytest requires it, but I would still add it to the base DaCe dependencies:
```
Collecting packaging (from pytest>=7.0.0->pytest-xdist)
  Downloading packaging-24.1-py3-none-any.whl.metadata (3.2 kB)
  ```

alexnick83 (approved):
Please add the `packaging` dependency and merge.

BenWeber42:
I think it's best to merge this set of minimal fixes to avoid blocking other PRs. We can take care of the more thorough fixes in a separate PR.

dace/symbolic.py
BenWeber42: Yes, I was thinking that most likely one of our dependencies needs it anyway, but definitely better to explicitly add it. It's a rather small dependency.

PR #1621, opened 2024-07-24
Improving the Usage of `#pragma unroll`

philip-paul-mueller:
Before it was only possible to specify if a Map should be unrolled or not, i.e. if `#pragma unroll` should be added or not.
However. the pragma accepts an argument, the most interesting use case for this is to _prevent_ unrolling, i.e. adding `#pragma unroll 1`.

I solved this by adding a new property `unroll_factor` to the map that takes an integer, i.e. the argument.
I am open for other suggestions to solve this problem.
As a wrote the most interesting use case (and in fact our motivation for this fix) is the ability to prevent unrolling explicitly.


tbennun (approved):
Looks great, thank you!

PR #1622, opened 2024-07-26
Made `TransientReuse` Less Verbose

philip-paul-mueller:
Added a flag to the `TransientReuse` transformation that prevents the transformation from unconditional printing of memory savings.
This commit changes the default behaviour, because now the report is not printed, but must be explicitly activated.


tbennun (approved):
LGTM if tests pass.

PR #1629, opened 2024-08-22
New and Improved MapFusion

philip-paul-mueller:
This PR introduces a new and improved version of `MapFusion`.
A summary of the changes can also be found [here](https://github.com/user-attachments/files/18516299/new_map_fusion_summary_of_changes.pdf), it compares the resulting SDFGs generated by the old and new transformation of some unit tests.

#### Fixed Bugs and removed Limitations
- The subsets (not the `.subset` member of the Memlet; I mean the concept) of the new intermediate data descriptor were not computed correctly in some cases, especially in presence of offsets. See the `test_offset_correction_range_read()`, `test_offset_correction_scalar_read()` and the `test_offset_correction_empty()` tests.
- Upon the propagation of the subsets, due to the changed intermediate, was not handled properly. Essentially, the transformation only updated `.subset` and ignored `.other_subset`. Which is correct in most cases but not always. See the `test_fusion_intrinsic_memlet_direction()` for more.
- During the check if two maps could be fused the `.dynamic` property of the Memelts were fully ignored leading to wrong code.
- The read-write conflict checks were refined, before all arrays needed to be accessed the wrong way, i.e. before a fusion was rejected when one map accessed `A[i, j]` and the other map was accessing `B[i + 1, j]`. Now this is possible as long as every access is point wise. See the `test_fusion_different_global_accesses()` test for an example.
- The shape of the reduced intermediate is cleaned, i.e. unnecessary dimensions of size 1, are removed, except they were present in the original shape. To make an example, the intermediate array, `T`, had shape `(10, 1, 20)` and inside the map was accessed `T[__i, 0, __j]`, then the old transformation would have created an reduced intermediate of shape `(1, 1, 1)`, new its shape is `(1)`. Note if the intermediate has shape `(10, 20)` instead and would be accessed as `T[__i, __j]` then a `Scalar` would have been created. See also the `struct_dataflow` flag below.

#### New Flags
- `only_toplevel_maps`: If `True` the transformation will only fuse maps that are located at the top level, i.e. maps inside maps will not be merged.
- `only_inner_maps`: If `True` then the transformation will only fuse maps that are inside other maps.
- assume_always_shared`: If `True` then the transformation will assume that every intermediate is shared, i.e. the referenced data is used somewhere else in the SDFG and has to become an output of the fused maps. This will create dead data flow, but avoids a scan of the full SDFG.
- `strict_dataflow`: This flag is enabled by default. It has two effects, first it will disable the cleaning of reduced intermediate storage. The second effect is more important as it will preserve a much stricter data flow. Most importantly, if the intermediate array is used downstream (this is not limited to the case that the array is the output of the second map) then the maps will not be fused together. This is mostly to work around some other bugs in DaCe, where other transformations failed to pink up the dependency. Note that the fused map would be correct, the problem are other transformations.

#### `FullMapFusion`
This PR also introduced the `FullMapFusion` pass, which makes use of the `FindSingleUseData` pass that was introduced in [PR#1906](https://github.com/spcl/dace/pull/1906).
The `FullMapFusion` applies MapFusion as long as possible, i.e. fuses all maps that can be fused.
But instead of scanning the SDFG every time an intermediate node has to be classified, i.e. can it be deleted or not, it is done once and then reused which will speed up fusion process as it will remove the need to traverse the full SDFG many times.
This new pass also replaced the direct application of MapFusion in `auto_optimizer`.

#### References
Collection of known issues in other transformation:
- [RedundantArrayRemoval (auto optimize)](https://github.com/spcl/dace/issues/1644)
- [Bug in `RefineNestedAccess` and `SDFGState._read_and_write_sets()](https://github.com/spcl/dace/issues/1643)
- [Error in Composite Fusion (auto optimize)](https://github.com/spcl/dace/issues/1642)


philip-paul-mueller:
Currently the auto opt uses strict dataflow

dace/sdfg/state.py
```diff
@@ -756,37 +762,59 @@ def _read_and_write_sets(self) -> Tuple[Dict[AnyStr, List[Subset]], Dict[AnyStr,
...
+
             # Union all subgraphs, so an array that was excluded from the read
             # set because it was written first is still included if it is read
             # in another subgraph
             for data, accesses in rs.items():
                 read_set[data] += accesses
             for data, accesses in ws.items():
                 write_set[data] += accesses
-        return read_set, write_set
+        return copy.deepcopy((read_set, write_set))
```
acalotoiu, line 817: Why is it necessary to make a copy here?

dace/sdfg/state.py
```diff
@@ -756,37 +762,59 @@ def _read_and_write_sets(self) -> Tuple[Dict[AnyStr, List[Subset]], Dict[AnyStr,
...
-                    for e in in_edges:
-                        # skip empty memlets
-                        if e.data.is_empty():
-                            continue
-                        # Store all subsets that have been written
-                        ws[n.data].append(e.data.subset)
-                    for e in out_edges:
-                        # skip empty memlets
-                        if e.data.is_empty():
+                if not isinstance(n, nd.AccessNode):
```
acalotoiu, line 767: Can you please add more comments to make it easier to follow what is being done?

dace/transformation/dataflow/__init__.py
acalotoiu, line 1: I am not 100% about the naming solution - having 3/5 different kinds of MapFusion seems like a poor solution that will lead to confusion - between Serial,Parallel, OTF and the original MapFusions versions are proliferating. Would it not be preferrable to fix/choose one?

acalotoiu:
I have some trouble understanding the difference between -Serial and -Parallel and why both are necessary?

acalotoiu (changes requested):
I agree that MapFusion has issues - but I am slightly confused by the current approach to provide multiple versions - it seems as though fixes to the original should be preferrable. Why should we maintain many different implementations, and not have one maintained one, and deprecate the rest? @tbennun - what is your view?

dace/sdfg/state.py
philip-paul-mueller: Because, the subsets are still linked to the Memlets, so if you modify them then you change them at the Memlets.
This might be useful but since it is not possible to determine which subset belongs to which memlet, it does not make sense to maintain this link.
You could potentially copy them above, because some of them are constructed on demand anyway, however:
- The on demand construction is the minority case.
- Doing it here allows to do it in one big sweep.

dace/transformation/dataflow/__init__.py
philip-paul-mueller: I agree, but the original map fusion was unable to do parallel fusing, the available solutions are:
1. Combining serial and parallel map fusion into one
2. Doing the above one but making parallel an opt in option.

The best solution is probably 2. because then everything will work as before. But I do not care what do you think @tbennun @acalotoiu

dace/sdfg/state.py
philip-paul-mueller: I added more comments.

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,537 +1,1583 @@
...
-            Other than the removal of the second map entry node (SME), and the first
-            map exit (FME) node, it has the following side effects:
+        map_exit_1: nodes.MapExit = state.exit_node(map_entry_1)
+        map_exit_2: nodes.MapExit = state.exit_node(map_entry_2)
+
+        # Get the read and write sets of the different maps, note that Views
+        #  are not resolved yet.
+        access_sets: List[Dict[str, nodes.AccessNode]] = []
+        for scope_node in [map_entry_1, map_exit_1, map_entry_2, map_exit_2]:
+            access_set: Set[nodes.AccessNode] = self.get_access_set(scope_node, state)
```
phschaad, line 1004: This only seems to collect access nodes which are connected as outputs to the map exits and inputs to the map entries, right? If so, are map-internal writes / reads taken care of?

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,537 +1,1583 @@
...
+            map_entry_2: The entry node of the second map.
+        """
+        # The three outputs set.
+        pure_outputs: Set[graph.MultiConnectorEdge[dace.Memlet]] = set()
+        exclusive_outputs: Set[graph.MultiConnectorEdge[dace.Memlet]] = set()
+        shared_outputs: Set[graph.MultiConnectorEdge[dace.Memlet]] = set()
+
+        # Compute the renaming that for translating the parameter of the _second_
+        #  map to the ones used by the first map.
+        repl_dict: Dict[str, str] = self.find_parameter_remapping(
```
phschaad, line 299: It appears to me that this function is called multiple times in a single call to `can_be_applied`. Is that necessary, and if so, why? Can this not be computed once and reused during the entire `can_be_applied` check?

phschaad (changes requested):
Thank you for the work on this transformation. MapFusion is in need of some improvements. There are a few general concerns I need to have addressed or at least discussed first in addition to the specific review comments though.

## Complete Change
You advertise this in the name and description of the PR (the later of which seems to be outdated, if I am not mistaken?), but this is essentially a completely re-written `MapFusion`. I have nothing against that, but this naturally worries me for a transformation this central, often used, and consequently battle tested.

There are some things, such as checking for no WCR outputs of the first map that serve as inputs to the second map, that I can no longer detect in the new `can_be_applied` method. If these are all correctly handled by the new transformation now, that is great. But I believe this drastic change at least needs a very concise and clear discussion in the PR to explain conceptually exactly what changed that now allows for different situations for the transformation to apply. Currently this is missing to me, and I was unable to piece this together from the doc comments alone.

## Difference to OTF Map Fusion and SubgraphFusion
This new version of the transformation seems to me like it contains components of both OTF map fusion and `SubgraphFusion` (or `CompositeFusion`). I am not categorically against bringing transformations together to reduce the size of the transformation repository if there is one transformation that can take the job of 2. However, currently I don't believe this transformation is meant to replace either of the two. Can you elaborate not just on how this new `MapFusion` is different from the old one (see previous point), but also on how this is different from the other existing transformations in this category (`OTFMapFusion` and `SubgraphFusion`)?

## Complexity
By seemingly being a much more capable fusion transformation, this `can_be_applied` method appears to perform a lot more checks in general, which is great and improves robustness if all previously identified issues are still covered. However, `can_be_applied` is also run _a lot_, particularly on large graphs. Can you elaborate on the performance difference between this and the old `MapFusion` transformation, or is there no measurable difference?

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,537 +1,1583 @@
...
     """
-    first_map_exit = transformation.PatternNode(nodes.ExitNode)
-    array = transformation.PatternNode(nodes.AccessNode)
-    second_map_entry = transformation.PatternNode(nodes.EntryNode)
 
-    @staticmethod
-    def annotates_memlets():
-        return False
+    # Pattern Nodes
+    map_exit_1 = transformation.transformation.PatternNode(nodes.MapExit)
```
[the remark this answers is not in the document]
philip-paul-mueller: I agree the naming is better.
I also agree that a transformation as central as this one should not change its public interface unless it is really needed.

I used search+replace, I think I got all, but if you spot a stray, please indicate it.

tests/transformations/mapfusion_test.py
```diff
@@ -265,7 +428,13 @@ def fusion_with_nested_sdfg_0(A: dace.int32[10], B: dace.int32[10], C: dace.int3
             A[i] = tmp[i] * 2
     
     sdfg = fusion_with_nested_sdfg_0.to_sdfg(simplify=True)
-    sdfg.apply_transformations(MapFusion)
+
+    # Because the transformation refuses to fuse dynamic edges.
+    #  We have to eliminate them.
+    for state in sdfg.states():
+        for edge in state.edges():
+            edge.data.dynamic = False
```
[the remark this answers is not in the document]
philip-paul-mueller: If you look at the SDFG, you see that the data connection is not really dynamic, but there will always be a write.
However, my transformation can not handle dynamic memlets (I forbid them because according to my understanding it does not make sense to include them).
The old transformation just went along with them, but I think this was just because it was not checked and it worked here because the memlet is not really dynamic.

However, as the name of the test is indicating, it is more about the nested SDFG, so I will rewrite the test to make it clear.

TODO(phimuell): Construct the graph in a way that there is no dynamic memlet.

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,537 +1,1583 @@
...
+    def __init__(
+        self,
+        only_inner_maps: Optional[bool] = None,
+        only_toplevel_maps: Optional[bool] = None,
+        strict_dataflow: Optional[bool] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        if only_toplevel_maps is not None:
+            self.only_toplevel_maps = bool(only_toplevel_maps)
```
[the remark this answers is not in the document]
philip-paul-mueller: The reason is, that I do not like how Python implement variables, that if you write `a = b`, you do not copy `b` into `a` but `a` and `b` refers to the same object.
I also _know_ that the booleans `True` and `False` are singeltons, they are not keywords.
I am not fully sure why I do it myself. But I will remove them.

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,537 +1,1583 @@
...
+        if scope[map_entry_1] != scope[map_entry_2]:
+            return False
+        elif self.only_inner_maps:
+            if scope[map_entry_1] is None:
+                return False
+        elif self.only_toplevel_maps:
+            if scope[map_entry_1] is not None:
+                return False
+
+        # We will now check if there exists a remapping that of the map parameter
```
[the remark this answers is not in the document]
philip-paul-mueller: I do not fully understand your question.
Do you refer to the checks that ensures that the maps are in the same scope or that a particular kind of check is missing at all?

dace/transformation/dataflow/map_fusion.py
philip-paul-mueller: At first I had trouble understanding your question, because I thought you refers to the data that is used to exchange data between the maps, i.e. what `self.array` matches (which is not examined in this function, but in the partition function).
However, then I realized that you mean writes that are fully withing the map, i.e. the dataflow does not pass through the map entry/exit.   
Then you are right, this is not checked, because until now I have not realized that this might be an issue, since I assumed that every data has to pass through the maps.
However, now I realize that this is wrong (for example I know that `CompositeFusion` may put intermediates fully inside a map.

To handle this case I have made the following additions:
- Fusion is refused if both map bodies consist of an access node that refers to the same data.
- Fusion is refused if a map body consists of an access node that refers to global data (this restriction could be lifted).

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,537 +1,1583 @@
...
+        self,
+        node: nodes.AccessNode,
+        sdfg: SDFG,
+    ) -> bool:
+        """Tests if `node` points to a view or not."""
+        node_desc: data.Data = node.desc(sdfg)
+        return isinstance(node_desc, data.View)
+
+
+    def track_view(
```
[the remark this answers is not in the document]
philip-paul-mueller: It seems that `track_view()` can be implemented by `get_last_view_node()`.

dace/transformation/dataflow/map_fusion.py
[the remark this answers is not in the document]
phschaad: I meant that the comment seems like something is missing or wrong, or maybe I'm reading it wrong :-)

phschaad (approved):

philip-paul-mueller:
Here is a summary that compares the behaviour of the old `MapFusion` transformation with the new one.
They are based on the unit tests.
[new_map_fusion_summary_of_changes.pdf](https://github.com/user-attachments/files/18516299/new_map_fusion_summary_of_changes.pdf)

acalotoiu (approved):
Thank you! Looks Good to me now.

PR #1631, opened 2024-08-26
Add UUIDs to SDFG elements

phschaad:


tbennun:
The diff looks good, the error is weird (because `super().__init__()` is called). I would refactor the uuid function to a central location so that it is configurable (e.g., for determinism).

phschaad:
I tried debugging the error a bit and I am genuinely super confused what is going on here. It doesn't even matter what I name the property, it really is like it is never defined for some instances of objects, but it is for others - despite `super().__init__()` seemingly being called everywhere, as you state.

phschaad:
It's `__deepcopy__`...

tbennun:
@phschaad so we already have a validation check that tests the uniqueness of `id(obj)` for objects for bad references. I wonder if we should also verify that for `obj.id` (probably secondary)

dace/sdfg/nodes.py
```diff
@@ -574,6 +580,9 @@ def __deepcopy__(self, memo):
         result = cls.__new__(cls)
         memo[id(self)] = result
         for k, v in self.__dict__.items():
+            # Skip GUID.
+            if k in ('guid'):
```
tbennun, line 584: Missing comma (elem,)

tbennun (changes requested):

tbennun (approved):
LGTM, great addition!

PR #1637, opened 2024-09-04
Specified behaviour of `Subset.covers()` for different dimensionality

philip-paul-mueller:
Before all cover functions ignored the case if the subsets have different lengths. The effect was that the subsets `A = Range.from_string("i, k, 0:N, 1")` and `B = Range.from_string("i, k")` would cover each other. This case was considered undefined behaviour.

This commit changes the behaviour of these function such that for covering it is required that the subsets must have the same dimensions.

### DIFF
--- a/dace/subsets.py
+++ b/dace/subsets.py
@@ -28,7 +28,10 @@ def bounding_box_cover_exact(subset_a, subset_b) -> bool:
 
     # Covering only make sense if the two subsets have the same number of dimensions.
     if len(min_elements_a) != len(min_elements_b):
-        return False
+        return ValueError(
+                f"A bounding box of dimensionality {len(min_elements_a)} can not"
+                f" cover a bounding box of dimensionality {len(min_elements_b)}."
+        )
 
     return all([(symbolic.simplify_ext(nng(rb)) <= symbolic.simplify_ext(nng(orb))) == True
                 and (symbolic.simplify_ext(nng(re)) >= symbolic.simplify_ext(nng(ore))) == True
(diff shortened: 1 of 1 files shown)


dace/subsets.py
```diff
@@ -28,7 +28,10 @@ def bounding_box_cover_exact(subset_a, subset_b) -> bool:
 
     # Covering only make sense if the two subsets have the same number of dimensions.
     if len(min_elements_a) != len(min_elements_b):
-        return False
+        return ValueError(
+                f"A bounding box of dimensionality {len(min_elements_a)} can not"
+                f" cover a bounding box of dimensionality {len(min_elements_b)}."
```
tbennun, line 33:
```suggestion
                f"A bounding box of dimensionality {len(min_elements_a)} cannot"
                f" test covering a bounding box of dimensionality {len(min_elements_b)}."
```

dace/subsets.py
```diff
@@ -43,7 +46,10 @@ def bounding_box_symbolic_positive(subset_a, subset_b, approximation = False)->
...
```
tbennun, line 51:
```suggestion
                f"A bounding box of dimensionality {len(min_elements_a)} cannot"
                f" test covering a bounding box of dimensionality {len(min_elements_b)}."
```

dace/subsets.py
```diff
@@ -73,7 +79,9 @@ def covers(self, other):
 
         # Subsets of different dimensionality can never cover each other.
         if self.dims() != other.dims():
-            return False
+            return ValueError(
+                    f"A subset of dimensionality {self.dim()} can not cover a subset of dimensionality {other.dims()}"
```
tbennun, line 83:
```suggestion
                    f"A subset of dimensionality {self.dim()} cannot test covering a subset of dimensionality {other.dims()}"
```

dace/subsets.py
```diff
@@ -97,7 +105,9 @@ def covers_precise(self, other):
...
```
tbennun, line 109:
```suggestion
                    f"A subset of dimensionality {self.dim()} cannot test covering a subset of dimensionality {other.dims()}"
```

tbennun (changes requested):
Please change error message to reflect issue. Otherwise looks good

tbennun (approved):

PR #1639, opened 2024-09-05
Add state replication and if raising transformations

luigifusco:
This PR adds the StateReplication and IfRaising transformations.
StateReplication performs a normalization step. It takes a state with `n` incoming edges and creates an equivalent SDFG with `n` replicas of the original state, each with a single incoming edge.

IfRaising takes an if guard and 'raises' the evaluation of the conditional. The condition needs to be independent of the computations done in the if guard. The if guard is replicated in each branch and the conditional is evaluated earlier in a new empty state.

The combination of these two transformations unlocks extra dataflow in the case of "diamond" patterns that can be found in some weather codes. In particular, it targets the case where some flag changes what computation is performed.

Example before:
image
After:
image
After State, Map, and Tasklet fusion:
image

### DIFF
--- a/dace/transformation/interstate/if_raising.py
+++ b/dace/transformation/interstate/if_raising.py
@@ -1,0 +1,67 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
+""" If raising transformation """
+
+from dace import data as dt, sdfg as sd
+from dace.sdfg import InterstateEdge
+from dace.sdfg import utils as sdutil
+from dace.sdfg.state import SDFGState
+from dace.transformation import transformation
+from dace.properties import make_properties
+
+@make_properties
+class IfRaising(transformation.MultiStateTransformation):
+    """
+    Duplicates an if guard and anticipates the evaluation of the condition
+    """
+
+    if_guard = transformation.PatternNode(sd.SDFGState)
+
+    @staticmethod
+    def annotates_memlets():
+        return False
+
+    @classmethod
+    def expressions(cls):
+        return [sdutil.node_path_graph(cls.if_guard)]
+
+    def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
+        if_guard: SDFGState = self.if_guard
+
+        out_edges = graph.out_edges(if_guard)
+
+        if len(out_edges) != 2:
+            return False
+
+        if if_guard.is_empty():
+            return False
+
+        # check that condition does not depend on computations in the state
+        condition_symbols = out_edges[0].data.condition.get_free_symbols()
+        _, wset = if_guard.read_and_write_sets()
+        if len(condition_symbols.intersection(wset)) != 0:
+            return False
+
+        return True
+
+
+    def apply(self, _, sdfg: sd.SDFG):
+        if_guard: SDFGState = self.if_guard
+
+        raised_if_guard = sdfg.add_state('raised_if_guard')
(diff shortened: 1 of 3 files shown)


dace/transformation/interstate/state_replication.py
```diff
@@ -0,0 +1,78 @@
...
+
+        if len(in_edges) < 2:
+            return False
+
+        # avoid useless replications
+        if target_state.is_empty() and len(out_edges) < 2:
+            return False
+
+        # make sure this is not a loop guard
+        if len(out_edges) == 2:
```
luigifusco, line 44: Application on loops results in the addition of useless states (but the SDFG is still correct). This will not get rid of the loop, meaning that `apply_transformations_repeated` will never halt.

dace/transformation/interstate/if_raising.py
```diff
@@ -0,0 +1,67 @@
...
+
+        out_edges = graph.out_edges(if_guard)
+
+        if len(out_edges) != 2:
+            return False
+
+        if if_guard.is_empty():
+            return False
+
+        # check that condition does not depend on computations in the state
```
luigifusco, line 38: might be missing something I do not know about here. Also I don't know if this can result in spurious failures

acalotoiu (changes requested):
LGTM, however, this is missing a couple of "negative" tests. Please add tests where the expected behavior is that the transformation is not applied.

phschaad:
Apart from @acalotoiu 's comment, this looks good to me. However, I want to add: I believe that the presence of more and more niche transformations like this should make us prioritize the idea of a 'community transformation repository' a bit more.

luigifusco:
> Apart from @acalotoiu 's comment, this looks good to me. However, I want to add: I believe that the presence of more and more niche transformations like this should make us prioritize the idea of a 'community transformation repository' a bit more.

I really see this as a member of a potential set of "transformation building blocks", towards having composable transformations. This also lies in between a utility method (e.g. in the SDFG class) and an actual transformation

tests/transformations/raise_and_duplicate_test.py
```diff
@@ -0,0 +1,62 @@
...
+        for i in dace.map[0:N]:
+                arr[i] = tmp2[i]
+
+    sdfg = program.to_sdfg()
+    sdfg.apply_transformations([IfRaising, StateReplication])
+    sdfg.simplify()
+    sdfg.apply_transformations_repeated([OTFMapFusion])
+
+    assert len(sdfg.nodes()) == 4
+    assert sdfg.start_state.is_empty()
```
luigifusco, line 31: the tests fails here but it works on my machine. Is there something specific to the setup in the pipeline?

pratyai:
Added one "negative test" (besides the existing one) where the expected behaviour is that the transformation is not applied. Gave each of the two transformations its own test file. PTAL.

phschaad:
hank you for addressing the comments @pratyai! In general, this looks good to me. However, I would prefer if this PR could be part of a brief discussion in the DaCe weekly this week before it is merged. For that reason I am not pressing 'accept' yet. I would prefer to discuss it due to two key points:
1. Preparation for release 1.0 which is currently underway for the next few days
2. Potential duplication of work for when the https://github.com/spcl/dace/pull/1676 is merged, as that should change the pattern the transformation matches from an if-guard to a conditional block and consequently slightly changes the raising being performed.

phschaad:
@acalotoiu  ping for re-review.

PR #1640, opened 2024-09-05
Enable user-defined map parameter types

ThrudPrimrose:
Adding `param_types` variable I can set the loop params to what I want.

I also remove the "auto" type to prefer the inferred type.

Using the inferred type instead of auto fixes the discrepancy of C++/CUDA inferred auto type and the inferred type between DaCe, where the discrepancy causes unnecessary implicit type conversion, resulting with a loss of performance.


dace/codegen/targets/cpu.py
```diff
@@ -1925,6 +1927,11 @@ def _generate_MapEntry(
             # Emit nested loops
             for i, r in enumerate(node.map.range):
                 var = map_params[i]
+                if var in map_param_types.keys():
+                    var_type = map_param_types[var]
+                else:
+                    var_type = dtypes.result_type_of(infer_expr_type(r[0], sdfg.symbols), infer_expr_type(r[1], sdfg.symbols))
```
tbennun, line 1933: I think it's better to use symbols_defined_at

dace/codegen/targets/cpu.py
```diff
@@ -1933,9 +1940,10 @@ def _generate_MapEntry(
                         unroll_pragma += f" {node.map.unroll_factor}"
                     result.write(unroll_pragma, cfg, state_id, node)
 
+
                 result.write(
-                    "for (auto %s = %s; %s < %s; %s += %s) {\n" %
-                    (var, cpp.sym2cpp(begin), var, cpp.sym2cpp(end + 1), var, cpp.sym2cpp(skip)),
+                    "for (%s %s = %s; %s < %s; %s += %s) {\n" % # it is like this before
```
tbennun, line 1945:
```suggestion
                    "for (%s %s = %s; %s < %s; %s += %s) {\n" %
```

dace/sdfg/nodes.py
```diff
@@ -793,7 +793,10 @@ def new_symbols(self, sdfg, state, symbols) -> Dict[str, dtypes.typeclass]:
         result = {}
         # Add map params
         for p, rng in zip(self._map.params, self._map.range):
-            result[p] = dtypes.result_type_of(infer_expr_type(rng[0], symbols), infer_expr_type(rng[1], symbols))
+            if p in self.map.param_types.keys():
```
tbennun, line 796:
```suggestion
            if p in self.map.param_types:
```

dace/sdfg/nodes.py
```diff
@@ -793,7 +793,10 @@ def new_symbols(self, sdfg, state, symbols) -> Dict[str, dtypes.typeclass]:
...
+                result[p] = dtypes.typeclass(self.map.param_types[p])
```
tbennun, line 797: why isn't the value type typeclass?

dace/sdfg/nodes.py
```diff
@@ -877,6 +880,7 @@ class Map(object):
     # List of (editable) properties
     label = Property(dtype=str, desc="Label of the map")
     params = ListProperty(element_type=str, desc="Mapped parameters")
+    param_types = DictProperty(key_type=str, value_type=str, desc="Types of mapped parameters")
```
tbennun, line 883: shouldn't param types have a typeclass value?

dace/sdfg/nodes.py
```diff
@@ -925,6 +929,8 @@ class Map(object):
                                  optional=True,
                                  optional_condition=lambda m: m.schedule in dtypes.GPU_SCHEDULES)
 
+    gpu_force_syncthreads = Property(dtype=bool, desc="Force a call to the __syncthreads for the map", default=False)
+
```
tbennun, line 933: Unrelated

ThrudPrimrose: Must have sneaked in from the other PR, will delete it

dace/sdfg/nodes.py
ThrudPrimrose: I had saved the typestrings set by the transformation as a Dict(str,str), such that if a user wants to give the type info as a part of a transformation then the user can just write the str, but in a sense, having the type be a typeclass has information value in the "param_types" parameter, I will update.

ThrudPrimrose:
There is one issue that I am currently trying to solve. 
I can know change the map parameters' types, and can set them to be for example int, but I am having problems implementing the calls to the nested SDFGs seeing the correct types.

(I need to find a way for the method "new_symbols" of an interstate edge not to infer the type of the parameters if the parameter was introduced by a map parameter set by the user)

dace/codegen/targets/cpu.py
```diff
@@ -1925,6 +1927,12 @@ def _generate_MapEntry(
             # Emit nested loops
             for i, r in enumerate(node.map.range):
                 var = map_params[i]
+                if var in map_param_types.keys():
+                    var_type = map_param_types[var]
+                else:
+                    var_type = dtypes.result_type_of(infer_expr_type(r[0], state_dfg.symbols_defined_at(node)), 
+                                                     infer_expr_type(r[1], state_dfg.symbols_defined_at(node)))
```
tbennun, line 1934: Please compute this once, preferably outside the loop of node.map.range. `symbols_defined_at` can be an expensive function. Thanks!

dace/sdfg/state.py
```diff
@@ -1601,6 +1601,7 @@ def add_nested_sdfg(
         schedule=dtypes.ScheduleType.Default,
         location=None,
         debuginfo=None,
+        symbol_type_mapping: Dict[str, dtypes.typeclass] = None
```
tbennun, line 1604:
```suggestion
        symbol_type_mapping: Dict[str, dtypes.typeclass] = None,
```

tbennun (changes requested):
Good addition, thank you! Two things are missing:
1. My comment on symbols_defined_at (which should be computed as seldomly as possible)
2. Missing unit test with types (maybe even an additional test that would fail with `auto`, as you now changed the behavior of codegen even in the default case)

Also, you do not have to keep merging with master, the merge queue will do that for you. Every merge puts an unnecessary burden on the CI queue.

ThrudPrimrose:
> Good addition, thank you! Two things are missing:
> 
>     1. My comment on symbols_defined_at (which should be computed as seldomly as possible)
> 
>     2. Missing unit test with types (maybe even an additional test that would fail with `auto`, as you now changed the behavior of codegen even in the default case)
> 
> 
> Also, you do not have to keep merging with the master; the merge queue will do that for you. Every merge puts an unnecessary burden on the CI queue.

I will update the changes according to your comments. Thanks for pointing out the merge queue.

Since auto was only used in a loop, the type of the loop variable depends on the type used in the loop condition. The auto case will never fail to produce code that does not compile unless we compile with `-Wconversion -Werror.`
However, I can create test cases where the SDFG will produce incorrect results. I will do that (and try to find some good test cases).

I have looked at the tests, no hard-coded SDFGs are used in the tests, I should create any SDFG I use through the API or a front-end, right?

tbennun:
@ThrudPrimrose First of all, many tests use the SDFG API (e.g., [add_edge_pair_test.py](https://github.com/spcl/dace/blob/master/tests/add_edge_pair_test.py), you should use that.

Second, for a wrong test case, a simple for loop with a mismatch in start/end types will run forever if not written correctly. See the following example (which one can write in dace): https://godbolt.org/z/YW5xsP51x

tbennun:
What I meant was to have a test that _would_ fail if we generated `auto`. Obviously all the tests will now pass. :)

PR #1641, opened 2024-09-06
Add if extraction transformation

luigifusco:
As part of the series of SDFG rewriting transformations, IfExtraction targets nested SDFGs that begin with an if statement whose condition is independent of symbols defined or updated in the outer state (these are either writes to arrays or definitions in maps). The if is "extracted": the outers state is duplicated and a new if guard is created in the outer SDFG with the appropriately translated condition. If the edges contain symbol definitions these become definitions in the nested SDFG `symbol_mapping`.

Example of application:
image
becomes:
image

### DIFF
--- a/dace/transformation/interstate/if_extraction.py
+++ b/dace/transformation/interstate/if_extraction.py
@@ -1,0 +1,136 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
+""" If extraction transformation """
+
+from dace import data as dt, sdfg as sd
+from dace.sdfg import utils as sdutil
+from dace.sdfg.state import SDFGState
+from dace.transformation import transformation
+from copy import deepcopy
+from dace.properties import make_properties
+
+
+def eliminate_branch(sdfg: sd.SDFG, initial_state: sd.SDFGState):
+    state_list = [initial_state]
+    while len(state_list) > 0:
+        new_state_list = []
+        for s in state_list:
+            for e in sdfg.out_edges(s):
+                if len(sdfg.in_edges(e.dst)) == 1:
+                    new_state_list.append(e.dst)
+            sdfg.remove_node(s)
+        state_list = new_state_list
+
+
+@make_properties
+class IfExtraction(transformation.MultiStateTransformation):
+    """
+    Detects an if statement as the root of a nested sdfg, and extracts it by computing it in the outer sdfg and
+    replicating the state containing the nested sdfg
+    """
+
+
+    root_state = transformation.PatternNode(sd.SDFGState)
+
+
+    @staticmethod
+    def annotates_memlets():
+        return False
+
+
+    @classmethod
+    def expressions(cls):
+        return [sdutil.node_path_graph(cls.root_state)]
+
+
+    def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
+        root_state: SDFGState = self.root_state
+
+        out_edges = graph.out_edges(root_state)
+        in_edges = graph.in_edges(root_state)
+
+        if len(in_edges) > 0:
+            return False
+
+        if len(out_edges) != 2:
+            return False
+
+        # needs to be a nested sdfg
+        if not sdfg.parent:
+            return False
+
+        # check if edges can be moved out (used symbols can be computed in the outer scope)
+        if_symbols = set(s for e in out_edges for s in e.data.free_symbols)
+        if not if_symbols.issubset(sdfg.parent_nsdfg_node.symbol_mapping.keys()):
+            return False
+
+        return True
+
+
+    def apply(self, _, if_sdfg: sd.SDFG):
+        if_root_state: SDFGState = self.root_state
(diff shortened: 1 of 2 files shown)


dace/transformation/interstate/if_extraction.py
```diff
@@ -0,0 +1,156 @@
...
+        outer_sdfg: sd.SDFG = if_branch.sdfg
+        if_nested_sdfg_node: sd.nodes.NestedSDFG = if_sdfg.parent_nsdfg_node
+
+        if_edge, else_edge = if_sdfg.out_edges(if_root_state)
+
+        # create new state to perform the if, and have it replace the state containing the nested SDFG
+        new_state = outer_sdfg.add_state()
+        sdutil.change_edge_dest(outer_sdfg, if_branch, new_state)
+
+        # take the old state as the if branch, and create a copy to act as the else branch
```
luigifusco, line 101: I feel like "duplicate_element" should become a utility method of the sdfg

dace/transformation/interstate/if_extraction.py
```diff
@@ -0,0 +1,156 @@
...
+
+from dace import data as dt, sdfg as sd, symbolic
+from dace.sdfg import utils as sdutil
+from dace.sdfg.state import SDFGState
+from dace.transformation import transformation
+from copy import deepcopy
+from dace.properties import make_properties
+
+
+def eliminate_branch(sdfg: sd.SDFG, initial_state: sd.SDFGState):
```
luigifusco, line 12: Is there an analogous method in the sdfg? Should I make it one?

phschaad: I do not know of one, if this has potential applicability anywhere else, I would add a good docstring and put it on the `ControlFlowRegion` or SDFG Utilities.

phschaad (changes requested):
In general happy to accept this, but we need at least one or two unit tests

acalotoiu:
Agreeing with @phschaad here - tests for when this applies and when it does not apply are needed. Apart from that, it LGTM.

dace/transformation/interstate/if_extraction.py
phschaad: How is this different from a deepcopy?

luigifusco: It is already used in some other transformation because according to some comment "it is faster than deepcopy". Consider that deepcopy will copy the entire SDFG as it is referenced in the state. Deepcopying nodes can also lead to invalid state -> deepcopy shoud be used on very small things (edge data, ...) or avoided completely, and some state/node duplication utility should likely be provided in the state/node itself

pratyai:
> In general happy to accept this, but we need at least one or two unit tests

Added more unit tests, added more explanation. PTAL.

phschaad:
Thank you for continuing this @pratyai :-) As with #1639, I would prefer to briefly discuss this in the DaCe meeting before accepting, for the same reasons listed in #1639. However, in general this looks good and I am happy to add this contribution to DaCe.

tests/transformations/if_extraction_test.py
```diff
@@ -0,0 +1,225 @@
...
+    # Then prepare the parent graph.
+    g = SDFG('prog')
+    g.add_array('A', (10,), dace.float32)
+    st0 = g.add_state('outer', is_start_block=True)
+    en, ex = st0.add_map('map', {'i': '0:10'})
+    body = st0.add_nested_sdfg(subg, None, {}, {'tmp'})
+    A = st0.add_access('A')
+    st0.add_memlet_path(en, body, memlet=Memlet())
+    st0.add_memlet_path(body, ex, src_conn='tmp', dst_conn='IN_A', memlet=Memlet(expr='A[i]'))
+    st0.add_memlet_path(ex, A, src_conn='OUT_A', memlet=Memlet(expr='A[0:10]'))
```
tbennun, line 138: This is really not the right use of `add_memlet_path`. The entire point of adding a memlet _path_ is that you can do it in one go:
`st0.add_memlet_path(body, ex, A, memlet=Memlet('A[i]')`

tbennun (changes requested):

tests/transformations/if_extraction_test.py
tbennun: @pratyai

pratyai: I don't think
```python
st0.add_memlet_path(body, ex, A, memlet=Memlet('A[i]'))
```
is equivalent to
```python
st0.add_memlet_path(body, ex, src_conn='tmp', dst_conn='IN_A', memlet=Memlet(expr='A[i]'))
st0.add_memlet_path(ex, A, src_conn='OUT_A', memlet=Memlet(expr='A[0:10]'))
```
I have replaced all the `add_memlet_path()` path calls regardless.

phschaad (approved):

phschaad:
Leaving a note here to revisit this transformation for slight adaptation after https://github.com/spcl/dace/pull/1676 is merged.

PR #1645, opened 2024-09-06
Add resolve condition pass

luigifusco:
This pass takes a Python condition as a string (e.g. "flag == True"), parses it and, assuming it always holds throughout the SDFG, it deletes all unsatisfiable branches.

This is a first step towards implementing "specialized SDFGs". While playing with some code-replication inducing transformations I noticed that the graph quickly became too big to work with. This pass can help users generate different specialized SDFGs and work on those. As an example, if `flag` is an input boolean that is never written two, two SDFGs, one with `flag` set to true and the other with `flag` set to false can be created and worked on indipendently.

Creating as draft since multiple design decisions are still unclear to me

### DIFF
--- a/dace/transformation/passes/resolve_condition.py
+++ b/dace/transformation/passes/resolve_condition.py
@@ -1,0 +1,88 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
+""" Resolve condition transformation """
+
+from typing import Any, Dict
+
+from dace.transformation import pass_pipeline as ppl
+import sympy as sy
+from dace.properties import CodeBlock
+from dace import sdfg as sd, symbolic
+from dace.properties import Property, make_properties, CodeBlock
+from sympy.parsing.sympy_parser import parse_expr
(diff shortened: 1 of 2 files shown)


dace/transformation/passes/resolve_condition.py
```diff
@@ -0,0 +1,88 @@
...
+from dace import sdfg as sd, symbolic
+from dace.properties import Property, make_properties, CodeBlock
+from sympy.parsing.sympy_parser import parse_expr
+from dace import symbolic
+from dace.frontend.python.astutils import unparse
+from dace.transformation.pass_pipeline import Pipeline
+import ast
+
+
+def eliminate_branch(sdfg: sd.SDFG, initial_edge: sd.graph.Edge):
```
phschaad, line 18: Looks like yes, it should become a reusable, documented utility function :-)

acalotoiu: Agreed! This one is a good candidate for "building block" used by multiple transformations

PR #1646, opened 2024-09-06
More robust loop detection

tbennun:
Generalizes the behavior of loop detection to support rotated loops, single-state loops, and as a result LLVM canonical loops. The PR also refactors the loop analysis methods and generalizes `DetectLoop` transformation subclasses, such as `LoopToMap` and `LoopPeeling`.


phschaad (approved):
Nice, this looks great - I'm hoping this will serve us well in a pass that lifts these into LoopRegions. Thanks!

PR #1647, opened 2024-09-09
Reworked Optional Serializing

philip-paul-mueller:
Before there were `optional` and `optional_condition` (which had the oposite meaning, i.e. not optional if the callable evaluates to `True`). This was all replaced by the `serialize_if` callable, which if it returns `True` what the default does the property is serialized and if it rerturns `False` the property is not serialized. This commit also introduces the `dtypes.CPU_SCHEDULES` collection.

### DIFF
--- a/dace/dtypes.py
+++ b/dace/dtypes.py
@@ -87,6 +87,12 @@ GPU_SCHEDULES = [
     ScheduleType.GPU_Persistent,
 ]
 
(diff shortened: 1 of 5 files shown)


phschaad (approved):
This seems like a good idea to me, makes it quite a bit simpler. LGTM, thank you!

PR #1648, opened 2024-09-09
Added  `can_be_applied_to()` To Transformation Interface

philip-paul-mueller:
Before there was only the `apply_to()` function which allowed to perform the transformation. However, sometimes when transformations are composed it is useful to be able to check if a transformation could be applied later. Sometimes this follows directly from the operation that is currently done, but sometimes it is not easy to check it.


tbennun (approved):
Nice addition to the API!

Issue #1493, opened 2024-01-02
f32 is not infered on basic mult


FlorianDeconinck:
Consider this snippet:
```
import numpy as np
import dace


@dace.program
def dace_float_cast(A, B, timestep: dace.float32):
    dt2: dace.float32 = 2 * timestep
    A[:] = B[:] * dt2


A = np.zeros(10, dtype=np.float32)
B = np.zeros(10, dtype=np.float32)
t = np.float32(250.0)

dace_float_cast(A, B, t)
```

SDFG & code generated (below) show `dt2` is consider a 64-bit float, despite the cast.

```
...
double dt2;
{
    float __in2 = timestep;
    double __out;

    ///////////////////
    // Tasklet code (_Mult_)
    __out = (dace::float64(2) * dace::float64(__in2));
    ///////////////////

    dt2 = __out;
}
...
```

SDFG (not a real zip, just going around GitHub) [program.sdfg.zip](https://github.com/spcl/dace/files/13814316/program.sdfg.zip)

FlorianDeconinck:
A follow up bug is that even if `dt2` is edited in SDFG to be a float32 and the SDFG is properly typed throughout, the generated code battles to be all float. E.g.:

```
{
        float dt2;

        {
            float __in2 = timestep;
            double __out;

            ///////////////////
            // Tasklet code (_Mult_)
            __out = (dace::float64(2) * dace::float64(__in2));
            ///////////////////

            dt2 = __out;
        }
        {
            #pragma omp parallel for
            for (auto __i0 = 0; __i0 < 10; __i0 += 1) {
                {
                    float __in1 = B[__i0];
                    double __in2 = *(double *)(&dt2);
                    float __out;

                    ///////////////////
                    // Tasklet code (_Mult_)
                    __out = (__in1 * dace::float32(__in2));
                    ///////////////////

                    A[__i0] = __out;
                }
            }
        }

    }
```

alexnick83:
The quick fix for anyone wanting to go ahead with a PR is to amend [line 3289 in frontend/python/newast](https://github.com/spcl/dace/blob/1393cb0a592cae42dde7b15d1b4c80dec8e2e9c4/dace/frontend/python/newast.py#L3289) as follows:
```python
                    elif (not result_data.transient or result in self.sdfg.constants_prop or
                          (dtype is not None and dtype != result_data.dtype)):
```
[...]

PR #1650, opened 2024-09-10
Extend TrivialTaskletElimination for map scope

edopao:
Extend the transformation `TrivialTaskletElimination` for the case where the input or output of the copy-tasklet is a map node.

The following SDFG:
image

is transformed to this SDFG:
image

### DIFF
--- a/dace/transformation/dataflow/trivial_tasklet_elimination.py
+++ b/dace/transformation/dataflow/trivial_tasklet_elimination.py
@@ -34,26 +34,34 @@ class TrivialTaskletElimination(transformation.SingleStateTransformation):
         read = self.read_map if expr_index == 1 else self.read
         tasklet = self.tasklet
         write = self.write_map if expr_index == 2 else self.write
-        # Do not apply on Streams
-        if expr_index != 1 and isinstance(sdfg.arrays[read.data], data.Stream):
-            return False
-        if expr_index != 2 and isinstance(sdfg.arrays[write.data], data.Stream):
+        if len(tasklet.in_connectors) != 1:
             return False
         if len(graph.in_edges(tasklet)) != 1:
             return False
-        if len(graph.out_edges(tasklet)) != 1:
-            return False
-        if graph.edges_between(tasklet, write)[0].data.wcr:
-            return False
-        if len(tasklet.in_connectors) != 1:
-            return False
         if len(tasklet.out_connectors) != 1:
+            return False
+        if len(graph.out_edges(tasklet)) != 1:
             return False
         in_conn = list(tasklet.in_connectors.keys())[0]
         out_conn = list(tasklet.out_connectors.keys())[0]
         if tasklet.code.as_string != f'{out_conn} = {in_conn}':
             return False
-
+        read_memlet = graph.edges_between(read, tasklet)[0].data
+        read_desc = sdfg.arrays[read_memlet.data]
+        write_memlet = graph.edges_between(tasklet, write)[0].data
+        if write_memlet.wcr:
+            return False
+        write_desc = sdfg.arrays[write_memlet.data]
+        # Do not apply on streams
+        if isinstance(read_desc, data.Stream):
+            return False
+        if isinstance(write_desc, data.Stream):
+            return False
+        # Keep copy-tasklet connected to map node if source and destination nodes
+        # have different data type (implicit type cast)
+        if expr_index != 0 and read_desc.dtype != write_desc.dtype:
+            return False
+    
         return True
 
     def apply(self, graph, sdfg):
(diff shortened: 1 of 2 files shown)


edopao:
The test failure highlights an aspect of codegen in baseline: the copy tasklet preceding the `MapExit` node is implicitly doing a cast from `float64` to `float32` before writing to the array. Here is the generated code for the failing testcase:
```
double __tmp2;
{
    float __in1 = gpu_B[((300 * i) + j)];
    double __out;

    ///////////////////
    // Tasklet code (_Add_)
    __out = (dace::float64(__in1) + 1.0);
    ///////////////////

    __tmp2 = __out;
}

dace::CopyND<float, 1, false, 1>::template ConstDst<1>::Copy(
&__tmp2, gpu_A + ((300 * i) + j), 1);
```

and the compilation error points to the type mismatch on the source pointer:
```
argument types are: (double *, float *, int)
                         dace::CopyND<float, 1, false, 1>::template ConstDst<1>::Copy(
```

I managed to make the test pass, in other word to fix the code generation, if I explicitly set the type of the decimal literal in the numpy test:
```
--- a/tests/transformations/gpu_grid_stride_tiling_test.py
+++ b/tests/transformations/gpu_grid_stride_tiling_test.py
@@ -39,7 +39,7 @@ def test_gpu_grid_stride_tiling():
     def dummy(A: dace.float32[M, N], B: dace.float32[M, N]):
         for i in dace.map[0:M]:
             for j in dace.map[0:N]:
-                A[i, j] = B[i, j] + 1.0
+                A[i, j] = B[i, j] + np.float32(1.0)
 
     sdfg = dummy.to_sdfg()
     sdfg.simplify()
```

Is this the right way to proceed?

edopao:
In case the usage of copy-tasklet is the intended (although questionable) solution to implicitly cast to the correct destination type, I could implement a check suggested by @philip-paul-mueller: only remove the copy-tasklet if the source and destination containers have the same data type.

Adding @acalotoiu and @BenWeber42

tbennun:
@edopao there are two issues here:
* If there is a cast, then the tasklet is not trivial and should not be eliminated.
* The Python frontend should use float32 for the `1.0` in the code there. It is strange that it doesn't

edopao:
Here is the generated CUDA code in baseline (thus keeping the copy tasklet):
```
double __tmp2;
{
    float __in1 = gpu_B[((300 * i) + j)];
    double __out;

    ///////////////////
    // Tasklet code (_Add_)
    __out = (dace::float64(__in1) + 1.0);
    ///////////////////

    __tmp2 = __out;
}
{
    double __inp = __tmp2;
    float __out;

    ///////////////////
    // Tasklet code (assign_42_12)
    __out = __inp;
    ///////////////////

    gpu_A[((300 * i) + j)] = __out;
}
```

Here you can see the SDFG:
image

tbennun:
> > @edopao there are two issues here:
> > ```
> > * If there is a cast, then the tasklet is not trivial and should not be eliminated.
> > 
> > * The Python frontend should use float32 for the `1.0` in the code there. It is strange that it doesn't
> > ```
> 
> Thank you for your reply. About the first bullet: the cast is implicitly generated in the C++/CUDA code, but the Python code of tasklet node is just `out = inp`.

It's not in the tasklet's contents, but the connector types (as you mentioned before). The equivalent tasklet-free code would be two access nodes rather than one (one that represents the input scalar and another output scalar).

tbennun:
> Here is the generated CUDA code in baseline (thus keeping the copy tasklet):
> 
> ```
> double __tmp2;
> {
>     float __in1 = gpu_B[((300 * i) + j)];
>     double __out;
> 
>     ///////////////////
>     // Tasklet code (_Add_)
>     __out = (dace::float64(__in1) + 1.0);
>     ///////////////////
> 
>     __tmp2 = __out;
> }
> {
>     double __inp = __tmp2;
>     float __out;
> 
>     ///////////////////
>     // Tasklet code (assign_42_12)
>     __out = __inp;
>     ///////////////////
> 
>     gpu_A[((300 * i) + j)] = __out;
> }
> ```
> 
> Here you can see the SDFG: image

I think this involuntary type promotion is something that should be resolved in the Python/numpy frontend. It's probably causing slowdowns in codes even without that elimination.

edopao:
The second issue is probably related to #1493.

edopao:
@tbennun When you have time, this PR is ready for a final review. The casting issue related to #1493 is outside of the PR scope.

edopao:
@acalotoiu A gentle reminder about the review, when you have some time.

acalotoiu (approved):
LGTM.

PR #1653, opened 2024-09-11
Modified `SetProperty`

philip-paul-mueller:
Because of the implementation of the `__set__()` and `__get__()` functions such properties modeled an immutable set, by essentilly copying the value that was assigned to it and returning a copy of the internal set. However, because `__get__()` returned a regiular `set` this object was mutable, but all mutation did not affect the property, which is a rather unintuitive behaviour.

The effect is, that every single `assert` in the code bellow fails.
```
import dace

@dace.properties.make_properties
class Test:
    member = dace.properties.SetProperty(element_type=int)

    def __init__(self):
        self.member = set()

a = Test()

assert a.member is a.member
a.member.add(10)
assert len(a.member) == 1
assert 10 in a.member
```
It is important that even if the first assert is removed the second two asserts will still fail and the insertion line, will succeed without issue, because the operation is performed on a copy.

The new implementation now stores a `frozenset` internally and returns it when accessed. Thus the first assert will fail and lines such as `42 in a.member` will not produce a copy.
Furthermore and more importantly, the line `a.member.add(10)` will now fail, since a `frozenset` can not be modified. As a side note the last two assert will still fail because the set can not be updated in place. The only way to perform an update is now `a.member = set(...)` but this was true before.

For backward compatibility the property pretends to be a regular set during serialization (`typestring`), however, `dtype` is reported as `frozenset` because the underlying implementation of `Property.__set__()` requires this.

### DIFF
--- a/dace/properties.py
+++ b/dace/properties.py
@@ -778,7 +778,11 @@ class DebugInfoProperty(Property):
 
 
 class SetProperty(Property):
-    """Property for a set of elements of one type, e.g., connectors. """
+    """Property for a set of elements of one type, e.g., connectors.
+
+    Despite its name, the property models a `frozenset`, this means that the set can
+    not be modified in place. Instead a new value has to be assigned to the property.
+    """
 
     def __init__(
             self,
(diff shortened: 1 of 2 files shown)


phschaad (approved):
Makes sense to me - I believe the backwards compatibility fix should be sufficient, otherwise I will make sure the visualization component is adjusted.

tbennun:
That's great, thank you. Does this also happen to dict (and list) properties?

PR #1659, opened 2024-09-16
[Draft PR] Graphcore backend support.

Sameeranjoshi:
This PR is a draft PR with lots of learnings from fpga.py and cuda.py backends. 
The goal of the PR is to add graphcore backend to DaCE and map from DaCE graphs onto GraphCore graphs. 
Helpful comments on parts I might be doing wrong would be super useful for next set of patches.


Sameeranjoshi:
The golden base code I plan to generate for IPU is [here](https://github.com/Sameeranjoshi/dace/wiki/IPU_Golden_base_code).

The current test I plan to support is (AccessNode + Library Node) - `tests/library/poplar/poplar_matmul.py`

dace/dtypes.py
ThrudPrimrose, line 1: For future users, using more descriptive names of the memory locations might be more beneficial.
I do not know the memory hierarchy, but if the Graphcore has only Global memory and on-chip L1, I would go with IPU_L1 and IPU_Global.

Using the naming can also be good if IPUs change their memory hierarchy across generations. Let's say they add a level called L2 -> Then it will be IPU_Memory and IPU_L2 (and in my opinion, this will be quite confusing)

Same for the schedule, IPU_Multicore can be better (this again depends on what GraphCore calls them).
For example:

```
    StorageType.IPU_L1: ScheduleType.IPU_Multicore,
    StorageType.IPU_Global: ScheduleType.IPU_Device
```

dace/codegen/targets/ipu_files/ipu_utils.py
ThrudPrimrose, line 1: There are, type maps such as  `_CTYPES` (C/C++) or `_OCL_TYPES` (OpenCL) in dtypes.py
I think having this map there with other type maps would be consistent with the previous implementations.

dace/libraries/poplar/environments/poplar.py
ThrudPrimrose, line 1: Init code is for the part of the code the use library requires in initialization (`MPI_Init`) and finalize then the code needed for the finalization (`MPI_finalize`)


From what I see, `engine.run(0)` is similar to a Cuda kernel launch. Wouldn't it be better to generate this in a `target/ipu.py`?

If implemented like this, wouldn't there be problems if you need to call `.run` more than once?

ThrudPrimrose: Due to how `graphcore` requires you to use the `poplar` library, using it as a library is unavoidable, I want to point that you can still put as much as you can in ipu code-gen and ipu code-gen makes this library mandatory. 

Good luck and success in your DaCe backend implementation!

PR #1661, opened 2024-09-18
Better Name Validation

philip-paul-mueller:
This PR adds checks to the SDFG to ensures that names from symbols, data descriptors and so on are unique.
Furthermore, it also ensures that the NestedSDFG validates correctly and ensures that no symbols can be written.

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -136,27 +136,6 @@ def create_datadescriptor(obj, no_custom_desc=False):
                     'adaptor method to the type hint or object itself.')
 
 
-def find_new_name(name: str, existing_names: Sequence[str]) -> str:
-    """
-    Returns a name that matches the given ``name`` as a prefix, but does not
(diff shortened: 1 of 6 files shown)


philip-paul-mueller:
Let's forget about this thing, it is just not worth it.

dace/data.py
```diff
@@ -136,27 +136,6 @@ def create_datadescriptor(obj, no_custom_desc=False):
                     'adaptor method to the type hint or object itself.')
 
 
-def find_new_name(name: str, existing_names: Sequence[str]) -> str:
```
alexnick83, line 139: I think that this is part of the external API, so we cannot remove it. It could be moved to a clearer utility file, but maybe this has to be discussed in a DaCe meeting.

alexnick83: Nvm, it is duplicate.

PR #1662, opened 2024-09-19
Various Cutout Fixes

phschaad:
- [x] Fix cutouts w.r.t. the use of UIDs, allowing them to be preserved or re-generated depending on an input parameter
- [x] Fix singlestate cutout extraction when memlets access struct members.


dace/sdfg/analysis/cutout.py
```diff
@@ -213,11 +227,24 @@ def singlestate_cutout(cls,
...
+            if '.' in dataname:
+                # This is an access to a struct memeber, which typically happens for the memlets between an access node
+                # pointing to a struct (or view thereof), and a view pointing to the member. Assert that this is indeed
+                # the case (i.e., only one '.' is found in the name of the data being accessed), and if so, clone the
+                # struct (or struct view) data descriptor instad.
+                parts = dataname.split('.')
+                if len(parts) == 2:
+                    dataname = parts[0]
+                else:
+                    raise RuntimeError('Attempting to add invalid multi-nested data ' + memlet.data + ' to a cutout')
```
tbennun, line 240: There is a get root data descriptor function as far as I'm aware. Please use it.

dace/sdfg/analysis/cutout.py
```diff
@@ -337,12 +365,16 @@ def multistate_cutout(cls,
...
+                               identical to the ones in their original graph. If False, new GUIDs will be generated.
+                               False by default - if make_copy is False, this has no effect by extension.
         :param override_start_block: If set, explicitly force a given control flow block to be the start block. If left
                                      None (default), the start block is automatically determined based on domination
                                      relationships in the original graph.
         :return: The created SDFGCutout or the original SDFG where no smaller cutout could be obtained.
         """
-        create_element = copy.deepcopy
+        def create_element(x: Union[ControlFlowBlock, InterstateEdge]) -> Union[ControlFlowBlock, InterstateEdge]:
+            return x.clone(keep_guid=preserve_guids)
```
tbennun, line 377: I would use an inline lambda, it's much cleaner

dace/sdfg/analysis/cutout.py
```diff
@@ -188,17 +193,26 @@ def singlestate_cutout(cls,
...
+        :param preserve_guids: If True, ensures that the GUIDs of graph elements contained in the cutout remain
+                               identical to the ones in their original graph. If False, new GUIDs will be generated.
+                               False by default - if make_copy is False, this has no effect by extension.
         :return: The created SDFGCutout.
         """
         if reduce_input_config:
             nodes = _reduce_in_configuration(state, nodes, use_alibi_nodes, symbols_map)
-        create_element = copy.deepcopy if make_copy else (lambda x: x)
+
+        def clone_f(x: Union[Memlet, InterstateEdge, nd.Node, ControlFlowBlock]):
```
tbennun, line 204: why not use a lambda for both

tbennun (changes requested):
I don't like the new clone methods, is there any other way? Also, see comments

dace/sdfg/analysis/cutout.py
phschaad: I can only find one defined as a property on `AccessNode`s (`AccessNode::root_data`). However, that appears to be doing something different than what this portion here attempts to do. `AccessNode::root_data` always grabs `dataname.split('.')[0]`, no matter how long the sub-member access chain is, whereas here we specifically want to make sure that there is only one nesting level.

dace/sdfg/analysis/cutout.py
phschaad: The goal was to be able to type-hint, which is very cumbersome with lambdas

dace/sdfg/analysis/cutout.py
phschaad: Changed - note that this drops type information

phschaad: After removing the `clone` function I believe it is cleaner this way since this way provides type information, while lambdas completely remove the ability to type hint things (cleanly)

dace/sdfg/analysis/cutout.py
phschaad: See above

tbennun (approved):

PR #1672, opened 2024-10-03
Inter-state edge assignment race test

tbennun:
Add a validation test for race conditions induced by assigning and reading the same value in a single interstate edge. For example: `i = j + 1; j = ...`


phschaad (approved):
Thank you! LGTM

PR #1673, opened 2024-10-04
[DO NOT REVIEW] Fixing a (likely) bug in `MapFusion`.

pratyai:
When checking for the "array usage" criteria that can prevent map-fusion, check only within the current state. Otherwise, any "use" of the array _globally_ (i.e., in the entire SDFG) will cancel the fusion.

### DIFF
--- a/dace/codegen/targets/cpp.py
+++ b/dace/codegen/targets/cpp.py
@@ -417,6 +417,7 @@ def reshape_strides(subset, strides, original_strides, copy_shape):
     dims = len(copy_shape)
 
     reduced_tile_sizes = [ts for ts, s in zip(subset.tile_sizes, original_copy_shape) if s != 1]
+    reduced_tile_sizes += [1] * (dims - len(reduced_tile_sizes))  # Pad the remainder with 1s to maintain dimensions.
 
     reshaped_copy = copy_shape + [ts for ts in subset.tile_sizes if ts != 1]
     reshaped_copy[:len(copy_shape)] = [s / ts for s, ts in zip(copy_shape, reduced_tile_sizes)]
--- a/dace/codegen/targets/cpp.py
+++ b/dace/codegen/targets/cpp.py
@@ -420,7 +420,7 @@ def reshape_strides(subset, strides, original_strides, copy_shape):
     reduced_tile_sizes += [1] * (dims - len(reduced_tile_sizes))  # Pad the remainder with 1s to maintain dimensions.
 
     reshaped_copy = copy_shape + [ts for ts in subset.tile_sizes if ts != 1]
-    reshaped_copy[:len(copy_shape)] = [s / ts for s, ts in zip(copy_shape, reduced_tile_sizes)]
+    reshaped_copy[:len(copy_shape)] = [s // ts for s, ts in zip(copy_shape, reduced_tile_sizes)]
 
     new_strides = [0] * len(reshaped_copy)
     elements_remaining = functools.reduce(sp.Mul, copy_shape, 1)
--- a/dace/transformation/dataflow/map_fusion.py
+++ b/dace/transformation/dataflow/map_fusion.py
@@ -105,7 +105,7 @@ class MapFusion(transformation.SingleStateTransformation):
                 intermediate_data.add(dst.data)
 
                 # If array is used anywhere else in this state.
-                num_occurrences = len([n for n in sdfg.data_nodes() if n.data == dst.data])
+                num_occurrences = len([n for n in graph.data_nodes() if n.data == dst.data])
                 if num_occurrences > 1:
                     return False
             else:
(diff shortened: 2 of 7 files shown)


tbennun:
This should not be the correct behavior. Eliminating an access node that corresponds to a data container has to be done safely: if a data container is used in another state (via an access node with the same name), its memory must be preserved because it can be reused in a subsequent read (which is a case that happens often when conserving memory). If you want map fusion to work, you would need to use a unique array name that can be removed, or rename arrays that you know they will be overwritten later.

Checking the SDFG for other instances of the access node is the correct way to check for a safe removal of an access node:
```python
>>> sdfg = dace.SDFG('test')
>>> state = sdfg.add_state()
>>> state2 = sdfg.add_state_after(state)
>>> state.add_access('A')
>>> state2.add_access('A')
# Checking access nodes:
>>> state.data_nodes()
[AccessNode (A)]
>>> sdfg.data_nodes()
[AccessNode (A), AccessNode (A)]
```

pratyai:
I see that my test case (of applying fusion twice) has this problem: the corresponding transient access nodes in two copies have the same name. I was assuming that the transient access nodes are scoped only within the state, and states cannot rely on other states' transient access nodes.
Just to confirm, must the transient access nodes have globally unique names?

Even then, I think that MapFusion's current way sometimes unnecessarily chooses to avoid the fusion. Could we just not remove such access nodes and just proceed with the fusion where both versions would produce identical output? In other words, what if we allow the fusion without any access-node-removal, and _then_ remove any redundant transient nodes?

[ By the way, I've been looking at this since only very recently, so chances are that I have some misunderstanding about access nodes. Apologies in advance, if that's indeed the case here. ]

pratyai:
I've come up with the following example to demonstrate. I want the MapFusion to work on the following graph:

```python
def f4():  
    g = dace.SDFG('f4')  
    g.add_array('A', [10], dace.float32)  
    g.add_array('B', [10], dace.float32)  
  
    st0 = g.add_state('st0')  
    # First map  
    wA_e, wA_x = st0.add_map('st0_wA', {'i': f"0:10"})  
    op0 = st0.add_tasklet('set_1', {}, {'Ai_out'}, 'Ai_out = 1')  
    st0.add_memlet_path(wA_e, op0, memlet=dace.Memlet())  
    wA = st0.add_write('A')  
    st0.add_memlet_path(op0, wA_x, wA, src_conn='Ai_out', memlet=dace.Memlet(data='A', subset='i'))  
    # Second map  
    wB_e, wB_x = st0.add_map('st0_wB', {'i': f"0:10"})  
    op1 = st0.add_tasklet('set_B', {'Ai_in'}, {'Bi_out'}, 'Bi_out = Ai_in')  
    st0.add_memlet_path(wA, wB_e, op1, dst_conn='Ai_in', memlet=dace.Memlet(data='A', subset='i'))  
    wB = st0.add_write('B')  
    st0.add_memlet_path(op1, wB_x, wB, src_conn='Bi_out', memlet=dace.Memlet(data='B', subset='i'))  
  
    st1 = g.add_state_after(st0, 'st1')  
    # First map  
    wA_e, wA_x = st1.add_map('st1_wA', {'i': f"0:10"})  
    op2 = st1.add_tasklet('set_2', {}, {'Ai_out'}, 'Ai_out = 2')  
    st1.add_memlet_path(wA_e, op2, memlet=dace.Memlet())  
    wA = st1.add_write('A')  
    st1.add_memlet_path(op2, wA_x, wA, src_conn='Ai_out', memlet=dace.Memlet(data='A', subset='i'))  
    g.validate()  
    return g
```

Which produces this SDFG: fusion_2state_before

This SDFG currently does not allow fusion inside state#1, since state#2 also contains an access to `A`. However, I would like to transform it to: fusion_2state_after_fix

I claim that:
1. Fusion should not remove the access node `A` from state#1.
2. State#2 can only rely on the data in `A` _after_ state#1 has completed in entirety. So, the fusion should still produce the same data in `A` and `B` at the end of state#1, but should be allowed to transform computation within it.

tbennun:
@philip-paul-mueller related to what you are working on?

philip-paul-mueller:
Yes. This problem should be solved by the new map fusion.

philip-paul-mueller:
The error you have looks familiar, I saw  it when I rewrote MapFusion.
Back then I had the feeling that subgraph fusion, the component that fails, highly depends on how the graph produced by MapFusion looks.
My solution was to tweak MapFusion in such a way to avoid some fusion operation during auto optimization.
This is not a nice solution, but it worked.

@pratyai 
I would also advice you to look at my [PR](https://github.com/spcl/dace/pull/1629) that reimplements MapFusion.
Since I have currently other, more urgent, stuff to do you are more than welcomed to work on it.
If you are interested contact me in Mattermost.

Issue #1674, opened 2024-10-07
GPU Code Generation Fails Because of  `CopyToMap`


philip-paul-mueller:
The CUDA code generation will call `CopyToMap` for certain memlets.
However, it seems that there is an error in certain cases.

We start with the initial SDFG.
It writes the result back in a state:
![initial_copy](https://github.com/user-attachments/assets/b6ad4614-e488-49f7-b92a-79800b686aa7)

If we try to compile the initial SDFG, attached as `initial_sdfg.json`, code gen will generate the code but the compilation will fail, with the error that `__vn_size_1` is not specified.
The output folder contains a copy of the SDFG (`program.sdfg`); Attached as `transformed_sdfg.json`.
We can visualize it and see that the memlet for writting back the result has been turned into a map.
![after_transformation](https://github.com/user-attachments/assets/8feba93d-2fad-4935-9cf9-7a040505e2da)

What is now very interesting if we load the transformed SDFG we can compile it without issues.

[initial_sdfg.json](https://github.com/user-attachments/files/17276111/initial_sdfg.json)
[transformed_sdfg.json](https://github.com/user-attachments/files/17276106/transformed_sdfg.json)

tbennun:
@philip-paul-mueller can we close this now that #1675 is merged? Please reopen if not

PR #1675, opened 2024-10-07
Better `CopyToMap`

philip-paul-mueller:
By default the transformation uses some linearization followed by a delinearization approach, while this is needed to copy certain shapes, it is unnecessarily complicated for memlets such as `a[0:10, 20:30] -> 40:50, 60:70`.
This PR adds special cases where the source and destination subset has the same size and transforms it to a simple copy.
It also supports the case where some dimensions are one, i.e. memlets such as `a[0:10, 0:10] -> 0:10, 1, 0:20`.

For all cases tests were added.
Most importantly the transformation now applies if the strides are the same.
Before this case was blocked.

This PR helps to _avoid_ errors that are related to [Issue#1674](https://github.com/spcl/dace/issues/1674), but it is not a fix or a solution to it.


dace/transformation/dataflow/copy_to_map.py
```diff
@@ -31,7 +36,12 @@ def can_be_applied(self, graph: SDFGState, expr_index: int, sdfg: SDFG, permissi
         if isinstance(self.b.desc(sdfg), data.View):
             if sdutil.get_view_node(graph, self.b) == self.a:
                 return False
-        if self.a.desc(sdfg).strides == self.b.desc(sdfg).strides:
+        if (not self.ignore_strides) and self.a.desc(sdfg).strides == self.b.desc(sdfg).strides:
+            return False
+        if self.a.data == self.b.data:
```
alexnick83, line 41: Why are we disallowing this? I understand that alarms should light up if we are copying from/to the same container. However, any potential damage is already done if the copy is already on the SDFG. Therefore, I would expect that, in a correct SDFG, to not actually have an overlap among the subsets.

dace/transformation/dataflow/copy_to_map.py
```diff
@@ -62,31 +72,69 @@ def delinearize_linearize(self, desc: data.Array, copy_shape: Tuple[symbolic.Sym
...
-        edge = state.edges_between(self.a, self.b)[0]
+        avnode = self.a
+        av = avnode.data
+        adesc = avnode.desc(sdfg)
+        bvnode = self.b
+        bv = bvnode.data
+        bdesc = bvnode.desc(sdfg)
+
+        edge = state.edges_between(avnode, bvnode)[0]
+        src_subset = edge.data.get_src_subset(edge, state)
```
alexnick83, line 83: Do you think we may need to call the `try_initialize` method before attempting to get src/dst subsets here?

alexnick83 (commented):
LGTM, in principle, but please check the two comments I left.

dace/transformation/dataflow/copy_to_map.py
philip-paul-mueller: I am actually not sure.
I was hit by this a few time in the past so I am in favour of keeping it.
However, the main reason why this function is used (I think I "discovered" it here) is because even in the original code it is used (see old line 70).
For that reason I decided to keep it.
But if you think we should remove that I will not object.

dace/transformation/dataflow/copy_to_map.py
philip-paul-mueller: I am not fully understand your comment about the same container.
Furthermore, I also do not understand why this test of the strides is, for me it does not make any sense.
I kept it because the original transformation had it so I maintained it.
I think that this check is there because of the history of this transformation, as it is used by the code generator.
But honestly I have no idea.

What we could do is change the default of `ignore_strides` to `True` and then inside the code generator to `False` to maintain the old behaviour.

alexnick83: My question is about checking that we are not copying from/to the same container in line 41. I have the impression that this wasn't there before but maybe I didn't check the diff properly.

dace/transformation/dataflow/copy_to_map.py
alexnick83: I mean, should we also add a call to the method that tries to initialize the edge/meet to ensure that the src/dst subsets are not None when they are not supposed to.

dace/transformation/dataflow/copy_to_map.py
philip-paul-mueller: No you checked the diff correctly, I am just stupid.

alexnick83 (approved):

Issue #1643, opened 2024-09-06
Bug in `RefineNestedAccess` and `SDFGState._read_and_write_sets()`


philip-paul-mueller:
During the implementation of the new [MapFusion](https://github.com/spcl/dace/pull/1629) transformation it became necessary to reimplement `_read_and_write_sets()` there several bugs were discovered and fixed. However, one bug could not be fixed that is related to how the read set is cleaned.

If the following patch is applied
```
From f61d4e142b2b2097250b77e5b56b636b67ac9219 Mon Sep 17 00:00:00 2001
From: Philip Mueller <philip.mueller@cscs.ch>
Date: Fri, 6 Sep 2024 15:23:59 +0200
Subject: [PATCH] Undo the compatibility mode in _read_and_write_sets

---
 dace/sdfg/state.py | 10 ----------
 1 file changed, 10 deletions(-)

diff --git a/dace/sdfg/state.py b/dace/sdfg/state.py
index 23ac2f763..0c912d4d1 100644
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -791,16 +791,6 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
                 # Filter out memlets which go out but the same data is written to the AccessNode by another memlet
                 for out_edge in list(out_edges):
                     for in_edge in in_edges:
-                        if out_edge.data.data != in_edge.data.data:
-                            # NOTE: This check does not make any sense, and is in my view wrong.
-                            #   If we consider a memlet between two access nodes, to which access
-                            #   node the `data` attribute of the memlet refers to is arbitrary and
-                            #   does not matter. However, the test will filter _some_ out but not
-                            #   all. See also the tests inside `tests/sdfg/state_test.py` for the
-                            #   wrong behaviour this check induces.
-                            #   This check is is retained for  compatibility with `RefineNestedAccess`,
-                            #   see `tests/numpy/ufunc_support_test.py::test_ufunc_add_accumulate_simple`.
-                            continue
                         if in_subsets[in_edge].covers(out_subsets[out_edge]):
                             out_edges.remove(out_edge)
                             break
-- 
2.46.0

```
then `tests/transformations/move_loop_into_map_test.py::MoveLoopIntoMapTest::test_apply_multiple_times` will fail with an invalid SDFG error. The bug is probably located somewhere inside `RefineNestedAccess`.

[...]

PR #1678, opened 2024-10-11
Fix to Read and Write Sets

philip-paul-mueller:
During my work on the [new map fusion](https://github.com/spcl/dace/issues/1643) I discovered a bug in `SDFGState._read_and_write_set()`.
Originally I solved it there, but it was decided to move it into its own PR.

Lets look at the first, super silly example, that is not useful on its own.
The main point here, is that the `data` attribute of the Memlet does not refer to the source of the connection but of the destination.

test_1

BTW: The Webviewer outputs something like `B[0] -> [0, 0]` however, the parser of the Memlet constructor does not understand this, it must be written as `B[0] -> 0, 0`, i.e. the second set of brackets must be omitted, this should be changed!

From the above we would expect the following sets:
- Reads:
	- `A`: `[Range (0, 0)]`
	- `B`: Should not be listed in this set, because it is fully read and written, thus it is excluded.
- Writes
	- `B`: `[Range (0)]`
	- `C`: `[Range (0, 0), Range (1, 1)]`

However, the current implementation gives us:
- Reads: `{'A': [Range (0)], 'B': [Range (1, 1)]}`
- Write: `{'B': [Range (0)], 'C': [Range (1, 1), Range (0)]}`

The current behaviour is wrong because:
- `A` is a `2x2` array, thus the read set should also have two dimensions.
- `B` inside the read set, it is a scalar, but the range has two dimensions, furthermore, it is present at all.
- `C` the first member of the write set (`Range(1, 1)`) is correct, while the second (`Range(0)`) is horrible wrong.

The second example is even more simple.

test_2

From the SDFG we expect the following sets:
- Reads:
	- `A`: `[Range(0, 0)]`
- Writes:
	- `B`: `[Range(0)]`

It is important that in the above example `other_subset` is `None` and `data` is set to `A`, so it is not one of these "crazy" non standard Memlets we have seen in the first test.
However, the current implementation gives us:
- Reads: `{'A': [Range (0, 0)]}`
- Writes: `{'B': [Range (0, 0)]}`

This clearly shows, that whatever the implementation does is not correct.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -766,7 +766,8 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
             subgraph_write_set = collections.defaultdict(list)
             # Traverse in topological order, so data that is written before it
             # is read is not counted in the read set
-            # TODO: This only works if every data descriptor is only once in a path.
+            # NOTE: Each AccessNode is processed individually. Thus, if an array appears multiple
+            #   times in a path, the individual results are combined, without further processing.
             for n in utils.dfs_topological_sort(subgraph, sources=subgraph.source_nodes()):
                 if not isinstance(n, nd.AccessNode):
                     # Read and writes can only be done through access nodes,
@@ -808,7 +809,8 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
                 #  exists a memlet at the same access node that writes to the same region, then
                 #  this read is ignored, and not included in the final read set, but only
                 #  accounted fro in the write set. See also note below.
-                # TODO: Handle the case when multiple disjoint writes are needed to cover the read.
+                # TODO: Handle the case when multiple disjoint writes would be needed to cover the
+                #   read. E.g. edges write `0:10` and `10:20` but the read happens at `5:15`.
                 for out_edge in list(out_edges):
                     for in_edge in in_edges:
                         if in_subsets[in_edge].covers(out_subsets[out_edge]):
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -811,12 +811,6 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
                 # TODO: Handle the case when multiple disjoint writes are needed to cover the read.
                 for out_edge in list(out_edges):
                     for in_edge in in_edges:
-                        if out_edge.data.data != in_edge.data.data:
-                            # NOTE: This check does not make any sense, and is in my (@philip-paul-mueller)
-                            #   view wrong. As it will filter out some accesses but not all, which one solely
-                            #   depends on how the memelts were created, i.e. to which container their `data`
-                            #   attribute is associated to. See also [issue #1643](https://github.com/spcl/dace/issues/1643).
-                            continue
                         if in_subsets[in_edge].covers(out_subsets[out_edge]):
                             out_edges.remove(out_edge)
                             break
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -761,13 +761,15 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
         read_set = collections.defaultdict(list)
         write_set = collections.defaultdict(list)
 
+        # NOTE: In a previous version a _single_ read (i.e. leaving Memlet) that was
+        #   fully covered by a single write (i.e. an incoming Memlet) was removed from
+        #   the read set and only the write survived. However, this was never fully
+        #   implemented nor correctly implemented and caused problems.
+        #   So this filtering was removed.
+
         for subgraph in utils.concurrent_subgraphs(self):
             subgraph_read_set = collections.defaultdict(list)  # read and write set of this subgraph.
             subgraph_write_set = collections.defaultdict(list)
-            # Traverse in topological order, so data that is written before it
-            # is read is not counted in the read set
-            # NOTE: Each AccessNode is processed individually. Thus, if an array appears multiple
-            #   times in a path, the individual results are combined, without further processing.
             for n in utils.dfs_topological_sort(subgraph, sources=subgraph.source_nodes()):
                 if not isinstance(n, nd.AccessNode):
                     # Read and writes can only be done through access nodes,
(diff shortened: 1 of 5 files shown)


philip-paul-mueller:
I now modified the test to expect success, since the bug with the `IF` is now solved.
However, the second failure is still there and as mentioend above, I am still unable to fix it.

philip-paul-mueller:
The only test that is still failing is `tests/transformations/move_loop_into_map_test.py::MoveLoopIntoMapTest::test_more_than_a_map` I took a look at it.
The error is that the transformation now applies, but the test expects that transformation not to apply.

The initial SDFG looks as follows.

more_than_map

The SDFG after the transformation looks like this

more_than_map_after


The map is now outside and the loop is inside a nested SDFG.
After some thinking I do not see a reason why this is not valid to do.
The mean reason is that before the map runs, the content of `out` is overwritten by `A`.
Further, we do not write directly into `out` but first into `tmp`.
Since neither `A` nor `B` is overwritten, the map will produce the same in every loop iteration.

I think that the test itself is wrong.
The description also says that "possible WR conflict".
So I propose the following things:
- We change the test such that it expects that the transformation applies.
- We modify the SDFG such that the transformation becomes invalid.

In either case I added code to ensure that the SDFG outputs the same before and after the transformation.
Which is the case for me locally.

What are you guys thinking?

phschaad:
The proposal on the second test seams reasonable to me, it appears to me too, that there is nothing wrong with applying the transformation there. And already having `test_more_than_a_map_1` means we do not need to include the inverse of the test.

dace/sdfg/state.py
```diff
@@ -745,51 +745,90 @@ def update_if_not_none(dic, update):
...
+                        sbs.Range.from_array(ac_desc)
+                        if out_edge.data.src_subset is None
+                        else out_edge.data.src_subset
+                    )
+
+                # If a memlet reads a particular region of data from the access node and there
+                #  exists a memlet at the same access node that writes to the same region, then
+                #  this read is ignored, and not included in the final read set, but only
+                #  accounted fro in the write set. See also note below.
+                # TODO: Handle the case when multiple disjoint writes are needed to cover the read.
```
phschaad, line 811: I assume from the code that the failure mode here is 'conservative', i.e., we overapproximate something as a read if we have disjoint writes that would cover it?

dace/sdfg/state.py
```diff
@@ -745,51 +745,90 @@ def update_if_not_none(dic, update):
...
-                    for e in out_edges:
-                        # skip empty memlets
-                        if e.data.is_empty():
-                            continue
-                        rs[n.data].append(e.data.subset)
-            # Union all subgraphs, so an array that was excluded from the read
-            # set because it was written first is still included if it is read
-            # in another subgraph
-            for data, accesses in rs.items():
+            # TODO: This only works if every data descriptor is only once in a path.
```
phschaad, line 769: What is the failure mode here if it appears multiple times?

dace/sdfg/state.py
philip-paul-mueller: Yes.
To be clear, lets say `e1` writes `0:10` and `e1` writes `10:20`, but at the same node `e3` reads `5:15` then it would still be considered as a read and a write and not just as a write.
In that sense it is an improvement.

I made the description a bit clearer.

dace/sdfg/state.py
philip-paul-mueller: It took my some time to understand it again, essentially each access node is processed individually and all are combined.
An example: Assume the first time the data is encountered the range `0:10` is written and read, so the read set is empty and the write set is `0:10` for that AN.
The second time, the subset `10:20` is written and `2:6` is read, therefore the read set is `2:6` and the write set is `10:20`.
The final write set is `{0:10, 0:20}` and the read set is `{0:6}`.
However, in this case it could the read set could be determined as `{}`.

But the more I think about this could actually catch some edge case were this would be incorrect.
In the end it is an over approximation.

I turned the TODO into a NOTE and added a better description.

philip-paul-mueller:
I updated the PR.
I mostly clarified the some comments.

Regarding the test:
- I changed the test to expect that it works.
- I added an additional test, with a different Memlet, this time it can not be applied.

phschaad (approved):
Thank you for the comments and changes! Looks good to me now if tests pass :-)

philip-paul-mueller:
The tests do not pass.
The error is probably the same that is mentioned in [issue#1643](https://github.com/spcl/dace/issues/1643).

phschaad (changes requested):
As per discussion

philip-paul-mueller:
This is something I have posted in the Matermost channel before.
It is replicated here for completeness:

I had some time so I trying to fix the last error inside the read and write sets, I have traced the error back to `RefineNestedAccess`.
I now found the minimum example.

Nested_min_example

In red you can see the initial subsets and in green what has changed (if nothing has changed it is still red).

I have conformed that the problem is that `T1` appear twice.
According to the definition of read and write sets the first `T1` AccessNode is not considered as a read node since everything that is read is also written.
Because of the second AccessNode for `T1` it is considered as a write access.

I confirmed that if the top `T1` is replaced by another array, let's say `T2` then it works.
Furthermore, if the filtering is disabled (i.e. `T1` is part of the read set) it works as well.

I have come up with the following proposals.

1. Considering the fact that the filtering never really worked correctly in the first place (since we get the errors _after_ I fixed them), I would propose to scrap it, without any replacement.
2. If the filtering should be kept, then we must precisely define what happens in cases such as above, i.e. there is a path in a subgraph that visits a data container multiple times.

If we want to keep the filtering, then we will have to fully define the behaviour of the function.
Since even now the function does not address all cases, such as 
- Multiple memlets are needed to cover a read (for we read the range `5:15`; However, Memlet 1 writes `0:10` and Memlet 2 writes `10:20`, thus the combination of both Memlets is needed to filter the range).
- What happens if the a data container appears multiple time in a subgraph? (this is our problem.)
- What happens if a data container appear multiple time, but each time in a different (concurrent) component?

From the above suggestions I would choose the first one, i.e. removing the filtering.
The main reason is, this _is_ the traditional defacto behaviour.

philip-paul-mueller:
To see what happens I have now disabled the mentioned filtering.
This was done in commit [b546b07](https://github.com/spcl/dace/pull/1678/commits/b546b0742104464de08a3ee2840c073be2a52605)

philip-paul-mueller:
I have now fixed the tests that were failing.
Each test is fixed in its own commit and in the description you can found a justification for this.

dace/sdfg/state.py
```diff
@@ -745,51 +745,82 @@ def update_if_not_none(dic, update):
...
+
     def _read_and_write_sets(self) -> Tuple[Dict[AnyStr, List[Subset]], Dict[AnyStr, List[Subset]]]:
         """
         Determines what data is read and written in this subgraph, returning
         dictionaries from data containers to all subsets that are read/written.
         """
+        from dace.sdfg import utils  # Avoid cyclic import
+
+        # Ensures that the `{src,dst}_subset` are properly set.
+        #  TODO: find where the problems are
```
phschaad, line 757: This is a very ugly hack, but I can replicate / encounter the issue too. I am fine with leaving this in, but please make sure there is an issue that keeps track of this TODO somewhere.

phschaad (approved):
This looks good to me, please see the comment in code.

PR #1682, opened 2024-10-15
Add host_map and host_data properties to control which maps and array…

ThrudPrimrose:
In the preprocessing step of velocity tendencies I have noticed that I need to prevent some stuff from being offloaded to the GPU. I suggest the following idea as a proposal.

(`to_gpu_transformations` always maps transient scalars to GPU even if it is not used in any map or used in interstate edge. This also let's avoid those errors)

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -1202,7 +1202,7 @@ class Scalar(Data):
                  host_data=False):
         self.allow_conflicts = allow_conflicts
         shape = [1]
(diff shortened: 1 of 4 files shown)


tbennun:
I would not do it this way. Instead, we mark arrays to skip in the transformation. For maps, you can probably add an option to not convert any nondefault schedule and it won't convert it, but add it to the transformation. These are not attributes of the nodes.

ThrudPrimrose:
I will create a new PR from a new branch according to the suggestions, as starting from  fresh master is easier than everything everything in this case

PR #1691, opened 2024-10-16
Fix a bug in `RedundantArray`

### LINKED ISSUE #1690: `RedundantArray` can crash during `compile()` call because the memlet path update is wrong in some cases.
**Describe the bug**

Demo SDFG: https://polybox.ethz.ch/index.php/s/EFQ4e4Kl4L1va0I

Applying `RedundantArray` on this graph a single time updates the memlet path from `A` to the map entry with data `c[i, j] -> [i, j]`. This broken graph passes `validate()` call (which it should not), but crashes during `compile()` call.

**To Reproduce**

(After appropriate prelude to import libraries and going to the right directory)
```python
g = SDFG.from_file('redundant-array-0.sdfg'))
g.apply_transformations(RedundantArray)
```
This crashes.

**Expected behavior**

It should produce a compilable SDFG.

pratyai:
Fixes #1690.

The problem was mishandling of the memlet path. Unfortunately, the difference is quite difficult to spot, since the vscode UI does not show it correctly.

### DIFF
--- a/dace/transformation/dataflow/redundant_array.py
+++ b/dace/transformation/dataflow/redundant_array.py
@@ -24,7 +24,7 @@ from dace.transformation import transformation as pm
 def _validate_subsets(edge: graph.MultiConnectorEdge,
                       arrays: Dict[str, data.Data],
                       src_name: str = None,
-                      dst_name: str = None) -> Tuple[subsets.Subset]:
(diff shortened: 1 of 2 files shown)


pratyai:
#1692 covers this already, so not needed anymore.

PR #1692, opened 2024-10-17
Fix `cpp.reshape_strides()`, which currently cannot handle inputs that tries to increase the number of dimensions after the reshape.

### LINKED ISSUE #1690: `RedundantArray` can crash during `compile()` call because the memlet path update is wrong in some cases.

pratyai:
Fix the issue with cpp codegen, where it currently cannot handle inputs:
```c++
cpp.reshape_strides(Range([(0, 4, 1), (0, 5, 1)]), None, None, [2, 3, 5])
```
and crashes with an index error.

Also fixes #1690 where `RedundantArray` was producing a (valid) graph that triggered this.


acalotoiu (commented):
LGTM -please add the test as discussed

pratyai:
@acalotoiu,
- The commit ec029bb introduces a test that verifies that there is an exception raised when compiling the transformed graph.
- There are also additional unit tests that covers various cases (I could add even more if we want to be really through).
- The fix commit 2960b9f fixes the compiling problem, but now it verifies that even now the transformed graph is not correct.
- Thanks for insisting on having this case written down. @alexnick83 pointed out earlier that my original input graph (which I started the bug with) had a couple of memlet's ill-defined. So, now I have come up with a slightly different graph to trigger the problem again. Perhaps you would like to see the input graph to confirm that it is indeed a valid SDFG, even if it looks contrived.
```python
    def original_graph_with_redundant_array():
        g = SDFG('prog')
        g.add_array('A', (5, 5), dace.float32)
        g.add_array('b', (1,), dace.float32, transient=True)
        g.add_array('c', (5, 5), dace.float32, transient=True)

        st0 = g.add_state('st0', is_start_block=True)
        st = st0

        # Make a single map that copies A[i, j] to a transient "scalar" b, then copies that out to a transient array
        # c[i, j], then finally back to A[i, j] again.
        A = st.add_access('A')
        en, ex = st.add_map('m0', {'i': '0:1', 'j': '0:1'})
        st.add_memlet_path(A, en, dst_conn='IN_A', memlet=Memlet(expr='A[0:1, 0:1]'))
        b = st.add_access('b')
        st.add_memlet_path(en, b, src_conn='OUT_A', memlet=Memlet(expr='A[i, j] -> b[0]'))
        c = st.add_access('c')
        st.add_memlet_path(b, c, memlet=Memlet(expr='b[0] -> c[i, j]'))
        st.add_memlet_path(c, ex, dst_conn='IN_A', memlet=Memlet(expr='c[i, j] -> A[i, j]'))
        A = st.add_access('A')
        st.add_memlet_path(ex, A, src_conn='OUT_A', memlet=Memlet(expr='A[0:1, 0:1]'))
        st0.fill_scope_connectors()

        g.validate()
        g.compile()
        return g
```

tests/codegen/targets/cpp_test.py
```diff
@@ -0,0 +1,196 @@
...
+from operator import mul
+
+import dace
+from dace import SDFG, Memlet
+from dace.codegen.targets import cpp
+from dace.subsets import Range
+from dace.transformation.dataflow import RedundantArray
+
+
+class ReshapeStrides(unittest.TestCase):
```
tbennun, line 12: please use pytest as we use it in the other tests (i.e., as free functions), and place the file in tests/codegen.

tests/codegen/targets/cpp_test.py
```diff
@@ -0,0 +1,196 @@
+import unittest
```
tbennun, line 1: Missing copyright notice.

tests/codegen/targets/cpp_test.py
```diff
@@ -0,0 +1,196 @@
...
+        g.compile()
+        return g
+
+    def test_removal(self):
+        g = self.original_graph_with_redundant_array()
+        g.apply_transformations(RedundantArray)
+        g.validate()
+        g.compile()
+
+        # TODO: The produced graph still has bug. So, let's test for its existence.
```
tbennun, line 184: Why is there a TODO in the test

tests/codegen/targets/cpp_test.py
```diff
@@ -0,0 +1,196 @@
...
+from dace.transformation.dataflow import RedundantArray
+
+
+class ReshapeStrides(unittest.TestCase):
+    def test_multidim_array_all_dims_unit(self):
+        r = Range([(0, 0, 1), (0, 0, 1)])
+
+        # To smaller-sized shape
+        target_dims = [1]
+        self.assertEqual(reduce(mul, r.size_exact()), reduce(mul, target_dims))
```
tbennun, line 18: you should use the pytest `assert reduce(...) == reduce(...)`

tbennun (changes requested):

tests/codegen/targets/cpp_test.py
pratyai: Done.

tests/codegen/targets/cpp_test.py
pratyai: Done.

tests/codegen/targets/cpp_test.py
pratyai: Replaced the text TODO with a NOTE. However, as I commented, the produced graph is still wrong because of a different bug. Since I don't address that here, I just wanted to highlight the problem anyway.

tests/codegen/targets/cpp_test.py
pratyai: Done.

Just curious: Is there a downside of `unittest` or is it just to be consistent with the rest of the code?
As far as I can tell, they are compatible with `pytest`, offers better organisation of code, and the better error messages. Then again, I'm not very familiar with python testing frameworks, so perhaps I've missed some disadvantages.

tbennun (approved):

PR #1697, opened 2024-10-20
fix(codegen/prettycode): Use base_indentation as intended

iBug:
The diff should be self-explanatory


tbennun:
@iBug Thank you for the PR. Where does the issue manifest itself in the generated code?

iBug:
I'm working on a local fork and find this parameter useful except for not being effective. I thought it was intended like so, judging by variable naming, so I figured I'd send a PR for that.

tbennun (approved):

PR #1702, opened 2024-10-23
Add a `SubrangeMapper` helper class which maps a _subrange_ of `src` range to its counterpart in `dst` range, if possible.

pratyai:
Equipped with a `src` and a `dst` range of equal volumes but possibly different shapes or even dimensions (i.e.,
there is an 1-to-1 correspondence between the elements of `src` and `dst`), maps a subrange of `src` to its
counterpart in `dst`, if possible (which is not always the case).

It remains dead-code in this PR and not used anywhere at all. But it should help with correctly reshaping memlet subsets in a following PR (which will aim to fix `RedundantArray` transform by simplying it to something like this: https://github.com/pratyai/dace/blob/be571d21aa8abbc465daa11d04941040fbf93401/dace/transformation/dataflow/redundant_array.py#L426-L480).

### DIFF
--- a/dace/subsets.py
+++ b/dace/subsets.py
@@ -1393,6 +1393,9 @@ class SubrangeMapper:
         while src_i < self.src.dims():
             assert dst_i < self.dst.dims()
 
+            # Find the next smallest segments of `src` and `dst` whose volumes matches (and therefore can possibly have
+            # a mapping).
+            # TODO: It's possible to do this in a O(max(|src|, |dst|)) loop instead of O(|src| * |dst|).
             src_j, dst_j = None, None
             for sj in range(src_i + 1, self.src.dims() + 1):
                 for dj in range(dst_i + 1, self.dst.dims() + 1):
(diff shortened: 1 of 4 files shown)


dace/subsets.py
```diff
@@ -895,6 +909,10 @@ def size(self):
     def size_exact(self):
         return self.size()
 
+    def volume_exact(self) -> int:
+        """ Returns the total number of elements in all dimenssions together. """
```
alexnick83, line 913: Typo in dimensions (double s).

dace/subsets.py
```diff
@@ -895,6 +909,10 @@ def size(self):
...
+        return reduce(operator.mul, self.size_exact())
```
alexnick83, line 914: Is this compatible with all supported Python versions? Note that we also have another method that does the same in `dace.data`: https://github.com/spcl/dace/blob/057a6804ea2da60b053895e490cf230d0ef90225/dace/data.py#L139
Maybe it is a good idea to move (one of them) to a utility method.

alexnick83 (commented):
`SubrangeMapper.map` looks a bit convoluted but seems like a useful method overall. I will re-review when the tests pass.

dace/subsets.py
pratyai: Fixed.

dace/subsets.py
pratyai: As far as I can tell, it does work for 3.7 and 3.12. Replaced with lambda function regardless, since that's how `_prod` works.

I don't think it's necessary to make `reduce(lambda a, b : a * b, sequence)` itself an utility function. It's really short, easy to see what it does. I've added the `volume_exact()` function to just give that operation a meaningful name (i.e. why it does that). But of course, if you rather have it separately, I can move it too.

pratyai:
> `SubrangeMapper.map` looks a bit convoluted but seems like a useful method overall.

I can try and rephrase the comments. If you have any suggestion on phrasing, please let me know.
The gist is: We were writing to some range in `A`, and then some superset of that range was written out to `B`. Now that we're removing `A` and will directly write to `B` instead, what should be the range for the new write.
I'll also do a cleanup later with the comments, removing any garbage print statements etc.

> I will re-review when the tests pass.

The unit tests pass. For the CI tests, please check the last commit to see what I'm aiming for (note: this commit is incomplete, but is there just to demonstrate that with a few additional bug fixes, redundant array can be simplified a lot while at the same time correcting its behaviour).

I'll try to trigger a CI run later.

PR #1704, opened 2024-10-24
Deferred Allocation

ThrudPrimrose:
This is possibly the start of a long PR that will evolve, and we should discuss it at every step. 

The main idea is as follows (so far in the prototype):

Idea 1:
When an array is created, let's say "A", then also a size array is created "A_size". The Size array is always one-dimensional and contiguously stored. When there is an A_size -> A (IN connector needs to be _size), we trigger a reallocation operation and not a copy operation. I am not sure if we should have an A_size array, I would like to discuss this.

I have a small SDFG that shows what it looks like.
```
import dace

def one():
    sdfg = dace.sdfg.SDFG(name="deferred_alloc_test")

    sdfg.add_array(name="A", shape=("__dace_defer", "__dace_defer"), dtype=dace.float32, storage=dace.dtypes.StorageType.Default, transient=True)

    state = sdfg.add_state("main")

    an_1 = state.add_access('A')
    an_1.add_in_connector('IN_size')

    an_2 = state.add_scalar(name="dim0", dtype=dace.uint64)
    an_3 = state.add_scalar(name="dim1", dtype=dace.uint64)

    s_an_1 = dace.nodes.AccessNode(data="A_size")
    state.add_node(s_an_1)
    s_an_2 = dace.nodes.AccessNode(data="A_size")
    state.add_node(s_an_2)
    state.add_edge(s_an_1, None, s_an_2, None,
                dace.Memlet(None) )

    state.add_edge(s_an_2, None, an_1, 'IN_size',
                dace.Memlet(expr="A_size[0:2]") )

    state.add_edge(an_2, None, s_an_1, None,
                dace.Memlet(expr="A_size[0]") )
    state.add_edge(an_3, None, s_an_2, None,
                dace.Memlet(expr="A_size[1]") )


    sdfg.save("def_alloc_1.sdfg")
    sdfg.validate()

    return sdfg

s = one()
s.generate_code()
s(dim0=512, dim1=512)
```
Generated code looks like follows:
image
And SDFG:
image

Next Step:
Not 100% of this array will always be used; for example, when calling realloc, only the "__dace_defer" dimensions will be read from the array. And the dimensions that are passed using symbols will be read from the respective symbols (this can't be changed throughout the SDFG). (I am currently working on this step)


tbennun:
I would not use IN_size (i.e., a throughput connector), maybe use just `size`.

ThrudPrimrose:
> I would not use IN_size (i.e., a throughput connector), maybe use just `size`.

About the connector, we can also read the size, then this will be "OUT_size" I had planned. Would still suggest both connectors to be named size?

tbennun:
> > I would not use IN_size (i.e., a throughput connector), maybe use just `size`.
> 
> About the connector, we can also read the size, then this will be "OUT_size" I had planned. Would still suggest both connectors to be named size?

yes, exactly. Since those connectors are endpoints, and not flowing _through_ the access node (as they would in a scope node).

ThrudPrimrose:
> > > I would not use IN_size (i.e., a throughput connector), maybe use just `size`.
> > 
> > 
> > About the connector, we can also read the size, then this will be "OUT_size" I had planned. Would still suggest both connectors to be named size?
> 
> yes, exactly. Since those connectors are endpoints, and not flowing _through_ the access node (as they would in a scope node).

I was updating the implementation according to a discussion we had with Torsten and Lex. I also changed the implementation to use "size" for both in and out connectors. Now the problem is that this SDFG does not validate because the access node has duplicate connectors. I think it is better to make the size connectors distinct rather than changing the validation rules. Or should I update the validation procedures with access nodes being an exception?

dace/codegen/targets/cpu.py
```diff
@@ -1098,15 +1209,23 @@ def process_out_memlets(self,
                             write_expr = f"*({ptr_str} + {array_expr}) = {in_local_name};"
                         else:
                             desc_dtype = desc.dtype
-                            expr = cpp.cpp_array_expr(sdfg, memlet, codegen=self._frame)
+                            # If the storage type if CPU_Heap or GPU_Global then it might be requiring deferred allocation
+                            # We can check if the array requires sepcial access using A_size[0] (CPU) or __A_dim0_size (GPU0)
+                            # by going through the shape and checking for symbols starting with __dace_defer
+                            deferred_size_names = self._get_deferred_size_names(desc, memlet)
+                            expr = cpp.cpp_array_expr(sdfg, memlet, codegen=self._frame, deferred_size_names=deferred_size_names)
```
alexnick83, line 1216: To make the API cleaner, would it make sense to move `_get_deferred_size_names` inside `cpp` or `cpp.cpp_array_expr`?

dace/sdfg/sdfg.py
```diff
@@ -496,6 +500,7 @@ def __init__(self,
         self._parent_sdfg = None
         self._parent_nsdfg_node = None
         self._arrays = NestedDict()  # type: Dict[str, dt.Array]
+        self._arrays = NestedDict()
```
alexnick83, line 503: double line

dace/sdfg/sdfg.py
```diff
@@ -418,6 +418,10 @@ class SDFG(ControlFlowRegion):
                        desc="Data descriptors for this SDFG",
                        to_json=_arrays_to_json,
                        from_json=_nested_arrays_from_json)
+    _arrays = Property(dtype=NestedDict,
```
alexnick83, line 421: Is this used anywhere?

dace/frontend/python/newast.py
```diff
@@ -3970,7 +3970,7 @@ def _parse_sdfg_call(self, funcname: str, func: Union[SDFG, SDFGConvertible], no
         # Change transient names
         arrays_before = list(sdfg.arrays.items())
         for arrname, array in arrays_before:
-            if array.transient and arrname[:5] == '__tmp':
+            if array.transient and arrname[:5] == '__tmp' and arrname not in sdfg.size_arrays():
```
alexnick83, line 3973: Is there any case where a size array can have a `__tmp` name instead of `__dace_deferred`?

dace/sdfg/sdfg.py
```diff
@@ -740,17 +752,33 @@ def replace_dict(self,
...
+                    if new_name != "__return":
+                        size_desc_map[new_name] = new_name + "_size"
+                        _replace_dict_keys(self.arrays, name + "_size", new_name + "_size", size_arrays)
                     _replace_dict_keys(self.symbols, name, new_name)
                     _replace_dict_keys(self.constants_prop, name, new_name)
                     _replace_dict_keys(self.callback_mapping, name, new_name)
                     _replace_dict_values(self.callback_mapping, name, new_name)
 
+        # Update size descriptors
+        # Return_size break things delete it from the arrays
```
alexnick83, line 775: Can `return` be a deferred array?

dace/sdfg/sdfg.py
```diff
@@ -1676,11 +1709,13 @@ def _find_new_name(self, name: str):
         """ Tries to find a new name by adding an underscore and a number. """
 
         names = (self._arrays.keys() | self.constants_prop.keys() | self._pgrids.keys() | self._subarrays.keys()
-                 | self._rdistrarrays.keys() | self.symbols.keys())
+                 | self._rdistrarrays.keys() | self.symbols.keys() | self._arrays.keys())
```
alexnick83, line 1712: Why are array keys included twice?

dace/sdfg/sdfg.py
```diff
@@ -1676,11 +1709,13 @@ def _find_new_name(self, name: str):
...
         return dt.find_new_name(name, names)
 
     def is_name_used(self, name: str) -> bool:
         """ Checks if `name` is already used inside the SDFG."""
+        if name in self._arrays:
```
alexnick83, line 1717: Double lines?

dace/sdfg/sdfg.py
```diff
@@ -2047,18 +2100,35 @@ def add_datadesc(self, name: str, datadesc: dt.Data, find_new_name=False) -> str
...
+
+        if (
+            datadesc.transient is True and
+            isinstance(datadesc, dt.Array) and
+            name != "__return"
+            ):
+            size_desc_name = f"{name}_size"
+            size_desc = dt.Array(dtype=dace.uint64,
+                                shape=(len(datadesc.shape),),
+                                storage=dtypes.StorageType.Default,
```
alexnick83, line 2115: Shouldn't this always be CPU Heap? Default can also be resolved to GPU Global depending on the context, unless special care is taken in code generation and GPU transform. I think you have done so, but maybe it is better to make it clear here too.

dace/sdfg/state.py
```diff
@@ -421,7 +421,8 @@ def memlet_path(self, edge: MultiConnectorEdge[mm.Memlet]) -> List[MultiConnecto
             # Trace through scope entry using IN_# -> OUT_#
             if isinstance(curedge.dst, (nd.EntryNode, nd.ExitNode)):
                 if curedge.dst_conn is None:
-                    raise ValueError("Destination connector cannot be None for {}".format(curedge.dst))
+                    #raise ValueError("Destination connector cannot be None for {}".format(curedge.dst))
```
alexnick83, line 424: Is there a reason for removing this check?

dace/sdfg/validation.py
```diff
@@ -325,7 +325,7 @@ def _accessible(sdfg: 'dace.sdfg.SDFG', container: str, context: Dict[str, bool]
     """
     Helper function that returns False if a data container cannot be accessed in the current SDFG context.
     """
-    storage = sdfg.arrays[container].storage
+    storage = sdfg.arrays[container].storage if container in sdfg.arrays else sdfg.arrays[container].storage
```
alexnick83, line 328: This line is confusing. Is the else branch correct?

dace/sdfg/validation.py
```diff
@@ -814,10 +901,11 @@ def validate_state(state: 'dace.sdfg.SDFGState',
 
         # Check dimensionality of memory access
         if isinstance(e.data.subset, (sbs.Range, sbs.Indices)):
-            if e.data.subset.dims() != len(sdfg.arrays[e.data.data].shape):
+            desc = sdfg.arrays[e.data.data] if e.data.data in sdfg.arrays else sdfg.arrays[e.data.data]
```
alexnick83, line 904: Confusing else branch

tests/deferred_alloc_test.py
```diff
@@ -0,0 +1,222 @@
...
+                                            (dace.dtypes.StorageType.GPU_Global, dace.dtypes.ScheduleType.GPU_Device)]:
+            print(f"Trivial Realloc with storage {storage_type} on transient:{transient} data")
+            test_trivial_realloc(storage_type, transient)
+            print(f"Trivial Realloc-Use with storage {storage_type} on transient:{transient} data")
+            test_realloc_use(storage_type, transient, schedule_type)
+
+    print(f"Realloc with incomplete write 1")
+    test_incomplete_write_dimensions_1()
+    print(f"Realloc with incomplete write 2")
+    test_incomplete_write_dimensions_2()
```
alexnick83, line 222: Missing newline

alexnick83 (commented):
LGTM overall. I have some questions but ignore the one that was already discussed in the chat.

dace/codegen/dispatcher.py
```diff
@@ -609,6 +629,20 @@ def dispatch_copy(self, src_node: nodes.Node, dst_node: nodes.Node, edge: MultiC
         self._used_targets.add(target)
         target.copy_memory(sdfg, cfg, dfg, state_id, src_node, dst_node, edge, function_stream, output_stream)
 
+    def dispatch_reallocate(self, src_node: nodes.Node, node: nodes.Node, edge: MultiConnectorEdge[Memlet], sdfg: SDFG,
+                      cfg: ControlFlowRegion, dfg: StateSubgraphView, state_id: int, function_stream: CodeIOStream,
+                      output_stream: CodeIOStream) -> None:
+        state = cfg.state(state_id)
+        target = self.get_reallocate_dispatcher(node, edge, sdfg, state)
+        assert target is not None
+        if target is None:
```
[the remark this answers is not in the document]
ThrudPrimrose: There should not be. And yes, I had copied the dispatch copy.
I will only keep the assert

dace/codegen/targets/cpu.py
```diff
@@ -650,6 +690,42 @@ def copy_memory(
...
+        data_name = dst_node.data
+        new_size_array_name = src_node.data
+
+        data = sdfg.arrays[data_name]
+        size_array_name = data.size_desc_name
+
+        dtype = sdfg.arrays[data_name].dtype
+
+        # Only consider the offsets with __dace_defer in original dim
+        mask_array = [str(dim).startswith("__dace_defer") for dim in data.shape]
```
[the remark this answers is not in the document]
ThrudPrimrose: To keep it simple, I think it is better to allow only "__dace_defer" as an atomic symbol and never as a part of an expression.
I will add this to the validation; thanks for pointing it out.

dace/sdfg/sdfg.py
ThrudPrimrose: I have independently noticed this. I decided to go with CPU_Heap. (But the codegen always allocates size arrays on CPU_Heap. I did not go with Registers due to the dangers of them considered accessible in GPU kernels).

dace/codegen/targets/cpu.py
ThrudPrimrose: I moved this function call to cpp as much as I can and also try to hide it as much as I can from function signatures.

dace/codegen/targets/cpu.py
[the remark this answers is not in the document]
ThrudPrimrose: I might still need to think of a fix on this.

dace/sdfg/state.py
ThrudPrimrose: I probably forgot it when I first trying to implement where size arrays and arrays had different containers. Must have forgotten it like that.

I reran some tests locally, does not seem to cause an issue.

dace/sdfg/validation.py
ThrudPrimrose: Same, fixed.

tests/deferred_alloc_test.py
ThrudPrimrose: Fixed.

dace/codegen/targets/cpu.py
```diff
@@ -336,6 +338,14 @@ def declare_array(self,
...
+
+            # Size desc is defined only for transient arrays
+            if nodedesc.transient and nodedesc.storage == dtypes.StorageType.CPU_Heap:
+                size_desc_name = sdfg.arrays[name].size_desc_name
+                if size_desc_name is not None:
+                    size_desc = sdfg.arrays[size_desc_name]
+                    size_ctypedef = dtypes.pointer(size_desc.dtype).ctype
+                    self._dispatcher.declared_arrays.add(size_desc_name, DefinedType.Pointer, size_ctypedef)
             return
         elif nodedesc.storage is dtypes.StorageType.CPU_ThreadLocal:
```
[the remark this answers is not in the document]
ThrudPrimrose: I added a lot of validation checks on the matter.

ThrudPrimrose:
I want to share the design document.

With the proposal, we support dynamic allocation and reallocation of GPU_Global and CPU_Heap arrays.
The only supported DaCe type for reallocation is dace.data.Array type. 

On the CPU_Heap array, reallocation is performed through a call to realloc; on GPU_Global storage, it is implemented through a sequence of malloc, copy, free as CUDA does not support realloc. 

Reallocation is only allowed on the host-side code and can be triggered when the scope is None. (Reallocation inside a map or nested SDFGs is, for example, impossible, as realloc / malloc are not usually thread-safe, and it would not have good performance if it had a thread-safe implementation.)

A deferred array is generated only upon the request of the user. This can be done by including the symbol __dace_defer in any of the expressions in the shape of the array when it is added. It is invalid to reshape an array that is deferred. If the expression has multiple appearances of the __dace_defer symbol, it is assumed to be the same symbol.

The array size is tracked by a unique size_array connected to a deferred array. It is stored in the _arrays dictionary of the SDFG, just like other arrays—the constructor of dace.data.Array sets is_size_arary to true for size arrays and is_deferred_array to true for deferred arrays. The members are set only for arrays, not other DaCe data types. The size_array of an array is tracked through the size_desc_name variable of an array. The size array is created by appending the  _size suffix to the end of the array’s name.

The size array is always one-dimensional. Its length matches the number of dimensions (length of the shape) of the deferred array. No size array is created if the array is not deferred.
The size array copies the dimensions that do not have __dace_defer as initial values of the size array. If an array has the shape A[2*N, 4*__dace_defer], then the size array is initialized as A_size[2]{2*N, 0}. The dimensions that do not have __dace_defer are not accessed by the codegen, yet they are written in the size array to allow the user to access them. When an offset expression is computed

The reallocation is triggered by writing to the special _write_size in the connector of an access node. As part of reallocation, the user provides only the size of the __dace_defer, and the new value of the __dace_defer symbol is used to compute the new to the size_array. If the new shape (of dimension 1) written by the user is 5, then for the A above,  the latest value of the __dace_defer will be written as 5, and the size will be A_size[1] = 5. But the dimension of A is [2*N][4*5], any call to the existing functions from the cpp module still calculates the dimensions using the shape member of the array. To support __dace_defer symbol and deferred allocation, some functions accept a variable called deferred_size_names  or automatically generate these parameters by detecting __dace_defer symbols in the shape. The offset expressions are matched against the __dace_defer_dim(d+) pattern and switched with accesses to the size_array on the host to process. On GPU kernels, the pattern matching is slightly different. 

The length of the write to the _write_size in the connector always needs to match the size of the array's shape. The values on the dimensions that do now have _dace_defer in their expressions are ignored.

The size of the array can be read from the _read_size out connector of an access node and used in maps.

Even though the size arrays are allocated on the stack, they require a call to cudamemcpy to transfer the size array to GPU to be transferred as pointers. To mitigate the issue (and have a more performant implementation than needing one more memcpy before every kernel), the size array is unpacked into integers. The name mangling pattern is as follows: the current value of the DimensionId’th _dace_defer symbol from the array A is read from A_size[i] and mangled as  __<ArrayName>_dim<DimensionId>_size for the example, name mangled from A_size[1]  would be __A_dim1_size , the integer unpacked sizes of deferred symbols are passed as integers to the kernel. The function _get_deferred_size_names method of the cpp module handles the generation of mangled names for array and offset accesses, and the CUDA codegen handles unpacking and instantiating mangled variable (integer) names.

My concerns:
_size names are misleading? Maybe I should mean that is _deferred_symbol_size?

ThrudPrimrose:
The bold text is not copied over I have google doc for the same purpose:

https://docs.google.com/document/d/1fBinC5d0gpBnYD9C4M3e0zyxVkGD94LZFzzTCydM2sQ/edit?usp=sharing

PR #1708, opened 2024-10-25
Modified `SDFGState.unordered_arglist()`

philip-paul-mueller:
This PR fixes the way how arguments are detected in scopes.

Technically this only affects GPU code generation, but it is a side effect of how the code is generated.
In GPU mode a `Map` is translated into one kernel, for this reason a signature must be computed (this is the reason why CPU code generation is not affected, no function call is produced).
To compute this signature the `unsorted_arglist()` function scans what is needed.
However, this was implemented not correctly.
Assume that AccessNode for array `A` is outside the map and inside the map a temporary scalar, `tmp_in` is defined and initialized to `tmp_in = A[__i0, __i1]`, see also this image:

argliost_situation

If the `data` property of the Memlet that connects the MapEntry with the AccessNode for `tmp_in` is referencing `A` then the (old) function would find that `A` is needed inside, although there is no AccessNode for `A`  inside the map.
If however, this Memlet referrers `tmp_in` (which is not super standard, but should be allowed), then the old version would not pick up.
This would then lead to a code generation error.

This PR modifies the function such that such cases are handled.
This is done by following all edges that are adjacent to the MapEntry (from the inside) to where the are actually originate.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -818,6 +818,8 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
         for node in self.nodes():
             if isinstance(node, nd.AccessNode):
                 descs[node.data] = node.desc(sdfg)
+                # NOTE: In case of multiple nodes of the same data this will
+                #   override previously found nodes.
                 descs_with_nodes[node.data] = node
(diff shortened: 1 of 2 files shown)


alexnick83 (approved):
Great catch!

PR #1709, opened 2024-10-25
Fix a bug in the `ExpandReducePure`

pratyai:
It was producing wrong indices for the initialization kernel, which would not work for some simple valid SDFGs (see the demo in the test).

### DIFF
--- a/dace/libraries/standard/nodes/reduce.py
+++ b/dace/libraries/standard/nodes/reduce.py
@@ -103,7 +103,7 @@ class ExpandReducePure(pm.ExpandTransformation):
                 'reduce_init', {'_o%d' % i: '0:%s' % symstr(d)
                                 for i, d in enumerate(outedge.data.subset.size())}, {},
                 '__out = %s' % node.identity,
-                {'__out': dace.Memlet.simple('_out', ','.join(['_o%d' % i for i in range(output_dims)]))},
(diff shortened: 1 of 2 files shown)


tests/library/reduce_test.py
```diff
@@ -1,35 +1,82 @@
...
+                strides=(C_out * H * W, C_out * W, C_out, 1))
+
+    st0 = g.add_state('st0', is_start_block=True)
+    st = st0
+
+    A = st.add_access('A')
+    C = st.add_access('C')
+    R = st.add_reduce('lambda x, y: x + y', [1, 2, 3], 0)
+    st.add_memlet_path(A, R, memlet=Memlet(expr='A[0:N, 0, 0, 0:C_in, 0:C_out]'))
+    st.add_memlet_path(R, C, memlet=Memlet(expr='C[0:N, 5, 5, 0:C_out]'))
```
tbennun, line 26: why use add_memlet_path for a path of length 1? You can just use `add_edge` or the shorthand for no connectors: `add_nedge`

tbennun (changes requested):

tests/library/reduce_test.py
pratyai: > why use add_memlet_path for a path of length 1?

Mostly because that was better (or at all) documented. Replaced now with `add_nedge()`, will also change this way for the other and future PRs.

tbennun (approved):

PR #1696, opened 2024-10-18
No-View Nested SDFGs

tbennun:
This PR implements the change proposed in #1695. See the discussion in that issue for more information as to the change.

- [x] Adapt code generation
- [x] Adapt validation
- [x] Adapt memlet propagation
- [x] Adapt frontends to use same dimensions for all arrays (i.e., no squeezing), potentially the same names too.
- [x] Ensure skipped tests in map_syntax_test.py pass (also fixes issue #650)
- [x] Ensure #1738 is fixed
- [x] ~~Remove `_temp_transients` and the temporary transient counter in the Python frontend~~ (done in #2121)

Summary of current errors:
- [ ] (memprop / Python frontend) By default, memlets are not propagated properly 
- [x] (Python frontend) When used in nested map ranges and control flow regions, views are not propagated inwards to nested SDFGs properly
- [ ] (dealias) Names of symbols and data containers may clash, this case is not handled properly
- [x] (dealias) Renaming symbols back and forth is excessive (?)

This PR will not unify the descriptor repositories unless necessary.

Related to (and contains the fixes of) PR #1611.

Issue #1562, opened 2024-04-26
sdfg.validate goes on infinite recursion with loopy sdfg


tim0s:
**Describe the bug**

For certain (erroneous) SDFGs, ```sdfg.validate()``` does not thrown an exception, instead it crashes with a RecursionError.
This may be undesirable, since it is unclear to the user if the SDFG is erroneous or he should increase the stack size in order to be able to validate the SDFG.

**To Reproduce**

Run the following Python code:

```
import dace
sdfg = dace.SDFG("foo")
state = sdfg.add_state()
mentry_2, mexit_2 = state.add_map("map_2", dict(i="0:9"))
mentry_6, mexit_6 = state.add_map("map_6", dict(i="0:9"))
mentry_8, mexit_8 = state.add_map("map_8", dict(i="0:9"))
state.add_edge(mentry_8, "OUT_0", mentry_6, "IN_0", dace.Memlet(data="bla", subset='0:9'))
state.add_edge(mentry_6, "OUT_0", mentry_2, "IN_0", dace.Memlet(data="bla", subset='0:9'))
state.add_edge(mentry_2, "OUT_0", mentry_6, "IN_0", dace.Memlet(data="bla", subset='0:9'))
sdfg.validate()
```

**Expected behavior**
Either sdfg.validate() throws an exception telling me I formed a loop, or the last add_edge() throws an exception telling me I cannot add this edge because it would form a loop.

tbennun:
Fixed in #1711

Issue #1189, opened 2023-02-15
Compilation error on GPU when using `ScheduleType.GPU_ThreadBlock_Dynamic`


lamyiowce:
**Describe the bug**
Encountered when implementing a Graph Attentional Operator. When using `ScheduleType.GPU_ThreadBlock_Dynamic`, incorrect code gets generated. The error is:
```
(dace14) [jbazinsk@ault24 bugs]$ python compile_error.py
Keeping schedule sequential for  _Mult__map[__i0=0:3]
Changing schedule to TB dynamic:  assign_47_8_map[__i0=0:3]
Traceback (most recent call last):
  File "/users/jbazinsk/miniconda3/envs/dace14/lib/python3.8/site-packages/dace/codegen/compiler.py", line 224, in configure_and_compile
    _run_liveoutput("cmake --build . --config %s" % (Config.get('compiler', 'build_type')),
  File "/users/jbazinsk/miniconda3/envs/dace14/lib/python3.8/site-packages/dace/codegen/compiler.py", line 407, in _run_liveoutput
    raise subprocess.CalledProcessError(process.returncode, command, output.getvalue())
subprocess.CalledProcessError: Command 'cmake --build . --config RelWithDebInfo' returned non-zero exit status 2.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "compile_error.py", line 92, in <module>
    main()
  File "compile_error.py", line 73, in main
    sdfg(node_features=node_features, factor=factor, rows=rows, columns=columns,
  File "/users/jbazinsk/miniconda3/envs/dace14/lib/python3.8/site-packages/dace/sdfg/sdfg.py", line 2289, in __call__
    binaryobj = sdfg.compile()
  File "/users/jbazinsk/miniconda3/envs/dace14/lib/python3.8/site-packages/dace/sdfg/sdfg.py", line 2198, in compile
    shared_library = compiler.configure_and_compile(program_folder, sdfg.name)
  File "/users/jbazinsk/miniconda3/envs/dace14/lib/python3.8/site-packages/dace/codegen/compiler.py", line 233, in configure_and_compile
    raise cgx.CompilationError('Compiler failure:\n' + ex.output)
dace.codegen.exceptions.CompilationError: Compiler failure:
[...]

PR #1711, opened 2024-10-28
Python frontend stability and inline storage specification

tbennun:
The PR adds a new syntax to support inline storage specification with the `@` operator, supporting the following statements: `a = np.ones(M) @ dace.StorageType.CPU_ThreadLocal`.

This PR also fixes multiple minor issues in the Python frontend:
- [x] `WarpTiling` did not respect sequential map schedules
- [x] Non-sequence inputs for `numpy.fill` variants (e.g., `numpy.zeros(N)`)
- [x] NumPy replacement syntax errors would sometimes not have source information
- [x] Fix type inference for nested scopes in Python frontend
- [x] ~~Dynamic map range related issues with simple modifications to the SpMV program~~ Fix deferred to #1696 
- [x] ~~Dynamic thread block scheduling would not pass object to nested functions~~ Fix deferred to future PR, see #1189
- [x] Dynamic thread block scheduling does not support multi-dimensional maps
- [x] Default schedule inference should use dynamic thread blocks if they exist
- [x] Type hints with storage type not being adhered to by the Python frontend
- [x] Validation issue #1562

### DIFF
--- a/dace/codegen/targets/cuda.py
+++ b/dace/codegen/targets/cuda.py
@@ -2215,7 +2215,7 @@ gpuError_t __err = {backend}LaunchKernel((void*){kname}, dim3({gdims}), dim3({bd
                 'auto {param}) {{'.format(fine_grained=('true' if Config.get_bool(
                     'compiler', 'cuda', 'dynamic_map_fine_grained') else 'false'),
                                           bsize=total_block_size,
-                                          kmapIdx=outer_scope.map.params[0],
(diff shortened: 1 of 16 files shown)


phschaad (approved):
Nothing to add, LGTM. Looks like one GPU test fails due to wrongful arglist construction for a nested SDFG. I have encountered a similar issue numerous times on https://github.com/spcl/dace/pull/1676 and fixed it there, but I suspect that the newly merged #1708 also fixed the issue? I have updated with main to verify.

PR #1715, opened 2024-10-30
Reduce dace package dependencies

tbennun:
Makes dace easier to install. DaCe's dependencies are almost all pure Python packages (only PyYAML is not, but it is likely easy to install).


tbennun:
Superseded by #1722

PR #1718, opened 2024-11-01
CI: Update checkout and setup-python actions

romanc:
GitHub Actions workflows use outdated versions of

- actions/checkout
- actions/setup-python

These actions are built for specific node versions, which are now end of life. While the workflows continue to run, GitHub issues a warning (visible in the online interface) and runs them with newer versions of node.

image

Since both, checkout and setup-python, are basic actions, none of the features that DaCe workflows are using changed. We might see slight speedup from out of the box caching added to setup-python in recent versions.

Parent issue: https://github.com/GEOS-ESM/SMT-Nebulae/issues/89


tbennun (approved):
LGTM

PR #1706, opened 2024-10-24
Add CFG to generate_scope in tutorials

ThrudPrimrose:
The DFG is missing from codegen tutorial, resulting with an error in codegen. Adding cfg as a param fixes it.

PR #1720, opened 2024-11-01
Fix broken codegen tutorial

romanc:
Quick fix follow-up from https://github.com/spcl/dace/pull/1706 which left a broken notebook.


tbennun (approved):
Thank you!

PR #1723, opened 2024-11-01
Fortran Frontend and External SDFGs

phschaad:


phschaad:
This has been taken care of through various other PRs

PR #1724, opened 2024-11-02
Various stability improvements and convenience APIs

### LINKED ISSUE #1315: Request for bitshift operators be added to symbols
# Current limitations
Currently, bitshift operators (e.g., `<<` and `>>`) throw the following error when applied to dace symbols:
```
TypeError: unsupported operand type(s) for >>: 'symbol' and 'int'
```
This might be something one would like to do on an iteration variable within a `dace.map`. For instance:
```py
@dace.program
def f(a):
    for i in dace.map[0:16:2]:
        a[i] = i >> 1
```
This function runs under the python interpreter:
```py
arr1 = np.zeros(16, dtype=np.int32)

f.f(arr1)
print(arr1) #[0 0 1 0 2 0 3 0 4 0 5 0 6 0 7 0]
```
But not under DaCe, giving the aforementioned `TypeError`.

# Current workarounds
Currently I have been using the following two workarounds:
```py
@dace.program
def f(a):
    for i in dace.map[0:16:2]:
        a[i] = int(i) >> 1
```
This workaround is a very low-effort workaround which doesn't require much explanation. Simply converting the symbol into an integer causes the code to run. Alternatively, integer division and multiplication _are_ implemented for symbols, so you could alternatively do:
```py
@dace.program
def f(a):
    for i in dace.map[0:16:2]:
        a[i] = i // 2
```
This is the workaround I have been using.

# Motivation
The workarounds are very easy to do, so I would consider this a very low-priority feature request. However, they can end up cluttering code a little bit. The symbol multiplication/division workaround may not be very self-explanatory and will also generate C++/CUDA code which does not use bitshifts. A good compiler should identify this and convert them into bitshifts, but this is not always guaranteed in my experience. The former workaround by converting a symbol to an integer will convert an `int`-represented symbol to a `long long` first via `dace::int64()`, which may be undesirable if doing a left bit shift, for instance.

# Proposed solution
I've not had the time to fully understand the DaCe codebase, nor is that a priority for me right now, but I propose adding an overload for `int`-represented symbols to implement bitshift operators natively on the 32-bit format, without having to first convert to a 64-bit integer and without generating multiplications or divisions in the target language.

tbennun:
* Various minor code generation and runtime fixes
* Minor API improvements to CompiledSDFG, sdfg.view(), and subset offsetting
* Minor memlet propagation fix
* Various simplify pass fixes that pertain to use of views, references, and tasklets with side effects
* Symbolic support for shift and ternary expressions (fixes #1315)
* Pass permissiveness into transformations

### DIFF
--- a/dace/properties.py
+++ b/dace/properties.py
@@ -329,7 +329,7 @@ def make_properties(cls):
         for name, prop in own_properties.items():
             # Only assign our own properties, so we don't overwrite what's been
             # set by the base class
-            if hasattr(obj, name):
+            if hasattr(obj, '_' + name):
                 raise PropertyError("Property {} already assigned in {}".format(name, type(obj).__name__))
             if not prop.indirected:
                 if prop.allow_none or prop.default is not None:
--- a/dace/cli/sdfgcc.py
+++ b/dace/cli/sdfgcc.py
@@ -48,7 +48,7 @@ def main():
         sdfg = SDFGOptimizer(sdfg).optimize()
 
     # Compile SDFG
-    sdfg.compile(outpath)
+    sdfg.compile(outpath, return_program_handle=False)
 
     # Copying header file to optional path
     if outpath is not None:
--- a/dace/cli/sdfv.py
+++ b/dace/cli/sdfv.py
@@ -44,7 +44,11 @@ def view(sdfg: dace.SDFG, filename: Optional[Union[str, int]] = None, verbose: b
         ):
             fd, filename = tempfile.mkstemp(suffix='.sdfg')
             sdfg.save(filename)
-            os.system(f'code {filename}')
+            if platform.system() == 'Darwin':
+                # Special case for MacOS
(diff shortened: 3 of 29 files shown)


dace/properties.py
```diff
@@ -329,7 +329,7 @@ def initialize_properties(obj, *args, **kwargs):
         for name, prop in own_properties.items():
             # Only assign our own properties, so we don't overwrite what's been
             # set by the base class
-            if hasattr(obj, name):
+            if hasattr(obj, '_' + name):
```
phschaad, line 332: Shouldn't we check for both?

phschaad (approved):
LGTM, see question before merging

dace/properties.py
tbennun: The reason it is not checked for is that it will return True if you override it as a `@property`, even if the type was not stored.

PR #1725, opened 2024-11-03
Fix type inference and code generation for typeclasses and numpy types

### LINKED ISSUE #1710: `simplify()` fails for SDFG that validates and compile

tbennun:
Fixes #1710 
Supersedes #1721


pratyai:
LGTM.
Thanks a lot for the fix! I can confirm that the patch solves my bug (I find it a bit difficult to track down how the patch affects the transform, but I think that's alright).

tbennun:
> LGTM. Thanks a lot for the fix! I can confirm that the patch solves my bug (I find it a bit difficult to track down how the patch affects the transform, but I think that's alright).

Specifically, it's this code: https://github.com/spcl/dace/pull/1725/files#diff-1081d65d9e2d7faf4c58d18c8652e775fb685d35ae6d4bf2e1f1234429210126L37

It would return the numpy type instead of the `typeclass`

phschaad (approved):
LGTM, thanks!

PR #1731, opened 2024-11-05
Quality of live: Improved error messages

romanc:
Small quality of live PR that improves two error messages. Especially if the state was called `state` (as per default), it was hard to parse at first sight

image

For some reason, `yapf` was also changing the formatting of other code parts. I thus made two commits

1. Format the two files that I was about to edit.
2. Improve the error messages in the freshly formatted files.

This allows to separate real changes from purely stylistic ones.

Would it be an idea to enforce linting with a GitHub Actions workflow? I'm happy to write an issue about this (and contribute early next year).

### DIFF
--- a/dace/codegen/targets/cpp.py
+++ b/dace/codegen/targets/cpp.py
@@ -983,9 +983,10 @@ def unparse_tasklet(sdfg, cfg, state_id, dfg, node, function_stream, callsite_st
     # To prevent variables-redefinition, build dictionary with all the previously defined symbols
     defined_symbols = state_dfg.symbols_defined_at(node)
 
-    defined_symbols.update(
-        {k: v.dtype if hasattr(v, 'dtype') else dtypes.typeclass(type(v))
-         for k, v in sdfg.constants.items()})
+    defined_symbols.update({
+        k: v.dtype if hasattr(v, 'dtype') else dtypes.typeclass(type(v))
+        for k, v in sdfg.constants.items()
+    })
 
     for connector, (memlet, _, _, conntype) in memlets.items():
         if connector is not None:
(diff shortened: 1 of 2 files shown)


romanc:
Not sure why GitHub Actions wasn't able to fetch the pace data. Works on my machine. Maybe just a network glitch. @tbennun would you mind re-running that workflow?

/cc @FlorianDeconinck FYI

phschaad:
This has happened quite a few times the last 2 days. An occasional re-run fixes the issue, but there may be an underlying issue.

romanc:
> This has happened quite a few times the last 2 days. An occasional re-run fixes the issue, but there may be an underlying issue. I've restarted the workflow

Hmm weird. Let me propose two things

1.`wget` would normally retry (as far as I know), but not with fatal errors (like the 503 that we've seen in this case). There's `--retry-connrefused` that we can leverage to even retry under these circumstances. I'll setup a quick PR for this since you say that this occurs often.
2. This data rarely changes. We could leverage GHA caches to avoid re-downloading the files and instead fetching them for a GHA cache. I'll have to discuss with @FlorianDeconinck time-budget wise. Shouldn't take too long, but is more than just adding a retry flag.

phschaad:
This seems reasonable, good idea. Thanks! Using GHA caches could make this more robust

phschaad (approved):
These changes LGTM, thank you.

As for the point about running a formatter through GHA when PRs are merged: I will bring this point up in our weekly DaCe meeting to discuss, thank you for the suggestion.

romanc:
> As for the point about running a formatter through GHA when PRs are merged: I will bring this point up in our weekly DaCe meeting to discuss, thank you for the suggestion.

Thanks for bringing this to the weekly meeting.

PR #1733, opened 2024-11-05
Used valid FORTRAN test program for a couple frontend tests + Made `floatlit2string()` convert the FORTRAN real literal strings into python floats.

pratyai:
Replace some of the invalid fortran test programs with valid ones (that passes under `gfortran -Wall`). This breaks some of the tests, so add fix for them too. `test_fortran_frontend_loop1()` still has an invalid test program, but the nearest valid one breaks for a different reason, and will be fixed later.


acalotoiu (commented):
LGTM

acalotoiu (approved):

PR #1738, opened 2024-11-07
Memlets propagated incorrectly out of nested SDFGs

phschaad:


tbennun:
@phschaad it's not memlet propagation, it's the Python frontend.

The culprit is in newast.py, line 3134-3135:
```python
            if ignore_indices:
                to_squeeze_rng = rng.offset_new(offset, True)
```
Upon squeezing, the offset of the outer memlet may change the offset of the internal memlet, so you need to offset it back to compensate. The offset for `rng` currently is the symbol's valid range. However, what you should be offsetting by (rather than the offset of the symbol's first value), is the final offset of the memlets on the outer SDFG. This is not yet computed at this point (as there may be multiple accesses, so the final offset may change).

This was introduced in #876, and the PR is nice because it includes several tests that demonstrate when you would want that offsetting. A correct fix should pass your test and also all of those tests.

As with similar issues, #1696 will fix it by removing this code altogether.

phschaad:
@tbennun I see, that makes sense. I was expecting this to be a non-issue in 1.x, but I suppose we will need to fix `newast.py` for 1.0 then.

PR #1741, opened 2024-11-07
Do not rename generated code files

tbennun:
If a compiled SDFG binary is already loaded, DaCe checks and generates a different binary name. The generated code files would unfortunately also be renamed, which caused some confusion among users. This only renames the output binary file rather than the code. This also clarifies some potential discrepancies between sdfg.name and the generated code's compiled sdfg.name.


tbennun:
Not a small modification, creates other issues.

PR #1743, opened 2024-11-09
Fixed MapExpansion Transformation

Berke-Ates:
The transformation did not consider maps that are connected with dependency edges.
Now, the transformation skips dependency edges if non-dependency edges exist.
Otherwise, it connects the inner and outer maps with a single dependency edge to maintain connectivity.


acalotoiu (approved):
LGTM

PR #1744, opened 2024-11-09
Instead of compiler specific align hint use C++ standard aligned alloc

ThrudPrimrose:
We can use `alignas, std::aligned_alloc and std::free` that are in the C++ standard instead of using `DACE_ALIGN` + delete[] which are compiler specific. Even though both Clang, GCC support this extension...


tbennun:
Also please remember that we are trying to generate as much C99 as possible. We only use C++ if we must

ThrudPrimrose:
> In which c++ standard was this added?

C11, and since it is in C standard library, C++ gets it too

tbennun:
> > Also please remember that we are trying to generate as much C99 as possible. We only use C++ if we must
> 
> Is there a specific reason for C99? We already use C++11 features like `std::mutex` and `std::atomics`.

Only because we have to and not on device code

PR #1746, opened 2024-11-10
Fix OpenMP dynamic loop bounds that use persistent memory

### LINKED ISSUE #1550: dace fails to compile a function with nested loops
**Describe the bug**
I have a simple program below:

```
N = dace.symbol('N')
@dace.program(auto_optimize=True)
def error_fn_dace(L:dace.float64[N, N], index:dace.uint64, active_size:dace.uint64):
    for i in range(index, active_size - 1):
        L[i + 1][i] = 1.0

        for j in range(i, active_size-1):
            L[j + 1][i] = 2.0
```

When trying to compile this function, I get an error, but if I change the j for loop to:
`for j in range(i, active_size)`
it compiles. 

Could you please let me know if I am doing something wrong? 

**To Reproduce**
1. I am using dace v0.15.1
2. I have attached the python code which produces the error
[error_report.txt](https://github.com/spcl/dace/files/14647667/error_report.txt)

**Expected behavior**
The compilation should not fail

**Desktop (please complete the following information):**
 - OS: [Linux]
 - DaCe v0.15.1

tbennun:
Fixes #1550

### DIFF
--- a/dace/codegen/targets/cpu.py
+++ b/dace/codegen/targets/cpu.py
@@ -1845,7 +1845,7 @@ class CPUCodeGen(TargetCodeGenerator):
 
         # Define all input connectors of this map entry
         for e in dynamic_map_inputs(state_dfg, node):
-            if e.data.data != e.dst_conn:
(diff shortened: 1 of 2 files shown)


acalotoiu (approved):
LGTM

PR #1753, opened 2024-11-13
[Multi SDFG branch, Fortran frontend] These tests relied on `normalize_offset=False` which has changed, since `True` is the default now.

pratyai:


mcopik (approved):
@pratyai LGTM!

PR #1754, opened 2024-11-13
[WIP, Do not review, Multi SDFG branch, Fortran frontend] Take another piece out of `create_sdfg_from_string()` and test it.

pratyai:


acalotoiu (approved):
Looks great to me - we will need to see if there are issues with bigger codes, but this should be a very useful refactor!

PR #1755, opened 2024-11-13
[Multi SDFG branch, Fortran frontend] Make `eliminate_dependencies()` support `use lib` without `only`.

pratyai:
Currently it does not work without explicitly adding everything to be used on an `only`-list.


acalotoiu (approved):
LGTM - although we will intersect the "everything" list with what is actually used. Given that this should happen anyway in the next step, it should work out of the box - but if we get extra bugs, this is a possible source of corner cases.

Issue #1129, opened 2022-10-25
DeadStateElimination on edge conditions with integers as boolean values


leporia:
**Describe the bug**
If we have a conditions with integers used as boolean values on an edge the DeadStateElimination pass removes edges that are actually needed. For example `(not ((N > 20) != 0))`.
I don't know if this is supposed behavior where we cannot use integers to compare boolean values. 

**To Reproduce**
```
import dace

sdfg = dace.SDFG("test")

sdfg.add_array('N', shape="1", dtype=dace.int32, transient=True)

state_init = sdfg.add_state()
state_middle = sdfg.add_state()
state_end = sdfg.add_state()

sdfg.add_edge(state_init, state_end,
              dace.InterstateEdge(
                  condition="(not ((N > 20) != 0))",
                  assignments={"result": "N"}
                  )
              )
sdfg.add_edge(state_init, state_middle, dace.InterstateEdge(
        condition="((N > 20) != 0)"
    ))
sdfg.add_edge(state_middle, state_end, dace.InterstateEdge(
        assignments={"result": "20"}
    ))

sdfg.save("untransformed.sdfg")
sdfg.simplify(verbose=True)
sdfg.save("transformed.sdfg")
```

After executing the following code we see that all states are removed because the conditional edge is removed during the DeadStateElimination pass. This can be seen by saving every intermediate SDFG after every pass done by the simplify method.
 
[pass-SDFG 0: Eliminated 0 states an.zip](https://github.com/spcl/dace/files/9858729/pass-SDFG.0.Eliminated.0.states.an.zip)

The attached SDFG is the one obtained after the DeadStateElimination pass is executed. We can see that the conditional edge was removed.

Note that if use `False` instead of 0 it works as expected.

**Expected behavior**
The two edges with conditions should be kept because it is not possible to solve the condition statically.

**Desktop (please complete the following information):**
 - OS: Fedora Linux 35 (Kernel: 5.19.9-100)
 - Dace Version: 0.14.1

PR #1756, opened 2024-11-14
Fix `pystr_to_symbolic` not correctly interpreting constants as boolean values in boolean comparisons

phschaad:
Strings like `not ((N > 20) != 0)` (== `Not(Ne(Gt(N, 20), 0))`) were incorrectly interpreted by `sympy.sympify` as constant "False". This is a limitation by sympy, which does not assume integer 0 to be a Falsy, and enforces exact equivalence (or difference) checks with `Ne`. To get around this limitation, the DaCe internal AST preprocessor now replaces constants with boolean values if they are arguments to Comparison operations, where the other operand is also a comparison operation, thus returning a boolean.

This fixes an issue with `DeadStateElimination`, closing issue #1129.

### DIFF
--- a/dace/symbolic.py
+++ b/dace/symbolic.py
@@ -1013,14 +1013,6 @@ class PythonOpToSympyConverter(ast.NodeTransformer):
     """ 
     Replaces various operations with the appropriate SymPy functions to avoid non-symbolic evaluation.
     """
-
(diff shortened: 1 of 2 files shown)


tbennun (approved):
I don't like the name truthy_falsy but I approve :)

PR #1758, opened 2024-11-14
[Multi SDFG branch, Fortran frontend] Change the interface of `recursive_ast_improver()` to hide the implementation-specific objects that helps the recursion, by defining a local function to manage the states involved in recursion internally.

pratyai:
Additionally:
- Some minor data structure change: if all the operations that we're doing on `exclude_list` is to check if an item is in it, and add it if it isn't, then a `set` is probably a better container than a `list`.
- Formatting fixes in `ast_components.py`.


acalotoiu (approved):
LGTM - please separate stylistic PRs in the future to make reviewing easier

PR #1760, opened 2024-11-14
`list_of_module_vars` is not actually a string list of the variables, but a structure that needs to be unfolded to get the list.

pratyai:
`list_of_module_vars` is not actually a string list of the variables, but a structure that needs to be unfolded to get the list.
  
 Also, added additional tests that trigger this case.


acalotoiu (approved):
LGTM

PR #1762, opened 2024-11-15
[Multi SDFG, Fortran] Remove the partial internal AST creation in topological order, since at this point (i.e., after the recent fixes), we should have full compilable AST after pruning anyway.

pratyai:
- If the pruned AST is not complete (i.e., won't compile with `gfortran -Wall -c`), then it is a bug in the pruning logic.
- This change affects only `create_sdfg_from_string()`, which is used in nearly all of the tests. All the 134 tests that were passing before this change, still passes after.


acalotoiu (approved):
LGTM

acalotoiu (commented):
LGTM

PR #1764, opened 2024-11-15
[Multi SDFG, Fortran] Add a test case where "a module contains a subroutine contains another subroutine" etc.

pratyai:
Just to make sure that this is handled correctly at the parsing+pruning level.


acalotoiu (approved):
LGTM

PR #1769, opened 2024-11-18
[Multi SDFG, Fortran] Make `functionStatementEliminator()` to modify the `node` itself and return it, so that it cannot forget to account for all the members of `node` when constructing a new one.

pratyai:
Additionally:
- Introduce explicit constructor for `Program_Node`, since it allows for better tracking of what are the member variables. We want to do this for all of the ~84 classes in the `ast_internal_classes.py`, but for now we start with just one.


acalotoiu (commented):
LGTM

PR #1776, opened 2024-11-19
[Multi SDFG, Fortran] 1) Account for `dep_graph.nodes.get(dep).get('info_list') is None`, since importing _everything_ from an empty module implies just that. 2) Keep a track of all the identifiers from the scope of the "use-site" of a module, then intersect with that list during pruning.

pratyai:


acalotoiu (commented):
LGTM

PR #1777, opened 2024-11-19
[Multi SDFG, Fortran, After PR#1774] New tests. In particular, now we can start testing the construction of internal SDFG.

pratyai:
Testing the internal SDFG construction will allow us to verify if the transforms that follow at least start from a valid input state. One of the added tests (`test_subroutine_contains_function()`) already demonstrates that the construction is not always correct.


acalotoiu (commented):
LGTM , let's hope it doesn't break bigger codes

PR #1779, opened 2024-11-20
[Multi SDFG, Fortran] Format the python code in `dace/frontend/fortran` so that the future PRs are not so noisy with formatting diffs.

pratyai:


acalotoiu (approved):
Thank you!

PR #1780, opened 2024-11-20
[Multi SDFG, Fortran, After PR#1779] Parse the internal subprograms of subprograms correctly. Currently, we do not parse them correctly.

pratyai:


acalotoiu (approved):
LGTM

PR #1781, opened 2024-11-20
[Multi SDFG, Fortran] Fix the array attribute tests: 1) Make the test Fortran codes `gfortran -Wall` certified, 2) Remove the unnecessary main program from them since we can set arbitrary entry points now, 3) Pass the extra arguments for array sizes that should have been there in the first place

pratyai:


acalotoiu (approved):
LGTM

PR #1783, opened 2024-11-20
[Multi SDFG, Fortran] Fix the `empty_test.py` by: 1) passing a valid test code, 2) removing redundant parts, 3) using a fresh list of functions everytime we need it, 4) avoiding the `transient_mode = False` trick, and 5) ensuring that some internal classes are initialized correctly.

pratyai:


acalotoiu (approved):
LGTM

PR #1784, opened 2024-11-20
[Multi SDFG, Fortran, After PR#1783] Fix `fortran_language_test.py`. Only one test was failing really, but we apply the same treatment to all of them anyway. The failing one was solved by just avoiding the `transient_mode = False` trick and picking the entry point explicitly (and dropping the `program` that should not be needed in the first place).

pratyai:


acalotoiu (approved):
LGTM

PR #1785, opened 2024-11-20
[Multi SDFG, Fortran] Fix `parent_test.py`: The first test was already passing, we just applied the "modern treatment". The second test was failing, because variable declarations in the module are now populated on a different member than before.

pratyai:


acalotoiu (approved):
LGTM

PR #1787, opened 2024-11-21
[Multi SDFG, Fortran] Fix `intrinsic_blas_test.py`: The `simplify()` pass is broken in `multi_sdfg` branch (probably because it didn't get the bug fixes from `main` branch). Most tests (if any) should not rely on the `simplify()` pass anyway.

pratyai:


acalotoiu (approved):
Disagree. Simplify should be allowed in tests! We should re-enable it once we merge it.

pratyai:
> Disagree. Simplify should be allowed in tests! We should re-enable it once we merge it.

Added a TODO about enabling it back. I agree that simplify should be "allowed". I don't agree that it should be "required". It's a design flaw of:
- Either DaCe if we cannot express things that must be expressed on SDFG.
- Or the frontend if we cannot correctly express things that are already possible to express on SDFG.

Regardless, I guess the matter can be considered resolved for the moment with the TODO comment.

PR #1788, opened 2024-11-21
Symbol redeclaration fix

ThrudPrimrose:
If a symbol is passed as an argument to the SDFG and assigned on an interstate edge, the framecode declares the scalar symbol again. This PR fixes that issue.

The case: If there is an interstate assignment on a variable, `N=5`, but if 'N` is also an input argument to the SDFG, then the generated code does not compile because `N` is redeclared.

### DIFF
--- a/tests/interstate_assignment_test.py
+++ b/tests/interstate_assignment_test.py
@@ -1,7 +1,5 @@
 from typing import Dict
-import warnings
 import dace
-import logging
 
 N = dace.symbol("N")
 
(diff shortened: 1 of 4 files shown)


tests/interstate_assignment_test.py
```diff
@@ -0,0 +1,52 @@
...
+        for m in [map_entry, map_exit]:
+            m.add_in_connector(f"IN_array{sid}")
+            m.add_out_connector(f"OUT_array{sid}")
+        s.add_edge(an, None, map_entry, f"IN_array{sid}", dace.memlet.Memlet(f"array{sid}[0:N]"))
+        s.add_edge(map_entry, f"OUT_array{sid}", t, "_in", dace.memlet.Memlet(f"array{sid}[i]"))
+        s.add_edge(t, "_out", map_exit, f"IN_array{sid}", dace.memlet.Memlet(f"array{sid}[i]"))
+        s.add_edge(map_exit, f"OUT_array{sid}", an2, None, dace.memlet.Memlet(f"array{sid}[0:N]"))
+
+    sdfg.add_edge(s1, s2, dace.InterstateEdge(None, assignments=assignments))
+    sdfg.save("s1.sdfg")
```
phschaad, line 36: remove please, or does this serve a purpose?

phschaad (changes requested):

tests/interstate_assignment_test.py
ThrudPrimrose: I forgot the save call in the code, removed.

PR #1789, opened 2024-11-21
[Multi SDFG, Fortran] Drop the "not implemented error" for the `ArrayLoopNodeLister`, and we have support for multiple ranges. Also, rewrote the logic to be more concise and readable.

pratyai:


acalotoiu (approved):
Awesome! Thank you!

PR #1790, opened 2024-11-21
Interstate assignment to map

ThrudPrimrose:
Maps simple interstate assignments to map encompassing the whole state.
Will complete after the bug fix on re-declaration of symbols is merge.

The pass only transforms if the assignment is a constant. (and the loops such as i = i + 1 should not be affected)

### DIFF
--- a/dace/transformation/passes/interstate_assignment_to_map.py
+++ b/dace/transformation/passes/interstate_assignment_to_map.py
@@ -1,0 +1,85 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
+from typing import Any, Dict, Optional, Set
+
+from dace import SDFG, SDFGState
+import dace
+from dace.sdfg.analysis.cutout import SDFGCutout
+from dace.transformation import pass_pipeline as ppl, transformation
+
+from dace.symbolic import symbol, SymExpr
+
+
+class InterstateAssignmentToMap(ppl.Pass):
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Edges
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        return modified & ppl.Modifies.Edges
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


ThrudPrimrose:
I should move this to a transformation. If the assigned symbol's value is used in more than one state, we need to wrap the whole subgraph where this value is necessary in a map and not just one state.

dace/transformation/passes/interstate_assignment_to_map.py
```diff
@@ -0,0 +1,82 @@
...
+        expr = SymExpr(expression_str)
+        simplified_expr = expr.simplify()
+        is_constant = simplified_expr.is_number
+        return is_constant, simplified_expr
+
+    def apply_pass(self, sdfg: SDFG, _: Dict[str, Any]) -> Optional[Dict[str, Set[str]]]:
+        for state in sdfg.states():
+            for in_edge in sdfg.in_edges(state):
+                assignments = in_edge.data._assignments
+                condition = in_edge.data._cond_sympy
```
tbennun, line 31: don't use the internal cached versions, use the methods that we provide.

tests/interstate_assignment_to_map_test.py
```diff
@@ -0,0 +1,51 @@
+from typing import Dict
```
tbennun, line 1: Missing copyright

tests/interstate_assignment_to_map_test.py
```diff
@@ -0,0 +1,51 @@
...
+    sdfg = _get_interstate_dependent_sdfg({"N": "5"}, False)
+    sdfg.validate()
+    sdfg()
+    iatm =  InterstateAssignmentToMap()
+    iatm.apply_pass(sdfg, {})
+    sdfg.validate()
+    sdfg()
+
+if __name__ == "__main__":
+    test_interstate_assignment()
```
tbennun, line 51: Please run yapf :)

tbennun (changes requested):

PR #1792, opened 2024-11-24
[Multi SDFG, Fortran] Account for the `UseAllPruneList` items in the graph simplification process (previously it missed it when working with interface blocks).

pratyai:


acalotoiu (approved):
LGTM

PR #1793, opened 2024-11-25
[Multi SDFG, Fortran] Deconstruct all the class procedures of a Fortran program, right after parsing it in.

pratyai:
This involves:
1. Find the bound subprogram for that procedure.
2. Replace the calls to the procedure with that subprogram (with the object as an extra argument).
3. Adding `use` statements as necessary.

This has many, many corner cases not addressed. Will be adding more tests as we refine it.


acalotoiu (approved):
LGTM

PR #1796, opened 2024-11-27
Add methods to replace `association`s and `generics`s.

pratyai:
1. `association`s with their targets (just replacing the node verbatim)
2. `generic`s with a target `procedure`. This is not really resolved well yet.


acalotoiu (approved):
LGTM

PR #1798, opened 2024-11-27
Add a special handling for `Part_Ref` too when deconstructing associations: _Try_ to find the consecutive array accesss pattern (e.g., `a(:, :, 1)(2, 2)` etc.) in such cases, and combine them to produce a single access.

pratyai:
Probably not the most elegant way to handle it, but needs least effort.


acalotoiu (approved):
LGTM

PR #1799, opened 2024-11-28
[Multi SDFG, Fortran] 1) Correctly sort the modules in topological order, 2) This allows propagating the aliasing easily, 3) Deconstruct enums into constant integers, 4) FParser sometimes misidentifies a function call as a structure constructor call, so we fix that too, 5) we can also drop the deconstructed procedures.

pratyai:


acalotoiu (approved):
LGTM, let's look for the big bugs

PR #1800, opened 2024-11-29
[Multi SDFG, Fortran] 1) Make recursively_improve_ast use a dictionary of pre-read source content. 2) Add a method to collect the floting subprograms that are possibly needed at the end of recursively_improve_ast.

pratyai:


acalotoiu (approved):
Please also add a test

PR #1801, opened 2024-11-29
[Multi SDFG, Fortran] Recognize interface blocks as symbols (if they have name). Also, when constructing the alias specs, consider that anonymous blocks essentially put their content onto their parent block.

pratyai:


acalotoiu (approved):
LGTM, so interface blocks will be gone before the internal AST now?

PR #1802, opened 2024-11-30
[Multi SDFG, Fortran] 1) Deconstruct functions, subroutines and procedures from interface blocks into calls to concrete implementation, 2) work around an FParser bug 3) work around optional arguments for type matching

pratyai:
Ideally (i.e., if no corner cases were missed and no bugs were made), after this there should be no indirect calls anymore in the AST (except for the explicit renames).


acalotoiu (approved):
lgtm

PR #1805, opened 2024-12-02
[Multi SDFG, Fortran] Have a better type matching to select concrete implementations for generic interfaces.

pratyai:
Specifically, handle the array cases better. There are a ton of other things to improve in there, so it's not a solved problem by any means -- just one step more correct than before.


acalotoiu (approved):
LGTM - would it make sense to separate the fparser processing , sdfg processing into separate files and leave the fortran_parser file as the framework?

PR #1806, opened 2024-12-03
[Multi SDFG, Fortran] Add a new, shorter pruning function that removes any function, subroutine, types etc. that are not used.

pratyai:
Not integrated anywhere yet. Just changed the tests as a proof of concept that the FParser AST stays the same with the new pruning.


acalotoiu (approved):
LGTM

PR #1811, opened 2024-12-04
[MultiSDFG, Fortran] 1) Assign global names to all the functions and subroutines 2) Bugfix about nested function calls being misidentified as part-refs, 3) random bug and test fixes.

pratyai:


acalotoiu (approved):
LGTM

PR #1818, opened 2024-12-07
[Multi SDFG, Fortran] Give variables unique names too. Refactor a whole bunch of code in the process.

pratyai:


acalotoiu (approved):
LGTM
As a suggestion, having multiple smaller tests might be easier to maintain than one test that checks all features.

PR #1820, opened 2024-12-09
Additional intrinsics for Fortran frontend

mcopik:
Added two intrinsics needed by ICON: gemm & transpose. Both use DaCe's library node, so they should generate optimal code.


pratyai (approved):

acalotoiu (approved):
Lgtm

PR #1827, opened 2024-12-13
Let LiftStructViews lift interstate edge struct accesses to views

phschaad:
`LiftStructViews` is a necessary pass for ensuring structure member accesses and container array accesses work correctly, by lifting accesses to them into a 'tower of views' that correctly traces their accesses to the root data container. However, this previously did not consider interstate edges. This PR fixes this by making two changes:
- Allow `LiftStructViews` to lift struct accesses on interstate edges to the correct tower of views, by inserting a lifting state right before the interstate edge, where views are constructed. The interstate edge is then re-written to access the correct view instead.
- Fix `ArrayElimination` to correctly handle these views, which are not directly connected to other nodes, by preventing it from incorrectly merging or removing them. This is done by allowing source/sink view nodes to be merged correctly, i.e., when they are 'the same view', i.e., their view memlets are identical.

This shows an example of the constructed tower of views for a complex interstate edge struct access:

image

### DIFF
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -867,16 +867,16 @@ def get_view_edge(state: SDFGState, view: nd.AccessNode) -> gr.MultiConnectorEdg
         return None
 
     in_edge = in_edges[0]
-    out_edge = out_edges[0]
+    out_edge = out_edges[0] if len(out_edges) > 0 else None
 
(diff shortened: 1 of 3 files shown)


acalotoiu (approved):
LGTM - is the test CI broken?

PR #1831, opened 2024-12-19
Address python frontend shape mismatch for indirect accesses in assignments

phschaad:


tbennun (approved):

tbennun:
@phschaad I'd add, just to be safe, similar tests where ind and/or B is multidimensional (and we access `B[0, ind[1], 2:4] += ...`). Thanks!

PR #1833, opened 2024-12-22
Ignore type hints in undecorated parsed function calls

tbennun:
Fixes cases in which the user has no control over undecorated nested function calls, which previously led to unclear and unrelated `pyobject` type errors.


phschaad (approved):
Very nice, thank you!

PR #1834, opened 2024-12-22
Python frontend fixes for return values and arguments

tbennun:
* Properly parse `return` statements without a return value
* Fix missing argument detection when no arguments are given

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -4738,15 +4738,17 @@ class ProgramVisitor(ExtNodeVisitor):
         # Modify node value to become an expression
         new_node = ast.copy_location(ast.Expr(value=node.value), node)
 
-        # Return values can either be tuples or a single object
+        # Return values can either be tuples, a single object, or None (no return value)
         if isinstance(node.value, (ast.Tuple, ast.List)):
             ast_tuple = ast.copy_location(
                 ast.parse('(%s,)' % ','.join('__return_%d' % i for i in range(len(node.value.elts)))).body[0].value,
                 node)
             self._visit_assign(new_node, ast_tuple, None, is_return=True)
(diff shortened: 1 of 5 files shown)


tbennun:
@phschaad when reviewing, keep the following code in mind (I couldn't reproduce such a situation, you are more qualified to do so): is there any case where the top-level CFG target is an SDFG, and we still create a terminator (`ReturnBlock`) state?
I am thinking of a situation where the python frontend may parse an SDFG, visit a Return statement, and then continue to visit other states/blocks. In that case, the "return" state might still be incorrectly connected to subsequent states.

I want to make sure that creating a terminator (like the `return` statement) is airtight. Thanks!

dace/frontend/python/newast.py
```diff
@@ -4738,15 +4738,17 @@ def visit_Return(self, node: ast.Return):
...
                 node)
             self._visit_assign(new_node, ast_tuple, None, is_return=True)
-        else:
+        elif node.value is not None:
             ast_name = ast.copy_location(ast.Name(id='__return'), node)
             self._visit_assign(new_node, ast_name, None, is_return=True)
+        else:  # No return value
+            pass
 
         if not isinstance(self.cfg_target, SDFG):
```
tbennun, line 4753: @phschaad See comment in PR.

phschaad:
@tbennun There isn't currently something in the frontend that consciously does it that way, but having a return block followed by other content is still valid and works similar to how it would work in Python - i.e., the program exits and everything following is dead code. On a separate branch I currently have even augmented dead state elimination to take that into account and kill more unreachable states. I have thus augmented the PR slightly to insert such a return block now to make the feature airtight.

phschaad (approved):

PR #1838, opened 2024-12-29
Refine condition for prohibiting scalar return values

tbennun:
The check disallowed valid cases where return values that are not returned to the Python frontend were disallowed. Scalar return values in nested functions take the form of a reference, which is not supported with our current Python bindings.


philip-paul-mueller (approved):
LGTM.

PR #1840, opened 2024-12-31
[Multi SDFG, Fortran] Prune redundant uses, which in turn allows pruning redundant objects more aggressively.

pratyai:


acalotoiu (changes requested):
Please add a test to check the functionality. LGTM otherwise

pratyai:
> Please add a test to check the functionality. LGTM otherwise

Thanks for pointing out. Turns out there was indeed a missing case.

acalotoiu (approved):
Nice! Thank you!

acalotoiu:
Have you also checked how it works when applied to ECRAD? (just as a side-note)

pratyai:
> Have you also checked how it works when applied to ECRAD? (just as a side-note)

Yeah, now it fails at a different place than `GET_COMMAND`; I guess that's a progress.

PR #1841, opened 2025-01-03
[Multi SDFG, Fortran] Add methods to inject configs (serialized to files) as constants into the parsed AST.

pratyai:
Naturally, this works only under the assumption that at the start of the program the configs are supplied with these injected values, and the values doe not change during the runtime either.


acalotoiu (approved):
LGTM

PR #1843, opened 2025-01-04
Python: Support complex left-hand side expressions, argmin/max

### LINKED ISSUE #1842: [Python Frontend]  Regression on attribute replacement when used in LHS assigns
```python
import numpy as np
import dace


@dace.program
def runtime(arr) -> None:
    arr.flat[:] += 1  # <--- Regression: left hand side assign fails
    return arr.flat[:]  # Right hand side still works


if __name__ == "__main__":
    arr = np.ones((3, 3))
    runtime(arr)
    assert (arr[:] == 2).all()
```

LHS of assign are failing at parsing times, probably in connetion with the `struct` work

```python
  File "/home/fgdeconi/work/git/ndsl/external/dace/dace/sdfg/sdfg.py", line 56, in __getitem__
    result = result.members[token]
             ^^^^^^^^^^^^^^
AttributeError: 'Array' object has no attribute 'members'
```

tbennun:
Fixes #1842

### DIFF
--- a/dace/frontend/python/memlet_parser.py
+++ b/dace/frontend/python/memlet_parser.py
@@ -273,9 +273,10 @@ def parse_memlet_subset(array: data.Data,
 def ParseMemlet(visitor,
                 defined_arrays_and_symbols: Dict[str, Any],
                 node: MemletType,
-                parsed_slice: Any = None) -> MemletExpr:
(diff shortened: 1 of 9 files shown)


phschaad (approved):
LGTM, thanks!

PR #1845, opened 2025-01-06
Fixes replacement of dict keys

alexnick83:
If the internal method `SDFG._replace_dict_keys` is called with the same `old` and `new` keys, then the method erroneously removes the entry for the specific key completely. Such a case occurs when calling `SDFG.replace_dict` with a dictionary that contains identity replacement matches.

This PR addresses the issue by:
- [x] Adding a check inside `SDFG._replace_dict_keys` and skipping if `old == new`. A warning is also thrown for debugging purposes.
- [x] Removing identity replacement matches from `repldict` and `symrepl` in `SDFG.replace_dict`.


acalotoiu (approved):
LGTM! Thank you for fixing this surprising issue !

PR #1846, opened 2025-01-06
[Multi SDFG, Fortran] Inject `size()` related configs in the internal AST.

pratyai:
This must happen after the intrinsic transforms, since they replace the intrinsic calls with the extra variables.

### DIFF
--- a/dace/frontend/fortran/ast_desugaring.py
+++ b/dace/frontend/fortran/ast_desugaring.py
@@ -2098,6 +2098,9 @@ class ConstInstanceInjection:
     value: Any  # Literal value to substitue with. The injected literal's type will match the type of the original.
 
 
+ConstInjection = Union[ConstTypeInjection, ConstInstanceInjection]
+
+
 def _val_2_lit(val: str, type_spec: SPEC) -> LITERAL_TYPES:
     val = str(val).lower()
     if type_spec == ('INTEGER1',):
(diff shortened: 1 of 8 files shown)


mcopik:
Is the PR completed and ready for review?

pratyai:
> Is the PR completed and ready for review?

Moved the position of the injector later in the pipeline since not all the array related variables are used by the earlier point. I think it's now ready enough.

But I've tested only with the `mk2` fortran (which has its own concerns) and without your last 4 commits (which break the multi sdfg). With that, I see all the `f2dace` variables gone from the generated sdfg (but even before it was only 3 interstate edges; really small potatoes really, so not sure if I missed something or it wasn't worth it in the first place).

pratyai:
> LGTM - but please test and make sure there are no ECRAD regressions, and it applies to ECRAD as expected before committing (if necessary, add additional tests to cover corner cases)

I've tested only with the mk2 fortran (which has its own concerns) and without Marcin's last 4 commits (which break the multi sdfg). With that, I see all the f2dace variables gone from the generated sdfg (but even before it was only 3 interstate edges; really small potatoes really, so not sure if I missed something or it wasn't worth it in the first place).

I'd leave fixing those 4 commits to Marcin (this PR can wait if it must). Also a bit hard to tell if ECRAD is regressing if the inputs are not pinned down (before the standard was running `icon_config_propagation.py`, but now that goalpost has silently moved a bit with mk2).

PR #1848, opened 2025-01-07
Revert "Make merge great again" which broke the fortran frontend (especially for ECRAD)

pratyai:
This reverts commit 9d592a47bb35259afc50a0f69007a5afee0d7b04.

### DIFF
--- a/dace/frontend/fortran/fortran_parser.py
+++ b/dace/frontend/fortran/fortran_parser.py
@@ -3010,30 +3010,14 @@ def create_sdfg_from_string(
         raise NameError("Not all functions were transformed to subroutines")
     program.function_definitions = []
     program = ast_transforms.SignToIf().visit(program)
+    program = ast_transforms.ArrayToLoop(program).visit(program)
+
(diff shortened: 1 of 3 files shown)


dace/frontend/fortran/fortran_parser.py
```diff
@@ -3010,30 +3010,14 @@ def create_sdfg_from_string(
         raise NameError("Not all functions were transformed to subroutines")
     program.function_definitions = []
     program = ast_transforms.SignToIf().visit(program)
-    program = ast_transforms.ArgumentExtractor(program).visit(program)
-
-    program = ast_transforms.TypeInference(program, assert_voids=False).visit(program)
+    program = ast_transforms.ArrayToLoop(program).visit(program)
```
mcopik, line 3013: Is it necessary to revert that? IIRC ecrad parser shouldn't even use that call inside fortran_parser, so this can stay. It only affects tests

pratyai: Undid this part in a new commit.

PR #1849, opened 2025-01-07
Python: Allow python-style unpacking assignments

phschaad:


dace/frontend/python/newast.py
```diff
@@ -3292,6 +3292,19 @@ def _visit_assign(self, node, node_target, op, dtype=None, is_return=False):
         if isinstance(node.value, (ast.Tuple, ast.List)):
             for n in node.value.elts:
                 results.extend(self._gettype(n))
+        elif isinstance(node.value, ast.Name) and node.value.id in self.sdfg.arrays and isinstance(
+                self.sdfg.arrays[node.value.id],
+                data.Array) and self.sdfg.arrays[node.value.id].total_size == len(elts):
```
tbennun, line 3297: Shouldn’t this have to do with the shape rather than total_size? This wouldn’t handle multidimensional arrays nor padded ones, because total size has to do with allocation.

tbennun: Also, shouldn’t this be the result of visiting rhs instead? Only if it returns a single array of course.

phschaad: Adjusted accordingly

tbennun (approved):
LGTM with modifications

PR #1853, opened 2025-01-08
Various Fixes and QOL improvements

phschaad:
- Allow the use of compression when using `SDFG.view`, allowing large SDFGs to be viewed
- Fix codegen issues related to out-of-date CFG lists
- Allow pass options to be passed to `SDFG.simplify`, passing them along to passes
- Implement the pass options API for `InlineSDFGs`
- Fix serialization issues for reduction library nodes where the identity function is `np.inf` or similar non-numeric numbers, which cause issues with other JSON deserializers that do not treat 'Infinity' (or similar) as a number.
- Fix missing symbol arguments when parsing a nested SDFG which takes a structure argument where a member's size is defined by a symbol not explicitly passed to the NSDFG, using symbol mappings
- Fix an error where the frontend reports missing symbols when a scalar struct member is being used in the range of a loop
- Fix replacements for `np.where` when using a member of a structure

### DIFF
--- a/dace/transformation/passes/fusion_inline.py
+++ b/dace/transformation/passes/fusion_inline.py
@@ -88,6 +88,16 @@ class InlineSDFGs(ppl.Pass):
     def report(self, pass_retval: int) -> str:
         return f'Inlined {pass_retval} SDFGs.'
 
+    def set_opts(self, opts):
+        opt_keys = [
+            'multistate',
+        ]
+
+        for k in opt_keys:
+            attr_k = InlineSDFGs.__name__ + '.' + k
+            if attr_k in opts:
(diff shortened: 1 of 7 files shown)


dace/transformation/passes/fusion_inline.py
```diff
@@ -88,6 +88,16 @@ def apply_pass(self, sdfg: SDFG, _: Dict[str, Any]) -> Optional[int]:
     def report(self, pass_retval: int) -> str:
         return f'Inlined {pass_retval} SDFGs.'
 
+    def set_opts(self, opts):
```
tbennun, line 91: what's this for?

tbennun (approved):
Approved except for one question

PR #1854, opened 2025-01-08
Allow for packing assignments in the python frontend

phschaad:


dace/frontend/python/newast.py
```diff
@@ -3296,7 +3296,65 @@ def _visit_assign(self, node, node_target, op, dtype=None, is_return=False):
             results.extend(self._gettype(node.value))
 
         if len(results) != len(elts):
-            raise DaceSyntaxError(self, node, 'Function returns %d values but %d provided' % (len(results), len(elts)))
+            if len(elts) == 1 and len(results) > 1 and isinstance(elts[0], ast.Name):
```
tbennun, line 3299: I would have self.visit-ed to get the name from other kinds of expressions (need to take care to visit the rhs once though

tbennun:
I'm not sure if we want this PR.

PR #1855, opened 2025-01-08
[Multi SDFG, Fortran] Track down `allocated()` calls on Fparser and resolve them to constants with config injection if possible.

pratyai:
Turns out this is enough to remove all `allocated()` calls that remains after pruning.


acalotoiu (approved):
Excellent!

PR #1857, opened 2025-01-08
[Multi SDFG, Fortran] 1) Refine the "practically const arg => const" 2) Add a "practically const global var => const" 3) Fix the pruning test

pratyai:
Now all `present()` calls are gone too. Post-pruning 8.6k LoC remains.


acalotoiu (approved):
LGTM

PR #1859, opened 2025-01-09
[Multi SDFG, Fortran] Track locally constant variables and struct members to certainly known literal values and use it to propagate more constants and prune more things.

pratyai:
With this, the single `allocate()` call is gone and ~8.5 LoC pruned code remains.

This is a pretty risky operation without proper SSA; hopefully nothing to terrible happens to the ECRAD graph.


acalotoiu (approved):
LGTM

PR #1861, opened 2025-01-09
[Multi SDFG, Fortran] 1) Handle part-refs better in the type deduction. 2) Cleanup empty modules and unused local variables inside functions.

pratyai:
1) There were some cases not handled that was found in the tests.
2) This is mostly cosmetic, but harmless pruning. Pruned code is at about ~8.3k LoC after this.


acalotoiu (approved):
LGTM

PR #1868, opened 2025-01-13
Validate tasklets subscript access dimensionality

tbennun:
This PR adds a validation check that ensures connector access in tasklets conforms to the SDFG IR rules. This gives a more verbose and clearer error than before, where it would either manifest in the code generator or remain undetected.

cc @acalotoiu @pratyai @mcopik


mcopik:
@tbennun Thank you, LGTM!

PR #1869, opened 2025-01-13
[Multi SDFG, Fortran] 1) Add initializers for each derived type with default values 2) Add global initializer for each global variable with default values.

pratyai:


acalotoiu (approved):
LGTM

PR #1870, opened 2025-01-14
[Multi SDFG, Fortran] Prune unused struct fields, i.e., if they are never referred anywhere in the AST.

pratyai:


acalotoiu (approved):
LGTM. Very compact!

PR #1872, opened 2025-01-14
[Multi SDFG, Fortran] 1) Convert "data statements" into regular assignments 2) Make tracking local constants more thorough (although not exhaustive yet)

pratyai:
- Caveat: there are some subtle differences with regular initialisation/assignment as per the standard.
- About ~8k LoC remains after pruning.


acalotoiu (approved):
LGTM

PR #1873, opened 2025-01-15
Fix a bug where dead state elimination leads to disconnected graphs

phschaad:


acalotoiu (approved):
LGTM - surprising that it was still a bug :)

PR #1875, opened 2025-01-16
Fix incorrect offsets during assignments with broadcasts

phschaad:


tbennun (approved):

tbennun (changes requested):
Please add test / parameterize split_test

tbennun (approved):

PR #1876, opened 2025-01-16
[f2dace/dev, fortran] 1) Use f-string w/o backslash. 2) Fix another case of test failure due to 3.9 compat.

pratyai:
Now all 30 tests pass at 3.9.


acalotoiu (approved):
Thanks! LGTM

PR #1877, opened 2025-01-16
Support non-contiguous arrays in Fortran

mcopik:
This PR adds support for non-contiguous array slicing in Fortran:

```
cols = (1, 3, 7)
func( arr(1, cols) )
```

It works on 1D and 2D arrays. Pending work: ECRAD issues.


acalotoiu (approved):
LGTM! Thanks!

PR #1881, opened 2025-01-18
[f2dace/dev, fortran] Fix the arg-spec handling of the call extractor. In fact, we can rewrite all of it more concisely and more generally.

pratyai:


acalotoiu (approved):
LGTM

PR #1886, opened 2025-01-20
Replace buggy elemental handling with new approach

mcopik:
We replace prior and problematic existing elemental expander with a new one, that uses a single loop expansion base for array2loop, elemental intrinsics, and now elementals + couple bugfixes.

Works on isolated ECRAD samples. This one is not fully verified yet because we fail earlier at SDFG generation due to missing symbols.


acalotoiu (approved):
LGTM - please check that no new tests are failing in f2dace/dev before merging!

PR #1887, opened 2025-01-20
[f2dace/dev, fortran] Instead of using `parfor` everywhere, give every transforms their unique short names.

pratyai:
Unsurprisingly, this "fix" brings out another two million bugs out of the woodwork. Surprisingly, this wasn't noticed earlier.

Before, the added variables had name collisions (possibly even type collisions), that passed the typical validation checks until one starts to really look at the built graphs.

### DIFF
--- a/dace/frontend/fortran/ast_transforms.py
+++ b/dace/frontend/fortran/ast_transforms.py
@@ -2239,7 +2239,7 @@ def par_Decl_Range_Finder(node: ast_internal_classes.Array_Subscript_Node,
                 newbody.append(
                     ast_internal_classes.Decl_Stmt_Node(vardecl=[
                         ast_internal_classes.Symbol_Decl_Node(
-                            name="tmp_parfor_" + str(count + len(rangepos) - 1), type="INTEGER", sizes=None, init=None,
+                            name="tmp_pdrf_" + str(count + len(rangepos) - 1), type="INTEGER", sizes=None, init=None,
                             parent=node.parent, line_number=node.line_number)
                     ]))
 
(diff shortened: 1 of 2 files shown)


dace/frontend/fortran/ast_transforms.py
```diff
@@ -2239,7 +2239,7 @@ def par_Decl_Range_Finder(node: ast_internal_classes.Array_Subscript_Node,
                 newbody.append(
                     ast_internal_classes.Decl_Stmt_Node(vardecl=[
                         ast_internal_classes.Symbol_Decl_Node(
-                            name="tmp_parfor_" + str(count + len(rangepos) - 1), type="INTEGER", sizes=None, init=None,
```
mcopik, line 2242: parDeclRangeFinder is called from many different transformations. Do you think this can still cause name collisions? if yes, then we can change it by passing a custom prefix from each transform

dace/frontend/fortran/intrinsics.py
```diff
@@ -714,20 +714,20 @@ def visit_Execution_Part_Node(self, node: ast_internal_classes.Execution_Part_No
                 initrange = i[0]
                 finalrange = i[1]
                 init = ast_internal_classes.BinOp_Node(
-                    lval=ast_internal_classes.Name_Node(name="tmp_parfor_" + str(self.count + range_index)),
+                    lval=ast_internal_classes.Name_Node(name="tmp_lbrt_" + str(self.count + range_index)),
```
mcopik, line 717: Same here - this is a base used by many intrinsics. Two different intrinsic replacements can emit the same `tmp_lbrt_0`, for example.

If this is a problem for name collision, then I will change the implementation to pass a custom prefix from each intrinsic.

PR #1892, opened 2025-01-22
[f2dace-dev, fortran] Add some basic validations for mistakes that have shown up before.

pratyai:
Of course we cannot enable it permanently right now because it breaks all the test. But still helps to raise at least some flags before SDFG construction begins, when enabled.


acalotoiu (approved):
LGTM! Good idea!

PR #1893, opened 2025-01-22
[f2dace-dev, fortran] Use a temporary name generator that will not run into name collision with its own generated names.

pratyai:


acalotoiu (approved):
LGTM! Please test to make sure no new tests fail

pratyai:
> LGTM! Please test to make sure no new tests fail

The failing tests are the same as the ones on the sheet at the moment; so I guess it's good to go.

PR #1894, opened 2025-01-22
[f2dace/dev, fortran] Project internal ast nodes into semi-readable, not-fully-encoded strings.

pratyai:


acalotoiu (approved):
LGTM - probably should have a test just do demo the behavior

pratyai:
> LGTM - probably should have a test just do demo the behavior

It already works in the debuggers by `print(node)` etc. typical str-able operations. I'll add some diff test for internal ast later when I see a good usecase.

PR #1897, opened 2025-01-22
Add skip to sdfg dot simplify

ThrudPrimrose:
I think this change would be convenient, allows us to skip some passes even if we call sdfg.simplify()

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -2517,7 +2517,9 @@ class SDFG(ControlFlowRegion):
             :note: This is an in-place operation on the SDFG.
         """
         from dace.transformation.passes.simplify import SimplifyPass
-        return SimplifyPass(validate=validate, validate_all=validate_all, verbose=verbose,
+        return SimplifyPass(validate=validate,
+                            validate_all=validate_all,
+                            verbose=verbose,
                             skip=options.get('skip', None),
                             pass_options=options).apply_pass(self, {})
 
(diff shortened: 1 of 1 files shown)


dace/sdfg/sdfg.py
```diff
@@ -2517,7 +2517,10 @@ def simplify(self, validate=True, validate_all=False, verbose=False, options=Non
             :note: This is an in-place operation on the SDFG.
         """
         from dace.transformation.passes.simplify import SimplifyPass
-        return SimplifyPass(validate=validate, validate_all=validate_all, verbose=verbose,
+        return SimplifyPass(validate=validate,
+                            validate_all=validate_all,
+                            verbose=verbose,
+                            skip=options.get('skip', None),
```
phschaad, line 2523: you'll have to make sure `options` is not None first.

phschaad (changes requested):

dace/sdfg/sdfg.py
tbennun: Why doesn't `SDFG.simplify` have a skip argument? It would be nicer for a user.

phschaad: that's probably even better, yes - we should just propagate the argument up to the next API level

ThrudPrimrose: I agree, creating a skip argument is better IMO too.

phschaad (approved):

PR #1898, opened 2025-01-22
[f2dace/dev, fortran] 1) Add declarations only to the specification-part. 2) Simplified a `a + 0` expression into just `a`

pratyai:


acalotoiu (approved):
LGTM

PR #1903, opened 2025-01-24
[f2dace-dev, fortran] Clean up a few unused pieces of code that are 1) either better to deal with on the fparser preprocessing 2) or can be written more concisely with `mywalk` instead of creating a single-use visitor class. 

pratyai:
The benefit of this (or any) cleanup is that we don't have to update/refactor in these code anymore when we refactor things elsewhere.


acalotoiu (approved):
Excellent! Thank you! LGTM

PR #1904, opened 2025-01-24
[f2dace-dev, fortran] Look at all the branches, then aggregate what can certainly be known, and certainly be unknown, and decide based on that.

pratyai:


acalotoiu (approved):
LGTM

PR #1905, opened 2025-01-24
[f2dace/dev, fortran] Write "1" or "0" instead of "True" or "False" as preferred by dace semantics.

pratyai:
- Allows enabling `empty_test`: fixes writing "true" or "false" into SDFG branches.
- Fixes a bug in the fparser processing, where an optional argument might have been fixed to one value, even if it would not be present some of the times.
- Fixes mishandling of internal subprograms by the optional argument transform.


pratyai:
Screenshot 2025-01-27 at 05 56 13

currently failing tests^

pratyai:
ECRAD seems good (6 of 9) when rebased on top an older commit (but not on top the current f2dace/dev)

acalotoiu (approved):
LGTM

PR #1906, opened 2025-01-27
Added `FindSingleUseData` Pass

philip-paul-mueller:
Added `FindSingleUseData` analysis pass

This new pass scans the SDFG and computes the data that is only used in one single place.
Essentially, this boils down to the check how often a data descriptor is accessed
A data descriptor is classified if the following statements are true:
- There is exactly one AccessNode referring to the data descriptor.
- The data descriptor is not referred to on an interstate edge, nor in the condition parts of `ConditionBlock`s or `LoopRegion`s.

This pass will be needed to speed up fusion passes.

### DIFF
--- a/dace/transformation/passes/analysis/analysis.py
+++ b/dace/transformation/passes/analysis/analysis.py
@@ -352,13 +352,13 @@ class FindExclusiveData(ppl.Pass):
         # If anything was modified, reapply
         return modified & ppl.Modifies.AccessNodes & ppl.Modifies.States
 
-    def apply_pass(self, sdfg: SDFG, _) -> Dict[SDFG, Dict[str, Set[SDFGState]]]:
+    def apply_pass(self, sdfg: SDFG, _) -> Dict[SDFG, Set[str]]:
         """
         :return: A dictionary mapping SDFGs to a `set` of strings containing the name
             of the data descriptors that are exclusively used.
         """
         # pschaad: Should we index on cfg or the SDFG itself.
-        exclusive_data: Dict[SDFG, Dict[str, Set[SDFGState]]] = {}
+        exclusive_data: Dict[SDFG, Set[str]] = {}
         for nsdfg in sdfg.all_sdfgs_recursive():
             exclusive_data[nsdfg] = self._find_exclusive_data_in_sdfg(nsdfg)
         return exclusive_data
--- a/dace/transformation/passes/analysis/analysis.py
+++ b/dace/transformation/passes/analysis/analysis.py
@@ -335,6 +335,69 @@ class FindAccessStates(ppl.Pass):
 
 @properties.make_properties
 @transformation.explicit_cf_compatible
+class FindExclusiveData(ppl.Pass):
+    """
+    For each SDFG find all data descriptors that are referenced in exactly one location.
+
+    This means that for every data descriptor there exists exactly one AccessNode that
+    refers to that data. In addition the data is also not read on an interstate edge.
+    """
+
+    CATEGORY: str = 'Analysis'
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Nothing
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        # If anything was modified, reapply
+        return modified & ppl.Modifies.AccessNodes & ppl.Modifies.States
+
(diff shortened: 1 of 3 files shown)


dace/transformation/passes/analysis/analysis.py
```diff
@@ -333,6 +333,71 @@ def apply_pass(self, top_sdfg: SDFG, _) -> Dict[int, Dict[str, Set[SDFGState]]]:
...
+    """
+
+    CATEGORY: str = 'Analysis'
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Nothing
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        # If anything was modified, reapply
+        return modified & ppl.Modifies.AccessNodes & ppl.Modifies.States
```
phschaad, line 356: This should include interstate edges too, no? I think to be on the safe side, given that we can access things on loop ranges and conditionals as well, this may need to be `AccessNodes & CFG`

dace/transformation/passes/analysis/analysis.py
```diff
@@ -333,6 +333,71 @@ def apply_pass(self, top_sdfg: SDFG, _) -> Dict[int, Dict[str, Set[SDFGState]]]:
...
+                data_name: str = dnode.data
+                if data_name in exclusive_data:
+                    exclusive_data.discard(data_name)  # Classified too early; Undo
+                elif data_name not in previously_seen:
+                    exclusive_data.add(data_name)  # Never seen; Assume it is exclusive.
+                previously_seen.add(data_name)
+
+        # Compute the set of all data that is accessed, i.e. read, by the edges.
+        interstate_read_symbols: Set[str] = set()
+        for edge in sdfg.edges():
```
phschaad, line 393: Two things here, because of Control flow regions:
1. You need to check `sdfg.all_interstate_edges()` to go through all possible interstate edges - `sdfg.edges()` only checks the top-level edges.
2. You must also check conditionals/loops - for this you can go over all control flow regions (`sdfg.all_control_flow_regions`) and query their used symbols with the `used_symbols` method, passing `with_contents=False`. That gives you the used symbols in things like conditional branch conditions, loop conditions, etc, without recursing into the graph.

phschaad (changes requested):
Very nice addition, thank you!

Needs some adaptation to control flow regions, but otherwise looks good

phschaad:
I vote for "Removable"

philip-paul-mueller:
I agree the name is not good.
But I do not like "Shared" or the idea of computing the inverse (although this is the very original name that was used in MapFusion).
I also do not like "Removable" because I think the concept is too broad.

So I propose "SingleUseData", because this is the core, it is used at one single place, i.e. to transmit data between two Maps, nowhere else.

@phschaad @tbennun

dace/transformation/passes/analysis/analysis.py
philip-paul-mueller: You are right.

dace/transformation/passes/analysis/analysis.py
philip-paul-mueller: Thanks for the info (and lesson) about them.
I have definitely to dig more into them.
Do you know where I can find more information about that, I checked the documentation, but it is not so verbose in that regard?

dace/transformation/passes/analysis/analysis.py
```diff
@@ -333,6 +333,79 @@ def apply_pass(self, top_sdfg: SDFG, _) -> Dict[int, Dict[str, Set[SDFGState]]]:
...
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        # If anything was modified, reapply
+        return modified & ppl.Modifies.AccessNodes & ppl.Modifies.CFG
+
+    def apply_pass(self, sdfg: SDFG, _) -> Dict[SDFG, Set[str]]:
+        """
+        :return: A dictionary mapping SDFGs to a `set` of strings containing the name
+            of the data descriptors that are only used once.
+        """
+        # TODO(pschaad): Should we index on cfg or the SDFG itself.
```
phschaad, line 367: For something related to data containers and the question of 'can I remove this' I think it makes sense to index on the SDFG level

phschaad (approved):
LGTM, thanks!

PR #1908, opened 2025-01-27
CFG (post)dominator SDFGStates

alexnick83:
Adds extraneous SDFGStates before/after CFGs when inferring the allocation lifetime.


acalotoiu:
Agreed, and merged, with my thanks!

PR #1910, opened 2025-01-28
Update the webclient to point to the dist branch again

phschaad:


philip-paul-mueller (approved):
LGTM.
The issue in my machine was resolved.

PR #1913, opened 2025-01-30
[f2dace-dev, fortran] Replace `recursive_ast_improver()` with something that reads and parses everything in one go.

pratyai:
1. The replacement parses every f90 file, discarding the ones with syntax errors and tracking the top-level objects. Then at the end combines all the top-level objects, sort them by dependencies, and the FParser preprocessing progresses as usual with pruning etc.
2. Now it does not depend on module names to find files, since all files are read and the module names are tracked anyway.
3. Now it is slightly faster (~75s to ~50s), since we are not reading the files multiple times (which we did before to collect floating subprograms that are outside any module).
4. There is also a CLI script to preprocess entire code bases (a bit more generalized version of what we have for ECRAD) and write into a single f90 file.
5. Handle calls "missing" functions slightly better, since there are certain included header files in the fortran code for which we simply do not have the file itself (not entirely sure how those parts of ICON get built in the first place).

Additionally:
1. A bug in local constant tracking was fixed (i.e., we should not consider array contents "constants", even if they are practically so -- at least until we really need to and we can handle it better).
2. Handle `=> null()` initialization of pointers.
3. Handle a bug in making the identifier lower cases.


pratyai:
Screenshot 2025-01-30 at 12 58 49

The usual 9 tests still failing.

acalotoiu (approved):
LGTM

PR #1915, opened 2025-01-30
[f2dace/dev, fortran] Minor fixes in tests + replace some tests that do not use "valid" programs with two reasonable ones.

pratyai:


pratyai:
Screenshot 2025-01-30 at 22 00 51

Usual 9 test failures + ECRAD 6/9 built.

acalotoiu (approved):
LGTM! Thank you!

PR #1916, opened 2025-01-31
[f2dace/dev, fortran] Now that we don't look up files from module names, and as such don't have the no-more-than-one-module-per-file rule, we can relax the test helper and also not pass a `main` in `ParseConfig`.

pratyai:
Benefit:
In tests, now we can just `add_file()` once with the whole test code with all the modules and programs. Before, we had to `add_file()` each module separately, because `recursive_ast_improver()` looked up files by module names, which didn't work if a file contained multiple modules. Now that we've replaced that function that does not assume anything about file names, we can add everything in one file, and it's valid (as long as it's a valid Fortran program).
That in turn makes it easier to copy the test code in a file, verify if it compiles, edit it until it does, then copy it back.


pratyai:
Screenshot 2025-01-31 at 14 37 40

The usual 9.

acalotoiu (approved):
LGTM

PR #1919, opened 2025-02-02
[f2dace/dev, fortran] 1) Handle more Fortran language constructs in the preprocessors, e.g., bindings, operators, includes 2) Right after constructing the full AST, run a coarse pruning to discard completely unused modules.

pratyai:
1. This mostly affects the ability to process rest of ICON (e.g., for velocity tendencies). ECRAD did not have these constructs, so we didn't have them before.
2. This also is more useful for rest of the ICON, where it is really easy to pull in a lot of code in the AST just by giving a wide filter, but those additional codes sometimes contain language constructs we cannot process, and in some cases may even be broken.


pratyai:
Screenshot 2025-02-02 at 22 19 51

pratyai:
Screenshot 2025-02-05 at 10 38 08
Still 9.

acalotoiu (approved):
LGTM!

PR #1920, opened 2025-02-03
Fix read_and_write_sets() for ControlFlowRegion

edopao:
The `PruneConnectors` transformation relies on `read_and_write_sets()` to identify connectors on a nested SDFG that refer to unused data containers. With the introduction of `ControlFlowRegion` nodes, the data containers can now be accessed by symbolic expressions used as conditions in such nodes. This case was not considered in baseline.

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -1376,12 +1376,20 @@ class SDFG(ControlFlowRegion):
         read_set = set()
         write_set = set()
         for state in self.states():
-            for edge in state.parent_graph.in_edges(state):
-                read_set |= edge.data.free_symbols & self.arrays.keys()
             # Get dictionaries of subsets read and written from each state
             rs, ws = state._read_and_write_sets()
             read_set |= rs.keys()
             write_set |= ws.keys()
+
+        array_names = self.arrays.keys()
+        for edge in self.all_interstate_edges():
+            read_set |= edge.data.free_symbols & array_names
+
+        # By definition, data that is referenced by the conditions (branching condition,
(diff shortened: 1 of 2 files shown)


dace/sdfg/sdfg.py
```diff
@@ -1376,12 +1376,20 @@ def read_and_write_sets(self) -> Tuple[Set[AnyStr], Set[AnyStr]]:
...
             # Get dictionaries of subsets read and written from each state
             rs, ws = state._read_and_write_sets()
             read_set |= rs.keys()
             write_set |= ws.keys()
+
+        array_names = self.arrays.keys()
+        for edge in self.all_interstate_edges():
+            read_set |= edge.data.free_symbols & array_names
+
+        # By definition, data that is referenced by the conditions (branching condition,
```
phschaad, line 1388: This comment appears copied from somewhere else, does this really apply here?

phschaad (changes requested):

dace/sdfg/sdfg.py
edopao: I think it does. We are removing from `single_use_data` the symbols used by condition expressions in control flow regions.

phschaad: But what is `single_use_data` here, in this context? It appears that all we are doing here - based on the code - is adding symbols used on control flow region's meta accesses (conditions, loop increments, etc.) to the `read_set` of the SDFG?

edopao: Now I get it, the misleading information is "single use data". I will change the code comment.

phschaad (approved):
Thank you, LGTM!

PR #1924, opened 2025-02-04
save pruned ast if path provided

ThrudPrimrose:
Generating SDFGs involve saving the pruned ast
I added these lines:
    import os
    FORTRAN_AST_PATH = os.getenv("FORTRAN_AST_PATH")
    if FORTRAN_AST_PATH is not None:
        with open(FORTRAN_AST_PATH, 'w') as f:
            f.write(ast.tofortran())

(my editor automatically removes empty lines that only have whitespace)

It can streamline the process of parsing SDFGs

### DIFF
--- a/dace/frontend/fortran/fortran_parser.py
+++ b/dace/frontend/fortran/fortran_parser.py
@@ -757,7 +757,7 @@ class AST_translator:
         if len(body_list)>1:
             print("Case making here!")
 
-        else:    
+        else:
 
             name = f"Conditional_l_{str(node.line_number[0])}_c_{str(node.line_number[1])}"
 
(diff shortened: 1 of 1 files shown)


ThrudPrimrose:
If we want to comment out a function, we need to save this and call it again.
So saving this always if a path is provided is a good idea IMHO

dace/frontend/fortran/fortran_parser.py
```diff
@@ -3187,6 +3187,12 @@ def create_sdfg_from_fortran_file_with_options(
     else:
         ast = correct_for_function_calls(ast)
 
+    import os
```
pratyai, line 3190: **Generally**, I'd rather advice against looking for environment variable in the middle of a library code, and insist on adding yet another `Optional[Path]` flag to the `ParserConfig` instead. Or even better, I'd suggest taking the FParser processing part out to the CLI and let it decide if it wants to save the file or not.

**However**, `create_sdfg_from_fortran_file_with_options()` is sadly already in quite an "anything goes" state. So, I'm okay with this too if Lex accepts it.

PR #1926, opened 2025-02-06
[f2dace/dev] Generate a compatible header.

pratyai:
Currently, the generated `.h` is "uncompilable" together with the generate `.cpp`. Here we are:
1. Declaring the functions in the header the exact same way as the definition inside.
2. Putting everything (including the struct definition that alexnick added yesterday) in the header behind a header guard.

TODO later:
- We should remove the struct definition in the `.cpp` and include the `.h`.

### DIFF
--- a/dace/codegen/codegen.py
+++ b/dace/codegen/codegen.py
@@ -3,25 +3,28 @@ import functools
 from typing import List
 
 import dace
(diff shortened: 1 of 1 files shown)


acalotoiu (approved):
LGTM

PR #1927, opened 2025-02-06
[f2dace/dev, fortran] Serializer+Deserializer generator

pratyai:
It works like this:
- It takes a (hopefully preprocessed) working F90 AST, and an SDFG (from which the C++ code is generated without additional transformations).
- It generates an F90 module that can serialise the structures defined in the F90 AST into a file. But it only serializes the components that are also present in the SDFG (partly to avoid unnecessary data-dumping, and partly to avoid having to be too smart when deserializing. Although the latter reason is not a good one and should be removed, that's a TODO for now).
- It generates a C++ header that can deserialise the data from the F90 module (as noted before, it is not too smart and does not know what to do if expected fields are not there or unexpected fields are there).
- The C++ header can also generate the config injection data for latter usage.
- *Then*, we add the F90 module on top of the original released *buildable* ICON, and *manually* instrument the code to dump various data into files.
- We write a `main.cc` that uses the generated C++ library from the SDFG. There we read those data into their corresponding structs, and call the 3 SDFG functions with appropriate arguments.
- If everything goes right by some miracle, we will have the correct output in the output arguments.


pratyai:
Screenshot 2025-02-06 at 18 20 50

The usual broken 9.

acalotoiu (approved):
LGTM - but I think it doesn't test nested structures and AOS/SOA

PR #1929, opened 2025-02-07
[f2dace/dev, fortran] Always default-initialize struct definitions, which is possible because we never have non-builtin types there (always builtin types or pointers). This avoids one source of undefined behaviours that is the use of uninitialized values.

pratyai:


acalotoiu (approved):
LGTM - no test?

pratyai:
Screenshot 2025-02-11 at 12 41 08

The usual 9 tests fail.

Issue #1928, opened 2025-02-07
State Fusion does not work in `if True` `ConditionalBlock`


philip-paul-mueller:
State Fusion does not work in `if True` `ConditionalBlock`

Consider the following SDFG, with two nodes.

![Image](https://github.com/user-attachments/assets/e0fcb7de-d2b5-485b-bed2-50cc348a5edf)

The first node is a regular `SDFGState` with a single map, the second "state" is `ConditionalBlock` with a single branch whose condition is `True`, thus the branch is always entered and the Map within it is executed.
For optimization it makes sense to fuse these two nodes into one and run MapFusion.
However, running simplify on that SDFG has no effect.

I suspect that `StateFusion` does not consider this cases.

Here is a reproducer:
```python
import dace

def make_sdfg() -> dace.SDFG:
    sdfg = dace.SDFG("conditional_sdfg")

    for name in "abc":
        sdfg.add_array(
                name,
                shape=(10,),
                dtype=dace.float64,
                transient=False,
        )
    sdfg.arrays["b"].transient = True

    first_state = sdfg.add_state("first_state")
    first_state.add_mapped_tasklet(
            "first_comp",
            map_ranges={"__i0": "0:10"},
            inputs={"__in1": dace.Memlet("a[__i0]")},
            code="__out = __in1 + 10.0",
            outputs={"__out": dace.Memlet("b[__i0]")},
            external_edges=True,
    )

    # create states inside the nested SDFG for the if-branches
    if_region = dace.sdfg.state.ConditionalBlock("if")
    sdfg.add_node(if_region)
    sdfg.add_edge(
            first_state,
            if_region,
            dace.InterstateEdge()
    )

    then_body = dace.sdfg.state.ControlFlowRegion(
            "then_body",
            sdfg=sdfg
    )
    then_state = then_body.add_state("true_branch", is_start_block=True)
    if_region.add_branch(
            dace.sdfg.state.CodeBlock("True"),
            then_body
    )
    then_state.add_mapped_tasklet(
            "second_comp",
            map_ranges={"__i0": "0:10"},
            inputs={"__in1": dace.Memlet("b[__i0]")},
            code="__out = __in1 + 1.0",
            outputs={"__out": dace.Memlet("c[__i0]")},
            external_edges=True,
    )
    sdfg.validate()
    return sdfg

def test():
    sdfg = make_sdfg()
    sdfg.view()
    sdfg.simplify()
    sdfg.view()

if __name__ == "__main__":
    test()
[...]

PR #1930, opened 2025-02-10
Fix DSE pass for conditional block with single branch and unconditional execution

edopao:
This PR addresses #1928 by fixing some code in `DeadStateElimination` (DSE) pass. This code was supposed to handle exactly the case reported in the issue, but the check for single branch with unconditional execution was not entirely correct. Test case added.

Includes a change in fpga tests to address a CI error not related to this change:
https://github.com/spcl/dace/actions/runs/13242670532/job/36961472662?pr=1930


phschaad (approved):
LGTM, thank you!

PR #1932, opened 2025-02-11
[f2dace-dev, fortran] 1) Relax config injection to allow missing (i.e., pruned) references, 2) Enforce renaming keywords (like `this`) even when they are on the interface 3) Remove the logic to "skip" everything but radiation function as entry point.

pratyai:


acalotoiu (approved):
LGTM

PR #1933, opened 2025-02-11
Fix bug in cutout

ThrudPrimrose:
This fixes a bug in cutout for me.

### DIFF
--- a/dace/sdfg/analysis/cutout.py
+++ b/dace/sdfg/analysis/cutout.py
@@ -164,7 +164,7 @@ class SDFGCutout(SDFG):
                 cutout.translate_transformation_into(transformation)
             return cutout
         raise Exception('Unsupported transformation type: {}'.format(type(transformation)))
-                    
+
     @classmethod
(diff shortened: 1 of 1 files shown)


dace/sdfg/analysis/cutout.py
```diff
@@ -1059,7 +1059,7 @@ def _cutout_determine_output_configuration(ct: SDFG, cutout_reach: Set[SDFGState
                 check_for_read_after.add(dn.data)
 
         original_state: SDFGState = out_translation[state]
-        for edge in original_state.parent.out_edges(original_state):
+        for edge in original_state.parent.out_edges(original_state.parent_graph):
```
phschaad, line 1062: @ThrudPrimrose Off the top of my head, I think this actually needs more adaptation: `original_state.parent` still is an 'old-style' reference to the SDFG the original state belongs to. I believe this may need to be `original_state.parent_graph.out_edges(original_state.parent_graph)`. can you see if that fixes the tests?

PR #1934, opened 2025-02-11
[f2dace/dev, fortran] A single pass at const propagation + pruning does not always prune everything that is transitively possible to prune right away. So try a bit (so far 3-4 times is enough) to remove more and more things until the AST stabilises.

pratyai:
+ a bug fix
+ some cleanup


pratyai:
Screenshot 2025-02-12 at 09 44 00

The same 9.

acalotoiu (approved):
LGTM

PR #1936, opened 2025-02-14
[f2dace/dev, fortran] Avoid writing into (constantly reallocated) strings before writing into file. Also, improve the interface.

pratyai:
Now the F90 instrumentation is:
- As before, generate `serde.F90` and put it in ECRAD codebase.
- Add a `use serde` statement inside the `radiation` subroutine from `radiation_interface` module. This will import the following functions/subroutines: `tic()`, `at()`, `serialize()`.
- Add a `call tic()` at the beginning of the execution of `radiation`. This will increase the internal version counter.
- Add `call serialize(at("some/absolute/or/relative/pathprefix"), my_obj)` whenever we need to serialize `my_obj` into a file `some/absolute/or/relative/pathprefix.{VERSION}.data`.  The `at()` function will automatically open a versioned file and return the handler, leaving it open. Then `serialize()` subroutine will write into it, then close it before returning.
- NOTE: serializing arrays directly will also write metadata now.

And the C++ instrumentation is:
- Mostly as before, but now there is a `auto [m, arr] = read_array<T>(data);` will  first read the metadata from `data`, then read the actual array content based on that metadata.

### DIFF
--- a/dace/frontend/fortran/fortran_parser.py
+++ b/dace/frontend/fortran/fortran_parser.py
@@ -3145,13 +3145,6 @@ def name_and_rename_dict_creator(parse_order: list, dep_graph: nx.DiGraph) \
     return name_dict, rename_dict
 
 
-@dataclass
-class FindUsedFunctionsConfig:
-    root: str
-    needed_functions: List[str]
-    skip_functions: List[str]
-
-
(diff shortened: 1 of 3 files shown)


pratyai:
Screenshot 2025-02-14 at 02 23 23

PR #1938, opened 2025-02-15
[f2dace/dev, fortran] 1) Config injection module that does not rely on SDFG (so fairly stable), 2) A bunch of such config injection files, 3) Two scripts to produce the config injection and serde module without fiddling with the code too much.

pratyai:
- The produced config injection files have been checked for validity (it's mostly valid but with some caveats), but we are not actually using them yet.
- The boolean representations are somewhat inconsistent (sometimes 'true/false' and sometimes '0/1').


pratyai:
Only `serde_test.py` is run, because this PR doesn't touch anything else.

acalotoiu (approved):
LGTM

PR #1939, opened 2025-02-17
[f2dace/dev, fortran] 1) Switch to the new config injection files, 2) Add more "type injections" from ecrad to cover nested structures.

pratyai:


acalotoiu (approved):
LGTM *(Assuming we fix the issues/failing functions first)

PR #1940, opened 2025-02-18
Rewrite type inference to reapply after expression size changes

mcopik:
It fixes the well-known issue of determining the size of a complex expression involving transposition. Required multiple changes across AST transformations.

No new tests broken, and `get_albedos` compiles.

### DIFF
--- a/dace/frontend/fortran/ast_transforms.py
+++ b/dace/frontend/fortran/ast_transforms.py
@@ -2363,8 +2363,13 @@ class TypeInference(NodeTransformer):
         should_be_updated = False
         if node.lval.type == 'VOID':
             should_be_updated = True
+
         if self._get_sizes(node.lval) is None:
             should_be_updated = True
+
+        # We do NOT overwrite size of a varible if it has been defined by the user.
+        # We only do it for temporaries introduced by our transformations.
+        # At this moment, the only way to distinguish between them is the `tmp_` prefix.
         if isinstance(node.lval, ast_internal_classes.Name_Node) and node.lval.name.startswith("tmp_"):
             should_be_updated = True
 
@@ -3037,6 +3042,7 @@ class ReplaceStructArgsLibraryNodes(NodeTransformer):
                             ])
                         )
 
+                        # No need to create array subscript node - Array2Loop will catch that.
                         dest_node = ast_internal_classes.Name_Node(
                             name=tmp_var_name,
                             parent=call_node.parent,
--- a/dace/frontend/fortran/ast_transforms.py
+++ b/dace/frontend/fortran/ast_transforms.py
@@ -2484,14 +2484,14 @@ class TypeInference(NodeTransformer):
 
     def visit_Call_Expr_Node(self, node: ast_internal_classes.Call_Expr_Node):
 
-        from dace.frontend.fortran.intrinsics import MathFunctions
+        from dace.frontend.fortran.intrinsics import FortranIntrinsics
 
         new_args = []
         for arg in node.args:
             new_args.append(self.visit(arg))
         node.args = new_args
 
-        sizes, offsets, return_type = MathFunctions.output_size(node)
+        sizes, offsets, return_type = FortranIntrinsics.output_size(node)
         if sizes is not None:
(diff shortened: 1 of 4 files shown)


dace/frontend/fortran/ast_transforms.py
```diff
@@ -2365,7 +2358,22 @@ def visit_BinOp_Node(self, node: ast_internal_classes.BinOp_Node):
...
+        if node.lval.type == 'VOID':
+            should_be_updated = True
+
+        if self._get_sizes(node.lval) is None:
+            should_be_updated = True
+
+        # We do NOT overwrite size of a varible if it has been defined by the user.
+        # We only do it for temporaries introduced by our transformations.
+        # At this moment, the only way to distinguish between them is the `tmp_` prefix.
+        if isinstance(node.lval, ast_internal_classes.Name_Node) and node.lval.name.startswith("tmp_"):
```
pratyai, line 2373: Out of curiosity: In what scenario we would need to change the size of a temporary variable? The extracted variables' shapes are basically fully determined by the expression itself, aren't they?

dace/frontend/fortran/ast_transforms.py
```diff
@@ -3655,6 +3668,17 @@ def visit_Execution_Part_Node(self, node: ast_internal_classes.Execution_Part_No
             self.nodes_to_process = {}
             n = self.visit(child)
 
+            # Restart type inference to ensure that we reassign size
```
pratyai, line 3671: Out of curiosity without a full understanding of the type inference logic: Is it possible to get into an infinite loop of constantly reassigning the size of the same temporaries over and over?

pratyai (approved):

dace/frontend/fortran/ast_transforms.py
mcopik: No, they aren't. Some of the transformations extract the expression when size is unknown, e.g., recursive calls to intrinsics. Another example is the processing of non-contiguous slices where the size of the result changes.

dace/frontend/fortran/ast_transforms.py
mcopik: I don't think so, this part is not done in a loop.

dace/frontend/fortran/ast_transforms.py
pratyai: So, in the first example: size can go from an unknown state to a known state, IIUC. That sounds reasonable. Does the second example mean that the size can go from a known state to a different known state? Could you perhaps give a fortran example of when that change is needed?

mcopik: The famous case of transposing is in `get_albedos`. The argument's size changes once we process the non-contiguous slicing.

pratyai: (Please feel free to submit; I'm happy as long as I get my answers)

So, say we have this expression `tmp = TRANSPOSE(this_var_378 % lw_emissivity(istartcol_var_380 : iendcol_var_381, config_var_379 % i_emiss_from_band_lw(config_var_379 % i_band_from_reordered_g_lw)))`. Do you mean that the size of `tmp` will depend on the **content** of the operands of `transpose()`, e.g., `config_var_379 % i_emiss_from_band_lw(config_var_379 % i_band_from_reordered_g_lw)`?

mcopik: It will depend on the shape of that array, not the contents. This, in turn, will recursively depend on the shape of `config_var_379 % i_band_from_reordered_g_lw`. We need to extract each component and process it sequentially to get the size of `tmp`.

pratyai: Great! So `tmp`'s shape is really in an unknown state until we know all the nested shapes, IIUC. Is there any case where we would want to change a known shape into a different want (e.g. `(1:5, 4:7)` to `(1:15)`, or `(b0:e0)` to `(b1:e1)` where `b0,e0,b1,e1` are symbols)?

acalotoiu:
LGTM!

PR #1943, opened 2025-02-19
Cutensor

ThrudPrimrose:
I have added tensor transpose node for cutensor.
I would like you to review the structure of the changes. 
I want to use this change to  improve the transpose calls in ECRAD.


dace/libraries/linalg/__init__.py
```diff
@@ -1,5 +1,4 @@
 # Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
 from dace.library import register_library
-from .nodes import *
```
tbennun, line 3: you should not remove this statement from a library init, we use it for querying nodes

tbennun (changes requested):

alexnick83 (commented):
Is the corresponding tensordot implementation compatible with CuTENSOR v2?

ThrudPrimrose:
> Is the corresponding tensordot implementation compatible with CuTENSOR v2?

No, but if you approve the design of tensor transpose, I will update the tensordot too. It is possible that I will do it after CS, and first focus on optimizing index permutations on f2dace/dev branch,

dace/libraries/linalg/__init__.py
ThrudPrimrose: I see I will reintroduce it. I had a circular import issue with them and looked unnecessary in the first glance. 
I will reintroduce them and handle the circular import differently.

PR #1948, opened 2025-02-22
SoA Flattening Pass

ThrudPrimrose:
SoA Flattening Pass


ThrudPrimrose:
It also has the additional checks on Array elimination such that simplify does not break after flattening.

PR #1951, opened 2025-02-24
[f2dace/dev, fortran] 1) Consolidate global data objects into a single struct, 2) Then pass it around in the processed AST as function arguments, 3) **Try** to keep the global data serialisation and deserialisation in sync.

pratyai:


pratyai:
Screenshot 2025-02-24 at 21 57 52

acalotoiu (approved):
LGTM!

PR #1954, opened 2025-02-26
`StateFusion` misses read-write conflict due to early return

FlorianDeconinck:
Check for all potential match in `_check_paths` to not miss potential `memlets_intersect` failure.

Unit tests for checks the internal `_check_paths` function since it triggers on a non deterministic networkx search.

:warning:  This code exists as-is in `main` as of Feb 26 and showcase the same issue, cherry-picking fix will have to happen :warning:

### DIFF
--- a/dace/transformation/interstate/state_fusion.py
+++ b/dace/transformation/interstate/state_fusion.py
@@ -153,12 +153,15 @@ class StateFusion(transformation.MultiStateTransformation):
                 path_to = nx.has_path(first_state._nx, node, match)
                 if not path_to:
                     continue
-                path_found = True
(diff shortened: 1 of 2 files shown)


tbennun:
This looks reasonable, makes sense. I would add a test though with the SDFG API and a certain node addition ordering for sanity’s sake.

FlorianDeconinck:
Unit tests in. Instead of figuring out a test that would pass most check but the one we are fixing, I am testing that one check (`_check_paths`). The bug is subtle and depends on what `networkx` respond on a leaf search, which doesn't seem to be deterministic...

PR #1958, opened 2025-02-28
[f2dace/dev, fortran] 1) Refine the scripts in `tools/`, 2) Add config injection data for velocity tendencies, 3) Resolve "locally constant pointers" when possible, 4) If we cannot determine a pointer's resolution in serialisation module, dump the pointer's content instead.

pratyai:


acalotoiu (approved):
LGTM!

PR #1964, opened 2025-03-05
Fix: DDE removing read from access_set in read/write nodes

FlorianDeconinck:
Cherry picking fix to the data flow elimination path from V1/maintenance. Original PR: #1955

### DIFF
--- a/tests/passes/dead_code_elimination_test.py
+++ b/tests/passes/dead_code_elimination_test.py
@@ -58,8 +58,8 @@ def test_dse_edge_condition_with_integer_as_boolean_regression():
     state_init = sdfg.add_state()
     state_middle = sdfg.add_state()
     state_end = sdfg.add_state()
-    sdfg.add_edge(state_init, state_end, dace.InterstateEdge(condition='(not ((N > 20) != 0))',
-                                                             assignments={'result': 'N'}))
+    sdfg.add_edge(state_init, state_end,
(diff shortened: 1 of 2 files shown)


tests/passes/dead_code_elimination_test.py
```diff
@@ -266,6 +266,54 @@ def nested(a: dace.float64[20], b: dace.float64[20]):
...
+    next_state.add_memlet_path(second_tasklet, s2_write_tmp, src_conn="write_tmp", memlet=dace.Memlet(data="tmp"))
+    next_state.add_memlet_path(second_tasklet,
+                               s2_write_computed,
+                               src_conn="write_computed",
+                               memlet=dace.Memlet(data="computed"))
+
+    results = {}
+    Pipeline([DeadDataflowElimination()]).apply_pass(sdfg, results)
+
+    dde_results = results["DeadDataflowElimination"]
```
romanc, line 311:
```suggestion
    dde_results = results["DeadDataflowElimination"][0]
```

The structure of pipeline results on `main` is slightly different compared to `v1/maintenance`.

Can't push to your fork, but this is all it should (TM) take to get the "General Tests" green again. Not sure about the FPGA tests. That seems odd and might hint that this fix is too simplistic.

FlorianDeconinck: Let's push this to begin with

Issue #1967, opened 2025-03-07
Inconsistent Behaviour between GPU and CPU Maps With "funny" Sizes


philip-paul-mueller:
The executive summary would be something like:
- The behaviour of the CPU backend is not consistent with the one of the GPU backend.
- The CPU backend is able to handle Maps with zero or a negative amount of iterations.
- The GPU backend is able to handle Maps with zero iterations only if the size is symbolically.
- It is possible to create a GPU Map with zero or a negative number of iterations, known at code generation time, validation will not complain and it can even be launched, but CUDA will complain (fun fact it does not complain because the value is negative, but because it is too large, probably someone casts an `int` to an `unsigned int`).

The long answer is that we should look at the following (partial) SDFG:
```python
def make_sdfg(lower_bound, upper_bound, be):
    sdfg = dace.SDFG(...)
    state = sdfg.add_state(is_start_block=True)
    state.add_mapped_tasklet(
        "comp",
        map_range={"__i": f"({lower_bound}):({upper_bound})"},
        schedule=be,
        ...
    )
    ...
    return sdfg
```

Now let's consider the following SDFGs on CPU
- `make_sdfg("1", "1", DeviceType.CPU)`: This is a zero sized range on CPU which will run as expected, i.e. do nothing.
- `make_sdfg("1", "0", DeviceType.CPU)`: This is a for loop with a negative number of iterations, which will work as expected, i.e. do nothing.
- `make_sdfg("a", "b", DeviceType.CPU)`: This will give an SDFG in which the bounds of the loop depends on the parameter `a` and `b`, so
	- `sdfg(a=1, b=1)`: The loop has a size of zero and do nothing.
	- `sdfg(a=1, b=0)`: The loop has a negative size and will do nothing.

The reason why zero and negative sized loops works on CPU is, because they are essentially lowered to the following C construct:
```c++
for(auto __i = a; __i < b; __i += increment)
```
Because the check is performed using `<`, which is afaik a requirement from OpenMP, it can cope with zero and negative sized loops.

On GPU things are a bit different (who need consistency), let's look at some examples:
- `make_sdfg("1", "1", DeviceType.GPU)`: This is a loop with zero iterations, which is known at code generation time. It will result in a launch error.
[...]

PR #1968, opened 2025-03-07
Fix GPU kernel launch for non-symbolic-empty and negative grid size

philip-paul-mueller:
This commit partially addresses [issue#1967](https://github.com/spcl/dace/issues/1967), in the sense that it turns GPU kernels into a no-ops (i.e. does not start it at all) if a dimension of the iteration space has zero or negative size.

The code generator contained such a check, which was called "empty grid".
Thus in a sense this commit only clarifies the notion of what an empty grid is.
Before this was only detected if the size was dependent on runtime values, but not if the size was known at compilation time.
As explained in the issue, this leads to strange behaviour, for example the map range `{"__i": "a:b", "__j": "10:0"}` the code generator will add checks for `__i` but not for `__j`.
This was inconsistent with CPU backend.


phschaad (approved):
LGTM, thanks for the formatting

PR #1970, opened 2025-03-12
[f2dace/dev, fortran] Better type array test that actually tests things beyond compilation

pratyai:
~1. Validation is disabled because that's what breaks. But if we disable it the actual computation graph turns out just fine.
2. A C++ change is needed, because my `clang++` is convinced that it's not right.~

### DIFF
--- a/tests/fortran/type_array_test.py
+++ b/tests/fortran/type_array_test.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 
+from dace.frontend.fortran import fortran_parser
 from dace.frontend.fortran.fortran_parser import create_singular_sdfg_from_string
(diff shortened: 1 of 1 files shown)


dace/sdfg/validation.py
acalotoiu, line 1: We cannot disable this. We must fix/ or merge and fix later.

dace/runtime/include/dace/reduction.h
acalotoiu, line 1: Can you please separate this -  or have this be a local patch (can you check if the CI passes with these changes on main DaCe?)

acalotoiu (changes requested):
Thank you!

acalotoiu (approved):
LGTM!

acalotoiu (approved):

PR #1971, opened 2025-03-12
Codegen: Fix delete for non-array objects

pratyai:


dace/codegen/targets/cpu.py
```diff
@@ -578,14 +578,17 @@ def deallocate_array(self, sdfg: SDFG, cfg: ControlFlowRegion, dfg: StateSubgrap
...
+                callsite_stream.write(f"delete {alloc_name};\n", cfg, state_id, node)
         elif nodedesc.storage is dtypes.StorageType.CPU_ThreadLocal:
             # Deallocate in each OpenMP thread
             callsite_stream.write(
-                """#pragma omp parallel
+                f"""#pragma omp parallel
                 {{
-                    delete[] {name};
-                }}""".format(name=alloc_name),
+                    delete[] {alloc_name};
```
tbennun, line 590: Why not do it here as well?

tbennun (changes requested):
Thanks! I have this change in another PR, too, so it looks good to me. Quick nitpick on thread local storage but otherwise good to go

tbennun:
Note my other version of it here:  https://github.com/spcl/dace/pull/1431/files#diff-297f0b170ebd5a6798af2d60f65fed22f3171cfcf8eaf2fc17bcee1af255ba24R650 

@pratyai  your condition is different here but I think it should be ok.

dace/codegen/targets/cpu.py
pratyai: Done; PTAL.

tbennun (approved):

PR #1977, opened 2025-03-14
[f2dace/dev, fortran] Sort the order of processing variables. Since arrays may depend on other arguments, they should go last.

pratyai:


acalotoiu (approved):
LGTM! Thank you! Great fix!

acalotoiu (approved):
LGTM

PR #1978, opened 2025-03-17
[f2dace/dev, fortran] Extract the consistent part of the global data objects and display them 1) as plain keyword-value entries 2) as constant injections 3) an f90    module that asserts for those values

pratyai:


acalotoiu (approved):
Wow - that is fewer changes than I expected! Nice! LGTM.

PR #1979, opened 2025-03-17
Updated/fixed/finalized the `set_opts()` mechanism of the passes.

philip-paul-mueller:
This PR properly implements the `Pass.set_opts()` mechanism.
The format is such that the key must be prefixed with the name of the transformation for which the key should be set.
According to Philipp Schaad this was the intended behaviour, but it was not implemented properly.
Now the function is implemented once in the `Pass` and sub classes does not have to provide their own implementation.


phschaad (approved):
LGTM, thank you!

Issue #1959, opened 2025-03-03
`InlineMultistateSDFG` Destroys SSA Invariant


philip-paul-mueller:
This is not a bug per se, as the resulting SDFGs are correct, however, the `InlineMultistateSDFG` transformation destroys the SSA invariant of SDFGs, which is a less than optimal behaviour.

Consider the following SDFG:

![Image](https://github.com/user-attachments/assets/b844a85b-4180-4f09-b530-f32d45934434)

The first nested SDFG writes into `T1` which is then written into `T3[0:a]`, the second nested SDFG writes its output into `T2` which is then copied into `T3[a:N]`.
`T3` is then finally written into the global memory `G`.
As you can see al _transients_ are written to exactly once, for certain reason (I am looking at you FORTRAN) we can not enforce this on global data, but this is not the issue here.

After we run `InlineMultistateSDFG` we get the following SDFG:

![Image](https://github.com/user-attachments/assets/b2c07258-adbf-4283-b770-c6212caaa2e2)

As you can see now the writing of the output, i.e. `{(T1), (T2)} -> (T3) -> (G)` is now split and put into different states.
Thus the transients are no longer SSA.

Since SSA is a very nice and useful property I would think/expect/hope that the transformation would maintain this invariant.
Thus from this perspective it is a bug.

PR #1980, opened 2025-03-19
Updated MultistateInlining

philip-paul-mueller:
This PR addresses [issue#1959](https://github.com/spcl/dace/issues/1959), the issue was about that the inlining transformation destroys the SSA invariant we maintain inside GT4Py.
This invariant states that in every code path there is exactly one AccessNode with that is used to write to a data container.

The PR solves this issue by modifying how the nested SDFG is isolated, before this was done in two steps and is now performed in a single step.
The main idea is that the number of writes is preserved.

### DIFF
--- a/dace/transformation/helpers.py
+++ b/dace/transformation/helpers.py
@@ -730,7 +730,7 @@ def isolate_nested_sdfg(
     pre_nodes: Set[nodes.Node] = set()
     to_visit: List[nodes.Node] = []
     for iedge in state.in_edges(nsdfg_node):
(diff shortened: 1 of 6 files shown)


acalotoiu (approved):
This looks really nice and useful! Good idea with the three states! Thank you!

phschaad (approved):
LGTM, thanks! Should be orthogonal to my fixes too mostly

phschaad:
Weird failing test - can that be reproduced locally? It seems like a possible missing reshape / view when inlining, but seems a bit strange that only one test fails then.

PR #1982, opened 2025-03-20
[f2dace/dev, fortran] A bunch of small changes: 1) Fix `nproma=32` as a global, 2) Drop the obsolete `ti.F90` file, 3) Allow specifying certain variables and functions as "do not rename"

pratyai:


acalotoiu (approved):
Thanks!

PR #1986, opened 2025-04-28
Add NVTX range in CUDA GPU kernel call of program

philip-paul-mueller:
This changes were made by [Ioannis Magkanaris](https://github.com/iomaganaris), I only opened the PR.
It adds NVTX ranges around the kernel call generated by DaCe, this allows to easily distinguish it from other CUDA activity such as CuPy.


tbennun:
I agree. This should be possible to implement in a nicer way with the instrumentation API, instrumenting the SDFG. This could even be implemented in Python with the SDFG call hooks. 
If it is part of the instrumentation, it is also not going to be enabled always by default (the calls might add overhead for very short microsecond-scale SDFGs). Here is an example of how to do it in CuPy, which should be even more portable towards AMD GPUs:
https://docs.cupy.dev/en/latest/reference/generated/cupy.cuda.nvtx.RangePush.html

PR #1965, opened 2025-03-06
Added Parallel Map Fusion

philip-paul-mueller:

PR #1987, opened 2025-04-28
MapFusion Multi Producer Intermediate

philip-paul-mueller:
Until now it was only possible to fuse two Maps if the intermediate node has a single producer (the top Map).
But this is not strictly required, it can have as multiple producer, the important thing is that the second Map only consumes data that have been written by the first Map.


alexnick83 (changes requested):
It is a great idea to have the standalone capability to fuse parallel Maps rather than relying on SubgraphFusion. Relaxing the single-consumer constraint for the serial fusion's intermediate also makes sense to me.

My main question is whether an implementation of parallel fusion as a separate transformation has been considered, and what the trade-offs are. My understanding is that the fused implementation was chosen to better accommodate the FullMapFusion pass, but I wonder if the alternative would be more flexible. If this has been discussed and decided already internally, then I am fine. Otherwise, I would also ask for a review from someone currently actively working on transformations/auto-optimizer for a second opinion.

philip-paul-mueller:
The parallel fusion of Maps (currently an [open PR](https://github.com/spcl/dace/pull/1965)) is from an implementation point of view, a subset of the functionality you need to handle serial Map fusion.
Thus to avoid code duplication, you will end up with some kind of base class containing the common aspects.

In the very first version of the [PR that introduced the new version of MapFusion](https://github.com/spcl/dace/pull/1629) there was this distinction between parallel and serial MapFusion.
However, it was decided that the parallel aspects should be removed to make the PR reviewable, or make it more managable.
Therefore no decision was made if they should be split to not.

This "subset aspect of the implementation" makes a split implementation very strange, you have a base, which essentially contains everything that is needed to implement `relocate_node()`.
Then you have a class that implements parallel Map fusion, which essentially alias `relocate_node()` to `apply()` the only new thing it essentially adds is `is_parallel()`.
And the idea to derive the serial Map fusion from the parallel version is just awkward.

However, my issue with splitting the implementation is, that I think there is no reason why a class that is called `MapFusion` should be unable to handle parallel Maps (okay the current default is to reject such cases).
If we split the implementation, then we should remove the name `MapFusion` completely and only retain `SerialMapFusion` and `ParallelMapFusion` (or variants thereof).

But I think we should discuss this aspect at the next DaCe meeting.

PS: I will not apply your correction yet, because I have multiple PRs about that subject open, so it get's complicated.

dace/sdfg/propagation.py
```diff
@@ -1263,6 +1263,40 @@ def propagate_memlets_scope(sdfg, state, scopes, propagate_entry=True, propagate
...
+    The main difference to `propagate_memlets_scope()` is that this function operates on Maps
+    instead of `ScopeTree` it is thus much more accessible.
+    The function will first propagate the Memlets of the nested SDFGs that are enclosed by the
+    Map. Then the propagation will start bit only for those Melets that starts with `map_entry`.
+
+    :param sdfg: The SDFG in which the scopes reside.
+    :param state: The SDFG state in which the scopes reside.
+    :param map_entry: Defining the Map scope to which propagation should be restricted.
+    """
+    assert isinstance(map_entry, nodes.MapEntry)
```
alexnick83: I suppose the assertions are there for debugging or as sanity checks. However, if they actually fail during execution, the error message will not be immediately helpful. I suggest either converting me to exceptions (e.g., the above should probably be ValueError) or removing them, if there is no way to trigger them in practice.

dace/transformation/dataflow/map_fusion_helper.py
```diff
@@ -0,0 +1,460 @@
...
+        return
+
+    second_map_scope = state.scope_subgraph(entry_node=second_map_entry)
+    # Why is this thing in symbolic and not in replace?
+    symbolic.safe_replace(
+        mapping=repl_dict,
+        replace_callback=second_map_scope.replace_dict,
+    )
+
+    # For some odd reason the replace function does not modify the range and
```
alexnick83, line 158: Because these attributes belong to the Map object, while the scope contains the entry/exit nodes. The Map object can be accessed from the entry/exit nodes, but we do not do "recursive" replacement in attributes. In retrospect, it may make sense to handle Maps and their entry/exit nodes in a special manner, but I would raise this in a meeting to discuss whether there are any pitfalls to such an approach.

PR #1988, opened 2025-04-28
LoopNormalization Transformation

Berke-Ates:
This PR adds a loop normalization transformation, transforming loops to start at zero with a step size of one.


tests/transformations/interstate/loop_normalize_test.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad, line 1: Outdated

phschaad (changes requested):
I am not a huge fan of having this kind of functionality be a pattern transformation. I am of the opinion that something like this is general-purpose enough that it warrants having a utility function somewhere that performs this operation. Whether the utility function can then be exposed through a pattern matching transformation can be discussed. However, there are a few reasons I am not in favor of it being a pattern matching transformation:
1) The pattern is a single node - that's often not a good sign
2) The operation is in my eyes primarily an optimization preparation, rather than an actual optimization itself. That means that when gathering applicable transformations for a program to improve performance, a transformation like this unnecessarily clutters the results.
3) The operation is difficult to reuse. Given that it serves as an optimization preparation, that makes it so it is useful to apply from many other transformations / passes. Having it be a transformation means that doing so requires a lot of boiler-plate (setup a match, etc.).

As such, I would be in favor of having the same functionality given through 2 loop utility functions:
1) Can a loop be normalized
2) Normalize a loop

This would also make it possible to expose these functions through the `LoopRegion`'s own API easily.

dace/transformation/interstate/loop_normalize.py
```diff
@@ -0,0 +1,180 @@
...
+from dace.sdfg import utils as sdutil
+from dace.sdfg.state import ControlFlowRegion, LoopRegion
+from dace.transformation import transformation as xf
+from dace.transformation.passes.analysis import loop_analysis
+from dace.sdfg.nodes import CodeBlock
+from dace.symbolic import pystr_to_symbolic
+
+
+# Returns a list of all defined symbols in the loop body
+def _defined_symbols(loop: LoopRegion) -> Set[str]:
```
phschaad, line 19: This is performed twice per `can_be_applied`, that is unnecessarily expensive.

phschaad (approved):
Thank you for the change, this is exactly what I had in mind. Looks good to me!

PR #1989, opened 2025-04-28
SymbolPropagation Pass

Berke-Ates:
This PR adds a symbol propagation pass, which propagates symbolic expressions through the graph.

### DIFF
--- a/dace/transformation/passes/symbol_propagation.py
+++ b/dace/transformation/passes/symbol_propagation.py
@@ -1,0 +1,248 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
+
+import copy
+from dataclasses import dataclass
+from dace.sdfg.state import (
+    ControlFlowBlock,
+    ControlFlowRegion,
+    ConditionalBlock,
+    LoopRegion,
+)
+from dace.transformation import pass_pipeline as ppl, transformation
+from dace import SDFG, properties, SDFGState
+from typing import Any, Dict, Set, Optional
+from dace import data as dt
+from dace.symbolic import pystr_to_symbolic
+
+
+@dataclass(unsafe_hash=True)
+@properties.make_properties
+@transformation.explicit_cf_compatible
+class SymbolPropagation(ppl.Pass):
+    """
+    Propagates symbols that were assigned to one value forward through the SDFG, reducing the number of overall symbols.
+    """
+
+    CATEGORY: str = "Simplification"
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Symbols | ppl.Modifies.Edges | ppl.Modifies.Nodes
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        # If anything was modified, reapply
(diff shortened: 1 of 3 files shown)


acalotoiu (approved):
LGTM

dace/transformation/passes/symbol_propagation.py
```diff
@@ -0,0 +1,214 @@
...
+                for sym in all_syms:
+                    if sym in new_in_syms:
+                        new_in_syms[sym] = None
+
+        return new_in_syms
+
+    # Given a CFGB, builds the outgoing set of symbols
+    def _get_out_syms(
+        self,
+        cfgb: ControlFlowBlock,
```
phschaad, line 127: For disambiguation / avoiding confusion it may be a good idea to call these kinds of variables `block` / `node` or something similar, `cfgb` immediately made me think it's a control flow graph / region with a typo.. may be a me issue, but a suggestion anyway ;-)

dace/transformation/passes/symbol_propagation.py
```diff
@@ -0,0 +1,214 @@
...
+        parent: ControlFlowRegion,
+        in_syms: Dict[ControlFlowBlock, Dict[str, Any]],
+        out_syms: Dict[ControlFlowBlock, Dict[str, Any]],
+    ) -> Dict[str, Any]:
+        if isinstance(cfgb, LoopRegion):
+            # Any symbol that is assigned in the loop region is not propagated out
```
phschaad, line 133: Wait, you mean assigned through the update/init statements? Because symbols assigned on interstate edges definitely propagate out, we don't consider a nested control flow graph an isolated symbol scope.

dace/transformation/passes/symbol_propagation.py
```diff
@@ -0,0 +1,214 @@
...
+
+            return new_out_syms
+
+        elif isinstance(cfgb, SDFGState):
+            # Cannot change symbols in SDFGStates
+            return in_syms[cfgb]
+
+        else:
+            # Use sink symbols as outgoing symbols
+            sink_nodes = [n for n in cfgb.nodes() if cfgb.out_degree(n) == 0 and isinstance(n, ControlFlowBlock)]
```
phschaad, line 160: 2 things:
1. This seems to not consider non-sink exits, such as return, continue, break, correct?
2. Why is this handled separately from loop regions?

phschaad (changes requested):
Some questions I need to have answered first

dace/transformation/passes/symbol_propagation.py
Berke-Ates: What is meant by this comment is that SymbolPropagation does not propagate the assignments inside loops to the outside. Imagine a loop with `a = some_val` in an interstate edge inside and `b = a` in an interstate edge after the loop. Performing the replacement `b = some_val` would be incorrect, since we don't know if the loop will be executed at all and thus, we don't know which value `a` has.

dace/transformation/passes/symbol_propagation.py
Berke-Ates: Done :)

PR #1990, opened 2025-04-28
ContinueToCondition Transformation

Berke-Ates:
This PR adds a transformation which replaces continue blocks with conditional blocks.

### DIFF
--- a/dace/transformation/interstate/continue_to_condition.py
+++ b/dace/transformation/interstate/continue_to_condition.py
@@ -25,7 +25,6 @@ class ContinueToCondition(xf.MultiStateTransformation):
 
         pg = self.cb.parent_graph.parent_graph
 
-
         if not isinstance(pg, ConditionalBlock):
             return False
         if len(pg.branches) != 1:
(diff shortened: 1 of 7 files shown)


acalotoiu (changes requested):
LGTM! I think this one also needs more negative tests and more complex scenarios such as interleaved conditions and loops

dace/transformation/interstate/continue_to_condition.py
```diff
@@ -0,0 +1,86 @@
...
+            return False
+
+        outer_cfg = pg.parent_graph
+        if len(outer_cfg.successors(pg)) > 1:
+            return False
+
+        return True
+
+    def apply(self, graph: ControlFlowRegion, sdfg: sd.SDFG):
+        # If there are no successors, we can just remove the continue node
```
phschaad, line 46: No, certainly not always! A continue block and the conditional it is contained in and pattern matched to can be nested N levels deep inside of a loop and / or further conditionals. That means that you need to check for any 'logical' successors too. This is only applicable if the conditional is nested _directly_ inside of a loop region, and the conditional has no successors.

phschaad (changes requested):
Aside from the limitation pointed out in the comment, I am again not in favor of having this be a pattern transformation, for reasons similar to the ones outlined in #1988. This here sounds like:
1) The pattern checked would need to be "A loop, directly containing a conditional, directly containing a single, lonesome continue block".
2) This would be part of a "control flow simplification" component in simplify, much rather than a pattern transformation.

phschaad (approved):
LGTM, thanks!

Please run the pre-commit to format :)

PR #1991, opened 2025-04-28
ConditionFusion Transformation

Berke-Ates:
This PR adds the condition fusion transformation which fuses directly consecutive or nested conditional blocks.


acalotoiu (changes requested):
I think this needs more tests - both with more branches, more nesting and "negative" tests

acalotoiu (approved):
Thanks! LGTM!

phschaad (approved):
LGTM.

We may want to consider amending / extending our pattern matching a bit to consider these kinds of cases where we have region X nested inside of region Y - or in general to consider some amount of nesting chains. Expressing such things as pattern-matching on a single node (as is the case with the second pattern expr. here) is always a bit sad. But in this case there's no nice way around that (yet).

PR #1995, opened 2025-04-29
Make CUDACodeGenerator reentrant

edopao:
GT4Py-next is introducing support for parallel build of programs with static arguments. In combination with the dace backend, this feature results in multiple SDFG variants that differ for the specialization of some static scalar arguments. By means of a thread pool, the SDFGs should be optimized and compiled in parallel.

The global attribute `CUDACodeGen._in_device_code` currently makes the dace backend non-reentrant. This PR makes `_in_device_code` an instance attribute.

### DIFF
--- a/dace/codegen/targets/cuda.py
+++ b/dace/codegen/targets/cuda.py
@@ -1531,10 +1531,9 @@ void __dace_alloc_{location}(uint32_t {size}, dace::GPUStream<{type}, {is_pow2}>
                     if not defined_type:
                         defined_type, ctype = self._dispatcher.defined_vars.get(ptrname, is_global=is_global)
 
-                    oldval = self._in_device_code
                     self._in_device_code = True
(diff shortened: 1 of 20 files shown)


dace/codegen/targets/cuda.py
```diff
@@ -60,7 +61,7 @@ class CUDACodeGen(TargetCodeGenerator):
     """ GPU (CUDA/HIP) code generator. """
     target_name = 'cuda'
     title = 'CUDA'
-    _in_device_code = False
+    _in_device_code = contextvars.ContextVar('_in_device_code')
```
edopao, line 64: Before merging, I will add a comment here:
`# Use a ContextVar attribute below in order to make CUDACodeGen thread-safe.`

tbennun (approved):
Even better, I like it a lot more!

PR #1996, opened 2025-04-30
Updated the validation regarding `other_subset`

philip-paul-mueller:
If the validation founds an out of bound access for in the subset of a Memlet that is marked as `dynamic` it will only output a warning. If it however, finds this oob behaviour in `other_subset` it will generate an hard error.

This commit changes this behaviour such that also for `other_subset` only a warning is generated, makes them more uniform.


alexnick83 (approved):
LGTM

tbennun (approved):

PR #2001, opened 2025-05-07
Add NVTX ranges to top level SDFG

iomaganaris:
- Added NVTX markers to top level SDFG in the CPU code to distinguish and measure easily the DaCe runtime using NVIDIA profiling tools
- Added related test and documentation
- Selected the `GPU_TX_MARKERS` name since there is also [rocTX](https://rocm.docs.amd.com/projects/roctracer/en/latest/reference/roctx-spec.html) for AMD GPUs
  - I didn't implement the changes for it though since I wasn't able to test them with the latest `ROCm`. AMD changed the way the library is distributed from the `roctracer` package which is deprecated to `ROCm` in latest versions. I left this as a todo for myself once I get access to a system with it to test it


alexnick83 (commented):
I like the implementation. I have a few questions/comments:
- In the documentation, it says that it "wraps with NVTX markers the CUDA kernels launched from the top level SDFG." However, with the current implementation, there is only going to be a single NVTX range for the whole top-level SDFG. Which exactly is the intended functionality?
- Do we really want this functionality on the GPUEventProvider at the same time (instead of running twice with separate instrumentation)? What is the use case here?
- Regarding ROCm, I think it would make sense to also support at least v.6.2/6.3, since many supercomputers are still in that version.

I would also ask @tbennun to also have a look since he wrote the original instrumentation API.

iomaganaris:
Hi @alexnick83 ,

> In the documentation, it says that it "wraps with NVTX markers the CUDA kernels launched from the top level SDFG." However, with the current implementation, there is only going to be a single NVTX range for the whole top-level SDFG. Which exactly is the intended functionality?

Maybe I should rephrase this to "wraps with NVTX markers the GPU activity of the top level SDFG.". The intention of this PR is to mark whatever happens inside the `__program_stencil_name_internal()` function so that all the CUDA kernels launched from the top level SDFG are inside this single NVTX range. What this PR does is the adding the following to the CPU code:
```
#include <nvtx3/nvToolsExt.h>
...
void __program_stencil_name_internal(args)
{
    __state->report.reset();
    nvtxRangePushA("stencil_name");

    {
       ...
    }
    nvtxRangePop();
    __state->report.save(...);
}
```

> Do we really want this functionality on the GPUEventProvider at the same time (instead of running twice with separate instrumentation)? What is the use case here?

No strong preference here or a solid use case. My idea was that since we might want to gather more detailed information per map with `GPUEventProvider` why not add the NVTX markers there since they shouldn't add much of an overhead? If you prefer to have things separate I can change this behavior.

> Regarding ROCm, I think it would make sense to also support at least v.6.2/6.3, since many supercomputers are still in that version.

Okay, I can look into this but I must say it's a bit painful to do atm since we have to add an extra dependency of this library that might or might not be installed in the system since it doesn't come with ROCm.

tbennun:
1. Part of the point of using instrumentation is to allow any SDFG/region to use custom nvtx region. This would be a very strong PR if it did
2. I second the AMD comment. The roctx API is stable between versions, it would be the same code to include roctracer.h (or roctracer/roctracer.h) and add the markers. Assume rocm 6.2-6.3 (as Alex said) for linkage instructions. If you don’t have access to an AMD machne, let me know if you want to test a configuration and on which version of rocm to do so.

iomaganaris:
Hi @tbennun ,

> Part of the point of using instrumentation is to allow any SDFG/region to use custom nvtx region. This would be a very strong PR if it did

Could you elaborate a bit more what you have in mind? The maps are already annotated since they are CUDA/HIP kernels. Would you like to add NVTX/rocTX ranges for example in states or nested SDFGs?
Already the current changes in this PR are pretty helpful for us and allow to to identify exactly how much time is spent in which DaCe SDFG.

Also I've added the ability to use `rocTX` markers as well. I tested it with `rocm 6.2.1` and it worked. The way to add link and include the necessary files it a bit ugly admittedly though 🙈

dace/codegen/instrumentation/gpu_events.py
```diff
@@ -3,20 +3,15 @@
...
-from dace.codegen import common
-from dace.codegen.instrumentation.provider import InstrumentationProvider
+from dace.codegen.instrumentation.gpu_tx_markers import GPUTXMarkersProvider
 from dace.sdfg.sdfg import SDFG
 from dace.sdfg.state import ControlFlowRegion, SDFGState
 
 
 @registry.autoregister_params(type=dtypes.InstrumentationType.GPU_Events)
-class GPUEventProvider(InstrumentationProvider):
+class GPUEventProvider(GPUTXMarkersProvider):
```
tbennun, line 12: Why is the old marker extending the new one?

tbennun (changes requested):
first of all, great work on roctx support!
I think there is a huge overlap between this and events, and an opportunity to make this code both shorter and more capable.

What I meant in my previous message is that ranges can be pushed recursively, which can allow people to potentially track both SDFGs and the internal states, or give one map an opportunity to appear with a marker range in the timeline. It is about making the code more general rather than just implementing the top level SDFG you want to achieve.

If you look at the GPU events instrumentation class, you can see that we already support the beginning and end of all the relevant SDFG constructs (map, state, SDFG) and we use some internal methods to generate the recording functions.

What if the GPUTX instrumentation provider overrode the GPU event provider and, for example, implemented `_record_event` and `_report`?
You can also provide empty implementations for `_create_event` and `_destroy_event`.

dace/codegen/instrumentation/gpu_events.py
iomaganaris: In the end I have deduplicated the `GPU_Event` and the `GPU_TX_MARKERS` providers since they should be used for different purposes

iomaganaris:
Thank you very much for your detailed explanation. Based on your comments I have also marked the different `state` and `SDFG` regions with `NVTX`/`rocTX` markers. I have also completely decoupled the `GPU_Event` and `GPU_TX_MARKERS` providers since they don't really share anything. `GPU_Event` should be used for profiling the GPU code without any profiler while `GPU_TX_MARKERS` are useful only when using `NVIDIA NSight Systems/Compute` or `rocprofiler-systems/compute/rocprof`. Those profilers also can mark the CUDA/HIP kernels executed on the GPU without any additional code.
What I haven't really understood is what the `copy`, `node` and `scope` functions from the `InstrumentatorProvider` annotate. I could only create an example for `copy` where I saw that the only related code can be printed in the `GPU` side while the `NVTX` markers reside only on the `CPU` code.
Also I have the impression that we can't determine if the code that is generated for `SDFG`s for `reduce` and `if` statements is compiled for `CPU` or `GPU`. That's why I had to wrap the `NVTX/rocTX` functions in macros so that they are not compiled in the device code.
I hope that I improved this PR towards what you have in mind. Please let me know if you have any other suggestion.

alexnick83:
It looks pretty good, but you are missing the instrumentation checks. Have a look at [TimerProvider](https://github.com/spcl/dace/blob/508323e63031f0faa3da469a791059068075b189/dace/codegen/instrumentation/timer.py). For every "element," you need the following check:
```python
if element.instrument == dtypes.InstrumentationType.GPU_TX_MARKERS:
    <do-something>
```
This way, you can control which states/scopes get the markers.

Other comments:
- The only copies (direct edges between two access nodes) that you can effectively mark are those where at least one of {src, dst} is on the host. Essentially, you could have a range for copies between host and GPU-global memory. Asynchronous calls will probably only measure the invocation, though.
- Example of scopes are Maps. If a Map defines a GPU kernel, then you would create a range around the (I guess) invocation of the kernel. Therefore, the usability may be limited in practice. @tbennun I guess we do not have a way to move `on_scope_exit` to the synchronization point, or is there some solution here?
- I think nodes cover anything else, so you could target CodeNodes (see TimerProvider again), for example, a cuBLAS call. Asynchronous calls will probably only measure the invocation, though.
- If you want to figure out whether an element is within a GPU kernel or not, you may want to check the `is_devicelevel` helper methods, for example, [`is_devicelevel_gpu_kernel`](https://github.com/spcl/dace/blob/d8b0347725204341c656c9be72d5221b736aa4af/dace/sdfg/scope.py#L235).

tbennun:
I agree with @alexnick83 's comments

> * Example of scopes are Maps. If a Map defines a GPU kernel, then you would create a range around the (I guess) invocation of the kernel. Therefore, the usability may be limited in practice. @tbennun I guess we do not have a way to move `on_scope_exit` to the synchronization point, or is there some solution here?

You're right, we don't have a way to get the synchronization point of a specific map. There are two ways to handle this:
1. Use what GPU_Events does and add an event-synchronize (and event-record on scope entry) if a marker range was requested for a specific map
2. Not support scopes

iomaganaris:
Hi Alex and Tal,

Thank you very much for your feedback. I will apply the necessary changes. I have some suggestions regarding `node`s, `scope`s and `copies`.
1. `node`: `CodeNode` that are CUDA/HIP API are automatically annotated by the profilers (using `--trace cuda` or `--hip-trace`) so there's no additional marking necessary. If `CodeNode`s are only calls to these libraries I would avoid adding new markers for them since there's no added benefit
2. `scope`: Maps become CUDA kernels so they are also automatically annotated by the profiler as well so I would skip the markers for them as well
3. `copy`: All types of copies that are done using `cuda/hip` API are also tracked by the profilers. I guess the only added benefit of adding a marker on top of them is associating the API call with the names of the variables. Is there somewhere an example to test this? I could only find examples where the `cuda/hip memcpy` functions were generated inside a state block.

> If you want to figure out whether an element is within a GPU kernel or not, you may want to check the is_devicelevel helper methods, for example, [is_devicelevel_gpu_kernel](https://github.com/spcl/dace/blob/d8b0347725204341c656c9be72d5221b736aa4af/dace/sdfg/scope.py#L235).

I found a solution using `is_devicelevel_gpu_kernel` but I'm not sure if going through of all the states and nodes of the SDFGs is a good idea to determine if an SDFG is in device. Also I'm not checking about NestedSDFGs inside NestedSDFGs. Are there cases like that?

dace/codegen/instrumentation/gpu_tx_markers.py
```diff
@@ -0,0 +1,162 @@
...
+                stream.write('roctxRangePop();')
+        else:
+            raise NameError(f'GPU backend "{self.backend}" not recognized')
+
+    def _is_sdfg_in_device_code(self, sdfg: SDFG) -> bool:
+        """ Check if the SDFG is in device code and not top level SDFG. """
+        if sdfg.parent is None:
+            return False
+
+        for state in sdfg.parent.sdfg.states():
```
tbennun: you can get those with `sdfg.parent_nsdfg_node` and `sdfg.parent_graph`, without any potentially expensive loops involved.

dace/codegen/instrumentation/gpu_tx_markers.py
```diff
@@ -0,0 +1,162 @@
...
+            self.print_range_pop(local_stream)
+
+    def on_copy_begin(self, sdfg: SDFG, cfg: ControlFlowRegion, state: SDFGState, src_node: nodes.Node,
+                      dst_node: nodes.Node, edge: MultiConnectorEdge[Memlet], local_stream: CodeIOStream,
+                      global_stream: CodeIOStream, copy_shape, src_strides, dst_strides) -> None:
+        if state.instrument == dtypes.InstrumentationType.GPU_TX_MARKERS:
+            if is_devicelevel_gpu_kernel(sdfg, state, src_node):
+                # Don't instrument device code
+                return
+            self.print_range_push(f'copy_{src_node.label}_to_{dst_node.label}', sdfg, global_stream)
```
tbennun: global_stream? Shouldn't this be local_stream?

dace/codegen/instrumentation/gpu_tx_markers.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def on_copy_end(self, sdfg: SDFG, cfg: ControlFlowRegion, state: SDFGState, src_node: nodes.Node,
+                    dst_node: nodes.Node, edge: MultiConnectorEdge[Memlet], local_stream: CodeIOStream,
+                    global_stream: CodeIOStream) -> None:
+        if state.instrument == dtypes.InstrumentationType.GPU_TX_MARKERS:
+            if is_devicelevel_gpu_kernel(sdfg, state, src_node):
+                # Don't instrument device code
+                return
+            self.print_range_pop(global_stream)
```
tbennun: same

dace/codegen/instrumentation/utils.py
```diff
@@ -0,0 +1,19 @@
+from dace import dtypes, nodes as dace_nodes
```
tbennun: Missing copyright

dace/codegen/instrumentation/utils.py
```diff
@@ -0,0 +1,19 @@
...
+from dace.sdfg.sdfg import SDFG
+
+
+def instrument_sdfg(instrumentation_type: dtypes.InstrumentationType, sdfg: SDFG) -> None:
```
tbennun: is this necessary, given that `dace.instrument` exists? https://github.com/spcl/dace/blob/main/dace/builtin_hooks.py#L109

dace/codegen/targets/cuda.py
```diff
@@ -995,6 +995,11 @@ def _emit_copy(self, state_id: int, src_node: nodes.Node, src_storage: dtypes.St
                 else:
                     raise NotImplementedError('2D copy only supported with one stride')
 
+            for instr in self._dispatcher.instrumentation.values():
+                if instr is not None:
+                    instr.on_copy_begin(sdfg, cfg, state_dfg, src_node, dst_node, edge, None, callsite_stream,
```
tbennun: passing the callsite_stream as the global stream instead of local (related to issue in `gpu_tx_markers.py`)

dace/codegen/targets/cuda.py
```diff
@@ -1098,6 +1103,10 @@ def _emit_copy(self, state_id: int, src_node: nodes.Node, src_storage: dtypes.St
 
             self._emit_sync(callsite_stream)
 
+            for instr in self._dispatcher.instrumentation.values():
+                if instr is not None:
+                    instr.on_copy_end(sdfg, cfg, state_dfg, src_node, dst_node, edge, None, callsite_stream)
```
tbennun: same as above

dace/codegen/targets/framecode.py
```diff
@@ -257,6 +258,13 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
         # Write closing brace of program
         callsite_stream.write('}', sdfg)
 
+        if sdfg.instrument is dtypes.InstrumentationType.GPU_TX_MARKERS:
+            # Need to make sure that the necessary includes for GPU_TX_MARKERS are present
+            # in the generated code.
+            gpu_tx_markers_provider = self._dispatcher.instrumentation.get(dtypes.InstrumentationType.GPU_TX_MARKERS)
+            if gpu_tx_markers_provider:
+                gpu_tx_markers_provider.print_include(callsite_stream)
```
tbennun: Is this section necessary anymore, now that you are calling the include printout in the instrumentation provider directly?

dace/codegen/targets/framecode.py
```diff
@@ -282,11 +290,18 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
...
-{{
-    int __result = 0;
-    {mangle_dace_state_struct_name(sdfg)} *__state = new {mangle_dace_state_struct_name(sdfg)};
+{{""", sdfg)
+
+        if sdfg.instrument is dtypes.InstrumentationType.GPU_TX_MARKERS:
+            gpu_tx_markers_provider = self._dispatcher.instrumentation.get(dtypes.InstrumentationType.GPU_TX_MARKERS)
+            # cannot push rocTX markers before initializing HIP
+            if gpu_tx_markers_provider and not gpu_tx_markers_provider.enable_rocTX:
+                gpu_tx_markers_provider.print_range_push(f'init_{sdfg.name}', sdfg, callsite_stream)
```
tbennun: HIP will be initialized below in line 292/307. Why is this section necessary?

dace/codegen/targets/framecode.py
```diff
@@ -307,17 +322,30 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
...
     if (__result) {{
         delete __state;
         return nullptr;
     }}
+""", sdfg)
+        if sdfg.instrument is dtypes.InstrumentationType.GPU_TX_MARKERS:
+            gpu_tx_markers_provider = self._dispatcher.instrumentation.get(dtypes.InstrumentationType.GPU_TX_MARKERS)
+            # cannot push rocTX markers before initializing HIP
+            if gpu_tx_markers_provider and not gpu_tx_markers_provider.enable_rocTX:
+                gpu_tx_markers_provider.print_range_pop(callsite_stream)
```
tbennun: same question as before

dace/codegen/targets/framecode.py
```diff
@@ -307,17 +322,30 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
...
 }}
 
 DACE_EXPORTED int __dace_exit_{sdfg.name}({mangle_dace_state_struct_name(sdfg)} *__state)
 {{
+""", sdfg)
+        if sdfg.instrument is dtypes.InstrumentationType.GPU_TX_MARKERS:
+            gpu_tx_markers_provider = self._dispatcher.instrumentation.get(dtypes.InstrumentationType.GPU_TX_MARKERS)
+            if gpu_tx_markers_provider:
+                gpu_tx_markers_provider.print_range_push(f'exit_{sdfg.name}', sdfg, callsite_stream)
+        callsite_stream.write(f"""
```
tbennun: same

dace/codegen/targets/framecode.py
```diff
@@ -352,6 +380,10 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
                 callsite_stream.write("}")
 
         callsite_stream.write('delete __state;\n', sdfg)
+        if sdfg.instrument is dtypes.InstrumentationType.GPU_TX_MARKERS:
+            gpu_tx_markers_provider = self._dispatcher.instrumentation.get(dtypes.InstrumentationType.GPU_TX_MARKERS)
+            if gpu_tx_markers_provider:
+                gpu_tx_markers_provider.print_range_pop(callsite_stream)
```
tbennun: if you want to be able to instrument init/exit codes, you should add another `on_sdfg_{init,exit}_{begin,end}()` event API calls to `InstrumentationProvider`. Otherwise this is a special case that looks odd.

dace/codegen/targets/framecode.py
```diff
@@ -257,6 +258,13 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
         # Write closing brace of program
         callsite_stream.write('}', sdfg)
 
+        if sdfg.instrument is dtypes.InstrumentationType.GPU_TX_MARKERS:
```
tbennun: Please use `==` instead of `is` for future proofing. (applies to the rest of the cases in this file)

doc/optimization/profiling.rst
```diff
@@ -121,7 +121,8 @@ Instrumentation can also collect performance counters on CPUs and GPUs using `LI
 The :class:`~dace.dtypes.InstrumentationType.LIKWID_Counters` instrumentation type can be configured to collect
 a wide variety of performance counters on CPUs and GPUs. An example use can be found in the
 `LIKWID instrumentation code sample <https://github.com/spcl/dace/blob/main/samples/instrumentation/matmul_likwid.py>`_.
-
+There is also the :class:`~dace.dtypes.InstrumentationType.GPU_TX_MARKERS` instrumentation type which wraps in NVTX or rocTX markers the DaCe program executed on the GPU. Important parts of the execution of the program on the GPU as the different states, SDFGs and initialization and finalization phases are marked with these markers.
+These markers can be used to visualize and measure the GPU activity using the NVIDIA Nsight Systems or rocprof/rocprofile-systems profilers.
```
tbennun:
```suggestion
These markers can be used to visualize and measure the GPU activity using the NVIDIA Nsight Systems or ROCm Systems profilers.
```

tests/instrumentation_test.py
```diff
@@ -57,6 +58,19 @@ def onetest(instrumentation: dace.InstrumentationType, size=128):
         report = sdfg.get_latest_report()
         print(report)
 
+    # Check that the NVTX/rocTX range wrapper is present in the generated CPU code
+    if instrumentation is dace.InstrumentationType.GPU_TX_MARKERS:
```
tbennun: ==

tbennun (changes requested):
This looks much better! Two major comments and the rest are simple to address.
I really like this PR now.

dace/codegen/instrumentation/gpu_tx_markers.py
iomaganaris: I have updated this function but I couldn't find a way to test the updates. IIRC I had to perform this test in case we had a nested SDFG that was called from inside a GPU map. ~I couldn't reproduce the issue anymore though with our icon4py stencils.~ I tried to expand the test in the `instrumentation_test.py` but I don't know what's the proper API to call a simple python function or a `dace.program` within another `dace.program` which would trigger I believe the same thing. If you could provide an example it would be great.

Edit: I have managed to test the change with the `icon4py` stencils and it works okay however I still haven't managed to create a proper unit test in DaCe

dace/codegen/targets/framecode.py
iomaganaris: `rocTX markers` cannot be used before initializing HIP (AMD software restriction) so we have to skip them for the `init`

dace/codegen/targets/framecode.py
iomaganaris: The include is added to added to the `global_code` conditionally if it's necessary by the `on_*_{begin,end}` functions. Here's it's necessary to make sure that it's printed because it's used later and we don't have access to the `global_code` to use the same way to print it.
Rethinking this now it might be a better idea to simplify this behavior and just include it all the time at the beginning of the file if `GPUTXMarkersProvider` is enabled

dace/codegen/instrumentation/utils.py
iomaganaris: Removed the file completely

dace/codegen/instrumentation/utils.py
iomaganaris: I did a small fix and now it works 👍 Thanks for the suggestion

dace/codegen/instrumentation/provider.py
```diff
@@ -183,3 +183,75 @@ def on_node_end(self, sdfg: SDFG, cfg: ControlFlowRegion, state: SDFGState, node
...
+            :param local_stream: Code generator.
+        """
+        pass
+
+    def on_deallocation_end(self, sdfg: SDFG, scope: Union[nodes.EntryNode, SDFGState, SDFG],
+                            lstream: CodeIOStream) -> None:
+        """ Event called at the end of a deallocation code generation.
+
+            :param sdfg: The generated SDFG object.
+            :param local_stream: Code generator.
```
tbennun, line 259: docstring / argument mismatch

dace/codegen/targets/framecode.py
```diff
@@ -254,6 +255,13 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
         # Write closing brace of program
         callsite_stream.write('}', sdfg)
 
+        if sdfg.instrument == dtypes.InstrumentationType.GPU_TX_MARKERS:
+            # Need to make sure that the necessary includes for GPU_TX_MARKERS are present
+            # in the generated code.
+            gpu_tx_markers_provider = self._dispatcher.instrumentation.get(dtypes.InstrumentationType.GPU_TX_MARKERS)
+            if gpu_tx_markers_provider:
+                gpu_tx_markers_provider.print_include(callsite_stream)
```
tbennun: This should be generalized.

doc/optimization/profiling.rst
```diff
@@ -121,7 +121,8 @@ Instrumentation can also collect performance counters on CPUs and GPUs using `LI
 The :class:`~dace.dtypes.InstrumentationType.LIKWID_Counters` instrumentation type can be configured to collect
 a wide variety of performance counters on CPUs and GPUs. An example use can be found in the
 `LIKWID instrumentation code sample <https://github.com/spcl/dace/blob/main/samples/instrumentation/matmul_likwid.py>`_.
-
+There is also the :class:`~dace.dtypes.InstrumentationType.GPU_TX_MARKERS` instrumentation type which wraps in NVTX or rocTX markers the DaCe program executed on the GPU. Important parts of the execution of the program on the GPU as the different states, SDFGs and initialization and finalization phases are marked with these markers.
```
tbennun:
```suggestion
The :class:`~dace.dtypes.InstrumentationType.GPU_TX_MARKERS` instrumentation type wraps a DaCe program executed on the GPU with NVTX or rocTX markers. Important parts of the execution of the program on the GPU as the different states, SDFGs and initialization and finalization phases are marked with these markers.
```

tbennun (changes requested):
Excellent! Super minor docstring comments and a comment on where to put the include. Look at the other instrumentation providers, they use the global scope and any function to add the include.

A modular code generator should not have specific instrumentation type checks.

iomaganaris:
@tbennun @phschaad sorry for bothering you again. Could someone please relaunch the `cscs/GPU` CI and add the PR to the merge queue? Thank you very much 🙏

PR #2006, opened 2025-05-15
GPU Codegen Expansion of Empty Memlets

philip-paul-mueller:
Memlets that the GPU code generator can not be expressed as a `cudaMemcpy*()` call are turned into Maps.
However, there was an issue because codegen did not make sure that an Memlet it wants to promote is empty.
So it accessed the subsets of an empty Memlets, which are `None` and hits an error.


acalotoiu (approved):
Thank you! LGTM!

PR #2009, opened 2025-05-15
Fixed a Possible Bug in the Extended State Fusion

philip-paul-mueller:
The variable `connections_to_make` was a class variable and not marked as a dace property and hence shared between all instances. It is now an instance variable and thus private.
Also the variable was never cleared, so technically, the connections should have accumulated inside it.

Furthermore, the transformation assumed that `can_be_applied()` was called immediately before `apply()`, because the `can_be_applied()` function populates the `connections_to_make`. However, this behaviour is not guaranteed and now `apply()` explicitly calls `can_be_applied()` to make sure that `connections_to_make` is properly populated.


acalotoiu (changes requested):
LGTM, can you please add a new test to this PR to showcase the bug and how this fixed it? Thank you!

phschaad (approved):
Looks good from my side

acalotoiu (approved):
LGTM

PR #2013, opened 2025-05-22
GPU Codegen now FWD Stream Functions 

philip-paul-mueller:
While `__dace_init_cuda()` and `__dace_exit_cuda()` where FWD the `__dace_gpu_set_stream()` and `__dace_gpu_set_all_streams()` function where not which prevent their usage in user supplied init and global code fragments.
This PR FWD them at the same location as the init and exit functions.

PS: The names are a bit off, we should either settle on "CUDA" or "GPU" but not on both.


alexnick83 (approved):
LGTM

PR #2016, opened 2025-05-23
The `temporary_config` Contextmanager Uses an Anyonymous File

philip-paul-mueller:
The context manager stores its current configuration inside a file.
However, because it has a name it might be influenced from the outside.
Switching to an unnamed file reduces this possibility.


phschaad (approved):
Seems fine by me, thanks!

PR #2017, opened 2025-05-23
F2dace/static parse fix

Mittagskogel:
Fix 'Could not parse...' error caused by substitution of constant variables with expressions preceded by a unary negation operator - in ast_desugaring.py. This is done by inserting an fparser.two.Fortran2003.Parenthesis object in the AST to wrap the offending constant expression in parentheses.


pratyai:
Kindly verify that that related test file passes. Otherwise looks alright.
LGTM

Mittagskogel:
Full `ast_desugaring_test.py` passes :+1:

PR #2020, opened 2025-05-26
Add import functionality for experimental modules

ThrudPrimrose:
Add import functionality for experimental modules such that anybody can import anything that is checked out to `dace.transformation.experimental`. I added a test to create some transformations in that folder, load and apply them.

Also, after `__init__.py` is called / triggered by `import dace` and then users can import any experimental transformation like: `import dace.transformation.experimental as exp`.


tests/experimental_repository_registry_test.py
```diff
@@ -0,0 +1,67 @@
+import os
```
phschaad, line 1: Same

phschaad (changes requested):
Very nice feature! Only one concern: In some cases, depending on how DaCe is installed, the package file system location (and by extension anything in `dace/transformation/experimental`) may not be writable. In fact, that will probably quite often be the case - at least not for non-sudo processes.

As such, my suggestion would be to use a default local repository location that is _not_ in the dace package root, but rather somewhere under the user's `$HOME` path, or equivalent (needs to be OS agnostic). Ideally it's a path that can also be overridden through the DaCe config. This would mean that the repositories JSON and any downloaded transformations would live under that repository.

ThrudPrimrose:
> Very nice feature!
> 
> Only one concern: In some cases, depending on how DaCe is installed, the package file system location (and by extension anything in `dace/transformation/experimental`) may not be writable. In fact, that will probably quite often be the case - at least not for non-sudo processes.
> 
> As such, my suggestion would be to use a default local repository location that is _not_ in the dace package root, but rather somewhere under the user's `$HOME` path, or equivalent (needs to be OS agnostic). Ideally it's a path that can also be overridden through the DaCe config. This would mean that the repositories JSON and any downloaded transformations would live under that repository.

I guess I can mimic the package path by doing smth like:

```
$HOME/dace_extensions/
└── dace/
    └── transformation/
        └── experimental/
            └── layout/
                └── __init__.py
                └── my_custom_transform.py
# Then I need to do this, or the user needs to set PYTONPATH
import sys
sys.path.insert(0, '/home/youruser/dace_extensions')
```

What would be the best way to expose and keep the namespace `dace.transformation.experimental` ?
Do we assume the user should know how to handle PYTHONPATH etc. if the want to go for this option - as this change needs to be persistent across bash sessions if we want to do it

dace/codegen/targets/framecode.py
```diff
@@ -235,7 +235,6 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
             :param global_stream: Stream to write to (global).
             :param callsite_stream: Stream to write to (at call site).
         """
-        import dace.library
```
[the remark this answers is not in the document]
tbennun: The reasoning was: dace.library shouldn't be imported by default, because it takes a while and might load libraries that are not used or something.

However, if the code works as it is right now, I am super happy with removing it. I'm just worried that someone added a bad import to a dace.library somewhere else and it works by accident. Notice how there is no dace.library import anywhere else in the file...

tbennun:
@ThrudPrimrose I am not sure if it's experimental transformations or just external ones. This also sets the ground for separate github.com/dace-framework/transformations repos. Otherwise I like it.

dace/__init__.py
```diff
@@ -26,6 +26,14 @@
 # Import hooks from config last (as it may load classes from within dace)
 hooks._install_hooks_from_config()
 
+from pathlib import Path
+import sys
+from dace import config
+
+__external_transformations_path__ = config.Config.get("external_transformations_path").replace(
+    '$HOME', str(Path.home()))
```
tbennun, line 34: From Copilot:
There is a more robust and portable way to expand all environment variables and user directories in Python, which works across Windows, Mac, and Linux:
```suggestion
import os

raw_path = config.Config.get("external_transformations_path")
__external_transformations_path__ = os.path.expanduser(os.path.expandvars(raw_path))
```

dace/config_schema.yml
```diff
@@ -918,6 +918,15 @@ required:
             be called before every compiled SDFG's generated code is invoked. Used
             for functionality such as low-level profiling.
 
+    external_transformations_path:
+        type: str
+        default: '$HOME/dace_transformations/external_transformations'
```
tbennun, line 923:
```suggestion
        default: '$HOME/dace_transformations/external_transformations'
        default_Windows: '%USERPROFILE%\\dace_transformations\\external_transformations'
```

external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
+#!/usr/bin/env python3
```
tbennun, line 1: copyright year

external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+"""
+Dace Transformation Repository Manager
```
tbennun, line 3:
```suggestion
DaCe Transformation Repository Manager
```

external_transformation_registry.py
tbennun, line 1: why is this file in the root directory? Should this not be in dace/transformation or (better) dace/cli?

external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+
+class TransformationRepoManager:
+
+    def _set_path(self):
+        import dace
+        from dace import config
+        # Get default path for external transformations, eval $HOME if present
+        base_path_home_unevaluated_str = config.Config.get("external_transformations_path")
+        home_str = str(Path.home())
+        base_path_str = base_path_home_unevaluated_str.replace('$HOME', home_str)
```
tbennun, line 27: see other comment

external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+        if success:
+            print("All repositories loaded successfully.")
+        else:
+            print("Some repositories failed to load.")
+
+        return success
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Manage Dace transformation repositories",
```
tbennun, line 230:
```suggestion
    parser = argparse.ArgumentParser(description="Manage DaCe transformation repositories",
```

external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+                                     formatter_class=argparse.RawDescriptionHelpFormatter,
+                                     epilog="""
+Examples:
+  %(prog)s add https://github.com/user/my-transformations.git
+  %(prog)s add https://github.com/user/repo.git --name custom-name
+  %(prog)s list
+  %(prog)s remove my-transformations
+  %(prog)s load-all-local
+  %(prog)s load-all-suggested
+        """)
```
tbennun, line 240: definitely should be a CLI tool that is defined in `setup.py` as well

tests/external_repository_registry_test.py
```diff
@@ -0,0 +1,143 @@
+import os
```
tbennun, line 1: copyright

tests/external_repository_registry_test.py
```diff
@@ -0,0 +1,143 @@
...
+import subprocess
+import sys
+from pathlib import Path
+import importlib.util
+import shutil
+import pytest
+
+REPO_URL = "https://github.com/ThrudPrimrose/LayoutTransformations.git"
```
tbennun, line 9: please no personal repos here. Use something in SPCL or in github.com/dace-framework

tests/external_repository_registry_test.py
```diff
@@ -0,0 +1,143 @@
...
+    assert result.returncode == 0, "Failed to remove repository"
+    assert not repo_path.exists(), "Repository folder still exists after removal"
+
+
+def test_remove_repository():
+    _test_add_repository_and_check_file()
+    _test_remove_repository()
+
+
+# Calling an external process won't see the dace.config setting we have changed
```
tbennun, line 115: Why not modify `os.environ` before launching? Would achieve the same effect.

tbennun (changes requested):
Comments

dace/transformation/experimental/registry.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
ThrudPrimrose: Added, please check as I copied over from another file and updated the year

tests/experimental_repository_registry_test.py
ThrudPrimrose: Added, please check as I copied over from another file and updated the year

external_transformation_registry.py
ThrudPrimrose: Added to the CLI tools, and to setup py

tests/external_repository_registry_test.py
ThrudPrimrose: I added a testing repository in SPCL:
```
REPO_URL = "https://github.com/spcl/TransformationsTest.git"
```

is this OK?

tests/external_repository_registry_test.py
ThrudPrimrose: You mean re-doing: 
`envvar = 'DACE_' + '_'.join(key_hierarchy)`

to mimic the behaviour of `config_schema` parsing for this specific variable?

ThrudPrimrose: I added test cases using this feature.

dace/__init__.py
```diff
@@ -26,6 +26,16 @@
 # Import hooks from config last (as it may load classes from within dace)
 hooks._install_hooks_from_config()
 
+from pathlib import Path
+import sys
+from dace import config
+
+import os
+
+raw_path = config.Config.get("external_transformations_path")
```
tbennun: `Config` is already imported above, no need for re-importing
I would also argue that you want to keep all the imports high in the code.

dace/__init__.py
```diff
@@ -26,6 +26,16 @@
...
+__external_transformations_path__ = os.path.expanduser(os.path.expandvars(raw_path))
+sys.path.insert(0, __external_transformations_path__)
```
tbennun: maybe insert it if and only if there is a raw path to begin with? If raw_path is None, this will probably add some folder that we don't want.

dace/cli/external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
+#!/usr/bin/env python3
```
tbennun: copyright where?

dace/cli/external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+import json
+import os
+import subprocess
+import sys
+from pathlib import Path
+from typing import Dict, List, Optional
+from urllib.parse import urlparse
+
+
+class TransformationRepoManager:
```
tbennun, line 21: docstring where?

dace/cli/external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def _set_path(self):
+        import dace
+        from dace import config
```
tbennun: See CONTRIBUTING.md, inline imports need comments with adequate explanations on why they are performed. In this case these imports should be global IIUC

dace/cli/external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+import os
+import subprocess
+import sys
+from pathlib import Path
+from typing import Dict, List, Optional
+from urllib.parse import urlparse
+
+
+class TransformationRepoManager:
+
```
tbennun, line 43: not a single method has a docstring here.

dace/cli/external_transformation_registry.py
```diff
@@ -0,0 +1,292 @@
...
+    def _set_path(self):
+        import dace
+        from dace import config
+        # Get default path for external transformations, eval $HOME if present
+        base_path_home_unevaluated_str = config.Config.get("external_transformations_path")
+        home_str = str(Path.home())
+        base_path_str = base_path_home_unevaluated_str.replace('$HOME', home_str)
```
tbennun, line 52: expand envvars like in dace/init or use the `dace.__external_transformations_path__` here (preferable).

tests/external_repository_registry_test.py
tbennun: Please don't resolve comments if you want me to read them. Let the reviewer resolve them :)

tests/external_repository_registry_test.py
```diff
@@ -0,0 +1,180 @@
...
+    test_add_repository_and_check_file()
+    test_import_empty_transformation()
+    test_remove_repository()
+    test_add_repository_and_check_file_with_env_var()
+    test_import_empty_transformation_with_env_var()
+    test_remove_repository_with_env_var()
+    test_add_repository_and_check_file_non_default_path()
+    test_import_empty_transformation_non_default_path()
+    test_remove_repository_non_default_path()
+    print("All External-Repository-Registry tests passed.")
```
tbennun:
```suggestion

```

tbennun (changes requested):

dace/__init__.py
ThrudPrimrose: Done, removed reduntant import, change ``config.Config` to `Config`

dace/__init__.py
ThrudPrimrose: The default in the config schema is a valid path of form $HOME/.... Is it possible for this variable to be None?
Regardless I think it is good to check if it is None nevertheless

dace/cli/external_transformation_registry.py
ThrudPrimrose: Added

dace/cli/external_transformation_registry.py
ThrudPrimrose: I moved them to top-level.

dace/cli/external_transformation_registry.py
ThrudPrimrose: Done

tbennun (approved):

phschaad (approved):
LGTM too now, thank you :)

PR #2027, opened 2025-05-29
F2dace/case global var fix

Mittagskogel:
Fix const evaluation for CASE statements.

Using a parameter variable in a `Case_Selector` causes a crash because the variable cannot be found during type inference. This fix adds support for CASE statements in `ast_desugaring.py:_const_eval_nodes` to automatically replace parameter variables with integer literals during internal ast transformation.

### DIFF
--- a/dace/frontend/fortran/ast_desugaring.py
+++ b/dace/frontend/fortran/ast_desugaring.py
@@ -3138,10 +3138,7 @@ def const_eval_nodes(ast: Program) -> Program:
         assert not np.isnan(val)
         val = numpy_type_to_literal(val)
         if val.tostr().startswith('-'):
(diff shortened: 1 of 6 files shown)


dace/frontend/fortran/ast_desugaring.py
pratyai, line 1: Would be great if you could add a test that shows, on a code level, how this change is going to affect a small program that has a case statement (similar to many other tests for this file). Do you think that's reasonable?

pratyai (changes requested):
looks fine otherwise.

Mittagskogel:
Sure! I've added a simple test to `ast_desugaring_test.py`, is this what you intended?

PR #2031, opened 2025-06-05
Update Edge Consolidation in `MapFusion`

philip-paul-mueller:
In [PR#1993](https://github.com/spcl/dace/pull/1993) the possibility to consolidate or not consolidate edges was added to `MapFusion`.
However, it turned out (at least in GT4Py word) that this was not enough.
Thus, a new mode was added, which allows conditionally consolidation.
Assume, that there are two edges with the subset `[:, 0:10]` and `[:, 1:8]` there is no harm in consolidating the two because one is a superset of the other.

The `consolidate_edges` flag was removed, instead the following flags were added:
- `never_consolidate_edges`: Which blocks all edge consolidation, i.e. all edge will be maintained.
- `consolidate_edges_only_if_not_extending`: Which allows consolidation only if the this does not lead to an increase of the transfered subset.


acalotoiu (approved):
LGTM. Thank you!

PR #2032, opened 2025-06-06
Extend interstate edge to provide set of used SDFG symbols and used arrays

ThrudPrimrose:
I had some issues on the behaviour of `InterstateEdge::read_symbols` member.

Q1. `read_symbols` returns both the names of the data containers accessed as well as the SDFG symbols. In the SDFG, symbols and containers are separate, is this behavior intended? Shouldn't we filter the strings that are in `sdfg.arrays` from the return? 
A1. In case some functions rely on this functionality (IMO symbols are separate than data, and this is an artifact due to have sympy treats array names as symbols), I propose providing `used_sdfg_symbols` to filter symbols and `used_arrays` to filter arrays.

Q2. Documentation of `read_symbols` state that ` Returns a set of symbols read in this edge (including symbols in the condition and assignment values).`, but if I look at the return of the function I do not see the LHS side of the assignment in the returned set. There is also a variable called `all_symbols` which is not used and allows you to call `used_symbols` would it not make sense to use symbol to check if we should return the `lhs` side too?
A2. For this I add 2 lines to used symbols, by using `all_symbols` we can union the `lhs` and truly return all symbols. To not change existing behavior, by default this is set to `false`.


phschaad (approved):
That seems reasonable, LGTM

PR #2033, opened 2025-06-06
Updated How Codegen Handles Unsupported Memlets

philip-paul-mueller:
There was a very subtle error in codegen.
At the beginning codegen will scan all edges and translate anyone, that it can not handle by a native call, into a Map, for this the `CopyToMap` transformation was used.
The error arises when there is not one but multiple edges between the same AccessNodes, since the codegen passes source and destination node to the transformation and not the edge in question and `CopyToMap` only translated a single edge is translated, there is no guarantee that the right one was translated, which is an obviously wrong behaviour.

In this PR the functionality of turning an edge into a Map was moved to a function, `memlet_to_map()` inside the `memlet_utils` module, that operated on a edge, thus it is more targeted (or targeting the object in question at all).
The code generator was also patched to use this function (it is not possible to use the new `CopyToMap` transformation, because it is not possible to select the edge that should be translated and we only one to translate the edge that is not supported or the wrong one).

For compatibility it keeps the `CopyToMap` transformation around, however, it now uses the new helper functions.
Furthermore, it now transforms all edges between the two nodes, the only logical operation.

In GT4Py we have found that the implicit conversion of Maps by the code generator caused more harm than good.
Thus, this PR also adds a new configuration flag that allows to turn the implicit conversion of Memlets to Maps into errors,
by default it is still allowed.


dace/config_schema.yml
```diff
@@ -447,6 +447,17 @@ required:
                             index types are needed to address memory offsets that are beyond the 32-bit
                             range, or to reduce memory usage.
 
+                    allow_implicit_memlet_to_map:
+                        type: bool
+                        title: Allow the implicit conversion of Memlets to Maps during code generation.
+                        default: true
+                        description: >
+                            If ``true`` the code generator will implicitly convert Memlets that can not be
```
tbennun:
```suggestion
                            If ``true`` the code generator will implicitly convert Memlets that cannot be
```

tbennun (approved):

PR #2045, opened 2025-06-13
Schedule tree modernization for control flow regions

phschaad:


dace/sdfg/analysis/schedule_tree/treenodes.py
```diff
@@ -113,46 +119,6 @@ def as_string(self, indent: int = 0):
         return indent * INDENTATION + f'assign {self.name} = {self.value.as_string}'
 
 
-@dataclass
-class ForScope(ControlFlowScope):
```
tbennun, line 117: I'm not sure we should remove those at all.
Despite similar names, they serve different purposes than the control flow regions. What I think is that a ForScope should contain the controlflowregion

phschaad: They were not removed because of control flow regions, rather just because a single `GeneralLoopScope` (see line 157/123) took their functionality to have a closer 1:1 matching with the structure of the SDFG and the structured control flow tree. Or do they have an additional purpose over what the `GeneralLoopScope` provides?

tbennun: Initially I opted for simple classes, similar to AST definitions in other languages. Let’s leave it like this (since this is not an AST) and see how it works out. I would rename GeneralLoopScope to LoopScope, however.

tests/control_flow_test.py
```diff
@@ -329,27 +335,8 @@ def test_fsm():
 
     if dace.Config.get_bool('optimizer', 'detect_control_flow'):
         code = sdfg.generate_code()[0].clean_code
-        assert 'switch ' in code
-
-
-def test_nested_loop_detection():
```
phschaad, line 335: This is redundant since this kind of construct is now explicit control flow straight from the frontend (tested there), and there are separate loop detection tests that cover nested loops for control flow raising.

dace/codegen/control_flow.py
```diff
@@ -826,241 +33,169 @@ def _child_of(node: SDFGState, parent: SDFGState, ptree: Dict[SDFGState, SDFGSta
...
-                elif oe[1].dst is mergestate:
-                    branch_block = IfScope(dispatch_state, parent_block, False, node, oe[0].data.condition,
-                                           cblocks[oe[0]])
-                else:
-                    branch_block = IfScope(dispatch_state, parent_block, False, node, oe[0].data.condition,
-                                           cblocks[oe[0]], cblocks[oe[1]])
+    sdfg = region.sdfg
+    loop = region
+    cond = unparse_interstate_edge(loop.loop_condition.code[0], sdfg, codegen=codegen, symbols=symbols)
+    cond = cond.strip(';')
```
alexnick83: Why strip the semicolon and then re-add it? Maybe the unparser should skip it in the first place?

dace/codegen/control_flow.py
```diff
@@ -826,241 +33,169 @@ def _child_of(node: SDFGState, parent: SDFGState, ptree: Dict[SDFGState, SDFGSta
...
+
+    expr = ''
+
+    if loop.update_statement and loop.init_statement and loop.loop_variable:
+        lsyms = {}
+        lsyms.update(symbols)
+        if codegen.dispatcher.defined_vars.has(loop.loop_variable) and not loop.loop_variable in lsyms:
```
alexnick83, line 99: Isn't the loop variable local to the loop region? Why would it already be defined in the codegen? Is it a case of a while loop, where the variable init is detached from the region?

dace/frontend/python/astutils.py
```diff
@@ -1,6 +1,7 @@
 # Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
 """ Various AST parsing utilities for DaCe. """
 import ast
+from hmac import new
```
alexnick83: Why do we need this? I don't see it being used in the updated code.

dace/sdfg/analysis/cfg.py
```diff
@@ -59,6 +59,177 @@ def all_dominators(
...
+    if sink is None:
+        remove_sink = True
+        sinks = set()
+        for block in cfg.nodes():
+            if cfg.out_degree(block) == 0 or isinstance(block, (ContinueBlock, BreakBlock, ReturnBlock)):
+                sinks.add(block)
+        sink = ControlFlowBlock('__DACE_dummy_sink')
+        cfg.add_node(sink)
+        for sink in sinks:
+            cfg.add_edge(sink, sink, InterstateEdge())
```
alexnick83:
```suggestion
        for s in sinks:
            cfg.add_edge(s, sink, InterstateEdge())
```

dace/sdfg/analysis/cfg.py
```diff
@@ -59,6 +59,177 @@ def all_dominators(
...
+    if entry_node is None or exit_node is None:
+        return target_nodes.copy(), entry_node, exit_node
+
+    # The region includes all nodes on paths from entry to exit
+    # that are reachable from entry and can reach exit
+    region_nodes = set()
+
+    # Add all nodes reachable from entry that can also reach exit
+    reachable_from_entry = set()
+    if entry_node in graph:
```
alexnick83, line 207: What does it mean if the entry or exit node is not in the graph? Does that imply that the whole graph is in the SESE region, or is it considered an error?

dace/sdfg/analysis/cfg.py
```diff
@@ -59,6 +59,177 @@ def all_dominators(
...
+    if exit_node in graph:
+        # Find all nodes that can reach the exit
+        reverse_graph = graph.nx.reverse()
+        can_reach_exit = set(nx.descendants(reverse_graph, exit_node)) | {exit_node}
+
+    # Region is intersection of reachable from entry and can reach exit
+    region_nodes = reachable_from_entry & can_reach_exit
+
+    # Ensure all target nodes are included
+    region_nodes.update(target_nodes)
```
alexnick83: Does it make sense that any target nodes are not in the region nodes? Is this related to the case where the entry or exit node is outside the graph?

dace/sdfg/analysis/cfg.py
```diff
@@ -98,17 +269,33 @@ def branch_merges(
...
+            branch_descendants.add(oedge.dst)
             frontier = adf[oedge.dst]
             if not frontier:
-                frontier = {oedge.dst}
+                # If no dominance frontier is found for this edge, there are two possible scenarios under which this
+                # may still lead to a valid merge state:
+                # 1: The edge destination is itself the branch merge state. To cover this, the frontier consisits of
+                #    the destination block itself, and if there is a concrete merge state, that will result in a single
+                #    common frontier block.
+                # 2: The edge leads to a completely separate control flow path that does not reconnect to the branch
```
alexnick83, line 279: I guess this can be the case only when the control flow is state-machine-based and not using control flow regions. Right?

dace/sdfg/analysis/schedule_tree/sdfg_to_tree.py
```diff
@@ -583,9 +607,107 @@ def state_schedule_tree(state: SDFGState) -> List[tn.ScheduleTreeNode]:
...
+        goto_node = tn.GotoNode(target=edge.dst.label)
+        goto_node.sdfg = graph.sdfg
+        edge_body.append(goto_node)
+
+    if edge.data.is_unconditional():
+        result.extend(edge_body)
+    else:
+        if emit_gotos:
+            state_if_node = tn.StateIfScope(condition=edge.data.condition, children=edge_body)
+        else:
```
alexnick83, line 624: This is a bit confusing since GoTo nodes are generated regardless of whether `emit_gotos` is True or False. Maybe a comment explaining the logic would help.

dace/sdfg/analysis/schedule_tree/treenodes.py
```diff
@@ -114,54 +119,14 @@ def as_string(self, indent: int = 0):
...
     """
     General loop scope (representing a loop region).
     """
-    header: cf.GeneralLoopScope
+    loop: LoopRegion
 
     def as_string(self, indent: int = 0):
-        loop = self.header.loop
+        loop = self.loop
         if loop.update_statement and loop.init_statement and loop.loop_variable:
```
alexnick83: Perhaps it would be easier to maintain if we first identified the loop type (for, while, do-while) and then had a clear generation of the result based on that type. Nested if conditions can be a bit confusing.

dace/sdfg/analysis/schedule_tree/treenodes.py
```diff
@@ -196,6 +161,7 @@ class IfScope(ControlFlowScope):
     If branch scope.
     """
     condition: CodeBlock
+    cond_block: Optional[ConditionalBlock] = None
```
alexnick83: This parameter doesn't seem to be used. What is its purpose?

dace/sdfg/utils.py
```diff
@@ -2056,28 +2058,33 @@ def get_control_flow_block_dominators(sdfg: SDFG,
...
-                sink = cfg.add_state()
-                added_sinks[cfg] = sink
-                sinks_per_cfg[cfg] = sink
-                for snode in sink_nodes:
-                    cfg.add_edge(snode, sink, dace.InterstateEdge())
-            elif len(sink_nodes) == 0:
-                return None
+                sinks_per_cfg[cfg] = cfg
+                for _, b in cfg.branches:
+                    ipostdom[b] = cfg
```
alexnick83, line 2063: Does it work well to set the conditional block as both the immediate dominator and postdominator of the branches?

dace/transformation/interstate/loop_detection.py
```diff
@@ -189,7 +189,13 @@ def detect_loop(self,
 
         # All nodes inside loop must be dominated by loop guard
         dominators = nx.dominance.immediate_dominators(graph.nx, graph.start_block)
-        loop_nodes = sdutil.dfs_conditional(graph, sources=[begin], condition=lambda _, child: child != guard)
+        loop_nodes = list(sdutil.dfs_conditional(graph, sources=[begin], condition=lambda _, child: child != guard))
+        # If the exit state is in the loop nodes, this is not a valid loop
+        if self.exit_state in loop_nodes:
+            return None
+        elif nx.has_path(graph.nx, self.exit_state, guard):
+            # Simliarly, if there is a path from the exit state to the guard, this is not a valid loop.
```
alexnick83: If you have a loop nest, wouldn't this be the case for all nested loops, depending on the exact state-machine representation?

dace/transformation/passes/simplification/control_flow_raising.py
```diff
@@ -109,7 +110,17 @@ def _lift_conditionals(self, sdfg: SDFG) -> int:
...
 
         for region in cfgs:
-            if isinstance(region, ConditionalBlock):
+            if isinstance(region, (ConditionalBlock, UnstructuredControlFlow)):
+                continue
+
+            if region.has_cycles():
+                # Do not lift conditionals if there are cycles present, since lifting conditionals requires an acyclic
+                # dominance frontier for the analysis. This may lead to incorrect results if cycles are present.
+                # Note that the combinatio nof loop raising and unstructured control flow lifting should
```
alexnick83:
```suggestion
                # Note that the combination of loop raising and unstructured control flow lifting should
```

dace/transformation/passes/simplification/control_flow_raising.py
```diff
@@ -178,21 +212,95 @@ def _lift_conditionals(self, sdfg: SDFG) -> int:
...
+
+            if not cfg.has_cycles():
+                # No cycles, no unstructured control flow.
+                continue
+
+            # Compute immediate dominators
+            idom: Dict[ControlFlowBlock, ControlFlowBlock] = nx.immediate_dominators(cfg.nx, cfg.start_block)
+
+            # Find back edges (u, v) where v dominates u
+            back_edges = []
```
alexnick83: Would it make sense to use `cfg_analysis.back_edges` here?

dace/sdfg/analysis/cfg.py
phschaad: Correct, yes. However, this may come from certain frontends that way, so we still need to detect it properly I propose.

dace/frontend/python/astutils.py
phschaad: That was a mistake, I've removed it again - thanks!

dace/sdfg/analysis/cfg.py
phschaad: Not sure how that check made it through cleanup - that is not necessary as you correctly point out. Removed :)

dace/sdfg/analysis/cfg.py
phschaad: This is a leftover from before I added the entry/exit nodes explicitly to the region nodes above - removed it.

dace/codegen/control_flow.py
phschaad: I actually believe the strip is completely redundant. I double checked, and the interstate edge unparser tells CPPUnparse to not generate any semicolons for expressions. I've removed it.

dace/codegen/control_flow.py
phschaad: The loop variable doesn't need to be local to the loop region as of yet - this is something I am also in favor of for the DaCe 2.0 transition, though. For now, you may have a symbol already defined somewhere, and the loop will happily use it without redefining it in the init statement if that's the case

dace/sdfg/analysis/schedule_tree/treenodes.py
phschaad: Removed

dace/transformation/interstate/loop_detection.py
phschaad: If the state machine contains nested loops, the process is that control flow raising (the only user of loop detection as of the control flow regions transition) raises them step-by-step from the top down. That means that once the top-most/ outer-most loop has been lifted to a loop region, this condition is true for the next loop in the hierarchy and it can thus be lifted in the next application of loop raising. But yes, it is true that if there is a pure state machine nesteded loop with no loop regions, this will only detect the outer-most loop.

dace/transformation/passes/simplification/control_flow_raising.py
phschaad: True, changed

dace/sdfg/utils.py
phschaad: Yes this seems to be a robust solution thus far - this way the domination / postdomination relationships seamlessly carry over from the parent graph to the child when computing the transitive relationship.

phschaad: Note that it is still advisable to examine if we can find a better system for determining allocation lifetime either way - but I believe we will do that anyway as part of the DaCe 2.0 discussion.

dace/sdfg/analysis/schedule_tree/treenodes.py
phschaad: Good idea, done

dace/codegen/targets/framecode.py
```diff
@@ -484,35 +482,10 @@ def dispatch_state(state: SDFGState) -> str:
...
-                [cflow.BasicCFBlock(dispatch_state, None, s is last, s)
-                 for s in states_topological], [], [], [], [], False)
-
-        callsite_stream.write(cft.as_cpp(self, sdfg.symbols), sdfg)
+        callsite_stream.write(cflow.control_flow_region_to_code(sdfg, dispatch_state, self, sdfg.symbols), sdfg)
 
         opbar.done()
 
-        # Write exit label
-        callsite_stream.write(f'__state_exit_{sdfg.cfg_id}:;', sdfg)
```
tbennun, line 514: what happened to the exit state? I still see references to it in control flow

dace/frontend/python/astutils.py
```diff
@@ -428,9 +428,19 @@ def and_expr(node_a, node_b):
     if isinstance(expr_b, ast.Expr):
         expr_b = expr_b.value
 
-    newexpr = ast.Expr(value=ast.BinOp(left=copy_tree(expr_a), op=ast.And, right=copy_tree(expr_b)))
-    newexpr = ast.copy_location(newexpr, expr_a)
-    return ast.fix_missing_locations(newexpr)
+    if isinstance(expr_a, ast.BoolOp) and expr_a.op == ast.And:
```
tbennun, line 431: should this be == or isinstance?

tests/schedule_tree/nesting_test.py
```diff
@@ -200,20 +200,20 @@ def test_dealias_interstate_edge():
     sdfg.add_array('B', [20], dace.float64)
 
     nsdfg = dace.SDFG('nester')
-    nsdfg.add_array('A', [19], dace.float64)
-    nsdfg.add_array('B', [15], dace.float64)
+    nsdfg.add_array('C', [19], dace.float64)
+    nsdfg.add_array('D', [15], dace.float64)
```
tbennun: why :)

tests/control_flow_test.py
```diff
@@ -265,6 +266,8 @@ def test_ifchain_manual():
         assert 'else if' in code
 
 
+@pytest.mark.skip(reason="Switch-case are not allowed in the ConditionalBlock semantics, and are thus not " +
+                  "generated with the new ControlFlowRaising pass.")
```
tbennun, line 270: any plans to reinstate?

dace/sdfg/analysis/schedule_tree/sdfg_to_tree.py
```diff
@@ -583,9 +607,111 @@ def state_schedule_tree(state: SDFGState) -> List[tn.ScheduleTreeNode]:
...
+            else:
+                else_node = tn.ElseScope(children=_block_schedule_tree(branch_body))
+                else_node.sdfg = block.sdfg
+                result.append(else_node)
+        return result
+    elif isinstance(block, SDFGState):
+        return _state_schedule_tree(block)
+    elif isinstance(block, ReturnBlock):
+        # For return blocks, add a goto node to the end of the schedule tree.
+        goto_node = tn.GotoNode(target=None)
```
tbennun, line 692: I'm not sure that's the case. Return can go to the end of the function region, but target=None would leave the entire SDFG? Can we add a test for that?

dace/sdfg/analysis/schedule_tree/sdfg_to_tree.py
```diff
@@ -583,9 +607,111 @@ def state_schedule_tree(state: SDFGState) -> List[tn.ScheduleTreeNode]:
...
+            # regardless. No goto needs to be placed for the exit.
+            state_if_node = tn.StateIfScope(condition=edge.data.condition, children=edge_body)
+        else:
+            # If not emitting gotos for successors, this indicates that we are in structured control flow. In this case,
+            # rather than hiding any successors behind a condition, we create a return / exit goto node that is executed
+            # if the inverse of the condition holds, i.e., the successor is NOT executed.
+            exit_goto = tn.GotoNode(target=None)
+            exit_goto.sdfg = graph.sdfg
+            state_if_node = tn.StateIfScope(condition=CodeBlock(negate_expr(edge.data.condition)), children=[exit_goto])
+        state_if_node.sdfg = graph.sdfg
```
tbennun: I see this multiple times - why do we need the sdfg as a field?

dace/sdfg/analysis/cfg.py
```diff
@@ -59,6 +59,174 @@ def all_dominators(
...
+    for dom in common_dominators:
+        # Find maximum distance to any target node
+        max_dist_to_targets = 0
+        for target in target_nodes:
+            if target in dominators and dom in dominators[target]:
+                # Count nodes between dom and target
+                path_exists = nx.has_path(graph.nx, dom, target)
+                if path_exists:
+                    try:
+                        dist = nx.shortest_path_length(graph.nx, dom, target)
```
tbennun: Isn't running shortest_path_length enough? if there is no path it will raise the exception.

dace/transformation/passes/simplification/control_flow_raising.py
```diff
@@ -139,10 +150,33 @@ def _lift_conditionals(self, sdfg: SDFG) -> int:
...
                     # Populate branches.
+                    full_cond_expression: Optional[sympy.Basic] = None
+                    uncond_generated = False
                     for i, oe in enumerate(oedges):
                         branch_name = 'branch_' + str(i) + '_' + block.label
                         branch = ControlFlowRegion(branch_name, sdfg)
-                        conditional.add_branch(oe.data.condition, branch)
+
+                        if not oe.data.is_unconditional():
+                            if i == len(oedges) - 1 and oe.data.condition_sympy() == sympy.Not(full_cond_expression):
```
tbennun, line 160: You can use ast_negate here and the original condition. We do this in other places where `condition_sympy` is not always possible

dace/transformation/passes/simplification/control_flow_raising.py
```diff
@@ -139,10 +150,33 @@ def _lift_conditionals(self, sdfg: SDFG) -> int:
...
+                                    warnings.warn(
+                                        f'Control flow raising: Found multiple unconditional branches in {block.label}')
+                                uncond_generated = True
+                                cond = None
+                            else:
+                                cond = oe.data.condition
+                                if full_cond_expression is None:
+                                    full_cond_expression = oe.data.condition_sympy()
+                                else:
+                                    full_cond_expression = sympy.And(full_cond_expression, oe.data.condition_sympy())
```
tbennun, line 171: Unsure if this always works (sometimes it evaluates and simplifies wrongly), I'd rather use ast.BoolOp or something

dace/transformation/passes/simplification/control_flow_raising.py
```diff
@@ -167,32 +201,103 @@ def _lift_conditionals(self, sdfg: SDFG) -> int:
...
                 region.remove_node(dummy_exit)
 
         n_cond_regions_post = len([x for x in sdfg.all_control_flow_blocks() if isinstance(x, ConditionalBlock)])
         lifted = n_cond_regions_post - n_cond_regions_pre
         if lifted:
+            sdfg.reset_cfg_list()
             sdfg.root_sdfg.using_explicit_control_flow = True
         return lifted
 
+    def _lift_unstructured(self, sdfg: SDFG) -> int:
```
tbennun, line 216: I'm not sure that we have enough tests for this. e.g., two loops with one jumping to the middle of the next (sibling), branch jumping to the branch contents of another if in the else clause (tree-depth change), etc.

Another key thing to test is that the Unstructured CF region contains all the nodes that also may be the targets of the jumps, rather than just SESE bounds. However, I am not 100% certain that we must do this goto target boundaries; we do it in schedule trees.

dace/frontend/python/astutils.py
phschaad: No this was cause of a bug that this change uncovered, actually - `Expression.op` is storing the expression type, rather than an instance of the object.

tests/control_flow_test.py
phschaad: Yes, I have a PR planned to add this kind of functionality eventually. I will make an issue to track that feature request.

dace/codegen/targets/framecode.py
phschaad: This exit state is now generated by control flow for each control flow region (see https://github.com/spcl/dace/pull/2045/files#diff-71a7eb81ec5c9ea510fed5ab5c8e104e1ae12c3fec90072031566b63f8d1a750R260), so it has just moved

dace/sdfg/analysis/cfg.py
phschaad: You are right, changed it

dace/transformation/passes/simplification/control_flow_raising.py
phschaad: But the purpose here (and with the comment below) is to evaluate the conditions symbolically to check for symbolic equivalence - won't that be impossible (or I suppose significantly harder) if using Python AST objects instead?

dace/transformation/passes/simplification/control_flow_raising.py
phschaad: See above

tests/schedule_tree/nesting_test.py
phschaad: This is a remnant of a change (https://github.com/spcl/dace/pull/2045/files#diff-6454a828c0b26fcb8b1d616db3ee75ff0c61f6fdb45f0a8b2a4aaffad3b9ca5dL60) that was supposed to be part of a subsequent PR. I've reverted this for the time being and will reevaluate the change for the next PR :)

dace/sdfg/analysis/schedule_tree/sdfg_to_tree.py
phschaad: In general, yes, you are absolutely right. However, there needs to be a separate PR to address this, because the current semantics of the ReturnBlock is that it returns the current SDFG context. This is a simplification we've made for now because function regions are inlined at this point still and thus not quite feature complete yet. A PR that will make function regions feature complete will absolutely have to address this, both here and in code generation.

dace/sdfg/analysis/schedule_tree/sdfg_to_tree.py
phschaad: We do not (yet) - this is a remnant of a different feature addition. I will move this to a subsequent PR with proper explanation.

dace/transformation/passes/simplification/control_flow_raising.py
phschaad: Great point for missing tests - I have added one for lifting sibling loops into unstructured control flow, and that has in fact also uncovered a bug with loop raising that I fixed :)

Note for the second test case (intermixing branches), that this is not currently raised by unstructured control flow detection, since it does not contain any cycles. This is something that we certainly will want to expand in the future, however at the moment this is safe to simply ignore, since no other control flow raising component detects it as viable control flow, and the `control_flow.py` codegen component consequently just generates unstructured control flow with gotos, since there branches in the resulting raised graph.

phschaad: As a result, this PR now closes https://github.com/spcl/dace/issues/2062

tbennun (approved):

PR #2046, opened 2025-06-13
Updated How `PruneConnectors` Isolates the Nested SDFG

philip-paul-mueller:
`PruneConnectors` used `state_fission_after()` to isolate the nested SDFG node.
However, it should use the `isolate_nested_sdfg()` function, whch was specially designed for that case.


dace/transformation/helpers.py
```diff
@@ -723,6 +723,13 @@ def isolate_nested_sdfg(
         that it can be performed by returning either `True` or `False`.
     """
 
+    # We can only isolate the nested SDFG if it is on global scope, for example
+    #  inside a Map it is impossible to split the state.
+    if state.scope_dict()[nsdfg_node] is not None:
+        if test_if_applicable:
+            return False
+        raise ValueError(f'Can not isolate NestedSDFG "{nsdfg_node}" because it is not on global scope.')
```
tbennun:
```suggestion
        raise ValueError(f'Cannot isolate NestedSDFG "{nsdfg_node}" because it is within a scope.')
```

PR #2047, opened 2025-06-13
Modified How State Fission Works

philip-paul-mueller:
This PR changes how state fission is performed within DaCe.
Before, there were two functions for it `state_fission()` and `state_fission_after()` (which was used only in two places and did not have a unit test).

As a smaller change, this PR removes the `state_fission_after()` as it was used at only one place, had no description and no unit test and was replaced with `state_fission()`.

Furthermore, this PR changes how `state_fission()` works, the redesign was inspired by the change introduced by [PR#1981](https://github.com/spcl/dace/pull/1981) which adds `isolate_nested_sdfg()`, which ensures that the number of "writing nodes", i.e. AccessNodes with incoming connections is preserved.
The changes of introduced by this function goes into the same direction as `state_fission()` will now also conserve this property.
This has some behavioural change on the function, but also fixes some behavioral problems.

As argument `state_fission()` takes `subgraph`, which is a list of nodes that should be moved to a state that is sequenced before the state they are currently located in.
The function will now automatically add all dependencies of these nodes to the list of nodes that have to be relocated, currently this had to be done by the caller.
The other and bigger change is, that now also nodes that are downstream of the passed `subgraph` might be moved to the first state as well.
There are two reasons for this, the first is to ensure that the number of "writing nodes" does not increase.
The second reason is that the cut, i.e. which nodes can go to the first node, is not arbitrary and only Memlets whose source is a non view AccessNode or empty Memlets can be cut. 
Previously, it was the callers responsibility to ensure that, now the function does this.

Another reason why nodes are moved to the first state is scope.
This is because a split can only be done on the top level, for example a Map can not be distributed over the two state (for example the MapEntry and some nodes are placed in the first state, while the rest, including the MapExit ends up in the second state).
Thus the function will now make sure that this does not happens.
For this it will make sure that if a node is not at the global scope of the state, that its surrounding  scope is also moved to the first state; This happens recursively to ensure that nested scopes are handled properly.

### DIFF
--- a/dace/transformation/helpers.py
+++ b/dace/transformation/helpers.py
@@ -1375,22 +1375,31 @@ def scope_tree_recursive(state: SDFGState, entry: Optional[nodes.EntryNode] = No
     :param state: The state that contains the root of the scope tree.
     :param entry: A scope entry node to set as root, otherwise the state is
                   the root if None is given.
+
+    :note: This function adds a `state` attribute to the `ScopeTree` objects, it refers to
+        the state to which the scope was found.
     """
-    stree = state.scope_tree()[entry]
+    # We have to make a copy because we modify the `ScopeTree` objects that are returned and they
+    #  are cached. We can not use `deepcopy()` here because this would also copy scope nodes,
+    #  which we do not want. Thus we call `ScopeTree.copy()` which only copies the `ScopeTree`s
+    #  inside `children` but everything else is just assigned.
+    stree = copy.copy(state.scope_tree()[entry])
     stree.state = state  # Annotate state in tree
 
     # Add nested SDFGs as children
     def traverse(state: SDFGState, treenode: ScopeTree):
-        snodes = state.scope_children()[treenode.entry]
+        snodes = copy.copy(state.scope_children()[treenode.entry])  # See above why.
+
         for node in snodes:
             if isinstance(node, nodes.NestedSDFG):
                 for nstate in node.sdfg.states():
-                    ntree = nstate.scope_tree()[None]
+                    ntree = copy.copy(nstate.scope_tree()[None])  # See above why.
                     ntree.state = nstate
                     treenode.children.append(ntree)
+
         for child in treenode.children:
             if hasattr(child, 'state') and child.state != state:
-                traverse(getattr(child, 'state', state), child)
+                traverse(child.state, child)
 
     traverse(state, stree)
     return stree
@@ -1398,17 +1407,21 @@ def scope_tree_recursive(state: SDFGState, entry: Optional[nodes.EntryNode] = No
 
 def get_internal_scopes(state: SDFGState,
                         entry: nodes.EntryNode,
-                        immediate: bool = False) -> List[Tuple[SDFGState, nodes.EntryNode]]:
+                        immediate: bool = False,
+                        recursive_scope_tree: Optional[ScopeTree] = None) -> List[Tuple[SDFGState, nodes.EntryNode]]:
     """
     Returns all internal scopes within a given scope, including if they
     reside in nested SDFGs.
 
     :param state: State in which entry node resides.
     :param entry: The entry node to start from.
-    :param immediate: If True, only returns the scopes that are immediately
-                      nested in the map.
+    :param immediate: If True, only returns the scopes that are immediately nested in the map.
+    :param recursive_scope_tree: The recursive scope tree, see `scope_tree_recursive()` for more.
     """
-    stree = scope_tree_recursive(state, entry)
+
+    if recursive_scope_tree is None:
+        recursive_scope_tree = scope_tree_recursive(state, entry)
+
     result = []
 
     def traverse(state: SDFGState, treenode: ScopeTree):
@@ -1420,7 +1433,7 @@ def get_internal_scopes(state: SDFGState,
             else:  # Nested SDFG
                 traverse(child.state, child)
 
-    traverse(state, stree)
+    traverse(state, recursive_scope_tree)
     return result
 
 
@@ -1428,11 +1441,13 @@ def gpu_map_has_explicit_threadblocks(state: SDFGState, entry: nodes.EntryNode) 
     """
     Returns True if GPU_Device map has explicit thread-block maps nested within.
     """
-    internal_maps = get_internal_scopes(state, entry)
+    rstree = scope_tree_recursive(state, entry)
+    internal_maps = get_internal_scopes(state, entry, recursive_scope_tree=rstree)
     if any(m.schedule in (dtypes.ScheduleType.GPU_ThreadBlock, dtypes.ScheduleType.GPU_ThreadBlock_Dynamic)
            for _, m in internal_maps):
         return True
-    imm_maps = get_internal_scopes(state, entry, immediate=True)
+
+    imm_maps = get_internal_scopes(state, entry, immediate=True, recursive_scope_tree=rstree)
     if any(m.schedule == dtypes.ScheduleType.Default for _, m in imm_maps):
         return True
 
--- a/dace/codegen/targets/cpu.py
+++ b/dace/codegen/targets/cpu.py
@@ -1039,8 +1039,8 @@ class CPUCodeGen(TargetCodeGenerator):
             if node == dst_node:
                 continue
 
-            # Tasklet -> array
-            if isinstance(node, nodes.CodeNode):
+            # Tasklet -> array with a memlet. Writing to array is emitted only if the memlet is not empty
+            if isinstance(node, nodes.CodeNode) and not edge.data.is_empty():
                 if not uconn:
                     raise SyntaxError("Cannot copy memlet without a local connector: {} to {}".format(
                         str(edge.src), str(edge.dst)))
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -3086,13 +3086,13 @@ class ProgramVisitor(ExtNodeVisitor):
         parent_array = self.scope_arrays[parent_name]
 
         has_indirection = (_subset_has_indirection(rng, self) or _subset_is_local_symbol_dependent(rng, self))
+        strides = list(parent_array.strides)
         if has_indirection:
             # squeezed_rng = list(range(len(rng)))
             shape = parent_array.shape
-            # strides = [parent_array.strides[d] for d in squeezed_rng]
             # # TODO: Why is squeezed_rng an index in the first place?
             # squeezed_rng = subsets.Range([(i, i, 1) for i in squeezed_rng])
-            squeezed_rng = subsets.Range.from_array(parent_array)
+            nested_rng = subsets.Range.from_array(parent_array)
             non_squeezed = list(range(len(rng)))
         else:
             ignore_indices = []
@@ -3148,6 +3148,14 @@ class ProgramVisitor(ExtNodeVisitor):
             non_squeezed = squeezed_rng.squeeze(ignore_indices)
             # TODO: Need custom shape computation here
             shape = squeezed_rng.size()
+            nested_rng = subsets.Range([(0, s - 1, 1) for s in shape])
+            for i, r in enumerate(rng.ranges):
+                if i in ignore_indices:
+                    continue
+                _, _, step = r
+                if (step < 0) == True:
+                    step = -step
+                strides[i] *= step
             for i, sr in zip(ignore_indices, sym_rng):
                 iMin, iMax, step = sr.ranges[0]
                 if (step < 0) == True:
@@ -3157,6 +3165,7 @@ class ProgramVisitor(ExtNodeVisitor):
                 shape[sqz_idx] = ts * sympy.ceiling(((iMax.approx if isinstance(iMax, symbolic.SymExpr) else iMax) + 1 -
                                                      (iMin.approx if isinstance(iMin, symbolic.SymExpr) else iMin)) /
                                                     (step.approx if isinstance(step, symbolic.SymExpr) else step))
+                nested_rng.ranges[sqz_idx] = squeezed_rng[sqz_idx]
         dtype = parent_array.dtype
 
         if arr_type is None:
@@ -3168,7 +3177,7 @@ class ProgramVisitor(ExtNodeVisitor):
             self.sdfg.add_scalar(var_name, dtype)
         elif issubclass(arr_type, data.Array):
             if non_squeezed:
-                strides = [parent_array.strides[d] for d in non_squeezed]
+                strides = [strides[d] for d in non_squeezed]
             else:
                 strides = [1]
             self.sdfg.add_array(var_name, shape, dtype, strides=strides)
@@ -3179,7 +3188,7 @@ class ProgramVisitor(ExtNodeVisitor):
         else:
             raise NotImplementedError("Data type {} is not implemented".format(arr_type))
 
-        self.accesses[(name, rng, access_type)] = (var_name, squeezed_rng)
+        self.accesses[(name, rng, access_type)] = (var_name, nested_rng)
 
         inner_indices = set(non_squeezed)
 
@@ -3205,7 +3214,7 @@ class ProgramVisitor(ExtNodeVisitor):
                 self.outputs[var_name] = (state, new_memlet, inner_indices)
 
         self.variables[var_name] = var_name
-        return (var_name, squeezed_rng)
+        return (var_name, nested_rng)
 
     def _add_read_access(self,
                          name: str,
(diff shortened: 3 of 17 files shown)


dace/transformation/dataflow/wcr_conversion.py
```diff
@@ -270,6 +270,83 @@ def apply(self, state: SDFGState, sdfg: SDFG):
...
+        tlet: nodes.Tasklet = self.tasklet
+        newstate = state.parent_graph.add_state_after(state)
+
+        # Bookkeeping
+        nodes_to_move = set([tlet])
+        boundary_nodes = set()
+        orig_edges = set()
+
+        # Find all dependencies of the Tasklet, which are all access nodes that generates data.
+        # TODO: What about viewes?
```
alexnick83: Perhaps it is easier to handle views later below, where you define the boundary nodes (lines 307-318). If a boundary node is an access node to a view, then you should include the whole stack of views and the first non-view access node to the boundary nodes (and `nodes_to_move`). You can use the existing utility methods to find all such nodes (see [here](https://github.com/spcl/dace/blob/186d21de1936f651bba7e4763d3a9466b112c6b2/dace/sdfg/utils.py#L786)).

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+    AccessNodes with incoming connections, this behaviour is analogous to `isolate_nested_sdfg()`.
+    Another reason why more nodes are moved to the first state are scopes. It is only possible to
+    split a state at the top scope, not as an example, within a Map. Thus if a node inside `subgraph`
+    if not at the global scope the function will include its containing state. Thus if you only
+    pass a Tasklet, with a Map, as `subgraph` then the function will add the surrounding Map
+    to the first state as well.
+
+    Note, the split might result in a state where nodes become isolated inside a state. By default
+    the function allows this to happen in either state. However, it will cause a validation error.
+    By  etting `allow_isolated_nodes` to `False` the function will remove all isolated nodes in
```
alexnick83:
```suggestion
    By setting `allow_isolated_nodes` to `False` the function will remove all isolated nodes in
```

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+    possible, i.e. there must be a non view AccessNode to transmit data from the first to the
+    second state. Another reason is that the function maintains the number of "writing nodes", i.e
+    AccessNodes with incoming connections, this behaviour is analogous to `isolate_nested_sdfg()`.
+    Another reason why more nodes are moved to the first state are scopes. It is only possible to
+    split a state at the top scope, not as an example, within a Map. Thus if a node inside `subgraph`
+    if not at the global scope the function will include its containing state. Thus if you only
+    pass a Tasklet, with a Map, as `subgraph` then the function will add the surrounding Map
+    to the first state as well.
+
+    Note, the split might result in a state where nodes become isolated inside a state. By default
```
alexnick83, line 603: Is there an example where this occurs? From the description above, I would expect that the method avoids such cases altogether.

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+        while len(to_scan) != 0:
+            node_to_scan = to_scan.pop()
+            if node_to_scan in scanned_nodes:
+                continue
+            to_scan.update(
+                iedge.src for iedge in state.in_edges(node_to_scan)
+                if ((iedge.src not in scanned_nodes) and (True if follow_empty_memlets else not iedge.data.is_empty())))
+            scanned_nodes.add(node_to_scan)
+
+    # State fissions can not occur within a scope, i.e. the MapEntry of a Map scope can not end up
```
alexnick83:
```suggestion
    # State fissions can not occur within a scope, i.e., the MapEntry of a Map scope can not end up
```

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
-    for bn in boundary_nodes:
-        bn._in_connectors.clear()
+        elif containing_scope is None:
+            initial_first_nodes.add(node)  # Global node just add it.
 
-    return newstate
+        else:
+            # Note at global scope, find the top most scope node and add the defining subgraph.
+            while scope_dict[containing_scope] is not None:
+                assert isinstance(containing_scope, nodes.EntryNode)
```
alexnick83: I would remove or raise an appropriate exception (ValueError?) with a descriptive message (scope dictionary is malformed or something like that).

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+                containing_scope = scope_dict[containing_scope]
+            top_entry_node = containing_scope
+            assert isinstance(top_entry_node, nodes.EntryNode)
```
alexnick83: Same as above.

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
-            # Set the new parent state
-            n.sdfg.parent = newstate
+    sdfg: SDFG = state.sdfg
+
+    def find_generating_nodes(node_to_start: nodes.Node,
+                              state: SDFGState,
+                              follow_empty_memlets: bool,
+                              seen: Optional[Set[nodes.Node]] = None) -> None:
+        to_scan: Set[nodes.Node] = {node_to_start}
+        scanned_nodes: Set[nodes.Node] = set() if seen is None else seen
```
alexnick83: Just a note: you can also just write ` scanned_nodes = seen or set()` here. I think it works with all Python collections.

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
 
-def state_fission_after(state: SDFGState, node: nodes.Node, label: Optional[str] = None) -> SDFGState:
-    """
-    """
-    newstate = state.parent_graph.add_state_after(state, label=label)
+            # If we are here then `fnode` would need to be a boundary node, but does not qualify.
+            #  In that case we have to add all of its consumers as well as their dependencies to
+            #  `first_nodes`. We could now go on and test all nodes currently inside `first_nodes`
+            #  but instead we will stop here.
+            needs_rescan = True
```
alexnick83: It looks like you are rescanning all the first nodes. However, can you not restrict this to the "unqualified" fnodes?

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+            initial_first_nodes.update(state.scope_subgraph(top_entry_node).nodes())
+
+    # These are all nodes that should be placed into the first state. These are all nodes listed
+    #  in `subgraph` as well as their dependencies. It is important that in the backtracking of
+    #  the dependency we have to follow empty Memlets here, because if we wouldn't then some
+    #  nodes in the first state would be executed before their dependencies, which end up in the
+    #  second state.
+    first_nodes: Set[nodes.Node] = set()
+    for node in initial_first_nodes:
+        find_generating_nodes(
```
alexnick83: So basically, this method finds all the nodes before the "initial_first_nodes" in the same connected component. I would make this method a utility if we don't have something like this already (maybe networkx has something?). It also looks to me that you always want to follow empty memlets, so this parameter is unnecessary.

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
-                if iedge == edge:
+    # Ensure that all dependencies are assigned to the first state.
+    assert all(all(iedge.src in first_nodes for iedge in state.in_edges(first_node)) for first_node in first_nodes)
+    assert all(boundary_node in first_nodes for boundary_node in boundary_nodes)
+    assert all(all(iedge.src in first_nodes for iedge in state.in_edges(bnode)) for bnode in boundary_nodes)
+    assert len(first_nodes) >= 1
+
+    # Now create the new states on which we will operate. It is important that the newly created
+    #  state is at the top.
+    # NOTE: There is the case that `subgraph` contains the entire dataflow in `state`, in that case
```
alexnick83, line 717: Maybe I am misunderstanding this, but if the first nodes encompass all the nodes, then you can just abort/return the input state here, no?

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+    first_nodes_map: Dict[nodes.Node, nodes.Node] = {node: node for node in pure_first_nodes}
+    first_nodes_map.update({old_bnode: copy.deepcopy(old_bnode) for old_bnode in boundary_nodes})
+
+    # Now populate the first state. It is important that the `pure_first_nodes` are at this point
+    #  also inside the second state.
+    first_state.add_nodes_from(first_nodes_map.values())
+
+    # Recreate the edge structure. For that we go through all first node and create the edges,
+    #  but only if both belong to the first state.
+    for node in first_nodes:
```
alexnick83: You could also do the following (may be clearer):
```python
sgraph = StateSubgraphView(second_state, first_nodes)
for edge in sgraph.edges():
   first_state.add_edge(...)
```

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+        for oedge in second_state.out_edges(node):
+            if oedge.dst in first_nodes:
+                first_state.add_edge(
+                    first_nodes_map[oedge.src],
+                    oedge.src_conn,
+                    first_nodes_map[oedge.dst],
+                    oedge.dst_conn,
+                    oedge.data,
+                )
+            elif oedge.data.is_empty():
```
alexnick83: Is there any point in the `elif` and `else`? Can they be removed?

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+                #  `dst` will end up in the second state.
+                pass
+            else:
+                assert oedge.src in boundary_nodes
+
+    # Now we have to remove the nodes that were relocated into the first state from the second
+    #  state, with the exception of the boundary nodes which remains here.
+    # TODO: Investigate if there are side effect from the removal (mutating the nodes), since we
+    #   have not deepcopied them.
+    second_state.remove_nodes_from(pure_first_nodes)
```
alexnick83, line 736: You should probably do the same as you did in `wcr_conversion`, amend the attributes of any NestedSDFGs.

dace/transformation/helpers.py
```diff
@@ -576,126 +576,227 @@ def nest_state_subgraph(sdfg: SDFG,
...
+
+    # Cleaning the connectors of the boundary nodes. In the first state we have to remove all
+    #  outgoing connectors and in the second state all incoming.
+    # NOTE: I am actually not sure why this is here, it was in the original code, so I retained it.
```
alexnick83, line 751: I also don't think this is needed. Maybe ask others as well.

alexnick83 (changes requested):
Overall, nice work. High-level comments:
- `isolate_tasklet` needs to take into account views (see suggestion in comments). I would add a test to make sure it works properly.
- You have repeating patterns where you essentially construct a StateSubgraph and then move it (nodes and edges) to another State. I think these patterns are good targets for utility methods. This would also make the transformations easier to read and maintain.
- You mention that this PR is inspired by `isolated_nested_sdfg`, where there is a decomposition into three states. Would it make sense, perhaps in a future PR, to generalize that decomposition for a subgraph (instead of a NestedSDFG node) and then reuse it for all cases (calling StateFusion appropriately)?
- Right now, StateFission will implicitly move any other connected components to the second state always. Would it be interesting to have an option? We have utility methods for finding connected components (see [here](https://github.com/spcl/dace/blob/186d21de1936f651bba7e4763d3a9466b112c6b2/dace/sdfg/utils.py#L1006)).

dace/transformation/helpers.py
philip-paul-mueller: Essentially if you have the patter:
```
(A) -> [tlet] -> (B) -> (C)
```
and you pass want to split at `A` then only `A` will be moved to the first state and the second state will still have the same pattern.
I am actually not sure, how useful this example is, I found it by accident, see test `test_state_fission_with_access_nodes_1`.
The original behaviour was to not clean up (actually the original function kind of did not care at all) so I kept it.
Since this function is a helper function and if the user wants it, then it should be able to do, maybe he want to attach stuff to it?
What we could do is issue a warning.

dace/transformation/helpers.py
philip-paul-mueller: The two are not equivalent.
If you would rewrite the `or` expression you get something:
```
scanned_nodes = seen if seen else set()
```
And this is highly not the same as
```
scanned_nodes = seen if seen is None else set()
```
The difference happens that if you pass an empty `set` as `seen` then the first code, i.e. the `or`, will create a new `set` since an empty set evaluates to `False`.
However, the user passed an empty `set`, that even have a name outside, to the function in anticipation that it will be filled.
Which does not happen.

dace/transformation/helpers.py
philip-paul-mueller: It kind of looks nicer, thanks.

dace/transformation/helpers.py
philip-paul-mueller: Your suggestion with the `SubgraphView` allowed me to rewrite that part.
Thus we now remove the edges from one state before we add them to another state.
If that is not legal then I have no idea what is.

Thanks a lot for the suggestion.

dace/transformation/helpers.py
philip-paul-mueller: After some thinking I realized that you were write and I did not need it.

dace/transformation/dataflow/wcr_conversion.py
philip-paul-mueller: There was also the `state_fission_after()` function (with no documentation, nor unit test) which I removed.
It was only used in one place and I essentially copied the content there.
It also looks as if I have added the comment.

I have now removed it, but I do not have any idea what it does exactly (beside the obvious thing).

dace/transformation/helpers.py
alexnick83: I see. So this happens when the requested subgraph consists of access nodes (only) that are not connected. It would make sense to print a warning because it is highly likely that the subgraph was selected erroneously. If I am correct, the result of such an operation would be either an empty state or an invalid one, and there are easier ways to produce both.

dace/transformation/helpers.py
alexnick83: I see, true.

dace/transformation/helpers.py
philip-paul-mueller: The original design was because while the classification what a "pure first state node" will never switch from 'yes' to 'no' the classification if a node is a boundary node might change, because of the discovery of new nodes.
I was not doing it because I was afraid that it also involves some complex management chore.
However, I have found a quite nice solution.

dace/transformation/helpers.py
philip-paul-mueller: The only thing that is there is the BFS and DFS algorithm, which are implemented in pure Python, both in DaCe and NetworkX.
However, we do not need BFS or DFS order, we just need the upstream nodes.
These function also has the advantages, that they stop scanning when we reach something that we have encountered already, thus we only explore the graph at most once instead of several times.
I will make it into a utility function.

dace/transformation/helpers.py
philip-paul-mueller: I agree, I tried that, but it did not worked, for some reason it failed spectacular.
This is probably because some function assumes that this second state is created.

I agree that the interface is not that good, I also think that it should return a pair with both states.
The issue is that the caller of this function expect to get a new state back which contains whatever they specified.

You have to understand that this function also bends itself (to some extend) to get backwards compatibility.

philip-paul-mueller (commented):
Thanks for your quick review, I should have addressed all of your comments, please ping if if I forgot one.

I agree that in the long run we should refactor the `isolate_nested_sdfg()` function and the `state_fission()` into one or at least extract the common parts.
The main reason why I did not do it in this PR, is because I would have to touch code I have no idea what it exactly does and I do not have time at the moment (this is why MapFusion is still on hold).
I think the smartest thing would be to return a `pair` containing the two new state, this would also make this optimization simpler, in which you essentially have to move everything from one state to the other.
I would say there should be two transformations a split transformation, and an isolate transformation, which is basically the generalization of `isolate_nested_sdfg()`.

Regarding the other connected components I am not really sure what the best behaviour is, as they can technically end up in any state.
If you have more than one, there are in fact $2^N$ possibilities.
Looking at it from that side, I think that the current behaviour is kind of the best, because the first state only contains the bare minimum that is needed for executing the dataflow in `subgraph` and everything that you do not need to "run" `subgraph` is put in the second state.
Looking at its usage, I get the impression that this was its actual intention, a state that is as clean as possible that contains `subgraph`.

alexnick83 (approved):
LGTM

PR #2048, opened 2025-06-15
Add ThreadBlock Maps as  Preprocessing

ThrudPrimrose:
I am testing how many existing test cases would break if we were to add a "thread block" map to all gpu kernels, if it does not exist, to enable simplification on codegen by removing case distinctions. [WIP]

### DIFF
--- a/dace/transformation/dataflow/add_threadblock_map.py
+++ b/dace/transformation/dataflow/add_threadblock_map.py
@@ -1,0 +1,113 @@
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
+""" This module contains classes and functions that implement the grid-strided map tiling
+    transformation."""
+
+import dace
+from dace.sdfg import SDFG, SDFGState
+from dace.properties import make_properties, SymbolicProperty
+from dace.sdfg import nodes
+from dace.sdfg import utils as sdutil
+from dace.transformation import transformation
+from dace.transformation.dataflow.tiling import MapTiling
+from dace import dtypes
+
+@make_properties
+class AddThreadBlockMap(transformation.SingleStateTransformation):
+    """
+    Adds a thread block schedule to a device map scope
+    """
+
+    map_entry = transformation.PatternNode(nodes.MapEntry)
+
+    # Properties
+    thread_block_size_x = SymbolicProperty(dtype=int, default=32, desc="Number threads in the threadBlock X Dim")
+    thread_block_size_y = SymbolicProperty(dtype=int, default=1, desc="Number threads in the threadBlock Y Dim")
+    thread_block_size_z = SymbolicProperty(dtype=int, default=1, desc="Number threads in the threadBlock Z Dim")
+
+    @classmethod
+    def expressions(cls):
+        return [sdutil.node_path_graph(cls.map_entry)]
+
+    def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
+        if self.thread_block_size_x * self.thread_block_size_y * self.thread_block_size_z > 1024:
+            return False
+
+        if self.map_entry.map.schedule != dtypes.ScheduleType.GPU_Device:
+            return False
+
+        kernel_nodes = graph.all_nodes_between(self.map_entry, graph.exit_node(self.map_entry))
+        for node in kernel_nodes:
+            if isinstance(node, nodes.MapEntry) and node.map.schedule == dace.dtypes.ScheduleType.GPU_ThreadBlock:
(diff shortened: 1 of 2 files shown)


tbennun:
Excellent PR, thank you for doing it! Please add a comment in #2036 to mention that TB maps need to be added as a pass.
Also, I would use the Pass/Pipeline API (specifically passes) to apply the transformation effectively, but this one should be fine too.

Part of the reason it's crashing, I assume, is that you didn't use a proper recursive test for the existence of TB maps. Please check cuda.py (or infer_types.py) for how to use the existing functionality.

ThrudPrimrose:
> Excellent PR, thank you for doing it! Please add a comment in #2036 to mention that TB maps need to be added as a pass. Also, I would use the Pass/Pipeline API (specifically passes) to apply the transformation effectively, but this one should be fine too.
> 
> Part of the reason it's crashing, I assume, is that you didn't use a proper recursive test for the existence of TB maps. Please check cuda.py (or infer_types.py) for how to use the existing functionality.

Yes, it is probably that. I used a variation of this transformation for the auto-tiling passes, therefore, I think the transformation should exist. We could have a pass version of the same transformation such that it aligns in style with other preprocessing passes, then we can have both.

ThrudPrimrose:
> Excellent PR, thank you for doing it! Please add a comment in #2036 to mention that TB maps need to be added as a pass. Also, I would use the Pass/Pipeline API (specifically passes) to apply the transformation effectively, but this one should be fine too.
> 
> Part of the reason it's crashing, I assume, is that you didn't use a proper recursive test for the existence of TB maps. Please check cuda.py (or infer_types.py) for how to use the existing functionality.

I have a new question, I updated to check `gpu_block_size` variable to read if the user set the block size using that feature. Now the question is:

```
    def test_highdim_default_block_size():
    
        @dace.program
        def tester(a: dace.float64[1024, 1024] @ dace.StorageType.GPU_Global):
            for i, j in dace.map[0:1024, 0:1024] @ dace.ScheduleType.GPU_Device:
                a[i, j] = 1
    
        with dace.config.set_temporary('compiler', 'cuda', 'default_block_size', value='32, 8, 2'):
            with pytest.warns(UserWarning, match='has more dimensions'):
                sdfg = tester.to_sdfg()
                gpu_code = sdfg.generate_code()[1]
>               assert 'dim3(32, 16, 1)' in gpu_code.code
E               assert 'dim3(32, 16, 1)' in '\n#include <cuda_runtime.h>\n#include <dace/dace.h>\n\n                                                              ...ms[0]);    ////__DACE:0:0:4\n    DACE_KERNEL_LAUNCH_CHECK(__err, "KernelEntryMap_0_0_4", 32, 128, 1, 32, 8, 1);\n}\n\n'
E                +  where '\n#include <cuda_runtime.h>\n#include <dace/dace.h>\n\n                                                              ...ms[0]);    ////__DACE:0:0:4\n    DACE_KERNEL_LAUNCH_CHECK(__err, "KernelEntryMap_0_0_4", 32, 128, 1, 32, 8, 1);\n}\n\n' = <dace.codegen.codeobject.CodeObject object at 0x7a897435c680>.code
```

If we have a 2D map [0:N, 0:N] and pass block size [32, 8, 2] I would expect the generated code to be launched [32, 8]. 
The behaviour before is to collapse the past dimensions, should it be kept as it is?

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,141 @@
...
+                    if self.thread_block_size_z > 1:
+                        print_warning = True
+                    self.thread_block_size_z = 1
+                elif num_dims_in_map == 1:
+                    self.thread_block_size_x *= self.thread_block_size_y * self.thread_block_size_z
+                    if self.thread_block_size_y > 1 or self.thread_block_size_z > 1:
+                        print_warning = True
+                    self.thread_block_size_y = 1
+                    self.thread_block_size_z = 1
+                new_block = (self.thread_block_size_x, self.thread_block_size_y, self.thread_block_size_z)
```
alexnick83, line 84: I am thinking that it would be simpler to write something similar to:
```python
old_block = (self.thread_block_size_x, self.thread_block_size_y, self.thread_block_size_z)
new_block = list(old_block)
for d in range(3, num_dims_in_map, -1):
    new_block[d-1] *= new_block[d]
    new_block[d] = 1
new_block = tuple(new_block)
if new_block != old_block:
    warnings.warn ...
```

It may not be immediately as readable as the current code though, so this just a suggestion.

alexnick83 (approved):
LGTM

PR #2054, opened 2025-06-18
Avoid losing symbols in `traverse_sdfg_with_defined_symbols`

romanc:
## Description

From what I can tell, `traverse_sdfg_with_defined_symbols()` is supposed to traverse the SDFG, accumulating symbols as it runs down the tree. The current implementation resets the dict of "already known symbols" when recursing down into a state that is the source of an edge. That seems odd. In particular, not even the globally available `sdfg.symbols` are marked as defined anymore with that reset.

The proposed fix passes along already known symbols like in all other cases, ensuring that the dict of known symbols only ever grows.


phschaad (approved):
Good catch, not sure how this was able to slip by.. Thanks!

PR #2055, opened 2025-06-18
Only report cycles if we found some

romanc:
## Description

The issue here is that `find_cycles()` returns a generator, which is truthy even if there are no cycles. The proposed solution is to evaluate the generator. That way, if the generated list is empty, it will be falsy and we won't report cycles if there are none. This allows developers to see the other error message, which is currently unreachable.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -617,16 +617,16 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
 
             # Sanity checks
             if validate and len(eq) != 0:
-                cycles = self.find_cycles()
+                cycles = list(self.find_cycles())
                 if cycles:
-                    raise ValueError('Found cycles in state %s: %s' % (self.label, list(cycles)))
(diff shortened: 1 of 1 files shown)


romanc:
Note: while this PR targets `v1/maintenance`, the issue is also [present in the `main` branch](https://github.com/spcl/dace/blob/main/dace/sdfg/state.py#L627).

tbennun (approved):
How was this not triggered yet?

romanc:
> How was this not triggered yet?

What do you mean by "not triggered yet"? Some tests were running, they just got cancelled because of a testfailure in (3.7, autoopt).

tbennun:
> > How was this not triggered yet?
> 
> What do you mean by "not triggered yet"? Some tests were running, they just got cancelled because of a testfailure in (3.7, autoopt).

To clarify what I am surprised by: how has no SDFG in the past years since this check was written, ever trigger this exception? We have cycle check unit tests, and many many acyclic SDFG states…
This relates to my “please add test” comment in the PR to the main branch.

PR #2056, opened 2025-06-18
Only report cycles if we found some

romanc:
## Description

The issue here is that `find_cycles()` returns a generator, which is truthy even if there are no cycle. The proposed solution is to evaluate the generator. That way, if the generated list is empty, it will be falsy and we won't report cycles if there are none. This allows developers to see the other error message, which is currently unreachable.

Same as PR https://github.com/spcl/dace/pull/2055.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -624,16 +624,16 @@ class DataflowGraphView(BlockGraphView, abc.ABC):
 
             # Sanity checks
             if validate and len(eq) != 0:
-                cycles = self.find_cycles()
+                cycles = list(self.find_cycles())
                 if cycles:
-                    raise ValueError('Found cycles in state %s: %s' % (self.label, list(cycles)))
+                    raise ValueError('Found cycles in state %s: %s' % (self.label, cycles))
                 raise RuntimeError("Leftover nodes in queue: {}".format(eq))
 
             entry_nodes = set(n for n in self.nodes() if isinstance(n, nd.EntryNode)) | {None}
             if (validate and len(result) != len(entry_nodes)):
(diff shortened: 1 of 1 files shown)


tbennun:
@romanc could you please add a test that reproduces this issue? If it’s not simple, let me know

romanc:
We've seen it on an SDFG with dangling/mismatched nodes into a nested SDFG, which I reproduced with

```py
def test_state_scope_children():
    sdfg = dace.SDFG("tester")
    sdfg.add_array("A", [1], dace.float32)
    first_state = sdfg.add_state(is_start_block=True)

    next_state = sdfg.add_state_after(first_state)
    read_A = next_state.add_read("A")
    inner_sdfg = dace.SDFG("inner_tester")
    inner_sdfg.add_array("B", [2], dace.int32)
    inner_state = inner_sdfg.add_state()
    read_B = inner_state.add_read("B")
    tasklet = inner_state.add_tasklet("noop", {"b"}, {}, "")
    inner_state.add_edge(read_B, None, tasklet, "b", dace.Memlet())
    no_inputs = set("b")
    no_outputs = set()
    nested_sdfg = next_state.add_nested_sdfg(inner_sdfg, sdfg, no_inputs, no_outputs)
    next_state.add_edge(read_A, None, nested_sdfg, "undefined", dace.Memlet())

    bla = next_state.scope_children(validate=True)
```

However, that seems to be okay. So I'm not quite sure how to get an "external queue" with leftover items in it [here](https://github.com/spcl/dace/blob/7bc980365a667f9bc00b20265144928defb71d83/dace/sdfg/state.py#L623).

PR #2065, opened 2025-07-02
Python Frontend: Improved support for strided access

alexnick83:
The Python frontend does not always support strided accesses. This PR addresses the following issues:
- [x] Read access is strided, e.g., `v = src[0:40:4]`


tests/python_frontend/strided_copy_test.py
tbennun, line 1: There is a whole file dedicated to testing copies; I’d add the test there instead

alexnick83:
@tbennun please note that I explicitly fail the case where the stride is negative. I think that it is better (for now) to have users use an explicit Map to generate such accesses. Let me know what you think.

tbennun (approved):

PR #2066, opened 2025-07-02
Fixed unsafe modification of `ScopeTree.children`

philip-paul-mueller:
The error occurred with structure such as seen below (although they are inside a Map).

2025-07-02-151241_1104x827_scrot

The error was that an element that was in the list before was added again, probably because the thing was not properly separated. 
The underlying issue was that the SDFG could not be compiled, if one would long enough we would get an out of memory error. 

I will try to turn the SDFG below into an error.

### DIFF
--- a/dace/transformation/helpers.py
+++ b/dace/transformation/helpers.py
@@ -1301,15 +1301,18 @@ def scope_tree_recursive(state: SDFGState, entry: Optional[nodes.EntryNode] = No
     # Add nested SDFGs as children
     def traverse(state: SDFGState, treenode: ScopeTree):
         snodes = state.scope_children()[treenode.entry]
+        children_to_inspect = treenode.children.copy()  # Avoid modifying a cached structure.
         for node in snodes:
             if isinstance(node, nodes.NestedSDFG):
                 for nstate in node.sdfg.states():
                     ntree = nstate.scope_tree()[None]
                     ntree.state = nstate
-                    treenode.children.append(ntree)
-        for child in treenode.children:
+                    assert ntree not in children_to_inspect
+                    children_to_inspect.append(ntree)
+
+        for child in children_to_inspect:
             if hasattr(child, 'state') and child.state != state:
-                traverse(getattr(child, 'state', state), child)
+                traverse(child.state, child)
 
     traverse(state, stree)
     return stree


philip-paul-mueller:
A solution to the cleaning of caches is to implement kind of a `copy()` method for it.
The issue is that you can not simply call `deepcopy()` because you would also copy the scope nodes.
However, looking at `ScopeTree` implementing a sensible copy method should be simple.

Regarding the test.
I have not fully identified the conditions that lead to the issue.
What I can say is:
- It involves nested SDFGs with multiple levels.
- The function before was clearly wrong since the return value, which is cached was modified.

However, I have an SDFG that reproduces the issue.

[compute_ppm4gpu_courant_number.json](https://github.com/user-attachments/files/21051660/compute_ppm4gpu_courant_number.json)

@alexnick83

alexnick83:
I would suggest using Python's copy API:
```python
import copy

class ScopeTree(object):
    """ A class defining a scope, its parent and children scopes, and
        scope entry/exit nodes. """

    def __init__(self, entrynode: nd.EntryNode, exitnode: nd.ExitNode):
        self.parent: 'ScopeTree' = None
        self.children: List['ScopeTree'] = []
        self.entry: nd.EntryNode = entrynode
        self.exit: nd.ExitNode = exitnode
    
    def __copy__(self):
        cls = self.__class__
        result = cls.__new__(cls)
        result.parent = self.parent
        result.children = [copy.copy(child) for child in self.children]
        result.entry = self.entry
        result.exit = self.exit
        return result
```

Then, I think you only need to update `scope_tree_recursive` as follows:
```python
def scope_tree_recursive(state: SDFGState, entry: Optional[nodes.EntryNode] = None) -> ScopeTree:
    """
    Returns a scope tree that includes scopes from nested SDFGs.

    :param state: The state that contains the root of the scope tree.
    :param entry: A scope entry node to set as root, otherwise the state is
                  the root if None is given.
    """
    stree = copy.copy(state.scope_tree()[entry])
    stree.state = state  # Annotate state in tree

    # Add nested SDFGs as children
    def traverse(state: SDFGState, treenode: ScopeTree,):
        snodes = state.scope_children()[treenode.entry]
        for node in snodes:
            if isinstance(node, nodes.NestedSDFG):
                for nstate in node.sdfg.states():
                    ntree = copy.copy(nstate.scope_tree()[None])
                    ...
```

I would also remove the rest of the changes; at least with the above, I can compile your SDFG. I would keep the updated docstrings.

alexnick83 (approved):

PR #2068, opened 2025-07-03
Strip Mining and Therefore MapTiling ignores non-unit stride

ThrudPrimrose:


dace/transformation/dataflow/strip_mining.py
```diff
@@ -125,6 +125,7 @@ def calc_set_union(set_a, set_b):
 class StripMining(transformation.SingleStateTransformation):
     """ Implements the strip-mining transformation.
 
+        TODO: Update doc
```
phschaad, line 128: Please do

phschaad (changes requested):
Doc update missing

dace/transformation/dataflow/strip_mining.py
ThrudPrimrose: As discussed it is a TODO from long ago + we got no unit tests for strip mining or map tiling

phschaad (approved):
LGTM in that case, thanks

PR #2069, opened 2025-07-03
F2dace/fix broken dead dataflow

mcopik:
This fixes an edge case where an access node - in particular, a struct view - is used in a state (dead) and on an interstate edge. Previously, we would remove such a node entirely, which would break the edge.

We solve it by introducing a new variant of AccessSets analysis that splits read/write states between states and interstate edges.


acalotoiu (approved):

acalotoiu:
LGTM. Ideally please add a test to check functionality

PR #2074, opened 2025-07-08
More Tests for NestedSDFG Inlining

philip-paul-mueller:
This adds some more tests, that assert on the structure of the SDFG and not simply tests if they can be compiled or not.

### DIFF
--- a/tests/inline_symbol_test.py
+++ b/tests/inline_symbol_test.py
@@ -16,10 +16,12 @@ def inline_symbol(D: dace.float32[E, F]):
 
 def test_inline_symbol():
     sdfg = inline_symbol.to_sdfg(simplify=False)
-    oldval = dace.Config.get_bool('experimental', 'validate_undefs')
-    dace.Config.set('experimental', 'validate_undefs', value=True)
-    sdfg.validate()
-    dace.Config.set('experimental', 'validate_undefs', value=oldval)
+
+    with dace.config.set_temporary('experimental', 'validate_undefs', value=True):
+        assert sdfg.is_valid()
+
+    with dace.config.set_temporary('experimental', 'validate_undefs', value=False):
+        assert not sdfg.is_valid()
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


tests/inline_symbol_test.py
```diff
@@ -16,10 +16,14 @@ def inline_symbol(D: dace.float32[E, F]):
...
-    oldval = dace.Config.get_bool('experimental', 'validate_undefs')
-    dace.Config.set('experimental', 'validate_undefs', value=True)
-    sdfg.validate()
-    dace.Config.set('experimental', 'validate_undefs', value=oldval)
+
+    # NOTE: Instead of checking if the SDFG is valid one should first, with `validate_undefs`
+    #   set to `False` if validation passes. Then enable it and ensure that the validation
+    #   fails, because as it stands now even when the feature is disabled validation passes.
+    with dace.config.set_temporary('experimental', 'validate_undefs', value=False):
+        assert sdfg.is_valid()
```
tbennun: the use of `assert sdfg.is_valid()` is going to yield less clear results over calling `sdfg.validate()`. The former yields a boolean False and the latter shows what the validation issue is. This is why we call `sdfg.validate()` in our tests.

tests/inlining_test.py
```diff
@@ -645,3 +1300,4 @@ def test(A: dace.float64[96, 32], B: dace.float64[42, 32]):
     test_chain_reduction_1()
     test_chain_reduction_2()
     test_chain_reduction_all()
+    test_multistate_inline_same_symbol_used_on_inside_and_outside_without_writeback_state()
```
tbennun, line 1304: missing function calls to other tests added in this file.

philip-paul-mueller: I added them, but can you please explain to me what the mean of adding them here is?
They need pytest fixtures and all that stuff, meaning that pytest must be installed to just run this file.

acalotoiu (approved):
LGTM. Thank you!

PR #2075, opened 2025-07-08
Do not trigger emit copy on out memlets of a tasklet if memlet is none

ThrudPrimrose:
Do not trigger emit copy on out memlets of a tasklet if memlet is none - this prevents from validation to fail because it thinkgs an empty memlet triggers a copy to an access node without a dst conenctor.

### DIFF
--- a/tests/codegen/no_emit_copy_on_empty_memlet_test.py
+++ b/tests/codegen/no_emit_copy_on_empty_memlet_test.py
@@ -1,17 +1,22 @@
 import dace
 import pytest
 
+
 def _gen_sdfg() -> dace.SDFG:
     sdfg = dace.SDFG('test_no_emit_copy_on_empty_memlet')
     state = sdfg.add_state()
 
-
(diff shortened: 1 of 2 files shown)


tests/codegen/no_emit_copy_on_empty_memlet_test.py
```diff
@@ -0,0 +1,37 @@
+import dace
```
tbennun, line 2: Copyright

tbennun (changes requested):
Looks good. One comment

tests/codegen/no_emit_copy_on_empty_memlet_test.py
ThrudPrimrose: Thanks for reminding, I keep forgetting about that

ThrudPrimrose: Will add

PR #2079, opened 2025-07-10
Apply ConstantPropagation to AccessNodes

iomaganaris:
`ConstantPropagation` pass used to work only for symbols/scalars in expressions like `if` statements.
This PR extends the `replace_dict` functions which is called from the `ConstantPropagation` to handle `AccessNode`s in case the `data` of an `AccessNode` (if it's scalar) is passed in the replacement dictionary.
The `AccessNode` is replaced with a `tasklet` setting the constant value.


tests/python_frontend/constant_and_keyword_args_test.py
```diff
@@ -556,6 +556,35 @@ def conditional_val(A: dace.float64[20], val: dace.compiletime):
...
+            A[:] = val + 1
+
+    a = np.random.rand(20)
+    sdfg = conditional_val_with_access.to_sdfg(a, 3, simplify=True, use_cache=True)
+    from dace.transformation.passes.constant_propagation import ConstantPropagation
+    res = dict()
+    ConstantPropagation().apply_pass(sdfg, res, {'val': 3})
+    sdfg.simplify()
+    assert sdfg.number_of_nodes() == 1
+    sdfg.view()
```
phschaad: Please remove viewing the SDFG from the test (unless it serves a specific test purpose?)

phschaad (changes requested):
LGTM except for a small comment - thanks!

tests/python_frontend/constant_and_keyword_args_test.py
iomaganaris: Thanks for catching this out! I forgot it during debugging 🙈

dace/sdfg/replace.py
```diff
@@ -67,6 +71,47 @@ def replace_dict(subgraph: 'StateSubgraphView',
...
+                elif repl[node.data] in state.sdfg.arrays:
+                    desc = state.sdfg.arrays[repl[node.data]]
+                else:
+                    continue
+                if state.in_degree(node) == 0 and not desc.transient and isinstance(desc, data.Scalar):
+                    node_data_symbolic = dace.symbolic.pystr_to_symbolic(node.data)
+                    if node_data_symbolic in symrepl:
+                        tasklet = state.add_tasklet(name="constant",
+                                                    inputs={},
+                                                    outputs={f'{node.data}_cp_val'},
```
tbennun: What’s this naming convention in replace? Maybe use a  _value suffix

dace/sdfg/replace.py
```diff
@@ -67,6 +71,47 @@ def replace_dict(subgraph: 'StateSubgraphView',
...
+                                                    code=f'{node.data}_cp_val = {symrepl[node_data_symbolic]}')
+                        if f'{node.data}_cp' not in sdfg.arrays:
```
tbennun: Why not use add_transient with find_new_name=True? You can get the new name from the return value.

dace/symbolic.py
```diff
@@ -1675,3 +1677,39 @@ def symbols_in_code(code: str, potential_symbols: Set[str] = None, symbols_to_ig
     if symbols_to_ignore is None:
         return tokens
     return tokens - symbols_to_ignore
+
+
+def sympyexpr_to_python(val: Union[Integer, Float, Rational, Expr, int, float, complex]) -> Union[int, float, complex]:
```
tbennun: Did you see symbolic.evaluate?

tbennun (changes requested):
This is a great way to implement it. I left some important comments though.

dace/sdfg/replace.py
iomaganaris: The `AccessNode.data` will be `{node.data}_cp` where `cp` stands for `ConstantPropagation`.
The `{node.data}_cp_val` is the name of the variable in the tasklet.
I can replace `val` with `value`

dace/sdfg/replace.py
iomaganaris: That way for every instance of the same value that gets propagated we would need to add a new array to the SDFG. Not sure if there's an issue with that but if we replace the same value maybe it's not necessary to add a new array

dace/symbolic.py
iomaganaris: Good idea. I wasn't aware of that 👍

dace/sdfg/replace.py
tbennun: cp stands for constant propagation, right? Then the name has no place in “replace.py”, because other methods or passes can call it

tbennun: Remove cp please

dace/sdfg/replace.py
tbennun: More scalar transients are safer to add than one value. The compiler optimizes them away anyway.

dace/sdfg/replace.py
iomaganaris: Fair point. I have removed `cp` from the suffix

dace/sdfg/replace.py
```diff
@@ -67,6 +71,46 @@ def replace_dict(subgraph: 'StateSubgraphView',
...
+                else:
+                    continue
+                if state.in_degree(node) == 0 and not desc.transient and isinstance(desc, data.Scalar):
+                    node_data_symbolic = dace.symbolic.pystr_to_symbolic(node.data)
+                    if node_data_symbolic in symrepl:
+                        tasklet = state.add_tasklet(name="constant",
+                                                    inputs={},
+                                                    outputs={f'{node.data}_value'},
+                                                    code=f'{node.data}_value = {symrepl[node_data_symbolic]}')
+                        access_node_name, _ = sdfg.add_transient(f'{node.data}', [1],
```
iomaganaris, line 95: For reference
```
type(sympyexpr_to_python(symrepl[node_data_symbolic])),
```
is necessary because `symbolic.evaluate` didn't always return an `int` but it could return a `symbolic.Number.Integer` which was problematic with `sdfg.add_transient`

dace/sdfg/replace.py
```diff
@@ -91,23 +91,22 @@ def replace_dict(subgraph: 'StateSubgraphView',
...
-                            sdfg.add_array(f'{node.data}_cp', [1],
-                                           type(sympyexpr_to_python(symrepl[node_data_symbolic])),
-                                           transient=True)
-                        tmp_an = state.add_access(f'{node.data}_cp')
-                        state.add_edge(tasklet, f'{node.data}_cp_val', tmp_an, None,
-                                       Memlet.simple(f'{node.data}_cp', '0'))
+                                                    outputs={f'{node.data}_value'},
+                                                    code=f'{node.data}_value = {symrepl[node_data_symbolic]}')
+                        access_node_name, _ = sdfg.add_transient(f'{node.data}', [1],
+                                                                 type(sympyexpr_to_python(symrepl[node_data_symbolic])),
```
tbennun: in the way you are using this method, it would also not work if you are using a value such as np.float16 or some more explicit type. Perhaps consider using `data.create_datadescriptor(symrepl[node_data_symbolic]).dtype` ?
Here is, for example, how we handle constants in the Python frontend: https://github.com/spcl/dace/blob/main/dace/frontend/python/replacements/utils.py#L45

dace/sdfg/replace.py
tbennun: See other comment.

dace/sdfg/replace.py
iomaganaris: Based on your suggestion I replaced `sympyexpr_to_python` with a call to `dtypes.typeclass`, which is the part of `create_datadescriptor` we actually want to use. I just had to extend `dtypes.typeclass` to handle the `SymPy Number`s necessary.

dace/sdfg/replace.py
```diff
@@ -67,6 +70,47 @@ def replace_dict(subgraph: 'StateSubgraphView',
...
+                if state.in_degree(node) == 0 and not desc.transient and isinstance(desc, data.Scalar):
+                    node_data_symbolic = dace.symbolic.pystr_to_symbolic(node.data)
+                    if node_data_symbolic in symrepl:
+                        tasklet = state.add_tasklet(name="constant",
+                                                    inputs={},
+                                                    outputs={f'{node.data}_value'},
+                                                    code=f'{node.data}_value = {symrepl[node_data_symbolic]}')
+                        access_node_name, _ = sdfg.add_transient(f'{node.data}', [1],
+                                                                 dtypes.typeclass(type(
+                                                                     symrepl[node_data_symbolic])).dtype,
```
tbennun: The .dtype is not necessary as add_transient accepts a typeclass

dace/symbolic.py
```diff
@@ -10,6 +10,8 @@
 
 import sympy.abc
 import sympy.printing.str
+from sympy.core.numbers import Integer, Float, Rational
+from sympy.core.expr import Expr
```
tbennun:
```suggestion
```

dace/symbolic.py
```diff
@@ -10,6 +10,8 @@
 
 import sympy.abc
 import sympy.printing.str
+from sympy.core.numbers import Integer, Float, Rational
```
tbennun:
```suggestion
```

tbennun (changes requested):
Very minor change but otherwise approved!

PR #2080, opened 2025-07-11
Improved `PruneSymbols`

philip-paul-mueller:
Before for every symbol, that should be removed, the whole (nested) sdfg was scanned to test if it can be removed. Now, the set of symbols that can not be removed is determined and then used. Thus the SDFG is only scanned once.


dace/transformation/dataflow/prune_connectors.py
```diff
@@ -153,22 +153,40 @@ def _candidates(nsdfg: nodes.NestedSDFG) -> Set[str]:
...
+        #  that it can be implemented in terms of `used_symbols()` but I am unsure if it is
+        #  so simple as `return sdfg.used_symbols()`?
+
+        unremovable_symbols: Set[str] = set()
+
+        for desc in sdfg.arrays.values():
+            unremovable_symbols.update(map(str, desc.free_symbols))
+        for state in sdfg.states():  # Previously `nodes()` was used.
+            unremovable_symbols.update(state.free_symbols)
+        for e in sdfg.edges():  # Are this all edges?
```
phschaad: This iterates only over the top-level edges - what you are looking for here is `all_interstate_edges`, which grabs all edges recursively down control flow regions (not recursing into nested SDFGs)

dace/transformation/dataflow/prune_connectors.py
```diff
@@ -153,22 +153,40 @@ def _candidates(nsdfg: nodes.NestedSDFG) -> Set[str]:
...
+        # The implementation of this function is such that it has the same behaviour as before,
+        #  where `dace.transformation.helpers.is_symbol_unused()` was used. However, I am sure
+        #  that it can be implemented in terms of `used_symbols()` but I am unsure if it is
+        #  so simple as `return sdfg.used_symbols()`?
+
+        unremovable_symbols: Set[str] = set()
+
+        for desc in sdfg.arrays.values():
+            unremovable_symbols.update(map(str, desc.free_symbols))
+        for state in sdfg.states():  # Previously `nodes()` was used.
```
phschaad:
```suggestion
        for state in sdfg.states():
```

`nodes()` was definitely wrong, if operating on a per-SDFG level as opposed to a per-CFG level - good catch.

dace/transformation/dataflow/prune_connectors.py
```diff
@@ -153,22 +153,40 @@ def _candidates(nsdfg: nodes.NestedSDFG) -> Set[str]:
 
         return candidates
 
+    def _find_symbols_that_can_not_be_removed(self, sdfg: SDFG) -> Set[str]:
+        """Find the set of symbols that can not be removed."""
+
+        # The implementation of this function is such that it has the same behaviour as before,
+        #  where `dace.transformation.helpers.is_symbol_unused()` was used. However, I am sure
+        #  that it can be implemented in terms of `used_symbols()` but I am unsure if it is
+        #  so simple as `return sdfg.used_symbols()`?
```
phschaad: I believe `used_symbols` ends up returning a subset of what is returned by this function - at least if run with the parameter `all_symbols=False`. I would vote to keep it as-is for now, but would be good if you could add a TODO or NOTE here about investigating this in the future.

phschaad (changes requested):
In general LGTM, only small comments

dace/transformation/dataflow/prune_connectors.py
philip-paul-mueller: Thanks.

philip-paul-mueller:
@phschaad Thanks for your review.
I have applied your suggestions, I also applied them to the original function, on which the new one was based on.

PR #2081, opened 2025-07-11
Updated the `SubgraphView` mainly the `edges()` function.

philip-paul-mueller:
Before the edges were stored inside a `list` thus every check, i.e. `in` resulted in a linear time operation. This was particular devastating for `edges()` since it degraded to quadratic runtime. This PR changes the underlying data structure to a `dict` which is still ordered but allow much faster `in` tests, which is the main workflow. The caveat now is that `nodes()` result in the construction of a `list`.

This PR does not change the observable behaviour of the class, it just makes it faster. The only change is

Note:
I would propose to store the list of nodes inside a `set` to begin with. The only reason why they are now in a `dict` is because they were originally ordered by their ID in the parent graph. Since the nodes are not necessarily consecutive, I do not really see why this is needed.


tests/transformations/subgraph_fusion/smax_test.py
acalotoiu, line 1: What has changed here?

tests/transformations/subgraph_fusion/util.py
acalotoiu, line 1: Is this change related to this PR?

tests/transformations/subgraph_fusion/smax_test.py
philip-paul-mueller: If you look at `SubgraphView::add_nodes()` you see that it throws an exception as you are not supposed to modify you change them, however, this file was doing it.
Thus I had to change that to adapt to the new format.

The massive changes are because that file had Windows encoding before and I changed it to UNIX.
If you look for example on line 66 you see that there some parts are more green than the rest.
And this indicates the changes.

tests/transformations/subgraph_fusion/util.py
philip-paul-mueller: Same reason as for the changes in [smax_test.py](https://github.com/spcl/dace/pull/2081/files/3e13cc0ad4c5b79a2bca4f62d3fba8bb443babf8#diff-28534700de15660d03969c158f3528242e76c9dc1c8443e216b0a001560041fc), i.e. someone break encapsulation and I had to clean up after that person.

acalotoiu: Thank you

tests/transformations/subgraph_fusion/smax_test.py
acalotoiu: Thank you!

PR #2083, opened 2025-07-13
Improve Validation 

ThrudPrimrose:
Improve validation to:
- Detect writes to scalars on interstate edges. For this I added `_no_writes_to_scalars_or_arrays_on_interstate_edges` and a test to `interstate_edge_utils_test` to check that an SDFG where we write to a scalar on iedge is invalid.
- Do not mark tasklets with no incoming and outgoing memlets as sink nodes. Accroding to discussion with Tal earlier the image from the SDFG should be valid as it is not a sink node due to no incoming our outgoing memlets, I improve validation to not mark this tasklet as sink node + add a test case :
image

### DIFF
--- a/dace/sdfg/validation.py
+++ b/dace/sdfg/validation.py
@@ -870,8 +870,16 @@ def validate_state(state: 'dace.sdfg.SDFGState',
         # unless the memlet is empty in order to connect to a scope
         elif scope_contains_scope(scope, dst_node, src_node):
             if not isinstance(dst_node, nd.AccessNode):
-                if e.data.is_empty() and isinstance(dst_node, nd.ExitNode):
-                    pass
+                # It is also possible that edge leads to a tasklet that has no incoming or outgoing memlet
+                # since the check is to be performed for all edges leading to the dst_node, it is sufficient
+                # to check for the memlets of outgoing edges
+                if e.data.is_empty():
+                    if isinstance(dst_node, nd.ExitNode):
+                        pass
+                    if isinstance(dst_node, nd.Tasklet) and all(
+                        {oe.data.is_empty()
+                         for oe in state.out_edges(dst_node)}):
+                        pass
                 else:
                     raise InvalidSDFGEdgeError(
                         f"Memlet creates an invalid path (sink node {dst_node}"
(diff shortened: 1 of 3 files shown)


dace/sdfg/validation.py
```diff
@@ -353,6 +353,9 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG', references: Set[int] = None, **context
                             "Arrays that use a multibank access pattern must have the size of the first dimension equal"
                             f" the number of banks and have at least 2 dimensions for array {name}", sdfg, None)
 
+        # Check for interstate edges that write to scalars or arrays
+        _no_writes_to_scalars_or_arrays_on_interstate_edges(sdfg)
```
phschaad: I suggest moving this to `validate_control_flow_region`, as it more directly makes use of the hierarchical, top-down traversal / validation. Otherwise you need to recursively go over interstate edges in this method, which causes twice the number of traversals necessary.

dace/sdfg/validation.py
```diff
@@ -1156,3 +1167,14 @@ def validate_memlet_data(memlet_data: str, access_data: str) -> bool:
     memlet_tokens = memlet_data.split('.')
     mem_root = '.'.join(memlet_tokens[:len(access_tokens)])
     return mem_root == access_data
+
+
+def _no_writes_to_scalars_or_arrays_on_interstate_edges(sdfg: 'dace.sdfg.SDFG'):
+    from dace.sdfg import InterstateEdge
+    for edge, graph in sdfg.all_edges_recursive():
```
phschaad: There's also `all_interstate_edges` - though as the previous comment states, this should imo be changed to go over a single control flow region only and called per control flow region.

tests/codegen/tasklet_generation_test.py
```diff
@@ -0,0 +1,61 @@
+import dace
```
phschaad, line 1: Wait are these tests related to the validation?

phschaad (changes requested):
Small comments, but great addition, thanks!

dace/sdfg/validation.py
ThrudPrimrose: Done it, I now do it within control flow regions and do not call edges recursively

phschaad (approved):

PR #2088, opened 2025-07-17
GPU codegen crashes and generates incorrect code with dynamic inputs to seq. maps inside GPU kernels or gpu dev. maps

ThrudPrimrose:
GPU codegen crashes and generates incorrect code with dynamic inputs to seq. maps inside GPU kernels or gpu dev. maps


dace/codegen/targets/cuda.py
```diff
@@ -1557,7 +1557,18 @@ def generate_scope(self, sdfg: SDFG, cfg: ControlFlowRegion, dfg_scope: StateSub
 
         # Handle dynamic map inputs
         for e in dace.sdfg.dynamic_map_inputs(state, scope_entry):
-            kernel_args[str(e.src)] = e.src.desc(sdfg)
+            data_name = None
+            data_desc = None
+            if isinstance(e.src, nodes.AccessNode):
```
alexnick83: I was thinking that it would be cleaner to use the source of the Memlet path, but I wonder if there are other cases to consider (e.g., direct scalar Tasklet output --- unsure if that would be valid).

Further food for thought: Considering your fix, why not just query the Memlet data directly? What is the point of checking the access node in the first place?

dace/codegen/targets/cuda.py
```diff
@@ -1770,7 +1781,15 @@ def generate_scope(self, sdfg: SDFG, cfg: ControlFlowRegion, dfg_scope: StateSub
 
         # make sure dynamic map inputs are properly handled
         for e in dace.sdfg.dynamic_map_inputs(state, scope_entry):
+            if e.data is not None and e.data.data == e.dst_conn:
```
alexnick83, line 1819: We had this issue before, and we had a pass that renamed the connectors to avoid clashing, but it may be frontend-dependent. Maybe take note of whether we should handle the issue at an earlier stage (see code generator redesign).

dace/codegen/targets/cuda.py
ThrudPrimrose: I think there needs to be an intermediate access node and it can't come directly from a tasklet. A question: does memlet path return the source data type or the view if the first access node is a view? 
I think accessing the memlet data directly should be enough in this case - I guess I will update it to do so

dace/codegen/targets/cuda.py
ThrudPrimrose: Well, having an additional check would not hurt. But it should definitely be handled by the pass. I did this fix because I was not aware of it.

dace/codegen/targets/cuda.py
alexnick83: The memlet path stops at the first access or code node found, so it can very well be a view. If you want to find the underlying non-view container, you need to do a separate search, outside the memlet path.

ThrudPrimrose: Are there cases where the `memlet.data` can be different than the source node of the memlet path when the dst node (and the last edge) connects to a map entry? Nothing directly came to my mind

ThrudPrimrose: I check the edge data directly, in case of an dynamic in connector it does not make sense the data to be none.

alexnick83 (approved):
LGTM

Issue #1911, opened 2025-01-29
`PatternTransformation._pipeline_results` only set during calls to `apply()`.


philip-paul-mueller:
Because `PatternTransformation`, i.e. most of the data flow transformations, including `MapFusion` are called through the `apply()` function, that lacks the `pipeline_results` argument that `apply_pass()` has.
To make this argument available to such passes, `PatternTransformation` has this member `_pipeline_results`, that is set to the value of `pipeline_results` argument that was passed to the `apply_pass()` method.
However, this member is only set during the execution of `apply()`, but when `can_be_applied()` is run the value is `None`.

I do not see a reason why it has to be that way and would propose to change this, such that `_pipeline_results` is also set during `can_be_applied()`.

philip-paul-mueller:
I also realized that this also affects the `can_be_applied_to()` and `apply_to()` functions.

PR #2090, opened 2025-07-21
Added Fine Grain Control to `MapFusionVertical`.

philip-paul-mueller:
The transformation already offers the `assume_always_shared` flag, which tells the transformation to assume that the intermediate is shared, which allows to skip the scan of the SDFG.
Intermediate are classified into two categories:
- Shared: Which need to be retained, because they are used somewhere else.
- Exclusive: Only needed to exchange information between the two Maps.

This PR introduces the `require_exclusive_intermediates` flag, which instructs the transformation to only apply if the intermediates are exclusive, i.e. will be removed after the fusion.

It is important that `require_exclusive_intermediates` does not influence the "pure outputs" of the first Map, i.e. the outputs that are not used by the second Map.
To address this the `require_only_intermediates` flag was introduced.
The flag instructs the transformation to only apply if all outputs of the first Map are intermediates, i.e. are consumed by the second Map.

The main use case for these flags is if the transformation is used through the `apply_to()` and `can_be_applied_to()` functions.
Where one wants to make sure that some intermediate is removed.

This PR also introduced the undocumented `_single_use_data` argument to the initializer of the transformation.
The `_single_use_data` member of the transformation is a work around to handle [issue#1911](https://github.com/spcl/dace/issues/1911).
However, when used with `apply_to()` and `can_be_applied_to()` then it is not possible to specify the information, although it is present.
This is not nice and instead the underlying issue should be addressed.

### DIFF
--- a/dace/transformation/dataflow/map_fusion_vertical.py
+++ b/dace/transformation/dataflow/map_fusion_vertical.py
@@ -58,20 +58,13 @@ class MapFusionVertical(transformation.SingleStateTransformation):
     speed this up. The second way is to use the transformation inside a pipeline, which includes the
     `FindSingleUseData` analysis pass, see note below. If the result of this pass is present then
     the transformation will use it instead to determine if a intermediate can be removed.
-    The second way is to either specify `assume_always_shared` or `assume_always_single_use_data`
-    (see note below). Which instruct the transformation to assume that the intermediate is either
-    shared, i.e. will become an output of the fused Map or that it is not used anywhere else
-    and will be removed (which means that if it is specified but the data is used somewhere else,
-    an invalid SDFG will be produced). However, the function will always perform local checks,
-    i.e. checks the degree of the intermediate and if it refers to a non-transient data. It is
-    important that. Furthermore, if a pipeline is present it will take precedence.
+    The second way is to specify `assume_always_shared`, which instruct the transformation to
+    assume that the intermediate is shared, i.e. will become an output of the fused Map.
 
     :param only_inner_maps: Only match Maps that are internal, i.e. inside another Map.
     :param only_toplevel_maps: Only consider Maps that are at the top.
     :param strict_dataflow: Which dataflow mode should be used, see above.
     :param assume_always_shared: Assume that all intermediates are shared.
-    :param assume_always_single_use_data: Assume that all intermediates are single use data,
-        i.e. are no longer needed.
     :param require_exclusive_intermediates: If `True` then the transformation will only apply if
         all intermediates can be removed.
     :param consolidate_edges_only_if_not_extending: If `True`, the default is `False`,
@@ -82,12 +75,8 @@ class MapFusionVertical(transformation.SingleStateTransformation):
         goes to the same AccessNode.
 
     :note: This transformation modifies more nodes than it matches.
-    :note: While it is always "safe" to specify the `assume_always_shared`, this is not true
-        for `assume_always_single_use_data`. Specifying it when the data is used somewhere
-        else will lead to invalid behaviour. Thus only specify it if you know what you are doing.
-    :note: The flags `assume_always_shared` and `assume_always_single_use_data` are intended
-        when it is clear from the usage context what should happen to the intermediate. This is
-        most often, but not always the case if `can_be_applied_to()` or `apply_to()` are used.
+    :note: Specifying `assume_always_shared` is generally safe, as it will not
+        remove writes and global data. However, it will potentially lead to dead dataflow.
     :note: Because of [issue#1911](https://github.com/spcl/dace/issues/1911) the `can_be_applied()`
         can not use the pipeline result and will thus scan the whole SDFG. The `FullMapFusion`
         pass is not affected by this.
@@ -124,15 +113,10 @@ class MapFusionVertical(transformation.SingleStateTransformation):
         default=False,
         desc="If `True` then all intermediates will be classified as shared.",
     )
-    assume_always_single_use_data = properties.Property(
-        dtype=bool,
-        default=False,
-        desc="If `True` then all intermediates are classified as single use data.",
-    )
     require_exclusive_intermediates = properties.Property(
         dtype=bool,
         default=False,
-        desc="If `True` then all intermediates need to be 'exclusive', i.e. can be removed.",
+        desc="If `True` then all intermediates need to be 'exclusive', i.e. they will be removed by the fusion.",
     )
 
     never_consolidate_edges = properties.Property(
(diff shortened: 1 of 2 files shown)


philip-paul-mueller:
My first idea was to create the `assume_always_single_use_data` argument.
But then I realized that that thing is not useful and I needed the `require_exlclusive_intermediate` argument.

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -55,33 +55,36 @@ class MapFusionVertical(transformation.SingleStateTransformation):
...
-    `FindSingleUseData` analysis pass, see note below. If the result of this pass is present then the
-    transformation will use it instead to determine if a intermediate can be removed. Note that
-    `assume_always_shared` takes precedence. For this pattern the `FullMapFusion` pass is provided,
-    that combines the analysis pass and `MapFusionVertical`.
+    speed this up. The first way is to use the transformation inside a pipeline, which includes
+    the `FindSingleUseData` analysis pass, see note below. If the result of the pass is present
+    then the transformation will use it  to determine if a intermediate can be removed.
+    The second way is to specify `assume_always_shared`, which instruct the transformation to
+    assume that the intermediate is shared, i.e. will become an output of the fused Map.
+    The downside of this is, that dead dataflow, i.e. writes that are not needed are generated.
```
phschaad:
```suggestion
    The downside of this is, that dead dataflow, i.e., writes that are not needed, are generated.
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -55,33 +55,36 @@ class MapFusionVertical(transformation.SingleStateTransformation):
...
-    :param assume_always_shared: Assume that all intermediates are shared.
-    :param consolidate_edges_only_if_not_extending: If `True`, the default is `False`,
-        the transformation will only consolidate edges if this does not lead to an
-        extension of the subset.
+    :param assume_always_shared: Handle all intermediate nodes as if they were classified as
+        "shared" , see `partition_first_outputs()` for more.
+    :param require_exclusive_intermediates: If `True` then the transformation will only apply
+        if all intermediates are "eclusive", i.e can be removed, see `partition_first_outputs()`.
+    :param require_all_intermediates: If `True` then the transformation will only apply if
+        the if all outputs of the first Map are intermediate, i.e. are consumed by the second Map.
```
phschaad:
```suggestion
        all outputs of the first Map are intermediate, i.e., are consumed by the second Map.
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -55,33 +55,36 @@ class MapFusionVertical(transformation.SingleStateTransformation):
...
 
     :param only_inner_maps: Only match Maps that are internal, i.e. inside another Map.
     :param only_toplevel_maps: Only consider Maps that are at the top.
     :param strict_dataflow: Which dataflow mode should be used, see above.
-    :param assume_always_shared: Assume that all intermediates are shared.
-    :param consolidate_edges_only_if_not_extending: If `True`, the default is `False`,
-        the transformation will only consolidate edges if this does not lead to an
-        extension of the subset.
+    :param assume_always_shared: Handle all intermediate nodes as if they were classified as
+        "shared" , see `partition_first_outputs()` for more.
```
phschaad:
```suggestion
        "shared", see `partition_first_outputs()` for more.
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -55,33 +55,36 @@ class MapFusionVertical(transformation.SingleStateTransformation):
...
+    :param require_exclusive_intermediates: If `True` then the transformation will only apply
+        if all intermediates are "eclusive", i.e can be removed, see `partition_first_outputs()`.
```
phschaad:
```suggestion
        if all intermediates are "eclusive", i.e., can be removed, see `partition_first_outputs()`.
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -55,33 +55,36 @@ class MapFusionVertical(transformation.SingleStateTransformation):
...
     :note: This transformation modifies more nodes than it matches.
-    :note: If `assume_always_shared` is `True` then the transformation will assume that
-            all intermediates are shared. This avoids the problems mentioned above with
-            the cache at the expense of the creation of dead dataflow.
     :note: Because of [issue#1911](https://github.com/spcl/dace/issues/1911) the `can_be_applied()`
-            can not use the pipeline result and will thus scan the whole SDFG. The `FullMapFusion`
-            pass is not affected by this.
+        can not use the pipeline result and will thus scan the whole SDFG. The `FullMapFusion`
+        pass is not affected by this.
+    :note: `require_exclusive_intermediates` means that all intermediates, i.e. AccessNodes
```
phschaad:
```suggestion
    :note: `require_exclusive_intermediates` means that all intermediates, i.e., AccessNodes
```

dace/transformation/passes/full_map_fusion.py
```diff
@@ -120,6 +138,9 @@ def __init__(
         if consolidate_edges_only_if_not_extending is not None:
             self.consolidate_edges_only_if_not_extending = consolidate_edges_only_if_not_extending
 
+        # TODO(phimuell): Raise an error if a flag was specified and the component that
+        #   needs it is disabled.
```
phschaad: Is there a reason this is not in this PR yet?

dace/transformation/passes/full_map_fusion.py
```diff
@@ -33,16 +33,28 @@ class FullMapFusion(ppl.Pass):
...
+
     assume_always_shared = properties.Property(
         dtype=bool,
         default=False,
         desc="If `True` then all intermediates will be classified as shared.",
     )
+    require_exclusive_intermediates = properties.Property(
+        dtype=bool,
+        default=False,
+        desc="If `True` then all intermediates need to be 'exclusive', i.e. they will be removed by the fusion.",
```
phschaad:
```suggestion
        desc="If `True` then all intermediates need to be 'exclusive', i.e., they will be removed by the fusion.",
```

dace/transformation/passes/full_map_fusion.py
```diff
@@ -33,16 +33,28 @@ class FullMapFusion(ppl.Pass):
...
+    )
+    require_all_intermediates = properties.Property(
+        dtype=bool,
+        default=False,
+        desc="If `True` all outputs of the first Map must be intermediate, i.e. going into the second Map.",
```
phschaad:
```suggestion
        desc="If `True` all outputs of the first Map must be intermediate, i.e., going into the second Map.",
```

dace/transformation/passes/full_map_fusion.py
philip-paul-mueller: The reason is that I was not sure if this is a good idea.
Because then one would have to process the arguments appropriately.
But I think it might be a good idea.

phschaad: I believe it makes sense to do that, as it may eliminate a possibly large source of confusion if something is not run when a user sets a corresponding flag.

philip-paul-mueller: I have now updated it.

phschaad (approved):
Thank you, LGTM

PR #2091, opened 2025-07-21
Optimizations Related to Symbols

philip-paul-mueller:
Some operations that operates on symbols are very slow.
This PR optimizes some of the usage.


phschaad (approved):
LGTM - interesting improvement with the change in conversion, thanks

Issue #1703, opened 2024-10-24
Invalid Cache State of Memlets


philip-paul-mueller:
Memlets have two subset properties `subset` and `other_subset`.
While `subset` always refers to the data container the `data` attribute of the Memlet is pointing to, `other_subset` is option (it is also not super clear to me what `None` in this case means) and describes to "other" side of the connection.
However, `subset` does not necessarily describes the subset at the source of the connection nor the subset at the destination.
For this the properties `src_subset` and `dst_subset` are provided, that takes the direction into account.
However, this information is sometimes outdated and to overcome it `try_initialize()` of the Memlet must be called.
In addition there are also `get_{src, dst}_subset()` which first call `try_initialize()`.

In order to make [PR#1678](https://github.com/spcl/dace/pull/1678) work, we had to call `try_initialize()` on all edges in the state, which is a very expensive operation.

pratyai:
- The interpretation of `other_subset` deserves at least a detailed doc-string at this point, in my opinion.
- Do you recall what happened before using `try_initialize()`?
I've been trying to see when `_is_data_src` field is `None` when calling `src_subset` and `dst_subset`. So far, the only time that happens when two memlets are compared with `==` or `!=` in `newast.py` or `propagation.py` etc. files that work with partly initialized graph anyway (because they are building it).
Turns out they could just just be compared with `is` and `is not` --- as far as I can tell, the tests do not fail that way. But that is an entirely different comparison.

PR #2092, opened 2025-07-21
Refine `try_initialize()` in Map Fusion

philip-paul-mueller:
To ensure that we have consistent subsets inside the Memlets, see [issue#1703](https://github.com/spcl/dace/issues/1703), we have to call `try_initialize()` on the edges.
This PR restrict this to the edges that are actually affected and gives thus a speedup.


phschaad (approved):
LGTM, thanks

PR #2093, opened 2025-07-22
Better Use Of Simplify

philip-paul-mueller:
Simplifying expressions is a relatively expensive operation.
As a performance improvement the code was refactored in such a way to that calling simplify is avoid or deferred as much as possible.
The general idea was to first try doing it without simplify and then if no "positive" answer was obtained try it again with simplify.

The code refactoring mostly preserves the previous behaviour, with the exception that some `TypeError`s, which are generated if the truth value of SymPy expression was determined were removed, since they were never propagated to the user.
Instead they now stay within the function.

It was also discovered that the assumption of `symbol_positivity` was used in a very aggressive way.
It is even possible that it is used in places where it should not.
However, in the 2025-07-24 DaCe meeting it was decided to leave the current behaviour, since major changes are expected soon in that part of the code.

### DIFF
--- a/dace/subsets.py
+++ b/dace/subsets.py
@@ -23,26 +23,30 @@ def _no_simplify(expr):
     return expr
 
 
-def bounding_box_cover_exact(subset_a, subset_b) -> bool:
+def bounding_box_cover_exact(subset_a, subset_b, approximation=False) -> bool:
     """Test if `subset_a` covers `subset_b`.
 
-    Essentially the function computes the bounding box of each subset and checks if
-    the box from `subset_a` is at least as big as `subset_b`. The main difference to
-    `bounding_box_symbolic_positive()` is that it is not possible to use approximative
-    bounds and that this function does not assumes positive symbols, see note.
+    The function uses a bounding box to test if `subset_a` covers `subset_b`,
+    i.e. that `subset_a` is at least as big as `subset_b`. By default the
+    box is constructed using `{min, max}_element()` or if `approximation` is
+    `True` `{min, max}_element_approx()`. The most important difference compared
+    to `bounding_box_cover_exact()` is that this function does not assume
+    that the symbols are positive.
 
     The function returns `True` if it _can be shown_ that `subset_a` covers `subset_b`
     and `False` otherwise.
 
     :param subset_a: The first subset, the one that should cover.
     :param subset_b: The second subset, the one that should be convered.
+    :param approximation: If `True` then use the approximated bounds.
 
-    :note: The function might assume positive symbols.
+    :note: Although the function claims that it does not assumes positivity,
+        this is not guaranteed. This might be a bug.
     """
-    min_elements_a = subset_a.min_element()
-    max_elements_a = subset_a.max_element()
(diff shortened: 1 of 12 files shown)


phschaad:
Question based on the discussion from the DaCe weekly yesterday, 24th of July 2025: is the PR review ready or are further changes expected?

philip-paul-mueller:
@phschaad I hope and expect no bigger changes to this header, with the exception of an updated PR description.

philip-paul-mueller:
@phschaad updated the description.
I consider this PR now as done and ready for review.

dace/subsets.py
```diff
@@ -45,24 +95,44 @@ def bounding_box_symbolic_positive(subset_a, subset_b, approximation=False) -> b
...
-        # NOTE: We first test for equality, which always returns True or False. If the equality test returns
-        # False, then we test for less-equal and greater-equal, which may return an expression, leading to
-        # TypeError. This is a workaround for the case where two expressions are the same or equal and
-        # SymPy confirms this but fails to return True when testing less-equal and greater-equal.
+        # NOTE: Applying simplify takes a lot of time, thus we try to avoid it and try to do the test
+        #   first with the symbols we get and if we are unable to figuring out something, we run
+        #   simplify. Furthermore, we also try to postpone `nng()` as long as we can.
+        # NOTE: We have to use the `== True` test because of SymPy's behaviour. Otherwise we would
+        #   get an expression resulting in a `TypeError`.
+        # TODO: Figuring out why we use this two stage version, instead of `(y <= x) == True`.
```
phschaad: It used to be in the past - and intuitively I assume that still holds true - that an equality check has a much lower optimal runtime complexity than the less-than/greater-than comparison. Specifically for cases where equality is trivially given with no rearrangement / simplification. Note that I have not verified that this is in fact still the case.

TL;DR: speed

dace/subsets.py
```diff
@@ -107,14 +165,20 @@ def covers_precise(self, other):
...
-        except TypeError:
+        if symbolic_positive and (not bounding_box_cover_exact(self, other)):
             return False
+        elif not bounding_box_symbolic_positive(self, other):
+            return False
+
+        # TODO: The original implementation always called `nng()` on the argument before passing
+        #   it into the simplify function. This looks like an error, calling it should depend
+        #   on the value of `symbolic_positive`.
+        simplify = lambda expr: symbolic.simplify_ext(nng(expr))
```
phschaad: Have you tested what happens to the test cases when that is changed to what seems like a fix?

phschaad (changes requested):
Generally LGTM, small question and a TODO can be changed to a note (see comment)

dace/subsets.py
philip-paul-mueller: As discussed I removed the call to `nng()` in `bounding_box_cover_exact()` and in `conver_precise()` made its application dependent on the value of `symbolic_positive`.
Let's see what happens.

phschaad (approved):
It's looking good for `nng` having indeed been redundant. LGTM in that case :-)

PR #2096, opened 2025-07-23
Avoided costly copy operation

philip-paul-mueller:
During a profiling it was noticed that a huge fraction of the time was spend in `__deepcopy__` of Memlets.
After some investigation it turned out that this was caused by copying SymPy expressions.

In SymPy symbols are [immutable](https://docs.sympy.org/latest/explanation/glossary.html#term-Immutable) the documentation makes this quite clear.
Thus copying such an object should be trivial, however, for some reason it is not (apparently SymPy does not define the `__deepcopy__()` function).

Thus this PR adds a `__deepcopy__()` method to the:
- `Memlet`
- `Range`
- `Indices`
- `DebugInfo`

This method only copies the members that are mutable but does not touch the immutable parts.
The effect is a speedup of about >10%, the side effect, however, is that `deepcopy()`, at least for some object, is no longer deep, however, due to the immutability, there is no behavioral change.

Furthermore, this PR fixes the `__add__()` method of the `Range`.

### DIFF
--- a/dace/dtypes.py
+++ b/dace/dtypes.py
@@ -1425,6 +1425,15 @@ class DebugInfo:
         return DebugInfo(json_obj['start_line'], json_obj['start_column'], json_obj['end_line'], json_obj['end_column'],
                          json_obj['filename'])
 
+    def __deepcopy__(self, memo) -> 'DebugInfo':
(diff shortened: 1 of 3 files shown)


phschaad (approved):
LGTM, thanks

PR #2097, opened 2025-07-23
Conditional Map Interchange

Berke-Ates:
This PR adds a new ConditionalMapInterchange transformation accompanied by tests. The transformation looks for maps inside conditional blocks such as:
```py
if cond:
  for i in map[0:10]:
    A[i] += 1
```

and interchanges the conditional block with the map:
```py
for i in map[0:10]:
  if cond:
    A[i] += 1
```

This is particularly useful if there is an outer map, which can then be fused with the inner map thanks to the interchange, increasing the number of parallel threads.

### DIFF
--- a/tests/transformations/interstate/condition_map_interchange_test.py
+++ b/tests/transformations/interstate/condition_map_interchange_test.py
@@ -232,10 +232,10 @@ def test_condition_map_interchange_data_dependency3():
 def test_condition_map_interchange_symbol_dependency():
 
     @dace.program
-    def tester(A: dace.float32[10], i_sym: dace.int32):
+    def tester(A: dace.float32[20], i_sym: dace.int32):
         if i_sym < 10:
             for i in dace.map[0:10]:
-                A[i_sym] = A[i_sym] + 10
+                A[i_sym + i] = A[i_sym + i] + 10
 
     _test_for_unchanged_behavior(tester, 1)
 


tests/transformations/interstate/condition_map_interchange_test.py
```diff
@@ -0,0 +1,208 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
+"""Tests loop overwrite elimination transformations."""
```
phschaad: Please adjust

dace/transformation/interstate/condition_map_interchange.py
```diff
@@ -0,0 +1,225 @@
...
+    If one or multiple maps are surrounded by a conditional block, moves the conditional block into the nested SDFG of maps.
+    """
+
+    cond_block = transformation.PatternNode(ConditionalBlock)
+
+    @classmethod
+    def expressions(cls):
+        return [sdutil.node_path_graph(cls.cond_block)]
+
+    def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
```
phschaad, line 29: Shouldn't there be an additional check to prevent interchanging in the case of a dependency preventing it?
Examples of a dependency I can think of would be:
- based on a symbol:
```Python
if i < 10:
  for i in map[0:10]:
    A[i] += 1
```
- or based on a data dependency:
```Python
if A[4] < 10:
  for i in map[0:10]:
    A[i] += 1
```

phschaad (changes requested):

dace/transformation/interstate/condition_map_interchange.py
Berke-Ates: Good points! Covered these cases by precomputing the condition in a new symbol, breaking the dependency.

dace/transformation/interstate/condition_map_interchange.py
```diff
@@ -0,0 +1,242 @@
...
+        return True
+
+    def apply(self, graph: ControlFlowRegion, sdfg: sd.SDFG):
+        branch: ControlFlowRegion = self.cond_block.branches[0][1]
+        branch_cond = self.cond_block.branches[0][0]
+        cond_syms = set(branch_cond.get_free_symbols())
+        all_states = list(branch.all_states())
+
+        # Prepend the condition computation
+        cond_sym = graph.sdfg.add_symbol("cond", dtypes.bool, find_new_name=True)
```
phschaad: Good that `find_new_name` is `True`, but please make the starting name a bit more specific to the conditional block it corresponds to. Maybe concat with `'_' + self.cond_block.label` or something.

phschaad (approved):
LGTM - feel free to merge after the comment is addressed

Issue #2086, opened 2025-07-16
DeadDataflowElimination fails to "inline" non-scalarized outputs


romanc:
**Describe the bug**

`DeadDataflowElimination` removes transient write access nodes that aren't read anymore. If the source is a `Tasklet`, that `Tasklet` needs to be changed. The write can be "inlined", i.e. scoped to the `Tasklet` (since it's never read after the Tasklet).

This "linining" or localization of the write fails for non-scalized memlets, as in the attached SDFG where the `Tasklet` writes into a 2x2 array. The issue presents itself as `simplify()` / DDE complaining that 'pointer' objects have no attribute 'typename'.

**To Reproduce**

1. Load this sdfg: [dde-pointer-inline.sdfg.zip](https://github.com/user-attachments/files/21258703/dde-pointer-inline.sdfg.zip)
2. Run `simplify()` on the loaded sdfg.
3. See error

**Expected behavior**

I'd expect DDE to either correctly scope the write to the `Tasklet` or not attempt to remove the write access node.

**Screenshots**

<img width="2617" height="1667" alt="Image" src="https://github.com/user-attachments/assets/bf718fe2-5d88-47b9-9c56-cd9289828e3a" />

**OS / version info**

- OS: Linux
- DaCe version: present in both, `main` (as of 25-07-16) and the `v1/maintenance` branch
- Python version: 3.11.9

**Additional context**

```none
>>> SDFG.from_file("./dde-pointer-inline.sdfg")
SDFG (TestTypedTemporary_dacecpu_3)
>>> sdfg = SDFG.from_file("./dde-pointer-inline.sdfg")
>>> sdfg.validate()
>>> sdfg.simplify()
Traceback (most recent call last):
  File "[...]/dace/dace/transformation/passes/dead_dataflow_elimination.py", line 164, in apply
    ast_find.generic_visit(code)
  File "[...]/python/3.11.9/lib/python3.11/ast.py", line 494, in generic_visit
    value = self.visit(value)
            ^^^^^^^^^^^^^^^^^
  File "[...]/python/3.11.9/lib/python3.11/ast.py", line 418, in visit
    return visitor(node)
           ^^^^^^^^^^^^^
  File "[...]/python/3.11.9/lib/python3.11/ast.py", line 503, in generic_visit
    new_node = self.visit(old_value)
               ^^^^^^^^^^^^^^^^^^^^^
  File "[...]/python/3.11.9/lib/python3.11/ast.py", line 418, in visit
[...]

PR #2098, opened 2025-07-23
Fix DDE applying incorrectly on pointer / size > 1 tasklet data

phschaad:
This addresses issue #2086

### DIFF
--- a/dace/transformation/passes/dead_dataflow_elimination.py
+++ b/dace/transformation/passes/dead_dataflow_elimination.py
@@ -262,6 +262,22 @@ class DeadDataflowElimination(ppl.ControlFlowRegionPass):
                     if _has_side_effects(l.src, sdfg):
                         return False
 
+                    # If data is connected to a tasklet through a pointer and more than 1 element is accessed,
+                    # we cannot eliminate the connector, as it may require dataflow analysis inside the tasklet.
+                    if isinstance(l.src, nodes.Tasklet):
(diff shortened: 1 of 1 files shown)


phschaad:
Those are valid concerns

> * The change potentially weakens DDE in the sense that, if all Tasklet outputs were deemed redundant, then the whole subgraph could just be removed. This change now disallows this (correct me if I am wrong).

Based on my understanding, that is correct, yes. DDE attempts to remove the entire subgraph in this situation. However, consider a situation where in addition to the `tmp` output in the linked issue, there is a second output to the tasklet (say, to some array `other_tmp`), which is not considered to be dead dataflow. In this situation, only the writes to `tmp` would be considered dead and DDE would only attempt to remove those, instead performing the mentioned 'inlining' inside the tasklet to remove the correct writes from the tasklet outputs. That in turn requires the mentioned dataflow analysis I am concerned about.

> * The change addresses a problem that I am unsure should exist in the first place. Is the Tasklet shown in #2086 valid? If, as you already suggested in the issue, there was one connector per scalar access, would DDE still fail?

If there were one connector per scalar access, I do not believe DDE would fail. The tasklet is "valid" just simply because we do not have concrete semantics forbidding it, rather just an "ideology" this goes against..

tbennun:
Unfortunately we are better safe than sorry on this one. Can we leave a note in the code though, such as a TODO(later)? @phschaad

Issue #2100, opened 2025-07-23
Nested SDFG connector creation / Inplace operations


almaeder:
**Describe the bug**
The input/output connectors for a nested SDFG for the python frontend are determined  in `dace/frontend/python/newast.py` `ProgramVisitor._visitname`

```python
if isinstance(node.ctx, ast.Store):
    rname, _ = self._add_write_access(rname, rng, node)
else:
    rname, _ = self._add_read_access(rname, rng, node)
```

Thus for example for: 
```python
c.fill(0.0)
```
It is determined through ast as a read access. This then lead to:

```python
dace.sdfg.validation.InvalidSDFGNodeError: Data descriptor __tmp_30_16_r is written to, but only given to nested SDFG as an input connector 
```

**To Reproduce**
The behavior can be reproduced by running the following code:

```python
import dace
import torch
from dace.transformation.interstate import GPUTransformSDFG

N = 32

c_desc = dace.data.Array(
    dace.float32,
    [N, N],
    storage=dace.StorageType.GPU_Global,
    strides=[N, 1],
)

@dace.program #(recreate_sdfg=False)
def global_matmul(
    C: c_desc @ dace.StorageType.GPU_Global
):
    for i, j in dace.map[0:N:N,0:N:N] @ dace.ScheduleType.GPU_Device:

        for l in dace.map[0:64] @ dace.ScheduleType.GPU_ThreadBlock:

            c = dace.ndarray(
                [N, N],
                dtype=dace.float32,
                storage=dace.StorageType.Register,
                strides=(N, 1),
            )

            for k in dace.map[0:1] @ dace.ScheduleType.Sequential:
                c.fill(0.0)

            C[i:i+N, j:j+N] = c[:, :]


C = torch.rand((32,32), dtype=torch.float32, device="cuda")

sdfg = global_matmul.to_sdfg(simplify=False)

sdfg.apply_transformations(
    GPUTransformSDFG, options=dict(sequential_innermaps=False, register_trans=False, simplify=False)
)

sdfg(C=C)
```

**Desktop (please complete the following information):**
 - OS: Linux
 - DaCe: edba15443b6d7ef48eea00ec8b5a29bce113369b

almaeder:
It is the existing one.

tbennun:
The issue has been found and reproduced (`fill` does return the original array as an output, but that indeed is not registering the contents as outputs). It will be fixed; in the meantime, please use `c[:] = 0.0` instead of fill as a workaround.

[...]

PR #2101, opened 2025-07-23
Properly register input/output data containers for nested programs

tbennun:
The issue manifests in methods that mutate the original data container, such as `ndarray.fill`. 

Fixes Issue #2100

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -4741,7 +4741,14 @@ class ProgramVisitor(ExtNodeVisitor):
             result = result[1]
 
         if not isinstance(result, (tuple, list)):
-            return [result]
+            result = [result]
+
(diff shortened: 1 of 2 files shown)


alexnick83 (approved):

dace/frontend/python/newast.py
```diff
@@ -4741,7 +4741,14 @@ def visit_Call(self, node: ast.Call, create_callbacks=False):
             result = result[1]
 
         if not isinstance(result, (tuple, list)):
-            return [result]
+            result = [result]
+
+        # Register input/output data containers
+        for res in result:
```
alexnick83, line 4747: It is a great idea. There seems to be an issue with the dict keys containing (mutable) slices. I will have a look into that.

Issue #2094, opened 2025-07-22
Memlet subset out-of-bounds


almaeder:
**Describe the bug**
Incorrect temporary memlet sizes within a range in a GPU map.
Thus, the validation fails with "Memlet subset out-of-bounds"/

**Message**
```
"dace/dace/sdfg/validation.py", line 824, in validate_state
    raise InvalidSDFGEdgeError("Memlet subset out-of-bounds", sdfg, state_id, eid)
dace.sdfg.validation.InvalidSDFGEdgeError: <exception str() failed>
```

**To Reproduce**
The following code results in the error:

```
import dace
import torch
from dace.transformation.interstate import GPUTransformSDFG

def _get_strides_squared(size):
    return (1, size), (size, 1), (size, 1)

N = 32
n = 32

a_desc = dace.data.Array(
    dace.float32,
    [N, N],
    storage=dace.StorageType.GPU_Global,
    strides=[N, 1],
)
b_desc = dace.data.Array(
    dace.float32,
    [N, N],
    storage=dace.StorageType.GPU_Global,
    strides=[N, 1],
)
c_desc = dace.data.Array(
    dace.float32,
    [N, N],
    storage=dace.StorageType.GPU_Global,
    strides=[N, 1],
)


op_size = 32
a_storage, b_storage, c_storage = dace.StorageType.GPU_Shared, dace.StorageType.GPU_Shared, dace.StorageType.GPU_Shared
a_strides, b_strides, c_strides = _get_strides_squared(op_size)

@dace.program
def global_matmul(
    A: a_desc @ dace.StorageType.GPU_Global,
    B: b_desc @ dace.StorageType.GPU_Global,
    C: c_desc @ dace.StorageType.GPU_Global,
):
    for i, j in dace.map[0:N:op_size,0:N:op_size] @ dace.ScheduleType.GPU_Device:

        for l in dace.map[0:64] @ dace.ScheduleType.GPU_ThreadBlock:

            c = dace.ndarray(
                [op_size, op_size],
                dtype=dace.float32,
                storage=c_storage,
                strides=c_strides,
            )
            c.fill(0.0)

            for k in range(1):
                a = dace.ndarray(
                    [32, 32],
                    dtype=dace.float32,
                    storage=a_storage,
                    strides=a_strides,
                )
                b = dace.ndarray(
                    [32, 32],
                    dtype=dace.float32,
                    storage=b_storage,
                    strides=b_strides,
                )

[...]

PR #2102, opened 2025-07-24
Fix for #2094

alexnick83:
Shape inference of a symbol-dependent squeezed nested access is wrong for dimensions where the index is a range. For an example, see #2094


alexnick83:
The main issue here is that the frontend tries to do too many things, starting from adhering to the (obsolete?) principle of squeezing nested accesses. Symbol-dependent accesses further complicate the required logic and, ultimately, the maintainable solution is to remove squeezing altogether. However, this needs an almost full rewrite, and this PR attempts to patch bugs to resolve issues in the short term.

tbennun:
@alexnick83 
We can see how this works in #1696, where squeezing is already disabled. In the meantime, if you can please add the test of #2094 I'll approve it.

PR #2104, opened 2025-07-24
Expand serde.f90 exponent output.

Mittagskogel:
Fortran only uses two digits for the exponent in scientific notation, by default? If the exponent has three digits, Fortran removes the E to get more space: 0.14099200558255298680-270.

Serde parses this as two numbers, resulting in headaches.

This fix sets the exponent field to four digits to avoid this problem.


acalotoiu:
can you please add a test to showcase the behavior?

PR #2106, opened 2025-07-25
Fixed `Range.__add__()`

philip-paul-mueller:
The function was ignoring the `tiling_sizes` property of the `Range` object.
Also added unit tests.

Originally this fix was part of [PR#2096](https://github.com/spcl/dace/pull/2096) but it was wrong, so it was moved to its own PR.


philip-paul-mueller:
@phschaad yes it is ready.
In my initial version I wrote `self.tiling_sizes + self.tiling_sizes` instead of `self.tiling_sizes + other.tiling_sizes`.
The net effect was that all unit tests in which 1D ranges were concatenated passed.

phschaad (approved):
LGTM, thanks!

PR #2112, opened 2025-07-29
Fix some crashing loop in stream sync function

ThrudPrimrose:


ThrudPrimrose:
This while loop does not update its condition and will result in an infinite loop, but luckily, it constantly crashes. I added proper handling to prevent the program from crashing. This loop is triggered when I want to add "profiling" tasklets to maps to measure entry and exit times. (It was easier instrument SDFG with tasklets in Fortran branch, but I want to backport the small codegen fixes I do there)

PR #2116, opened 2025-07-30
Add threadblock map transformation

aydogdub:
**Description**

I implemented a transformation which adds an explicit `GPU_ThreadBlock`-scheduled map to a `GPU_Device`-scheduled map if
it does not already have one (in certain cases). Note that thread blocks are always present and used in CUDA kernels. Making 
them explicit in the SDFG is good practice and promotes modularity.

I implemented this transformation as part of my Master’s thesis work, aiming to closely replicate the previous behavior to ensure 
backwards compatibility. Since only minimal changes were needed to make it useful in combination with `CUDACodeGen`  as well,
I am submitting it as a pull request.

**Example**

Below is an example illustrating the effect of the transformation on a simple SDFG. The transformation adds an explicit `GPU_ThreadBlock`-scheduled map with default block sizes obtained from the configuration.

grafik

grafik

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -26,7 +26,7 @@ from dace.frontend.python.astutils import rname
 from dace.frontend.python import nested_call, replacements, preprocessing
 from dace.frontend.python.memlet_parser import DaceSyntaxError, parse_memlet, ParseMemlet, inner_eval_ast, MemletExpr
 from dace.sdfg import nodes
-from dace.sdfg.propagation import propagate_memlet, propagate_subset, propagate_states
+from dace.sdfg.propagation import propagate_memlet, propagate_subset, propagate_states, align_memlet
 from dace.memlet import Memlet
 from dace.properties import LambdaProperty, CodeBlock
 from dace.sdfg import SDFG, SDFGState
@@ -2937,7 +2937,12 @@ class ProgramVisitor(ExtNodeVisitor):
                     memlet.other_subset = op_subset
                     if op:
                         memlet.wcr = LambdaProperty.from_string('lambda x, y: x {} y'.format(op))
-                    state.add_nedge(op1, op2, memlet)
+                    if isinstance(self.sdfg.arrays[target_name], data.Reference):
+                        e = state.add_edge(op1, None, op2, 'set', memlet)
+                        # Align memlet to referenced array
+                        e.data = align_memlet(state, e, dst=False)
+                    else:
+                        state.add_nedge(op1, op2, memlet)
             else:
                 memlet = Memlet("{a}[{s}]".format(a=target_name,
                                                   s=','.join(['__i%d' % i for i in range(len(target_subset))])))
@@ -3444,9 +3449,10 @@ class ProgramVisitor(ExtNodeVisitor):
             storage = dtypes.StorageType.Default
             type_name = rname(node.annotation)
             warnings.warn('typeclass {} is not supported'.format(type_name))
-        if node.value is None and dtype is not None:  # Annotating type without assignment
+        if dtype is not None:
             self.annotated_types[rname(node.target)] = dtype
-            return
+            if node.value is None:  # Annotating type without assignment
+                return
         results = self._visit_assign(node, node.target, None, dtype=dtype)
         if storage != dtypes.StorageType.Default:
             self.sdfg.arrays[results[0][0]].storage = storage
@@ -3629,6 +3635,12 @@ class ProgramVisitor(ExtNodeVisitor):
                     result_data = self.sdfg.arrays[result]
                     if (name.startswith('__return') and isinstance(result_data, data.Scalar)):
                         true_name, new_data = self.add_temp_transient([1], result_data.dtype)
+                        self.variables[name] = true_name
+                        defined_vars[name] = true_name
+                    elif name in self.annotated_types and isinstance(self.annotated_types[name], data.Reference):
+                        desc = copy.deepcopy(self.annotated_types[name])
+                        desc.transient = True
+                        true_name = self.sdfg.add_datadesc(name, desc, find_new_name=True)
                         self.variables[name] = true_name
                         defined_vars[name] = true_name
                     elif (not name.startswith('__return')
--- a/dace/subsets.py
+++ b/dace/subsets.py
@@ -938,15 +938,20 @@ class Indices(Subset):
         N-dimensional data descriptor. """
 
     def __init__(self, indices):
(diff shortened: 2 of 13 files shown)


dace/codegen/targets/cuda.py
```diff
@@ -153,6 +158,18 @@ def preprocess(self, sdfg: SDFG) -> None:
                                       'CUDA',
                                       target_type=target_type)
 
+        # Identify kernels with inserted GPU_ThreadBlock-scheduled maps
```
alexnick83, line 164: It seems a bit wasteful to do multiple passes over all nodes recursively. I think this is fine for the initial implementation, but I would leave a note here to look into developing a custom pass. Such a pass could, for example, ignore nested contexts when already inside a GPU-scheduled map, and directly output the set of GPU-device maps that had a TB map added.

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,285 @@
+# Copyright 2019-2023 ETH Zurich and the DaCe authors. All rights reserved.
```
alexnick83: Update year

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,285 @@
...
+""" This module contains classes and functions that implement the grid-strided map tiling
```
alexnick83: Update docstring

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,285 @@
...
+        Determines whether the transformation can be applied to a map entry.
+
+        The transformation applies only to GPU_Device-scheduled maps ("kernel maps") that:
+        1. Are not already nested within any GPU-scheduled maps,
+        2. Do not themselves contain any inner GPU-scheduled maps.
+        """
+        map_entry = self.map_entry
+
+        # Find the state that contains the map entry
+        state = next(state for node, state in sdfg.all_nodes_recursive() if node == map_entry)
```
alexnick83: Seems a bit wasteful. Would it make sense to follow the `{graph, graph.parent_graph, ...}` chain until you find a state?

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,285 @@
...
+        if map_entry.map.schedule != dtypes.ScheduleType.GPU_Device:
+            return False
+
+        # Check if any inner maps are GPU-scheduled (e.g., GPU_ThreadBlock)
+        for _, inner_entry in helpers.get_internal_scopes(graph, map_entry):
+            if inner_entry.map.schedule in dtypes.GPU_SCHEDULES:
+                return False  # Already has GPU-scheduled inner scope — does not apply
+
+        # Check if the map is nested inside another GPU-scheduled map
+        parent_map_tuple = helpers.get_parent_map(state, map_entry)
```
alexnick83, line 133: I do not think that we support GPU-device map inside any other GPU-scheduled map, but I amy not be up-to-date with developments. However, in such a case, you can omit the following check.

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,285 @@
...
+            if parent_map.map.schedule in dtypes.GPU_SCHEDULES:
+                return False  # Nested inside a GPU scope — does not apply
+
+            parent_map_tuple = helpers.get_parent_map(parent_state, parent_map)
+
+        return True
+
+    def apply(self, state: SDFGState, sdfg: SDFG):
+        """
+        Ensures that `self.map_entry`, a `GPU_Device`-scheduled map, is explicitly nested
```
alexnick83: This is a bit confusing because a GPU thread block-map should be nested inside a GPU device-map. I understand where you are coming from based on the next paragraph in the docstring, which is fully correct, but I suggest you rephrase this sentence.

dace/transformation/dataflow/add_threadblock_map.py
```diff
@@ -0,0 +1,285 @@
...
+        # After tiling: kernel_map_entry is now the thread block map, configure its schedule
+        thread_block_map_entry = kernel_map_entry
+        thread_block_map_entry.map.schedule = dtypes.ScheduleType.GPU_ThreadBlock
+
+        # Set the new kernel_entry's gpu_block_size attribute
+        new_kernel_entry, *_ = helpers.get_parent_map(state, kernel_map_entry)
+        new_kernel_entry.map.gpu_block_size = gpu_block_size
+
+        # Catch any unexpected mismatches of inserted threadblock map's block size and the used block size
+        tb_size = to_3d_dims([symbolic.overapproximate(sz) for sz in thread_block_map_entry.map.range.size()[::-1]])
```
alexnick83: Is this a sanity/debug check, or can this actually fail? In the latter case, would it make sense to move this check to the `can_be_applied` method and just not apply the transformation?

dace/transformation/dataflow/add_threadblock_map.py
aydogdub: Done

dace/transformation/dataflow/add_threadblock_map.py
aydogdub: Done

dace/codegen/targets/cuda.py
aydogdub: I added a corresponding note

dace/transformation/dataflow/add_threadblock_map.py
aydogdub: Actually, nested GPU_Device maps are supported and are (rarely) used (See `tests/codegen/warp_specialization_test.py`)

dace/transformation/dataflow/add_threadblock_map.py
aydogdub: I agree. Is it a bit more understandable now?

dace/transformation/dataflow/add_threadblock_map.py
aydogdub: This was only a sanity check for me, it did never fail. I have removed it.

dace/transformation/dataflow/add_threadblock_map.py
aydogdub: Yes this is wasteful. It can be even solved simpler. I overlooked that the input `graph` is actually the state containing `map_entry` . I have corrected this and updated the function signature with type annotations (and renamed `graph` to `state`). Thank you!

PR #2118, opened 2025-08-01
Fix: Memlets in StripMining transformation

aydogdub:
**Bug Description:**

To put it simple, the `StripMining` transformation essentially performs tiling- that is, it adds a  **new outer** map which encloses
an existing one, where the new map runs over chunks (=tiles) of the original iteration space, while the inner map handles 
individual elements inside each chunk. The inner map is the original map, where the range just go adapted.

Now, the new (outer) map needs to be connected to the original one, and the dataflow between them must be captured correctly. 
This essentially involves analyzing which parts of a data descriptor should flow for each chunk. For the sake of simplicity, one 
can think of this process as: copying an existing memlet (flowing to and from the original inner map) and update it to create
the correct memlets between the two maps. The screenshots below help visualize this - you can clearly see the similarity 
between the two highlighted memlets.

However, this copying step is problematic when an `other_subset` is present - in such cases, it should be explicitly set to `None` for
the newly created memlet. Otherwise there can be a dimensionality mismatch between the two subsets leading to an invalid SDFG.

**Fix Description**

Set the `other_subset` to `None` for memlets created between the two maps. Additionally, a test case has been added which now passes and would have failed before this fix.

**Failing Example**

After the fix, ```[0] -> A[block_i : Min(63, block_i + 31) + 1]``` in the right image will become  
```A[block_i : Min(63, block_i + 31) + 1]```.

<p float="left">
  
  
</p>

### DIFF
--- a/tests/transformations/strip_miningy_test.py
+++ b/tests/transformations/strip_miningy_test.py
@@ -3,6 +3,7 @@ from dace.transformation.dataflow.strip_mining import StripMining
 
 import numpy as np
 
+
 def test_strip_mining():
     """
     Test a simple example example where Stripmining works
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tests/transformations/strip_miningy_test.py
Comment (tbennun) [inline tests/transformations/strip_miningy_test.py:1]: Rename to strip_mining_test.py

File: tests/transformations/strip_miningy_test.py [hunk 1]
```diff
@@ -0,0 +1,54 @@
+import dace
```
tbennun: Copyright statement

tbennun (changes requested):

tests/transformations/strip_miningy_test.py
aydogdub: Yes, small typo. Fixed it. Thank you!

tests/transformations/strip_miningy_test.py
aydogdub: I added:
"Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved."
on the top of the file.

Thank you!

tbennun (approved):
LGTM, thanks!

PR #2121, opened 2025-08-04
Python frontend: Better transient naming heuristics

tbennun:
Avoids adding `__tmp*` variables in the Python frontend. New API calls (`ProgramVisitor.get_target_name` and `ProgramVisitor.add_temp_transient`) are added to help replacements find better names for variables.

Also fixes discovered bugs in code generation and argument names / validation.

The heuristics are as follows:
1. If the expression is a direct descendant of an assignment (e.g., `a = f(x) + 1`), uses the appropriate left-hand side target of the assignment (or the matching one for tuples)
2. If the AST node is a sub-expression of a right-hand side expression, use the operators and function names (`a_plus_b`, `f_x_y_z`)
3. Otherwise, tries to find appropriate names from the AST node type, or uses `__tmp*` if all else fails.

This improves the readability of programs given to the Python/NumPy frontend.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -110,6 +110,24 @@ augassign_ops = {
     'BitAnd': '&'
 }
 
+# Mappings for determining variable name based on operator
+_UNOP_TO_NAME = {ast.UAdd: 'pos', ast.USub: 'neg', ast.Not: 'not', ast.Invert: 'inv'}
+_BINOP_TO_NAME = {
+    ast.Add: 'plus',
+    ast.Sub: 'minus',
+    ast.Mult: 'times',
+    ast.Div: 'div',
+    ast.FloorDiv: 'floordiv',
+    ast.Mod: 'mod',
+    ast.Pow: 'pow',
+    ast.LShift: 'lshift',
+    ast.RShift: 'rshift',
+    ast.BitOr: 'bitor',
+    ast.BitXor: 'bitxor',
+    ast.BitAnd: 'bitand',
+    ast.MatMult: 'matmul',
+}
+
 
 class AddTransientMethods(object):
     """ A management singleton for methods that add transient data to SDFGs. """
@@ -825,7 +843,6 @@ class TaskletTransformer(ExtNodeTransformer):
 
         if arr_type is None:
             arr_type = type(parent_array)
-            var_name = self.sdfg._find_new_name(var_name)
         if arr_type == data.Scalar:
             var_name, _ = self.sdfg.add_scalar(var_name, dtype, find_new_name=True)
         elif issubclass(arr_type, data.Array):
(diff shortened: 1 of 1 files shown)


dace/frontend/python/replacements/operators.py
```diff
@@ -45,41 +45,23 @@ def _unop(sdfg: SDFG, state: SDFGState, op1: str, opcode: str, opname: str):
     return name
 
 
-def _binop(sdfg: SDFG, state: SDFGState, op1: str, op2: str, opcode: str, opname: str, restype: dtypes.typeclass):
```
tbennun, line 48: unused

.github/workflows/fpga-ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         # and overflowed in the year 2022, run the FPGA tests pretending like it's January 1st 2021.
         # faketime -f "@2021-01-01 00:00:00" pytest -n auto --cov-report=xml --cov=dace --tb=short -m "fpga"
         # Try running without faketime
-        pytest -n auto --cov-report=xml --cov=dace --tb=short --timeout=300 -m "fpga"
+        pytest -n auto --cov-report=xml --cov=dace --tb=short --timeout_method thread --timeout=300 -m "fpga"
```
tbennun, line 64: Needed to detect hangs

dace/frontend/python/newast.py
```diff
@@ -824,12 +825,13 @@ def _add_access(
 
         if arr_type is None:
             arr_type = type(parent_array)
+            var_name = self.sdfg._find_new_name(var_name)
```
alexnick83: Since all the options below use the `find_new_name` flag, is this needed?

dace/frontend/python/newast.py
```diff
@@ -1356,6 +1360,155 @@ def defined(self):
...
+                return candidate
+
+        # If the current AST node is a variable name, return the name of the variable.
+        if isinstance(current_ast_node, ast.Name):
+            return current_ast_node.id
+
+        # If the current AST node is an expression (e.g., a binary operation),
+        # return the operands concatenated with the operator, e.g., "not_a", "a_times_b"
+        if isinstance(current_ast_node, ast.UnaryOp):
+            op_map = {ast.UAdd: 'pos', ast.USub: 'neg', ast.Not: 'not', ast.Invert: 'inv'}
```
alexnick83: Would it make sense to move those dictionaries to the global namespace of the submodule, so they are generated only once, when importing the submodule, similarly to op replacements?

dace/frontend/python/newast.py
```diff
@@ -1356,6 +1360,155 @@ def defined(self):
...
+            if arg_names:
+                result = f"{func_name}_{'_'.join(arg_names)}"
+            else:
+                result = func_name
+
+            if dtypes.validate_name(result):
+                return result
+
+        # If all else fails, try to see if unparsing yields a valid name.
+        if dtypes.validate_name(astutils.unparse(current_ast_node)):
```
alexnick83: Probably minor, but can avoid calling unparse twice.

dace/frontend/python/newast.py
tbennun: Good catch! I think I read the rest of the branches as `elif`s.

PR #2123, opened 2025-08-07
Add CopyND support to copy structs / data types that are not trivially copyable

ThrudPrimrose:


tests/codegen/non_trivally_copyable_copy.py
```diff
@@ -0,0 +1,194 @@
...
+
+    for s in [s0, s1, s2]:
+        for n in s.nodes():
+            if s.degree(n) == 0:
+                s.remove_node(n)
+
+    return sdfg
+
+
+def test_non_trivial_copy_compilation():
```
alexnick83, line 173: You could make a proper test with execution here by using ctypes to construct a struct. However, you should also use a struct in the SDFG too (instead of an opaque type), and maybe it is better to handle this in another PR.

alexnick83 (approved):
LGTM

tests/codegen/non_trivally_copyable_copy.py
ThrudPrimrose: I will address this in another PR once I have time (I am now crunching to get all fixes and features to main ASAP to use for SoftHier so this might need to wait a bit).

PR #2127, opened 2025-08-15
Fixed MapFusion with NestedSDFG

philip-paul-mueller:


philip-paul-mueller:
To better understand the issue consider the following scenario:

unfused

In the picture above the slice `[i_Vertex, 0:5, i_K]` of `__tmp1` is mapped into the nested SDFG.
On the inside this slice has the name `values` and is a 1d array of length 5.
For the sake of argument let's assume that `__tmp1` has the strides `(x, y, z)` and a shape of `(N, 5, M)`, then the stride of `values` must be `y`.

If we now merge the two Maps together, and by that eliminating `__tmp1` we get to the following situation:

fused

`__tmp1` has been replaced by `__s0_n7OUT__tmp1_n8None` who has a shape of `(5,)` and a strides of `(1,)`.
If we ignore the nested SDFG `values` would still use a strides of `(y,)` instead of `(1,)` thus Map fusion has to modify it.

tests/transformations/map_fusion_vertical_test.py
```diff
@@ -2282,3 +2742,4 @@ def test_map_fusion_consolidate_consume_not_same_range_default():
     test_map_fusion_consolidate_consume_same_range_not_allowed()
     test_map_fusion_consolidate_consume_same_range_if_not_extending()
     test_map_fusion_consolidate_consume_not_same_range_if_not_extending()
+    #test_map_fusion_nested_sdfg_slicing()
```
phschaad: Left over from debugging?

phschaad (approved):
LGTM, but please check the comment I left / fix it - thanks!

tests/transformations/map_fusion_vertical_test.py
philip-paul-mueller: I would say yes.
Thanks for the review.

PR #2128, opened 2025-08-15
used_symbols() call does not match argument list

aydogdub:
**Description**

The `used_symbols(...)` method is called incorrectly in the `used_symbols_within_scope(...)` function.  
From the `nodes.py` file, we can see that only the `NestedSDFG` class implements this method. However, it is currently invoked with two arguments instead of the single argument it expects. So we need to call it for `NestedSDFG` instances only with
one argument.

In addition, the `hasattr(n, "used_symbols")` check can be removed and replaced with a direct `isinstance` check for `NestedSDFG`, extending the preceding type-based logic for updating the `free_symbols` set.

### DIFF
--- a/dace/sdfg/nodes.py
+++ b/dace/sdfg/nodes.py
@@ -912,10 +912,10 @@ class MapEntry(EntryNode):
                     )
                     free_symbols |= code_symbols
                     continue
-            elif isinstance(n, dace.nodes.NestedSDFG):
+            elif isinstance(n, NestedSDFG):
                 free_symbols |= n.used_symbols(all_symbols)
-            else:
-                free_symbols |= n.free_symbols
+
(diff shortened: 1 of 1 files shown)


dace/sdfg/nodes.py
```diff
@@ -912,11 +912,10 @@ def used_symbols_within_scope(self, parent_state: 'dace.SDFGState', all_symbols:
                     )
                     free_symbols |= code_symbols
                     continue
+            elif isinstance(n, NestedSDFG):
```
ThrudPrimrose, line 915: CFGs, Data Descriptors, and NestedSDFGs can have the function `used_symbols` defined. And since we check the nodes between map entry and exit, it needs to be a nested SDFG. This seems to be a correct fix for it.

PR #2131, opened 2025-08-21
Broadcast Hoisting

Berke-Ates:
This PR adds a new transformation, which pushes a broadcast operation up (i.e. earlier). It does so by extending the dimensions of the array being broadcasted to turn the broadcast into a one-to-one opration. 
Example:
  ```py
  for n in range(...):
    x[n] = y[n] # one to one

  for n in range(...):
    for k in range(...):
      z[n, k] = x[n] # broadcast
  ```

  becomes:
  ```py
  for n in range(...):
    for k in range(...):
      x[n, k] = y[n] # broadcast

  for n in range(...):
    for k in range(...):
      z[n, k] = x[n, k] # one to one
  ```

### DIFF
--- a/dace/transformation/dataflow/__init__.py
+++ b/dace/transformation/dataflow/__init__.py
@@ -31,6 +31,7 @@ from .local_storage import InLocalStorage, OutLocalStorage
 from .double_buffering import DoubleBuffering
 from .streaming_memory import StreamingMemory, StreamingComposition
 from .reduce_expansion import ReduceExpansion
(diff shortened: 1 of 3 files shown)


tbennun:
I like the notion of the transformation, but it feels specialized. Should we pilot this for the new dace-transformations repo? @acalotoiu

Berke-Ates:
@acalotoiu Any updates on the status of the transformation repo?

PR #2132, opened 2025-08-24
Allocate arrays in scope after map variables are defined

aydogdub:
**Description**

In `CUDACodeGen`, arrays in a scope should be allocated **after** defining the map variables.
Currently, the allocation happens earlier, which can (and does) lead to code referencing undefined variables.

For example, consider the following part of an SDFG:

Screenshot 2025-08-24 195204

Which will generate:

```cpp
B_0 = &gpu_B[((N * (i + 1)) + j)];
```
**before** `j`is defined, resulting in a compiler error.

### DIFF
--- a/dace/codegen/targets/cuda.py
+++ b/dace/codegen/targets/cuda.py
@@ -2354,7 +2354,7 @@ gpuError_t __err = {backend}LaunchKernel((void*){kname}, dim3({gdims}), dim3({bd
                 callsite_stream.write(
                     f'auto {scope_map.params[0]} = {scope_map.range[0][0]} + {dynmap_step} * {dynmap_var};', cfg,
                     state_id, scope_entry)
-                
+
             # Emit internal array allocation (deallocation handled at MapExit)
             self._frame.allocate_arrays_in_scope(sdfg, cfg, scope_entry, function_stream, callsite_stream)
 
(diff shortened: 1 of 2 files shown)


ThrudPrimrose:
Can you add a unit-test where previous allocation strategy would fail and the new one does not?

ThrudPrimrose:
This is also a bigger question, AFAIK, this SDFG should be invalid. I remember the discussion that it is now allowed to have `MapEntry (OUT_*) -> (views) ViewNode` and the view nodes src needs to be an access node? @phschaad can you verify this?

aydogdub:
So I upload the sdfg in the screenshot (as a .txt since github does not allow to upload .sdfgz files) which Yakup requested.

[example.sdfgz.txt](https://github.com/user-attachments/files/21988293/example.sdfgz.txt)

phschaad:
> This is also a bigger question, AFAIK, this SDFG should be invalid. I remember the discussion that it is now allowed to have `MapEntry (OUT_*) -> (views) ViewNode` and the view nodes src needs to be an access node? @phschaad can you verify this?

That is correct, yes. An view `AccessNode`'s `views` connector must be directly connected to another `AccessNode`

phschaad (changes requested):
Same comment as @ThrudPrimrose - This PR should have a unit test to expose the issue and demonstrate the fix.

alexnick83 (approved):
LGTM

PR #2136, opened 2025-08-30
fix: delete big arrays that are moved to heap

romanc:
## Description

When arrays have storage type Register, but are bigger than `max_stack_array_size`, they are moved to the heap. In that situation they are dynamically allocated and need to be deleted again.

In the test, I removed `redundant_array_crashes_codegen_test_original_graph()` which is a setup function of a test that got removed in https://github.com/spcl/dace/pull/2105.

I wasn't sure where to put the test. `codegen/cpp_test.py` seemed logical. The new test is also very close to `tests/vla_test.py` (which tests the same but for variable length arrays). Since that test was explicitly for variable length arrays, I didn't wanna put it there. In any case, I'm happy to move the new test around if you want it somewhere else.

Backport on `v1/maintenace` is in PR https://github.com/spcl/dace/pull/2135.

### DIFF
--- a/dace/codegen/targets/cpu.py
+++ b/dace/codegen/targets/cpu.py
@@ -488,10 +488,10 @@ class CPUCodeGen(TargetCodeGenerator):
 
                 if symbolic.issymbolic(arrsize, sdfg.constants):
                     warnings.warn('Variable-length array %s with size %s '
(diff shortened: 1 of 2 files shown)


phschaad (approved):
Thank you!

PR #2138, opened 2025-09-04
Lift trivial Ifs

ThrudPrimrose:
Dead control flow elimination does not simplify by trying to evaluate trivial if conditions. I wrote a pass that does that:

It tries to eval if branches that are trivially true and then lifts the inner CFG upwards:
image
Becomes:
image

It supports branches of type:
```
if (cond)
```

and (empty else):
```
if(cond) 
...
else:
....
```

Example:
image
Becomes:
image


acalotoiu (approved):
LGTM!

dace/transformation/passes/lift_trivially_true_if.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def _trivial_cond_check(self, code: CodeBlock, val: bool):
+        if code.language != dace.dtypes.Language.Python:
+            return False
+        try:
+            node = ast.parse(code.as_string, mode='eval')
+            result = eval(compile(node, '<string>', 'eval'))
+            return bool(result) is val
+        except Exception as e:
+            return False
```
tbennun: I'm not sure about this. We already have `symbolic.evaluate` and `astutils.evalnode`, you should use one of those instead.

tbennun (changes requested):
minor change

dace/transformation/passes/lift_trivially_true_if.py
ThrudPrimrose: I have switched to `symbolic.evaluate`.

dace/transformation/passes/lift_trivial_if.py
```diff
@@ -26,14 +27,56 @@ def should_reapply(self, modified: ppl.Modifies) -> bool:
...
 
+    def _make_unique_names(self, sdfg: dace.SDFG):
+        all_blocks = {
+            n
+            for n, _ in sdfg.all_nodes_recursive()
+            if isinstance(n, dace.SDFGState) or isinstance(n, ControlFlowRegion) or isinstance(n, ControlFlowBlock)
+        }
+        all_labels = set()
+
+        def _find_new_name(cfg: ControlFlowRegion) -> str:
```
tbennun: Why not use data.find_new_name(cfg_label, all_labels)?

dace/transformation/passes/lift_trivial_if.py
```diff
@@ -26,14 +27,56 @@ def should_reapply(self, modified: ppl.Modifies) -> bool:
...
+                    "and": " * ",
+                    "or": "+",
+                    "False": "0"
+                }))
+            symbolic_expr = symbolic_expr.simplify()
+            pystring = pycode(symbolic_expr)
+            result = symbolic.evaluate(expr=dace.symbolic.SymExpr(pystring), symbols=dict())
             return bool(result) is val
         except Exception as e:
+            print(e)
```
tbennun: Do you need the printout?

dace/transformation/passes/lift_trivial_if.py
```diff
@@ -78,16 +126,32 @@ def _detect_trivial_ifs_and_rm_cfg(self, graph: Union[ControlFlowRegion, SDFG],
...
+        sdutil.set_nested_sdfg_parent_references(graph.sdfg)
+        graph.sdfg.reset_cfg_list()
+
+        return rmed_count
+
+    def _detect_trivial_ifs_and_rm_cfg(self, graph: Union[ControlFlowRegion, SDFG]):
         # We might now have trivial control flow blocks at top level, apply in fixpoint
+        rmed_count = self._detect_and_remove_top_level_trivial_ifs(graph)
         local_rmed_count = rmed_count
+        graph.sdfg.validate()
```
tbennun: Do we need this here? Validation happens after transformation by default

dace/transformation/passes/lift_trivial_if.py
```diff
@@ -78,16 +126,32 @@ def _detect_trivial_ifs_and_rm_cfg(self, graph: Union[ControlFlowRegion, SDFG],
...
         while local_rmed_count > 0:
-            local_rmed_count = self._detect_trivial_ifs_and_rm_cfg(graph, depth)
+            local_rmed_count = self._detect_and_remove_top_level_trivial_ifs(graph)
             rmed_count += local_rmed_count
 
-        # Now go one one more level recursive
-        for node in graph.nodes():
-            if isinstance(node, ControlFlowRegion):
-                rmed_count += self._detect_trivial_ifs_and_rm_cfg(node, depth + 1)
+        graph.sdfg.validate()
```
tbennun: Same

dace/transformation/passes/lift_trivial_if.py
```diff
@@ -26,14 +27,56 @@ def should_reapply(self, modified: ppl.Modifies) -> bool:
...
+            def _token_replace_dict(string_to_check: str, dict) -> str:
+                # Split while keeping delimiters
+                tokens = re.split(r'(\s+|[()\[\]])', string_to_check)
+
+                # Replace tokens that exactly match src
+                tokens = [dict[token.strip()] if token.strip() in dict else token.strip() for token in tokens]
+
+                return " ".join(tokens).strip()
+
+            symbolic_expr = dace.symbolic.SymExpr(
```
tbennun: Why use symexpr (aka the exact/approx expression) here rather than calling symbolic.pystr_to_symbolic? It also handles and/or AFAIR

tbennun (changes requested):
More comments, mostly about reusing existing infrastructure

dace/transformation/passes/lift_trivial_if.py
ThrudPrimrose: Fortran SDFGs have expressions such as `(a == 1) == 0`. Which makes sympy crash as due to comparing a comparison object `Eq` with an integer. And I use the token version as a fallback when this crash happens

ThrudPrimrose: That was the original motivation

ThrudPrimrose: Now, having a pass to replace those integer comparisons with proper booleans seem good to me, still I think I should have some fallback code just in case.

dace/transformation/passes/lift_trivial_if.py
ThrudPrimrose: I guess no

dace/transformation/passes/lift_trivial_if.py
ThrudPrimrose: I didn't know it was default. I usually sprinkle validate calls to confirm correctness during the development and then I just don't remove them again as they wouldn't hurt. Performance wise they are completely unnecessary So, I will remove them

dace/transformation/passes/lift_trivial_if.py
ThrudPrimrose: I'll check. I guess I should be able to use it. I don't remember why I didn't use it

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,45 @@
...
+from dace.sdfg.sdfg import ConditionalBlock
+from dace.sdfg.state import ControlFlowRegion
+
+
+def move_branch_cfg_up_discard_conditions(if_block: ConditionalBlock, body_to_take: ControlFlowRegion):
+    """
+    Moves a branch of a conditional block up in the control flow graph (if_block) (CFG),
+    replacing the conditional with the selected branch (body_to_take), discarding
+    the conditional check and other branches. Connects the body to previous in and out neighbors of the CFG.
+    """
```
tbennun: This is not formatted in the Sphinx documentation format

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,45 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```
tbennun: Needs a file docstring stating what is the purpose of this file and why this cannot just be in transformation helpers.

Honestly I would move this function there.

dace/transformation/passes/lift_trivial_if.py
```diff
@@ -0,0 +1,152 @@
...
+from dace.sdfg.state import ControlFlowBlock
+from dace.sdfg.construction_utils import move_branch_cfg_up_discard_conditions
+from dace.transformation import pass_pipeline as ppl, transformation
+import dace.sdfg.utils as sdutil
+import sympy
+from sympy import pycode
+
+
+@transformation.explicit_cf_compatible
+class LiftTrivialIf(ppl.Pass):
```
tbennun, line 19: needs docstring (also to appear properly in vscode!)

tests/passes/lift_trivial_if_test.py
```diff
@@ -0,0 +1,253 @@
...
+    "abs(-5) == -5",
+]
+
+cant_eval = ["a < 5", "c == 0", "d >= 1"]
+
+# Conditions where pystr_to_symbolic yields an unevaluated symbolic
+# expression (Function/Indexed/Symbol). bool() of such an expression is
+# truthy, which would mis-classify dynamic dataflow conditions like
+# `A[0]` as trivially true.
+dynamic_runtime_cond = ["A[0]", "tmp_r[0]", "x", "x[0] + 1", "A[i, j]"]
```
tbennun: nit: I'd use all-caps variable names such as `ALWAYS_TRUE` to ensure this is a test constant. Maybe even `_ALWAYS_TRUE` so it is not visible by default.

tbennun (changes requested):
Looks great, minor comments only

dace/transformation/passes/lift_trivial_if.py
ThrudPrimrose: I added the docstring.

dace/sdfg/construction_utils.py
ThrudPrimrose: I moved the function to the helpers file.

tests/passes/lift_trivial_if_test.py
ThrudPrimrose: Makes sense, I have updated the variable names.

tbennun (approved):

tbennun:
cscs-ci run

PR #2143, opened 2025-09-09
Add split tasklets pass

ThrudPrimrose:
Fortran Frontend results with tasklets that can have multiple ops inside (other than users being able to write arbitrary code inside them).

For the explicit vectorization pass I am building, I need tasklets to consist of a single operation (exception: FMA as it is often supported by vector processors).

This pass splits such tasklets into vectorizable tasklets.
For example this:
image

becomes:
image

I will write a separate pass that tries to detect and fuse an FMA operation.

### DIFF
--- a/dace/transformation/passes/split_tasklets.py
+++ b/dace/transformation/passes/split_tasklets.py
@@ -155,7 +155,6 @@ class SplitTasklets(ppl.Pass):
                     t.add_out_connector(lhs_var)
                 added_tasklets.append(t)
 
-
             added_accesses = dict()
             for i, t in enumerate(added_tasklets):
                 assert isinstance(t, dace.nodes.Tasklet)


dace/transformation/passes/split_tasklets.py
```diff
@@ -0,0 +1,227 @@
+# Copyright 2019-2025
```
phschaad: Non-standard copyright statement

dace/transformation/passes/split_tasklets.py
```diff
@@ -0,0 +1,227 @@
...
+                    inputs=set(rhs_vars),
+                    outputs=set(lhs_vars),
+                    code=ssa_statement,
+                )
+                for rhs_var in rhs_vars:
+                    t.add_in_connector(rhs_var)
+                for lhs_var in lhs_vars:
+                    t.add_out_connector(lhs_var)
+                added_tasklets.append(t)
+            #tasklet_new_connection_map[t] = (added_tasklets[0], added_tasklets[-1])
```
phschaad: remove

dace/transformation/passes/split_tasklets.py
```diff
@@ -0,0 +1,227 @@
...
+        # i3 -> |         |
+        # Now they need to be split e.g.:
+        # i1 -> |         |
+        # i2 -> | tasklet | -> tmp1 -> |         |
+        #                        i3 -> | tasklet | -> o1
+
+        # For the case a tasklet goes to a taskelt that needs to be split
+        # If we have t1 -> t2 but then split t1 to (t1.1, t1.2) -> t2
+        # For each tasklet we split we need to track the new input and output maps
+        #tasklet_new_connection_map = dict()
```
phschaad: leftover?

phschaad (changes requested):
The PR looks good, minor comments and changes.

However, I don't feel too great about this as a whole. This goes fully against the DaCe philosophy and very much seem like a solution to a problem that should not exist. There are two things that go against the DaCe philosophy with this:
1. Tasklets are a black box.
2. Tasklets are "minimal".
It is my opinion that it is a frontend's responsibility to produce tasklets in an as-atomic-as-possible fashion, which then allows full dataflow analysis (no dataflow is hidden) and helps other optimizations work.

Only when a _conscious decision_ is made by a human optimizer or a more in-depth optimization pass, are tasklets fused together, with the express purpose of enabling some specific optimization, and while accepting the fact that further analysis / optimization of that tasklet is significantly hindered.

If a frontend _already_ produces tasklets where the granularity is larger than one operation, then this is a frontend that does not amend itself well to DaCe optimization. If a pass like this is something that a frontend does after constructing an SDFG and before the SDFG ever enters the DaCe middleware, that may be something I am more happy with (not on a principle of good design, but simply accepting that the SDFG coming out of the frontend conforms to DaCe to begin with). Since the structure we are moving to with DaCe 2.0, involving schedule trees, does not allow for that to work with an actual DaCe SDFG pass, that would also not be the right solution for that, though.

dace/transformation/passes/split_tasklets.py
ThrudPrimrose: Yes, will remove unnecessary comments.

PR #2144, opened 2025-09-09
Add memcpy and memset library nodes

ThrudPrimrose:
I have added these to the memset and memcpy library nodes.
I will use these library nodes in a follow-up PR where I will be PRing a pass that replaces memcpy / memset maps wilibrary nodesdes, which can get expanded to the corresponding calls.


ThrudPrimrose:
> @ThrudPrimrose wmma fragment fill was just an example of a potential implementation. The key here is to have a Fill (what you call memset) library node that the frontends can use to more effectively set or clear memory.

I see, then yes I am doing exactly that :)

PR #2148, opened 2025-09-14
Add dace::float32sr type to DaCe

creavin:
This new type enables DaCe users to perform calculations with stochastic rounding in single precision.

This change is validated with unit tests.

### DIFF
--- a/dace/dtypes.py
+++ b/dace/dtypes.py
@@ -911,6 +911,43 @@ class pyobject(opaque):
         return ctypes.cast(obj_id, ctypes.py_object).value
 
 
+class Float32sr(typeclass):
+
+    def __init__(self):
+        self.type = numpy.float32
+        self.bytes = 4
+        self.dtype = self
+        self.typename = "float"
+        self.stochastically_rounded = True
+
+    def to_json(self):
+        return 'float32sr'
+
+    @staticmethod
+    def from_json(json_obj, context=None):
+        from dace.symbolic import pystr_to_symbolic # must be included!
+        return float32sr()
+
+    @property
+    def ctype(self):
+        return "dace::float32sr"
+
+    @property
+    def ctype_unaligned(self):
+        return self.ctype
+
+    def as_ctypes(self):
+        """ Returns the ctypes version of the typeclass. """
+        return _FFI_CTYPES[self.type]
+
+    def as_numpy_dtype(self):
+        return numpy.dtype(self.type)
+
+    @property
+    def base_type(self):
+        return self
+
+
 class compiletime:
     """
     Data descriptor type hint signalling that argument evaluation is
@@ -1257,6 +1294,7 @@ complex64 = typeclass(numpy.complex64)
 complex128 = typeclass(numpy.complex128)
 string = stringtype()
 MPI_Request = opaque('MPI_Request')
+float32sr = Float32sr()
 
 
 @undefined_safe_enum
--- a/dace/frontend/python/replacements/operators.py
+++ b/dace/frontend/python/replacements/operators.py
@@ -113,6 +113,36 @@ def _is_op_boolean(op: str):
     return False
 
 
+def _handle_casting_for_stochastically_rounded_types(input_datatypes, result_type, cast_types):
+    float_to_sr = {
+        dace.float32: dace.float32sr, 
+    }
+
+    for i, dtype in enumerate(input_datatypes):
+        if hasattr(dtype, "stochastically_rounded"):
+            if cast_types[i] and dtype.type == eval(cast_types[i]).type:
+                cast_types[i] = None
+
+    # check if stoc rounded inputs
+    stochastically_rounded = True
+    for i, dtype in enumerate(input_datatypes):
+        if not hasattr(dtype, "stochastically_rounded"):
+            stochastically_rounded = False
+            break
+
+    if stochastically_rounded:
+        # make the result SR
+        if result_type in float_to_sr:
+            result_type = float_to_sr[result_type]
+
+        # cast the intermediate types
+        for i, dtype in enumerate(cast_types):
+            if dtype in float_to_sr:
+                cast_types[i] = float_to_sr[dtype]
+    
+    return result_type
+
+
 def result_type(arguments: Sequence[Union[str, Number, symbolic.symbol, sp.Basic]],
                 operator: str = None) -> Tuple[Union[List[dtypes.typeclass], dtypes.typeclass, str], ...]:
 
@@ -144,12 +174,16 @@ def result_type(arguments: Sequence[Union[str, Number, symbolic.symbol, sp.Basic
             raise TypeError("Type {t} of argument {a} is not supported".format(t=type(arg), a=arg))
 
     complex_types = {dtypes.complex64, dtypes.complex128, np.complex64, np.complex128}
-    float_types = {dtypes.float16, dtypes.float32, dtypes.float64, np.float16, np.float32, np.float64}
+    float_types = {dace.float16, dace.float32, dace.float32sr, dace.float64, np.float16, np.float32, np.float64}
     signed_types = {dtypes.int8, dtypes.int16, dtypes.int32, dtypes.int64, np.int8, np.int16, np.int32, np.int64}
     # unsigned_types = {np.uint8, np.uint16, np.uint32, np.uint64}
 
     coarse_types = []
-    for dtype in datatypes:
+    for dt in datatypes:
+        dtype = dt
+        if hasattr(dt, "srtype"): # unwrap stochastically rounded vars
+            dtype = dt.srtype
+
         if dtype in complex_types:
             coarse_types.append(3)  # complex
         elif dtype in float_types:
@@ -347,6 +381,8 @@ def result_type(arguments: Sequence[Union[str, Number, symbolic.symbol, sp.Basic
         for i, t in enumerate(coarse_types):
             if t != coarse_result_type:
                 casting[i] = cast_str(restype)
+
+    result_type = _handle_casting_for_stochastically_rounded_types(datatypes, restype, casting)
 
     return restype, casting
 
--- a/dace/libraries/blas/nodes/dot.py
+++ b/dace/libraries/blas/nodes/dot.py
@@ -70,6 +70,7 @@ class ExpandDotOpenBLAS(ExpandTransformation):
         (desc_x, stride_x), (desc_y, stride_y), desc_res, sz = node.validate(parent_sdfg, parent_state)
         dtype = desc_x.dtype.base_type
         veclen = desc_x.dtype.veclen
+        cast = "(float *)" if dtype == dace.float32sr else ""
 
         try:
             func, _, _ = blas_helpers.cublas_type_metadata(dtype)
@@ -82,7 +83,9 @@ class ExpandDotOpenBLAS(ExpandTransformation):
         n = n or node.n or sz
         if veclen != 1:
             n /= veclen
-        code = f"_result = cblas_{func}({n}, _x, {stride_x}, _y, {stride_y});"
+
+
+        code = f"_result = cblas_{func}({n}, {cast} _x, {stride_x}, {cast} _y, {stride_y});"
         # The return type is scalar in cblas_?dot signature
         tasklet = dace.sdfg.nodes.Tasklet(node.name,
                                           node.in_connectors, {'_result': dtype},
@@ -545,7 +548,11 @@ class Dot(dace.sdfg.nodes.LibraryNode):
         if desc_x.dtype != desc_y.dtype:
             raise TypeError(f"Data types of input operands must be equal: {desc_x.dtype}, {desc_y.dtype}")
         if desc_x.dtype.base_type != desc_res.dtype.base_type:
-            raise TypeError(f"Data types of input and output must be equal: {desc_x.dtype}, {desc_res.dtype}")
+            input_types = (desc_x.dtype.base_type, desc_res.dtype.base_type)
+            if dace.float32sr in input_types and dace.float32sr in input_types:
+                pass  # ignore mismatch if it is stochastically rounded
+            else:
+                raise TypeError(f"Data types of input and output must be equal: {desc_x.dtype}, {desc_res.dtype}")
 
         # Squeeze input memlets
         squeezed1 = copy.deepcopy(in_memlets[0].subset)
--- a/dace/libraries/blas/nodes/gemv.py
+++ b/dace/libraries/blas/nodes/gemv.py
@@ -735,6 +735,8 @@ class ExpandGemvOpenBLAS(ExpandTransformation):
                                                                                    name_out="_y")
         dtype_a = outer_array_a.dtype.type
         dtype = outer_array_x.dtype.base_type
+        cast = "(float *)" if dtype == dace.float32sr else ""
+
         veclen = outer_array_x.dtype.veclen
         alpha = f'{dtype.ctype}({node.alpha})'
         beta = f'{dtype.ctype}({node.beta})'
@@ -783,7 +785,7 @@ class ExpandGemvOpenBLAS(ExpandTransformation):
             alpha = '&__alpha'
             beta = '&__beta'
 
-        code += f"""cblas_{func}({layout}, {trans}, {m}, {n}, {alpha}, _A, {lda},
+        code += f"""cblas_{func}({layout}, {trans}, {m}, {n}, {alpha}, {cast} _A, {lda},
                                 _x, {strides_x[0]}, {beta}, _y, {strides_y[0]});"""
 
         tasklet = dace.sdfg.nodes.Tasklet(node.name,
--- a/dace/libraries/linalg/nodes/cholesky.py
+++ b/dace/libraries/linalg/nodes/cholesky.py
@@ -16,6 +16,7 @@ def _make_sdfg(node, parent_state, parent_sdfg, implementation):
 
     inp_desc, inp_shape, out_desc, out_shape = node.validate(parent_sdfg, parent_state)
     dtype = inp_desc.dtype
+    cast = "(float *)" if dtype == dace.float32sr else ""
 
     sdfg = dace.SDFG("{l}_sdfg".format(l=node.label))
 
@@ -35,7 +36,7 @@ def _make_sdfg(node, parent_state, parent_sdfg, implementation):
     _, me, mx = state.add_mapped_tasklet('_uzero_',
                                          dict(__i="0:%s" % out_shape[0], __j="0:%s" % out_shape[1]),
                                          dict(_inp=Memlet.simple('_b', '__i, __j')),
-                                         '_out = (__i < __j) ? 0 : _inp;',
+                                         f'_out = (__i < __j) ? {cast}(0) : _inp;',
                                          dict(_out=Memlet.simple('_b', '__i, __j')),
                                          language=dace.dtypes.Language.CPP,
                                          external_edges=True)
--- a/dace/runtime/include/dace/reduction.h
+++ b/dace/runtime/include/dace/reduction.h
@@ -3,6 +3,7 @@
 #define __DACE_REDUCTION_H
 
 #include <cstdint>
+#include <dace/stocastic_rounding.h>
 
 #include "types.h"
 #include "vector.h"
@@ -114,6 +115,40 @@ namespace dace {
         template <typename WCR>
         static DACE_HDFI float reduce(WCR wcr, float *ptr, const float& value) {
             float old = *ptr;
+            *ptr = wcr(old, value);
+            return old;
+        }
+    };
+
+    template <>
+    struct wcr_custom<dace::float32sr> {
+        template <typename WCR>
+        static DACE_HDFI dace::float32sr reduce_atomic(WCR wcr, dace::float32sr *ptr, const dace::float32sr& value) {
+            #ifdef DACE_USE_GPU_ATOMICS
+                // Stochastic rounding version of atomic float reduction
+                int *iptr = reinterpret_cast<int *>(ptr);
+                int old = *iptr, assumed;
+                do {
+                    assumed = old;
+                    float old_val = __int_as_float(assumed);
+                    float new_val = static_cast<float>(wcr(static_cast<dace::float32sr>(old_val), value));
+                    old = atomicCAS(iptr, assumed, __float_as_int(new_val));
+                } while (assumed != old);
+                return static_cast<dace::float32sr>(__int_as_float(old));
+            #else
+                dace::float32sr old;
+                #pragma omp critical
+                {
+                    old = *ptr;
+                    *ptr = wcr(old, value);
+                }
+                return old;
+            #endif
+        }
+
+        template <typename WCR>
+        static DACE_HDFI dace::float32sr reduce(WCR wcr, dace::float32sr *ptr, const dace::float32sr& value) {
+            dace::float32sr old = *ptr;
             *ptr = wcr(old, value);
             return old;
         }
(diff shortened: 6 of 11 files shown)


acalotoiu (commented):
The tests would need to be fixed. I am also not sure if we should aim merging to main before trying to work on improving the performance of the SR implementation.

creavin:
Ack, will (attempt to) fix the tests this week

dace/frontend/python/replacements/operators.py
```diff
@@ -336,18 +370,20 @@ def result_type(arguments: Sequence[Union[str, Number, symbolic.symbol, sp.Basic
     else:  # Operators with 3 or more arguments
         restype = np_result_type(dtypes_for_result)
         coarse_result_type = None
-        if result_type in complex_types:
+        if restype in complex_types:
```
phschaad, line 373: What is/was `result_type`, why did this change? Was this a bug?

dace/libraries/blas/nodes/dot.py
```diff
@@ -204,7 +206,11 @@ def validate(self, sdfg, state):
         if desc_x.dtype != desc_y.dtype:
             raise TypeError(f"Data types of input operands must be equal: {desc_x.dtype}, {desc_y.dtype}")
         if desc_x.dtype.base_type != desc_res.dtype.base_type:
-            raise TypeError(f"Data types of input and output must be equal: {desc_x.dtype}, {desc_res.dtype}")
+            input_types = (desc_x.dtype.base_type, desc_res.dtype.base_type)
+            if dace.float32sr in input_types and dace.float32sr in input_types:
+                pass  # ignore mismatch if it is stochastically rounded
```
phschaad: I'm not an expert on SR types, so my question here is: can we safely do that, or rather - why can we safely do that?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,390 @@
+#ifndef __DACE_SROUND_H
```
phschaad, line 2: Please include the appropriate copyright comment

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,390 @@
...
+#include <thread>
+
+#ifdef __CUDACC__
+#define DACE_HOST_DEVICE __host__ __device__
+#include <curand_kernel.h>
+#else
+#define DACE_HOST_DEVICE
+#endif
+
+namespace dace {
```
phschaad, line 23: This file could in general use some more documenting comments

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,390 @@
...
+        uint64_t& rng_state = get_rng_state_64();
+        rng_state ^= rng_state << 13;
+        rng_state ^= rng_state >> 7;
+        rng_state ^= rng_state << 17;
+        return rng_state;
+    }
+    #endif
+
+    DACE_HOST_DEVICE static float stochastic_round(double x) {
+        // uint64_t rbits = lcg64();
```
phschaad: Do these commented out lines still carry significance?

tests/codegen/float32sr.py
```diff
@@ -0,0 +1,371 @@
...
+    print(f"LF: {lower_diff}")
+    upper_diff = abs(upper - higher_prec_val)
+    print(f"UF: {upper_diff}")
+    theoretical_prob = lower_diff / (lower_diff + upper_diff)
+
+    return lower_diff, upper_diff, theoretical_prob
+
+
+if __name__ == "__main__":
+    import pytest
```
phschaad: Please check the other test files on how we run tests from the main function. Note that this is done for debugging purposes and thus is important.

dace/frontend/python/replacements/operators.py
creavin: This is a bug introduced by this refactor: https://github.com/spcl/dace/commit/c909f8bf12337926156e23a0f3ed83238cf6e7f6#diff-b20441227a628465d1c6ca8819915b77cf036f8fc1a8e26fbceb8a930fde9d1dR353

The origional code, formerly in the replacements.py file, looked like
```
    else:  # Operators with 3 or more arguments
        result_type = _np_result_type(dtypes_for_result)
        coarse_result_type = None
        if result_type in complex_types:
            coarse_result_type = 3  # complex
        elif result_type in float_types:
            coarse_result_type = 2  # float
        elif result_type in signed_types:
            coarse_result_type = 1  # signed integer, bool
        else:
            coarse_result_type = 0  # unsigned integer
        for i, t in enumerate(coarse_types):
            if t != coarse_result_type:
                casting[i] = _cast_str(result_type)
```

I believe this var was renamed to restype to not conflict when the "result_type" function

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ack

dace/libraries/blas/nodes/dot.py
creavin: I have made a typo here -- thank you for catching that. Line 210 should be: 
`if dace.float32sr in input_types and dace.float32 in input_types:`. I will also rename `input_types` to `arg_types` and add more comments

What this does and why it's needed: when using stocastic rounding, a legitimate (i.e not a bug) mismatch between the input and output arguments may arise where one argument is a float32sr and the other is a float32 (round-to-nearest). The underlying data type is the same so this should not cause the validation to fail. This is a check for that edge-case

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ack

dace/runtime/include/dace/stocastic_rounding.h
creavin: So as part of my thesis, I make a comparison between five different rounding modes. I kept the functions and these commented out invocations partly for posterity and partly to allow future devs to play around with it. 

I think there is value in keeping this. I will make this into one compact comment to reduce its footprint. If you disgree, I can remove the comment and the associated functions

tests/codegen/float32sr.py
creavin: Makes sense. Updated to match the existing test calling style

creavin (commented):
Addressed all feedback item. Corrections to follow in next commit

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+    #else
+        // Default stochastic rounding mode: Linear congruential generator (lcg64)
+        // Alternative SR modes: Circular buffer (get_preloaded_random_u64), Xorshift64 generator (xorshift64)
+        return lcg64();
+    #endif
+    }
+
+    #if !defined(__CUDA_ARCH__)
+    static std::mt19937_64& get_rng_mt() {
+        static std::mt19937_64 rng(std::random_device{}());
```
tbennun, line 142: not repeatable with seed

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+        double x;
+        std::memcpy(&x, &bits, sizeof(x));
+        return x;
+    #endif
+    }
+
+    DACE_HOST_DEVICE static uint64_t get_random_u64() {
+    #if defined(__CUDA_ARCH__)
+        int tid = threadIdx.x + blockIdx.x * blockDim.x;
+        uint64_t seed = clock64() + tid;
```
tbennun, line 126: should the seed be different for every thread?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+#include <cstdint>
+#include <cstring>
+#include <iomanip>
+#include <iostream>
+#include <limits>
+#include <random>
+#include <thread>
+
+#ifdef __CUDACC__
+#define DACE_HOST_DEVICE __host__ __device__
```
tbennun, line 17: why not use existing macros from other headers?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+#include <curand_kernel.h>
```
tbennun, line 18: what about rocrand?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+#include <cmath>
+#include <cstdint>
+#include <cstring>
+#include <iomanip>
+#include <iostream>
+#include <limits>
+#include <random>
+#include <thread>
+
+#ifdef __CUDACC__
```
tbennun, line 16: hipcc?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+        return __longlong_as_double(bits);
+    #else
+        double x;
+        std::memcpy(&x, &bits, sizeof(x));
+        return x;
+    #endif
+    }
+
+    DACE_HOST_DEVICE static uint64_t get_random_u64() {
+    #if defined(__CUDA_ARCH__)
```
tbennun, line 124: mixture of `__CUDA_ARCH__`  and `CUDACC`

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+    #if !defined(__CUDA_ARCH__)
+    static std::mt19937_64& get_rng_mt() {
+        static std::mt19937_64 rng(std::random_device{}());
+        return rng;
+    }
+    #endif
+
+    #if !defined(__CUDA_ARCH__)
+    static std::mt19937& get_rng_single() {
+        static std::mt19937 rng(std::random_device{}());
```
tbennun, line 149: not repeatable/reproducible with seeds

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+
+public:
+    DACE_HOST_DEVICE float32sr() : value(0.0f) {}
+    DACE_HOST_DEVICE float32sr(int v) : value(static_cast<double>(v)) {}
+    DACE_HOST_DEVICE float32sr(float v) : value(v) {}
+    DACE_HOST_DEVICE float32sr(double v) : value(stochastic_round(v)) {}
+
+    DACE_HOST_DEVICE operator float() const { return value; }
+    DACE_HOST_DEVICE operator float*() { return &value; }
+    DACE_HOST_DEVICE operator const float*() const { return &value; }
```
tbennun, line 162: strange. is this necessary?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+            for (int i = 0; i < 1000; ++i) {
+                shared_randoms[i] = get_random_u64();
+            }
+            index = 0;
+            initialized = true;
+        }
+
+        __syncthreads();
+
+        int my_index = atomicAdd(&index, 1) % 1000;
```
tbennun, line 186: why not use the thread index to avoid that?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+
+    friend std::istream& operator>>(std::istream& is, float32sr& obj);
+    friend std::ostream& operator<<(std::ostream& os, const float32sr& obj);
+
+    // Circular buffer for random number generation
+    // The idea is a large buffer of rands is created once on init then
+    // reused by all threads for rounding.
+    DACE_HOST_DEVICE static uint64_t get_preloaded_random_u64() {
+    #if defined(__CUDA_ARCH__)
+        __shared__ uint64_t shared_randoms[1000];
```
tbennun, line 172: imposing shared memory (8KB) that does not get registered with the code generator for validation.

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+
+    DACE_HOST_DEVICE float32sr operator=(const int& v) {
+        value = stochastic_round(static_cast<double>(v));
+        return *this;
+    }
+
+    DACE_HOST_DEVICE float32sr operator=(const int64_t& v) {
+        value = stochastic_round(static_cast<double>(v));
+        return *this;
+    }
```
tbennun, line 267: do integers need stochastic rounding?

dace/runtime/include/dace/stocastic_rounding.h
```diff
@@ -0,0 +1,395 @@
...
+
+    //==================
+    // Primary operators
+    //==================
+    DACE_HOST_DEVICE float32sr operator+(const float32sr& other) const {
+        return float32sr(stochastic_round(static_cast<double>(value) + static_cast<double>(other.value)));
```
tbennun, line 273: so this is going up to double, adding with the double units, but storing as float32.
Neither `float32sr`  nor its behavior are documented anywhere!

dace/frontend/python/replacements/operators.py
```diff
@@ -144,12 +174,16 @@ def result_type(arguments: Sequence[Union[str, Number, symbolic.symbol, sp.Basic
             raise TypeError("Type {t} of argument {a} is not supported".format(t=type(arg), a=arg))
 
     complex_types = {dtypes.complex64, dtypes.complex128, np.complex64, np.complex128}
-    float_types = {dtypes.float16, dtypes.float32, dtypes.float64, np.float16, np.float32, np.float64}
+    float_types = {dace.float16, dace.float32, dace.float32sr, dace.float64, np.float16, np.float32, np.float64}
```
tbennun, line 177: no, you should use the module if it exists.

dace/frontend/python/replacements/operators.py
```diff
@@ -144,12 +174,16 @@ def result_type(arguments: Sequence[Union[str, Number, symbolic.symbol, sp.Basic
...
     signed_types = {dtypes.int8, dtypes.int16, dtypes.int32, dtypes.int64, np.int8, np.int16, np.int32, np.int64}
     # unsigned_types = {np.uint8, np.uint16, np.uint32, np.uint64}
 
     coarse_types = []
-    for dtype in datatypes:
+    for dt in datatypes:
```
tbennun, line 182: why the rename?

dace/dtypes.py
```diff
@@ -817,6 +817,43 @@ def to_python(self, obj_id: int):
         return ctypes.cast(obj_id, ctypes.py_object).value
 
 
+class Float32sr(typeclass):
```
tbennun, line 820: No documentation?

tests/codegen/float32sr.py
tbennun, line 1: this test will likely not run due to the filename not being test_*.py or *_test.py

dace/runtime/include/dace/stocastic_rounding.h
tbennun, line 1: Typo in file name: should be stochastic

dace/frontend/python/replacements/operators.py
```diff
@@ -113,6 +113,36 @@ def _is_op_boolean(op: str):
...
 
 
+def _handle_casting_for_stochastically_rounded_types(input_datatypes, restype, cast_types):
+    float_to_sr = {
+        dace.float32: dace.float32sr,
+    }
+
+    for i, dtype in enumerate(input_datatypes):
+        if hasattr(dtype, "stochastically_rounded"):
+            if cast_types[i] and dtype.type == eval(cast_types[i]).type:
```
tbennun, line 123: random eval, not good!!!

dace/frontend/python/replacements/operators.py
```diff
@@ -113,6 +113,36 @@ def _is_op_boolean(op: str):
...
+
+    if stochastically_rounded:
+        # make the result SR
+        if restype in float_to_sr:
+            restype = float_to_sr[restype]
+
+        # cast the intermediate types
+        for i, dtype in enumerate(cast_types):
+            if dtype in float_to_sr:
+                cast_types[i] = float_to_sr[dtype]
```
tbennun, line 141: why is cast_types not also a return value here, just to clarify that it gets mutated?

dace/frontend/python/replacements/operators.py
```diff
@@ -113,6 +113,36 @@ def _is_op_boolean(op: str):
     return False
 
 
+def _handle_casting_for_stochastically_rounded_types(input_datatypes, restype, cast_types):
```
tbennun, line 116: docstring? data type hints??

dace/libraries/linalg/nodes/cholesky.py
```diff
@@ -16,6 +16,7 @@ def _make_sdfg(node, parent_state, parent_sdfg, implementation):
 
     inp_desc, inp_shape, out_desc, out_shape = node.validate(parent_sdfg, parent_state)
     dtype = inp_desc.dtype
+    cast = "(float *)" if dtype == dace.float32sr else ""
```
tbennun, line 19: casting a scalar to a pointer!!!

dace/runtime/include/dace/reduction.h
```diff
@@ -313,6 +348,31 @@ namespace dace {
         DACE_HDFI float operator()(const float &a, const float &b) const { return ::max(a, b); }
     };
 
+
+    template <>
+    struct _wcr_fixed<ReductionType::Min, dace::float32sr> {
+
+        static DACE_HDFI dace::float32sr reduce_atomic(dace::float32sr *ptr, const dace::float32sr& value) {
+            return wcr_custom<dace::float32sr>::reduce_atomic(
+                _wcr_fixed<ReductionType::Min, dace::float32sr>(), ptr, value);
```
tbennun, line 357: shouldn't you just use the float32 atomicMin/Max here?

dace/libraries/blas/nodes/gemv.py
```diff
@@ -232,6 +232,8 @@ def expansion(node: 'Gemv', state, sdfg, m=None, n=None, **kwargs):
                                                                                    name_out="_y")
         dtype_a = outer_array_a.dtype.type
         dtype = outer_array_x.dtype.base_type
+        cast = "(float *)" if dtype == dace.float32sr else ""
```
tbennun, line 235: I think that C++ can handle those casts if necessary

dace/runtime/include/dace/reduction.h
```diff
@@ -313,6 +348,31 @@ namespace dace {
...
+        }
+
+
+        DACE_HDFI dace::float32sr operator()(const dace::float32sr &a, const dace::float32sr &b) const { return ::min(a, b); }
```
tbennun, line 361: does that work? should they not be cast to float explicitly first?
I think that breaks AMD GPU builds

dace/runtime/include/dace/reduction.h
```diff
@@ -313,6 +348,31 @@ namespace dace {
...
+
+    template <>
+    struct _wcr_fixed<ReductionType::Max, dace::float32sr> {
+
+        static DACE_HDFI dace::float32sr reduce_atomic(dace::float32sr *ptr, const dace::float32sr& value) {
+            return wcr_custom<dace::float32sr>::reduce_atomic(
+                _wcr_fixed<ReductionType::Max, dace::float32sr>(), ptr, value);
+        }
+
+        DACE_HDFI dace::float32sr operator()(const dace::float32sr &a, const dace::float32sr &b) const { return ::max(a, b); }
```
tbennun, line 372: same as above

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ack, will change

dace/runtime/include/dace/stocastic_rounding.h
creavin: Some background: this change came about because I wanted to experiment with using SR to enable lower precision calculations without incurring increased rounding error for a particular climate simulation. With this, we could port the simulation to GPU, enjoy a speed up, and optimise this code. In the end, because of the nature of the simulation, it did not benefit for SR and the research direction changed to explore the error properties and less about the performance. 

The CPU code here achieve reasonable performance: the Julia SR library notes a 5-10x penalty when using SR and we note a similar one for DaCe when SR is ran on NPBench. The GPU code on the otherhand is implemented insofar as it runs. There are no guarantees about performance. 

I considered not committing the GPU code but it is helpful for one to use as a starting point so I kept it. If you disagree, I can remove.

Anyways, to address your comment: I believe it should be. Take the case where you perform stochastic operations on blocks of uniform data. If all threads share the same seed, all values will be rounded in the same direction, which is not the correct behaviour. 

One improvement here would be to avoid calling `curand_init`. That, for me, is outside the scope of this change: either we keep this basic implementation, or we remove it and add an issue for someone to add a performant version later

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ack, not aware of them, will update

dace/runtime/include/dace/stocastic_rounding.h
creavin: I haven't worked with ROC and don't want to expand this PR further to include it. But also, I would wait until there is a performant CUDA impl in place before porting to ROC

Given that the cuda code presented here is experimental, I don't think it makes sense to expand to roc

dace/runtime/include/dace/stocastic_rounding.h
creavin: See previous

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ack, will make this consistent

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ack, will set a seed

dace/runtime/include/dace/stocastic_rounding.h
creavin: iirc this was needed to handle some error arising from the generated c code when the blas libraries were called

dace/runtime/include/dace/stocastic_rounding.h
creavin: If the thread index is used, then the thread will always read the same random value from the circular buffer which is not the desired effect. It should advance each time. This function is kept for posterity. The performance hasn't been evaluated. There is likely a lot of room for improvement. But seeing the shared memory issue comment you raised, best I remove it.

dace/runtime/include/dace/stocastic_rounding.h
creavin: Good catch, I have removed this function

dace/runtime/include/dace/stocastic_rounding.h
creavin: Yes, they do. Float32 has a 24 bit mantissa. int{8,16} can be stored exactly; int{32, 64} must be rounded

dace/runtime/include/dace/stocastic_rounding.h
creavin: Yes, this is necessary to minimise rounding error. You perform the operation in higher precision then stochastically round back down to lower precision. Here is an example of how it is done in another SR [library](https://github.com/milankl/StochasticRounding.jl/blob/ff86c801ddd15b4182cb551b84811eb24295a48f/src/general.jl#L116-L118): 

```julia
for f in (:+, :-, :*, :/, :^, :mod, :atan, :hypot)
    @eval Base.$f(x::T, y::T) where {T<:AbstractStochasticFloat} = stochastic_round(T,$(f)(widen(x), widen(y)))
end
```

The type is widened (so float -> double), the operation is f is performed, then SR'd back down.

Where is the appropriate place to document this?

dace/frontend/python/replacements/operators.py
creavin: Ack

dace/frontend/python/replacements/operators.py
creavin: Will change this back to dtype -- artefact from my earlier impl

dace/dtypes.py
creavin: Ack, will add a short doc string

tests/codegen/float32sr.py
creavin: True will rename

dace/runtime/include/dace/stocastic_rounding.h
creavin: Ouch, yes

dace/frontend/python/replacements/operators.py
creavin: Ack, will do a lookup instead

dace/frontend/python/replacements/operators.py
creavin: Ack, good call, will change

dace/frontend/python/replacements/operators.py
creavin: Ack

dace/runtime/include/dace/reduction.h
creavin: True would be simplier. Will change

dace/libraries/blas/nodes/gemv.py
creavin: I will double check it but I ran into implicit cast issues for these blas calls

dace/libraries/linalg/nodes/cholesky.py
creavin: Ah, true, will change

dace/runtime/include/dace/reduction.h
creavin: This worked so I didn't question it. I can't speak to it breaking AMD GPU builds. I trust you're right here

dace/runtime/include/dace/reduction.h
creavin: See previous

PR #2149, opened 2025-09-14
Add Type Change Pass

creavin:
This change adds a new DaCE Pass that can replace one simple DaCe dtype with another. This change was successfully tested on a complex ICON sdfg. Here is a sample usage:

```
import dace
import numpy as np

from dace.transformation import pass_pipeline as ppl
from dace.transformation.passes.type_change import TypeChange

N = dace.symbol('N')

@dace.program
def simple(a: dace.float64[N], b: dace.float64[1]):
    for i in range(N):
        a[i] = a[i] * 2.0 + b[0]

sdfg = simple.to_sdfg()
tc = TypeChange(dace.float64, dace.float32)

type_change_pipeline = ppl.Pipeline([tc])
print("Pipeline created")
results = type_change_pipeline.apply_pass(sdfg, {})

print(results) # {'TypeChange': 6}

A = np.ones(10, dtype=np.float32)
B = np.ones(1, dtype=np.float64)
sdfg(A, B, N=10)
print(A)
```

### DIFF
--- a/dace/transformation/passes/type_change.py
+++ b/dace/transformation/passes/type_change.py
@@ -1,0 +1,149 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
+import dace
+from dace.transformation import pass_pipeline as ppl
+from dace import SDFG, properties
+from typing import Optional
+
+from dace.transformation.transformation import explicit_cf_compatible
+
+@properties.make_properties
+@explicit_cf_compatible
+class TypeChange(ppl.Pass):
+    CATEGORY: str = 'Helper'
+
+    def __init__(self, from_type: dace.typeclass = None, to_type: dace.typeclass = None):
+        self._from_type = from_type
+        self._to_type = to_type
+        self._swaps_count = 0
+
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Nothing
+
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        return False
+
+
+    def apply_pass(self, sdfg: SDFG, _) -> Optional[int]:
+        if hasattr(sdfg, "orig_sdfg") and sdfg.orig_sdfg:  # apply the pass to orig cpu sdfg
+            if hasattr(sdfg.orig_sdfg, "all_sdfgs_recursive"):
+                for nested_sdfg in sdfg.orig_sdfg.all_sdfgs_recursive():
+                    self._change_sdfg_type(nested_sdfg)
+
+        for nested_sdfg in sdfg.all_sdfgs_recursive():
+            self._change_sdfg_type(nested_sdfg)
+            
+        return self._swaps_count
+
+
+    def report(self, pass_retval: int) -> str:
+        if pass_retval is None:
+            return "No arrays found to analyze."
+        return f"Analyzed {pass_retval} arrays and printed their types."
+
+
+    def _change_sdfg_type(self, sdfg: SDFG):
+        # Swap nodes
+        for node, parent in sdfg.all_nodes_recursive():
+            if hasattr(node, "in_connectors"):
+                for in_con_name, in_con_type in node.in_connectors.items():
+                    if in_con_type == self._from_type:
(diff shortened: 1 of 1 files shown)


dace/transformation/passes/type_change.py
```diff
@@ -0,0 +1,149 @@
...
+    CATEGORY: str = 'Helper'
+
+    def __init__(self, from_type: dace.typeclass = None, to_type: dace.typeclass = None):
+        self._from_type = from_type
+        self._to_type = to_type
+        self._swaps_count = 0
+
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Nothing
```
phschaad: That does not appear to be true..

dace/transformation/passes/type_change.py
```diff
@@ -0,0 +1,149 @@
...
+import dace
+from dace.transformation import pass_pipeline as ppl
+from dace import SDFG, properties
+from typing import Optional
+
+from dace.transformation.transformation import explicit_cf_compatible
+
+@properties.make_properties
+@explicit_cf_compatible
+class TypeChange(ppl.Pass):
```
phschaad: A documentation comment is missing, and I would appreciate the name being less ambiguous tbh, but that may be personal preference.

dace/transformation/passes/type_change.py
```diff
@@ -0,0 +1,149 @@
...
+        return ppl.Modifies.Nothing
+
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        return False
+
+
+    def apply_pass(self, sdfg: SDFG, _) -> Optional[int]:
+        if hasattr(sdfg, "orig_sdfg") and sdfg.orig_sdfg:  # apply the pass to orig cpu sdfg
+            if hasattr(sdfg.orig_sdfg, "all_sdfgs_recursive"):
```
phschaad: why is this check necessary? SDFG's should implement this function, always.

dace/transformation/passes/type_change.py
```diff
@@ -0,0 +1,149 @@
...
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Nothing
+
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        return False
+
+
+    def apply_pass(self, sdfg: SDFG, _) -> Optional[int]:
+        if hasattr(sdfg, "orig_sdfg") and sdfg.orig_sdfg:  # apply the pass to orig cpu sdfg
```
phschaad: Same comment here, this should always be present on an SDFG, it may just be None (i.e., second check should suffice)

dace/transformation/passes/type_change.py
creavin: Ack, will update

dace/transformation/passes/type_change.py
creavin: Ack, will fix

dace/transformation/passes/type_change.py
creavin: Okay, I have SDFGs where orig_sdfg is not present but yes, I can just use the second check

dace/transformation/passes/type_change.py
creavin: Ack, not aware of that. Will remove

creavin:
Thank you both for your comments

> I’m not sure this is a good candidate for a pass. Maybe if it checked for reductions and, e.g., transcendental ops, for higher precision it would be more appropriate. Otherwise, as it is right now it should be a utility function.

I'm happy to move this into a utility function then. Where is the correct place to put it / do you have a similar function I can mimic

> Additionally, please avoid type(x) == y checks
Yes, good point. Will use `isinstance` instead

phschaad (changes requested):
The change looks good to me, however the PR is missing unit tests. This is a new feature and as such is certainly not covered by existing tests. Please ensure there are at least 2-3 unit tests to check for correct type changing behavior w.r.t. the data types and places the utility function modifies.

phschaad (approved):

PR #2152, opened 2025-09-18
Fix yield edge for states

ThrudPrimrose:
The yield returns intersate edges for state nodes, but I do not think that can be correct. I try to return multiconnector edges, can you check this?


ThrudPrimrose:
> This is incorrect, as you can also see from the failing tests. Also, it is unclear to me whether we want to return multiedges in the first place or just one edge. If we do, then I think it would be better to use `self.edges_between(e[0], e[1]) for e in path`, which is agnostic to the edge type.

I'll update

PR #2154, opened 2025-09-22
Copy / Set Library Nodes and Passes that Lift Memset / Memcpy kernels or Acecss Node -> Access Node Paths to use Copy Library Nodes

ThrudPrimrose:
In this PR, I introduce copy and memset library nodes, as well as a pass that lifts contiguous memory copies and memset operations from kernels and places them in copy / memset library nodes.

The implementation currently supports all existing copy strategies and a map-based fallback.  I support lowering to CUDA API calls, memset/memcpy calls, or kernels/maps as a fallback.

Example library node for copy:
image

The default strategy is to generate maps:
image

There are implementations of CUDA and CPU to lower to memcpy API calls. It looks like this:
image

The pass detects paths in maps that perform a 0 memset or memcpy, and lifts them to use the library nodes. This will be convenient, for example, not to need to change DaCe frontends and to run this as a post-processing pass. The pass respects data dependencies between kernels. The image I show has none, so it is not visible.
Before:
image

After:
image

(The detection only works if a data path between maps is only a 0-memset or memcpy, 2 paths are not lifted here)


dace/codegen/targets/cpp.py
```diff
@@ -935,8 +938,8 @@ def unparse_tasklet(sdfg, cfg, state_id, dfg, node, function_stream, callsite_st
         # If this code runs on the host and is associated with a GPU stream,
         # set the stream to a local variable.
         max_streams = int(Config.get("compiler", "cuda", "max_concurrent_streams"))
-        if not is_devicelevel_gpu(sdfg, state_dfg, node) and (hasattr(node, "_cuda_stream")
-                                                              or connected_to_gpu_memory(node, state_dfg, sdfg)):
+        if not is_devicelevel_gpu_kernel(sdfg, state_dfg, node) and (hasattr(node, "_cuda_stream")
```
alexnick83: Marking the change from `is_devicelevel_gpu` to `is_devicelevel_gpu_kernel` to check again. I guess it fixes an issue with expanded library nodes?

dace/libraries/standard/environments/cpu.py
```diff
@@ -0,0 +1,21 @@
...
+    cmake_minimum_version = None
+    cmake_packages = []
+    cmake_variables = {}
+    cmake_includes = []
+    cmake_libraries = []
+    cmake_compile_flags = []
+    cmake_link_flags = []
+    cmake_files = []
+
+    headers = []
```
alexnick83: I guess we want this to include cstring (or whatever C/C++ header would be appropriate)?

alexnick83: But you actually include the headers (for both CPU and GPU) through the `code_global` attribute below.

dace/libraries/standard/nodes/copy_node.py
```diff
@@ -0,0 +1,201 @@
...
+
+from dace.libraries.standard.helper import collapse_shape_and_strides, add_dynamic_inputs
+
+
+@library.expansion
+class ExpandPure(ExpandTransformation):
+    environments = []
+
+    @staticmethod
+    def expansion(node, parent_state, parent_sdfg):
```
alexnick83, line 484: Do we always want a mapped tasklet here? How does this implementation measure performance-wise compared to the copy/copyND methods in the DaCe runtime?

alexnick83 (commented):
It looks good. I am not so warm on the pass, especially regarding the map-fission behavior, as it may undo the work of other transformations. However, this is fine to have as an option until the frontends use the new functionality consistently. My main question is on the future of the current copy/copyND methods. Do they provide any benefits compared to the mapped tasklet expansion? If yes, should we consider an alternate expansion that utilizes them somehow?

dace/libraries/standard/environments/cpu.py
ThrudPrimrose: I will move it hear, it will be cleaner.

dace/libraries/standard/nodes/copy_node.py
ThrudPrimrose: From what I remember from benchmarking things, CopyND mostly converts CPU<->CPU copies into memcpy calls most often. This can be slower than the default map-based lowering (if copy size is not big enough), but also slower in certain cases.

I think the default lowering to a kernel will not always yield the best performance. But if we do the default to be the CPU version (memcpy, memset) then it should be same as what CopyND currently does?

dace/codegen/targets/cpp.py
ThrudPrimrose: When I tested with unit tests, I did not find a difference in behavior between these two calls. (I will run all unit tests again after reverting this change and if it changes anything including the copy)

alexnick83 (commented):
LGTM. However, is InsertGPUStreams used in any capacity in this PR, i.e., is the stream array actually utilized? If not, maybe it would make sense to leave it for the PR that will actually introduce the use of the stream array?

dace/libraries/standard/environments/cpu.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```

dace/libraries/standard/nodes/copy_node.py
```diff
@@ -0,0 +1,199 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```

dace/libraries/standard/helper.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
+import dace
+import copy
+
+
+# Compute collapsed shapes and strides, removing singleton dimensions (length == 1)
+def collapse_shape_and_strides(subset, strides):
```
phschaad: Both functions in this file could use a doc comment

dace/transformation/passes/assignment_and_copy_kernel_to_memset_and_memcpy.py
```diff
@@ -0,0 +1,533 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```

dace/transformation/passes/assignment_and_copy_kernel_to_memset_and_memcpy.py
```diff
@@ -0,0 +1,533 @@
...
+from dace.transformation import pass_pipeline as ppl, transformation
+from dace.libraries.standard.nodes.copy_node import CopyLibraryNode
+from dace.libraries.standard.nodes.memset_node import MemsetLibraryNode
+from dace.transformation.passes.fusion_inline import InlineSDFGs
+from typing import Dict, Iterable, List, Set
+
+
+@properties.make_properties
+@transformation.explicit_cf_compatible
+class AssignmentAndCopyKernelToMemsetAndMemcpy(ppl.Pass):
```
phschaad, line 17: Please add a doc comment to the class - doesn't need to be super long, just explain the pass briefly.

dace/transformation/passes/assignment_and_copy_kernel_to_memset_and_memcpy.py
```diff
@@ -0,0 +1,533 @@
...
+    overapproximate_first_dimension = properties.Property(
+        dtype=bool,
+        default=False,
+        desc=
+        "If True, the first dimension of the map is overapproximated to be contiguous, even if it is not. This is useful for some cases where the first dimension is always contiguous, but the map range is not.",
+    )
+    apply_only_on_labels = properties.ListProperty(element_type=str, default=[], allow_none=False)
+
+    rmid = 0
```
phschaad, line 42: this should probably be "private" (`_rmid`)

dace/transformation/passes/assignment_and_copy_kernel_to_memset_and_memcpy.py
```diff
@@ -0,0 +1,533 @@
...
+@properties.make_properties
+@transformation.explicit_cf_compatible
+class AssignmentAndCopyKernelToMemsetAndMemcpy(ppl.Pass):
+    overapproximate_first_dimension = properties.Property(
+        dtype=bool,
+        default=False,
+        desc=
+        "If True, the first dimension of the map is overapproximated to be contiguous, even if it is not. This is useful for some cases where the first dimension is always contiguous, but the map range is not.",
+    )
+    apply_only_on_labels = properties.ListProperty(element_type=str, default=[], allow_none=False)
```
phschaad: bit confusing - from the usage I suggest a better name would be something like "node_label_whitelist" or something. Could also be addressed by just adding a description string

dace/transformation/passes/assignment_and_copy_kernel_to_memset_and_memcpy.py
```diff
@@ -0,0 +1,533 @@
...
+
+        for memcpy_path in memcpy_paths:
+            src_access_node = memcpy_path[0].src
+            map_entry = memcpy_path[0].dst
+            tasklet = memcpy_path[1].dst
+            map_exit = memcpy_path[2].dst
+            dst_access_node = memcpy_path[3].dst
+            if src_access_node not in state.nodes() or map_entry not in state.nodes() or tasklet not in state.nodes(
+            ) or map_exit not in state.nodes() or dst_access_node not in state.nodes():
+                raise Exception(
```
phschaad: is the exception caught somewhere so that this only equates to a "skipping" as opposed to a failure? Otherwise probably a warning is better?

dace/transformation/passes/assignment_and_copy_kernel_to_memset_and_memcpy.py
```diff
@@ -0,0 +1,533 @@
...
+
+            assert node not in rmed_memsets
+            assert node not in rmed_memcpies
+            rmed_memcpies[node] = rmed_memcpy
+            rmed_memsets[node] = rmed_memset
+
+        num_rmed_memcpies = sum(rmed_memcpies.values())
+        num_rmed_memsets = sum(rmed_memsets.values())
+
+        InlineSDFGs().apply_pass(sdfg, {})
```
phschaad: Hm, not a huge fan of this - we should maybe discuss extending our pass infrastructure to include not just required passes (dependencies) to run before, but also specify if we want to run something after a pass, if this is truly a necessity here. This kind of hides a non-trivial cost in a pass a bit in my view.

dace/transformation/passes/gpu_specialization/insert_gpu_streams.py
```diff
@@ -0,0 +1,153 @@
...
+from dace.sdfg import is_devicelevel_gpu
+from dace.sdfg.nodes import AccessNode, MapEntry, MapExit, Node, Tasklet
+from dace.transformation import pass_pipeline as ppl, transformation
+from dace.transformation.passes.gpu_specialization.gpu_stream_scheduling import NaiveGPUStreamScheduler
+from dace.transformation.passes.gpu_specialization.helpers.gpu_helpers import get_dace_runtime_gpu_stream_name, get_gpu_stream_array_name, get_gpu_stream_connector_name
+
+
+@properties.make_properties
+@transformation.explicit_cf_compatible
+class InsertGPUStreams(ppl.Pass):
```
phschaad: Wait is this related with the other changes here - or is this a required addition for something? I don't quite understand the connection

tests/library/copy_node_test.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```

phschaad: Change this in the other files too pls

phschaad (changes requested):
Nice, only small changes and questions, thanks

PR #2157, opened 2025-09-24
Updated MapFusion To Allow Multiple Connections Between Intermediate and the Second Map Entry

philip-paul-mueller:


dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -576,49 +584,44 @@ def partition_first_outputs(
...
                             return None
 
-            # Now we determine the consumer of nodes. For this we are using the edges
-            #  leaves the second MapEntry. It is not necessary to find the actual
-            #  consumer nodes, as they might depend on symbols of nested Maps.
-            #  For the covering test we only need their subsets, but we will perform
-            #  some scan and filtering on them.
+            # We now determine the consumers of the intermediate node. For this we are
+            #  looking at the edges that leave the node and enter the second Map.
+            #  This means that we ignore the actual consumer, which are inside the
```
alexnick83:
```suggestion
            #  This means that we ignore the actual consumers, who are inside the
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -446,6 +446,12 @@ def partition_first_outputs(
         # Set of intermediate nodes that we have already processed.
         processed_inter_nodes: Set[nodes.Node] = set()
 
+        debug = False
```
alexnick83: TODO: Remove debug code before merging.

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -576,49 +584,44 @@ def partition_first_outputs(
...
-            # Now we determine the consumer of nodes. For this we are using the edges
-            #  leaves the second MapEntry. It is not necessary to find the actual
-            #  consumer nodes, as they might depend on symbols of nested Maps.
-            #  For the covering test we only need their subsets, but we will perform
-            #  some scan and filtering on them.
+            # We now determine the consumers of the intermediate node. For this we are
+            #  looking at the edges that leave the node and enter the second Map.
+            #  This means that we ignore the actual consumer, which are inside the
+            #  second Map scope. Instead we are only looking at the subset that the
+            #  second Map consumes. Note that we allows that the second Map is found
```
alexnick83:
```suggestion
            #  second Map consumes. Note that we allow the second Map to be found
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -576,49 +584,44 @@ def partition_first_outputs(
...
+                        if i < j and psbs1.intersects(psbs2):
                             return None
 
-            # Now we determine the consumer of nodes. For this we are using the edges
-            #  leaves the second MapEntry. It is not necessary to find the actual
-            #  consumer nodes, as they might depend on symbols of nested Maps.
-            #  For the covering test we only need their subsets, but we will perform
-            #  some scan and filtering on them.
+            # We now determine the consumers of the intermediate node. For this we are
+            #  looking at the edges that leave the node and enter the second Map.
```
alexnick83:
```suggestion
            # We now determine the consumers of the intermediate node. For this, we
            #  look at the edges that leave the node and enter the second Map.
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -576,49 +584,44 @@ def partition_first_outputs(
...
                 # The output of the top Map can not define a dynamic Map range in the
                 #  second Map.
-                if not intermediate_node_out_edge.dst_conn.startswith("IN_"):
+                if not intermediate_consumer_edge.dst_conn.startswith("IN_"):
                     return None
 
                 # Now we look at all edges that leave the second MapEntry, i.e., the
-                #  edges that feeds the consumer and define what is read inside the Map.
-                #  We do not check them, but collect them and inspect them.
+                #  edges that feeds the data to the actual consumers and define what is
```
alexnick83:
```suggestion
                #  edges that feed the data to the actual consumers and define what is
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -1434,21 +1437,25 @@ def is_shared_data(
...
             return True
 
+        # If the data is used (read/write) by more than one entity then it is shared (in this
+        #  state) and we do not have to scan the whole SDFG.
+        # NOTE: We can not use `state.{in, out}_degree(data) > 1` because we allow that there
+        #   are multiple connections between the intermediates and the Maps, thus we have to
+        #   look at all adjacent nodes.
+        # NOTE: We have to do these checks before we consult the pipeline results or perform a
+        #   scan, because the `FindSingleUseData` pass and the scan, ignores the degree and we
+        #   need to check for "shared data" within the state separately.
```
alexnick83, line 1445:
```suggestion
        # If the data is used (read/write) by more than one entity, then it is shared (in this
        #  state) and we do not have to scan the whole SDFG.
        # NOTE: We cannot use `state.{in, out}_degree(data) > 1` because we allow
        #   multiple connections between the intermediates and the Maps, thus we have to
        #   look at all adjacent nodes.
        # NOTE: We have to do these checks before we consult the pipeline results or perform a
        #   scan, because the `FindSingleUseData` pass and the scan ignore the degree, and we
        #   need to check for "shared data" within the state separately.
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -1480,42 +1487,28 @@ def _scan_sdfg_if_data_is_shared(
...
-        A data descriptor is classified as shared if any of the following is true:
-        - `data` is non transient data.
-        - `data` has at most one incoming and/or outgoing edge.
-        - There are other AccessNodes beside `data` that refer to the same data.
-        - The data is accessed on an interstate edge.
-
-        This function should not be called directly. Instead it is called indirectly
-        by `is_shared_data()` if there is no short cut.
+        Note that it is not safe to use this function directly, instead the `is_shared_data()`
+        should be used. The function will scan the entire SDFG and looks if it found another
```
alexnick83, line 1485:
```suggestion
        should be used. The function scans the entire SDFG and tries to find another
```

dace/transformation/dataflow/map_fusion_vertical.py
```diff
@@ -1480,42 +1487,28 @@ def _scan_sdfg_if_data_is_shared(
...
-        #   We just need to know if we can delete it (early exit).
-        if not data.desc(sdfg).transient:
-            return True
-
-        # See description in `is_shared_data()` for more.
-        if state.out_degree(data) > 1:
-            return True
-        if state.in_degree(data) > 1:
-            return True
+        # These checks ensures that this function is not called directly.
```
alexnick83:
```suggestion
        # These checks ensure that this function is not called directly.
```

tests/transformations/map_fusion_vertical_test.py
```diff
@@ -2702,6 +2715,378 @@ def test_map_fusion_with_non_slicing_nsdfg(strict_dataflow: bool):
...
+    assert state.scope_dict()[inner_map_exit] is not None
+
+    initial_ac = {dnode.data for dnode in count_nodes(state, dace.nodes.AccessNode, return_nodes=True)}
+    assert len(initial_ac) == 4
+    assert "b" in initial_ac
+
+    ref, res = make_sdfg_args(sdfg)
+    compile_and_run_sdfg(sdfg, **ref)
+
+    # Here it is possible to apply the fusion, because the nested SDFG does not need
```
alexnick83: Is this comment correct? I don't see any nested SDFG.

tests/transformations/map_fusion_vertical_test.py
```diff
@@ -2702,6 +2715,378 @@ def test_map_fusion_with_non_slicing_nsdfg(strict_dataflow: bool):
...
+    assert state.scope_dict()[inner_map_exit] is not None
+
+    initial_ac = {dnode.data for dnode in count_nodes(state, dace.nodes.AccessNode, return_nodes=True)}
+    assert len(initial_ac) == 6
+    assert "b" in initial_ac
+
+    ref, res = make_sdfg_args(sdfg)
+    compile_and_run_sdfg(sdfg, **ref)
+
+    # Here it is possible to apply the fusion, because the nested SDFG does not need
```
alexnick83: Same comment as above.

tests/transformations/map_fusion_vertical_test.py
```diff
@@ -2702,6 +2715,378 @@ def test_map_fusion_with_non_slicing_nsdfg(strict_dataflow: bool):
...
+    assert len(initial_maps) == 2
+
+    initial_ac = {dnode.data for dnode in count_nodes(state, dace.nodes.AccessNode, return_nodes=True)}
+    assert len(initial_ac) == 4
+    assert "b" in initial_ac
+
+    ref, res = make_sdfg_args(sdfg)
+    compile_and_run_sdfg(sdfg, **ref)
+
+    # Here it is possible to apply the fusion, because the nested SDFG does not need
```
alexnick83: I suppose the comment here should say that (currently) MapFusion cannot be applied because ... (same justification as in the xfail statement above).

tests/transformations/map_fusion_vertical_test.py
philip-paul-mueller: I am actually not sure, maybe my brain was not fully up.
But I think I copied some part from `test_map_fusion_with_non_slice_nsdfg`.

PR #2158, opened 2025-09-24
Fix InlineMultistateSDFG: find new name for all control-flow blocks

edopao:
In `InlineMultistateSDFG` transformation, the names of all control flow blocks need to be made unique, after inlining. The issue was that the function only considered the names of the states, ignoring other blocks like `LoopRegion` and `ConditionalBlock`.


phschaad (approved):
LGTM, thanks! I will update the PR and merge once the FPGA test issues have been resolved (CI issue, not related to PR)

PR #2160, opened 2025-09-24
Loop Local Memory Reduction

Berke-Ates:
This PR adds the LoopLocalMemoryReduction transformation, also known as K-Caching.
This transformation looks for loops, which access arrays with a fixed alive window, then shrinks the array size to the window size by introducing a modulo operation to the index computations (i.e. a ringbuffer).

Example:
```py
for i in range(2, N):
        v[i] = a[i - 1] + a[i - 2]
        a[i] = b[i] * 2
```

turns into:
```py
for i in range(2, N):
        v[i] = a[(i - 1) % 3] + a[(i - 2) % 3]
        a[i % 3] = b[i] * 2
```

### DIFF
--- a/dace/transformation/passes/loop_local_memory_reduction.py
+++ b/dace/transformation/passes/loop_local_memory_reduction.py
@@ -12,7 +12,7 @@ from dace.transformation.passes.analysis import loop_analysis, StateReachability
 from dace.symbolic import pystr_to_symbolic, issymbolic
 from dace.subsets import Range
 import copy
-from typing import Union, Set, Optional
+from typing import Union, Set, Optional, Dict, Any
 
 
 @properties.make_properties
@@ -110,11 +110,22 @@ class LoopLocalMemoryReduction(ppl.Pass):
         # If anything was modified, reapply
         return modified != ppl.Modifies.Nothing
 
-    def apply_pass(self, sdfg: sd.SDFG, _) -> Optional[Set[str]]:
+    def depends_on(self):
+        return {StateReachability, FindAccessStates}
+
+    def apply_pass(self, sdfg: sd.SDFG, pipeline_results: Dict[str, Any]) -> Optional[Set[str]]:
         self.num_applications = 0
-        self.states_reach = StateReachability().apply_pass(sdfg, {})
-        self.access_states = FindAccessStates().apply_pass(sdfg, {})
         self.out_of_loop_states_cache = {}
+
+        if StateReachability.__name__ in pipeline_results:
+            self.states_reach = pipeline_results[StateReachability.__name__]
+        else:
+            self.states_reach = StateReachability().apply_pass(sdfg, {})
+
+        if FindAccessStates.__name__ in pipeline_results:
+            self.access_states = pipeline_results[FindAccessStates.__name__]
+        else:
+            self.access_states = FindAccessStates().apply_pass(sdfg, {})
 
         for node, _ in sdfg.all_nodes_recursive():
             if isinstance(node, LoopRegion):


Berke-Ates:
This transformation is orthogonal to Loop2Map. In order for this transformation to apply, there needs to be a loop, which writes into the "future" and reads from the "past". This behavior allows us to reuse the memory in the first place. However, this also means there is intra loop dependency, which should make it impossible for Loop2Map to apply.

dace/transformation/passes/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+    @classmethod
+    def expressions(cls):
+        return [sdutil.node_path_graph(cls.loop)]
+
+    def annotates_memlets(self) -> bool:
+        return True
+
+    def can_be_applied(self, graph: ControlFlowRegion, expr_index, sdfg: sd.SDFG, permissive=False):
+        arrays = set(acc_node.data for acc_node in self.loop.data_nodes())
+        self.can_reach = StateReachability().apply_pass(sdfg, {})
```
phschaad: uff this is very very expensive, state reachability is - to the best of my knowledge - the most expensive pass we have, and applying it in each instance of `can_be_applied`, for each pattern match, is going to kill pattern matching in a large program with many loops. We need a better solution for this.

tests/transformations/interstate/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,787 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
```

dace/transformation/passes/loop_local_memory_reduction.py
phschaad: In a pass this information could be cached.

dace/transformation/interstate/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+      - Interleaved reads and writes
+      - Gaps in the range of indices
+      - Iteration variable on multiple dimensions, e.g. a[i, i]
+      - Constant indexes (a*i + b) with a = 0
+      - Index expression (a*i + b) with a > 1
+      - Backwards loops step < 0 or a < 0
+      - Step > 1
+
+    Does not work with:
+      - Non-linear index expressions (a*i + b)
```
phschaad: that looks the same as the expressions in "works with"?

dace/transformation/interstate/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+            # If we have a span of one, it's enough that reads happen after writes in the loop.
+            if span == 0:
+                cond = all(
+                    st.in_degree(an) > 0 and st.out_degree(an) > 0 for st in self.loop.all_states()
+                    for an in st.data_nodes() if an.data == array_name)
+
+            # Take maximum from previous accesses into account
+            try:
+                k = max(span, max_indices[dim])
+                not cond  # XXX: Ensure condition can be evaluated
```
phschaad: The `XXX: ..` comment suggests this should be taken care of before merging, or what's the plan here?

dace/transformation/passes/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+            # Condition must hold
+            if not cond or k + 1 >= sdfg.arrays[array_name].shape[dim]:
+                k_values.append(None)
+            else:
+                k_values.append(k + 1)  # +1 because k is the highest index accessed, so size is k+1
+        return k_values
+
+    def _write_is_loop_local(self, array_name: str, write_indices: list[list[tuple]], sdfg: sd.SDFG) -> bool:
+        # The (overapproximated) written subset must be written before read or not read at all.
+        # TODO: This is overly conservative. Just checks if there are access nodes after the loop.
```
phschaad, line 287: Hm, if this were called in a pass this would be dead simple and would not require iterating over data nodes within states (can get expensive too) - since there's an access sets pass, which, combined with reach information can give you all you need.

dace/transformation/passes/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def _get_max_indices_before_loop(self, array_name: str, sdfg: sd.SDFG) -> list[int]:
+        # Collect all read and write subsets of the array before the loop.
+        loop_states = set(self.loop.all_states())
+        subsets = set()
+        for k1, v1 in self.can_reach.items():
+            for k2, v2 in v1.items():
+                if len(v2.intersection(loop_states)) == 0 or k2 in loop_states:
+                    continue
+                access_nodes = set(an for an in k2.data_nodes() if an.data == array_name)
```
phschaad, line 311: Similar situation w.r.t. pass information caching

dace/transformation/interstate/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+        # Must be transient (otherwise it's observable)
+        if not sdfg.arrays[array_name].transient:
+            return False
+
+        # Views and References are not supported
+        if isinstance(sdfg.arrays[array_name], dt.View) or isinstance(sdfg.arrays[array_name], dt.Reference):
+            return False
+
+        # Loop step expression must be constant.
+        has_const_loop_expr, step = self._has_constant_loop_expressions(sdfg)
```
phschaad: You're calling `_can_apply_for_array` multiple times, but if I read correctly, this function's return value doesn't change on a per-array basis but just per-loop? So I suggest doing this first, before checking each array, as a possible early-termination condition.

dace/transformation/interstate/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+        # All constants (a == 0) must be in the same dimension.
+        for dim in range(len(read_indices[0])):
+            if any(il[dim][0] == 0 for il in read_indices) and any(il[dim][0] != 0 for il in read_indices):
+                return False
+            if any(il[dim][0] == 0 for il in write_indices) and any(il[dim][0] != 0 for il in write_indices):
+                return False
+
+        # None of the write accesses must be within a conditional block. Reads are ok.
+        for st in self.loop.all_states():
+            for an in st.data_nodes():
```
phschaad: Similar to above, you can cache all data nodes within the loop before checking the per-array can-be-applied - these repeat traversals can become expensive.

dace/transformation/interstate/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,472 @@
...
+        if all(k is None for k in Ks):
+            return False
+
+        # Otherwise, we can apply the transformation.
+        return True
+
+    def _apply_for_array(self, array_name: str, sdfg: sd.SDFG):
+        read_indices, write_indices = self._get_read_write_indices(array_name)
+        _, step = self._has_constant_loop_expressions(sdfg)
+        Ks = self._get_K_values(array_name, read_indices, write_indices, step, sdfg)
```
phschaad: All of these 3 things are something you've already computed in the per-array can-be-applied, why not pass them along and avoid recomputation?

dace/transformation/interstate/loop_local_memory_reduction.py
Berke-Ates: This line seems like dead code but has it's purpose (checking if the condition can be evaluated). The comment is a warning to not remove it. I will adjust the comment to make that clearer.

dace/transformation/passes/loop_local_memory_reduction.py
```diff
@@ -0,0 +1,482 @@
...
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.Edges | ppl.Modifies.Descriptors
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        # If anything was modified, reapply
+        return modified != ppl.Modifies.Nothing
+
+    def apply_pass(self, sdfg: sd.SDFG, _) -> Optional[Set[str]]:
+        self.num_applications = 0
+        self.states_reach = StateReachability().apply_pass(sdfg, {})
```
phschaad: Almost there, but the caching can be improved to make use of the pipeline infrastructure, that's what I meant originally. Check [this pass here](https://github.com/spcl/dace/blob/main/dace/transformation/passes/array_elimination.py#L20) for reference (using `depends_on` to make sure dependent passes are run first in a pipeline, and then obtain their results here. If not inside a pipeline, there will be no pipeline results, or rather, `StateReachability` can not be found inside the pipeline results, and if that's not the case, THEN you can manually apply the pass here as a fallback)

Berke-Ates: Added `depends_on` and fallback, if the results are not present.

Issue #2176, opened 2025-10-14
LoopToMap Issues


ThrudPrimrose:
When LoopToMap is applied to the main-compatible CloudSC SDFG, multiple issues arise:

Firstly, memlet-propagation uses symbols that are only defined in the nested map below and thus result in missing symbols:
<img width="1141" height="732" alt="Image" src="https://github.com/user-attachments/assets/e0d12c6f-36d5-4ebf-a602-48068524339b" />

Multiple unused symbols are not deleted from SDFG symbols, resulting in missing symbol errors.

Also, the symbol mapping is not correct for some SDFGs:

For example, to the Nested SDFG, the scalar `nlev` connects to the in connector `sym_nlev_var_112`, and then in the symbol mapping has `sym_nlev_var_112: nlev` as a field, but then `nlev` has never been defined as a symbol in the nestedSDFG. For example, symbol mapping mismatches with the connectors:

<img width="225" height="201" alt="Image" src="https://github.com/user-attachments/assets/94a4d71f-8496-44a0-89b1-5ff39b851dfa" />

To run, remove the `.txt` suffix I added to the SDFG to make it uploadable and apply LoopToMap transformations to the SDFG:
[stage1.sdfgz.txt](https://github.com/user-attachments/files/22901228/stage1.sdfgz.txt)

```
import dace
from dace.transformation.interstate import LoopToMap
sdfg = dace.SDFG.from_file("stage1.sdfgz")
sdfg.apply_transformations_repeated(LoopToMap, validate=False)
sdfg.save("parallel.sdfgz", compress=True)
sdfg.validate()
sdfg.save("stage2.sdfgz", compress=True)
```

PR #2177, opened 2025-10-14
LoopToMap Fixes

Berke-Ates:
This PR adds some fixes to the LoopToMap transformation. Addresses the issue #2176.

### DIFF
--- a/tests/transformations/loop_to_map_test.py
+++ b/tests/transformations/loop_to_map_test.py
@@ -550,7 +550,7 @@ def test_thread_local_transient_single_state():
 
     sdfg.apply_transformations_repeated([LoopLifting])
     sdfg.apply_transformations_repeated(LoopToMap)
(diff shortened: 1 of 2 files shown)


tests/transformations/loop_to_map_test.py
```diff
@@ -550,7 +550,7 @@ def test_thread_local_transient_single_state():
 
     sdfg.apply_transformations_repeated([LoopLifting])
     sdfg.apply_transformations_repeated(LoopToMap)
-    assert not ('A' in sdfg.arrays)
+    assert 'A' in sdfg.arrays
```
phschaad: Wait, why this? How does this relate to the fix? If I read the test correctly, I would still assume `A` is removed after that?

Berke-Ates: Good point. Reverted these changes and fixed L2M.

PR #2178, opened 2025-10-15
Added EmptyLoopElimination pass

Berke-Ates:
This PR adds a new simplification pass, which eliminates loops with empty bodies.

### DIFF
--- a/dace/transformation/passes/empty_loop_elimination.py
+++ b/dace/transformation/passes/empty_loop_elimination.py
@@ -1,6 +1,6 @@
 # Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
 
-from typing import Set
+from typing import Set, Optional
 
 from dace import SDFG, InterstateEdge, properties
 from dace.sdfg.state import ControlFlowRegion, LoopRegion, ReturnBlock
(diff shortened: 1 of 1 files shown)


tests/passes/simplification/continue_to_condition_test.py
```diff
@@ -73,7 +73,6 @@ def tester(a: dace.float64[20]):
             a[i] = a[i] + 1
 
     sdfg = tester.to_sdfg(simplify=False)
-    sdfg.simplify(skip=["ContinueToCondition"])
```
phschaad, line 76: Why remove this, wasn't this on purpose?

dace/transformation/passes/empty_loop_elimination.py
```diff
@@ -0,0 +1,69 @@
...
+    CATEGORY: str = 'Simplification'
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.CFG
+
+    def should_reapply(self, modified: ppl.Modifies) -> bool:
+        # If connectivity or any edges were changed, some more loops might be dead
+        return modified & ppl.Modifies.CFG
+
+    def apply_pass(self, sdfg: SDFG, _) -> Optional[Dict[str, Set[str]]]:
```
phschaad: you're not currently returning anything, ever - so the type annotation is incorrect. I'd suggest counting the number of loops removed and returning the count, None if nothing was done

Berke-Ates: I don't think None is necessary as it's the same as returning a zero.

tests/passes/simplification/continue_to_condition_test.py
Berke-Ates: Removed, because the EmptyLoopElimination being part of simplify invalidates this test. But adjusted it now by added EmptyLoopElimination to the skipped transformations.

dace/transformation/passes/empty_loop_elimination.py
phschaad: It may alter behavior if run in a pipeline, we identify if a pass has modified nothing (if it reports that it modifies something) by the fact that it returns `None` as a result

Berke-Ates: Good point. Will adjust accordingly.

PR #2183, opened 2025-10-19
Fix type-punned pointer warning in DaCe runtime

tbennun:


acalotoiu (approved):
LGTM, but I don't understand the benefit - I guess it supresses the warning but there should be no semantic difference

tbennun:
> LGTM, but I don't understand the benefit - I guess it supresses the warning but there should be no semantic difference

@acalotoiu Correct. It just eliminates the annoying warning in all generated codes

PR #2188, opened 2025-10-21
Fix `RedundantSecondArray`

philip-paul-mueller:
This PR fixes a branch in `RedundantSecondArray`, where the copy is turned into a View.
When the View was created the strides were not set correctly.

I also discovered a possible other bug in `RedundantArray` but was unable to find a reproducer I added a note there.


phschaad (approved):
LGTM - the test you added is failing pre-fix, right?

PR #2189, opened 2025-10-21
Added `canonicalize_memlet_trees()`

philip-paul-mueller:
In the DaCe of old [sic] a Memlet only had `.subset` and there was no `.other_subset` and `.data` would always refer to what was on the outside.
However, these days are gone, but as it can be seen for example in [PR#2188](https://github.com/spcl/dace/pull/2188) not every part of the code has been updated/changed/fixed to honor this brave new world.
The `canonicalize_memlet_trees()` function inspects a Map and the Memlets that passing through them and modifies them such that they conform to the old way.

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -751,7 +751,7 @@ with open(r"{temp_path}", "wb") as f:
     def _convert_return_values(self):
         # Return the values as they would be from a Python function
         # NOTE: Currently it is not possible to return a scalar value, see `tests/sdfg/scalar_return.py`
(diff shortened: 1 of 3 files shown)


tests/canonicalize_memlet_tree_test.py
```diff
@@ -0,0 +1,234 @@
+# Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
```

phschaad (approved):
Approved, but please integrate the requested change

PR #2199, opened 2025-10-28
Fallback stuff for memlet propagation

ThrudPrimrose:
Weirdly, the memlet propagation crashes on the valid cloudSC SDFG on a memlet that has src_subset, but the data is None (even though in the visualized SDFG one can clearly see the memlet data)

With this fallback added, it no longer crashes (Honestly, I have no idea why memlet data would be None there).


ThrudPrimrose:
A small test I wrote for map collapse sneaked in, I think it can be merged too having more tests shouldnt be bad.

PR #2201, opened 2025-10-29
Branch Elimination + Construction Utils

ThrudPrimrose:
I am working on auto-vectorization. As part of the auto-vectorization, I have written the Branch Elimination pass, which also uses `construction utilities`. I am trying to make them into the atoms for SDFG manipulation.

I want to discuss the placement of the possible atoms.

I have implemented many helper functions for the passes I have been writing, and I extensively use them as helpers in the `offloading` and `auto-vectorization` passes I have been writing. They have been tested on the CloudSC SDFG, and I will write unit tests for most of them later.

I want to discuss where such atoms should live

I have SDFG-manipulation atoms - these functions aim to provide reusable functions when constructing and changing SDFGs. What I have so far:
 `move_branch_cfg_up_discard_conditions`,  `copy_state_contents`, `copy_graph_contents`, `generate_assignment_as_tasklet_in_state`, `insert_non_transient_data_through_parent_scopes`, `get_parent_map_and_loop_scopes`, `replace_length_one_arrays_with_scalars`

I think they should be in the file `sdfg/construction_utils.py`

Existing string replacements usually fail and error out in bigger SDFGs like `CloudsC` or `VelocityTendencies`
For that, I have multiple tasklet and interesting edge utilities written.

In tasklet utilities, I have stuff that analyzes tasklets and tasklet codes (tasklet blackbox is not possible for some passes):
`classify_tasklet`, which returns information on the code of the tasklet, which is essential for auto-vectorization or implementation of the `BlockedFP` (microscaling) formats. Some of the functions used for tasklets, like `tasklet_has_symbol`, `replace_tasklet_code`, are currently in the construction utilities, but should be moved there.

I think they should go to the file `sdfg/tasklet_utils.py`

For interstate edges, I have ended up writing string utilities: `token_replace_dict`, `token_split_variable_names`, `tasklet_has_symbol`, `replace_symbol`, `extract_bracket_tokens`, and `remove_bracket_tokens`. I guess we can enforce that interstate edges use string expressions that can be parsed into symbolic expressions, and then move these functions to `dace/symbolic.py`. Or should we make these into something like `string_utils.py` or `code_utils.py` ? 

I am also open to comments on which of them are necessary.

The PR is a draft, and I still have tasks to do:
- [ ] Fix imports for `BranchElimination` and `EliminateBranches` 
- [ ] Write tests for `copy_state_contents`
- [ ] Write tests for `copy_graph_contents`
- [ ] Write tests for `move_branch_cfg_up_discard_conditions`
- [ ] Write tests for `insert_non_transient_data_through_parent_scopes`
- [ ] Write tests for `token_replace_dict`, `token_split_variable_names`, `tasklet_has_symbol`, `replace_symbol`, `extract_bracket_tokens`, `remove_bracket_tokens`
- [ ] Write tests for `replace_length_one_arrays_with_scalars`
- [ ] Write tests for `connect_array_names`
- [ ] Write tests for `generate_assignment_as_tasklet_in_state`
- [ ] Write tests for `get_parent_map_and_loop_scopes`


dace/transformation/passes/eliminate_branches.py
```diff
@@ -0,0 +1,71 @@
...
+from dace import properties, SDFG, nodes
+from dace.sdfg.state import ConditionalBlock
+from dace.transformation import pass_pipeline as ppl
+from dace.transformation.transformation import explicit_cf_compatible
+from typing import Union, Optional
+
+
+@properties.make_properties
+@explicit_cf_compatible
+class EliminateBranches(ppl.Pass):
```
Berke-Ates, line 11: I would suggest renaming this pass to `BranchEliminationPass` to avoid confusion between `EliminateBranches` and `BranchElimination`.

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
+import re
```
alexnick83: In b4 @phschaad missing copyright comment

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+
+import sympy
+from sympy import symbols, Function
+
+from sympy.printing.pycode import PythonCodePrinter
+import dace.sdfg.utils as sdutil
+from dace.transformation.passes import FuseStates
+
+
+class BracketFunctionPrinter(PythonCodePrinter):
```
alexnick83: Where is this used?

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+        c_dst = node_map[e.dst]
+        new_graph.add_edge(c_src, c_dst, copy.deepcopy(e.data))
+
+    sdutil.set_nested_sdfg_parent_references(new_graph.sdfg)
+
+    return node_map
+
+
+def move_branch_cfg_up_discard_conditions(if_block: ConditionalBlock, body_to_take: ControlFlowRegion):
+    # Sanity check the ensure apssed arguments are correct
```
alexnick83:
```suggestion
    # Sanity check the ensure passed arguments are correct
```

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+        c_dst = node_map[e.dst]
+        new_state.add_edge(c_src, e.src_conn, c_dst, e.dst_conn, copy.deepcopy(e.data))
+
+    return node_map
+
+
+def copy_graph_contents(old_graph: ControlFlowRegion,
+                        new_graph: ControlFlowRegion) -> Dict[dace.nodes.Node, dace.nodes.Node]:
+    """
+    Deep-copies all nodes and edges from one SDFG state into another.
```
alexnick83, line 52: My understanding is that this method works for ControlFlowRegions in general. Perhaps the docstring should indicate that instead of naming only SDFGStates.

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+    sdutil.set_nested_sdfg_parent_references(new_graph.sdfg)
+
+    return node_map
+
+
+def move_branch_cfg_up_discard_conditions(if_block: ConditionalBlock, body_to_take: ControlFlowRegion):
+    # Sanity check the ensure apssed arguments are correct
+    bodies = {b for _, b in if_block.branches}
+    assert body_to_take in bodies
+    assert isinstance(if_block, ConditionalBlock)
```
alexnick83, line 114: If `if_block` is not a ConditionalBlock, won't the attribute access `if_block.branches` already fail in line 104?

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+
+    graph = if_block.parent_graph
+
+    node_map = dict()
+    # Save end and start blocks for reconnections
+    new_start_block = None
+    new_end_block = None
+
+    for node in body_to_take.nodes():
```
alexnick83, line 123: You introduced utility methods above for deep-copying graphs into other graphs, but they cannot be used here due to the need for special handling. Could we enhance those methods with function parameters for per-node or per-edge special handling to overcome the issue?

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+                                                    parent_graph: 'dace.SDFGState',
+                                                    parent_sdfg: 'dace.SDFG',
+                                                    add_to_output_too: bool = False,
+                                                    add_with_exact_subset: bool = False,
+                                                    exact_subset: Union[None, dace.subsets.Range] = None,
+                                                    nsdfg_connector_name: Union[str, None] = None):
+    """
+    Inserts non-transient data containers into all relevant parent scopes (through all map scopes).
+
+    This function connect data from top-level data
```
alexnick83, line 161: I would consider renaming top-level data to None-scope data (maybe a bad suggestion) or something that cannot be confused with **top-level SDFG** data.

alexnick83: (I was confused)

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+            if isinstance(src, dace.nodes.AccessNode):
+                return None
+            elif isinstance(src, dace.nodes.NestedSDFG):
+                return inner_data_access
+            else:
+                return state['cur_out_conn_name']
+
+        an = parent_graph.add_access(data_access)
+        src = an
+        for it_id, parent_scope in enumerate(reversed(parent_scopes)):
```
alexnick83, line 275: I didn't follow the logic here carefully, but couldn't we simplify the method by employing `add_memlet_path`? I see that there is also logic concerning the connectors, but I believe we also have utility methods/passes that fix them.

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+
+    # Recombine everything
+    return ''.join(tokens).strip()
+
+
+def token_match(string_to_check: str, pattern_str: str) -> str:
+    # Split while keeping delimiters
+    tokens = re.split(r'(\s+|[()\[\]])', string_to_check)
+
+    # Replace tokens that exactly match src
```
alexnick83: Comment seems to have been copied from the previous method but I don't think it makes sense here.

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+                if isinstance(node, dace.nodes.NestedSDFG):
+                    replace_length_one_arrays_with_scalars(node.sdfg, recursive=True, transient_only=True)
+
+
+def connect_array_names(sdfg: dace.SDFG, local_storage: dace.dtypes.StorageType, src_storage: dace.dtypes.StorageType,
+                        local_name_prefix: str):
+
+    array_name_dict = dict()
+    for state in sdfg.all_states():
+        for node in state.nodes():
```
alexnick83: There is a method that iterates over access nodes only; I think `data_nodes`.

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+            if isinstance(node, dace.nodes.AccessNode):
+                local_arr = state.sdfg.arrays[node.data]
+                print(local_arr.storage)
+                if local_arr.storage == local_storage:
+                    assert len(state.in_edges(node)) <= 1
```
alexnick83: Why would this hold?

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+                                array_name_dict[node.data] = oe.data.data
+
+    print(array_name_dict)
+    repldict = {k: f"{local_name_prefix}{v}" for k, v in array_name_dict.items()}
+
+    sdfg.replace_dict(repldict, replace_keys=True)
+    sdfg.validate()
+
+
+def tasklet_has_symbol(tasklet: dace.nodes.Tasklet, symbol_str: str) -> bool:
```
alexnick83: Maybe check the methods related to tasklets, symbols, and code replacement against ScalarToSymbolPromotion's helper methods and code

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,764 @@
...
+    num_parent_maps = 0
+    cur_node = node
+    while scope_dict[cur_node] is not None:
+        if isinstance(scope_dict[cur_node], dace.nodes.MapEntry):
+            num_parent_maps += 1
+        cur_node = scope_dict[cur_node]
+
+    # Check parent nsdfg
+    parent_nsdfg_node = parent_state.sdfg.parent_nsdfg_node
+    parent_nsdfg_parent_state = _find_parent_state(root_sdfg, parent_nsdfg_node)
```
alexnick83: Shouldn't this already be stored in `parent_state.sdfg.parent_state` (or something similar?

dace/sdfg/construction_utils.py
alexnick83: I am also unsure whether they belong in this file. They look to be general utilities, not something specific to SDFG construction.

alexnick83 (commented):
I just reviewed the construction_utils file. I will do the rest later. In general, I like this PR.

dace/sdfg/construction_utils.py
ThrudPrimrose: You are right

dace/sdfg/construction_utils.py
ThrudPrimrose: I was trying some stuff with the python printers, I might have forgotten it.

dace/sdfg/construction_utils.py
ThrudPrimrose: This is correct, that check is redundant.

dace/sdfg/construction_utils.py
ThrudPrimrose: This makes sense, I guess `scopeless` would work here, what do you think?

dace/sdfg/construction_utils.py
ThrudPrimrose: It is possible to pass set to filter nodes that will be copied over?
The default could be None and everything gets copied?

dace/sdfg/construction_utils.py
ThrudPrimrose: I  see the attributed `parent_graph`, `parent_nsdfg_node` and `parent_sdfg`. In this case `parent_graph` should be the same as parent state. I did not know it existed, I will use that.

dace/sdfg/construction_utils.py
ThrudPrimrose: I think tasklet-related and common python-ast related functionality should go to a file called `tasklet_utils`. I think I could move some stuff from `ScalarToSymbol` from that class to the utility file too,

dace/sdfg/construction_utils.py
alexnick83: Or a (lambda) function that returns `True` if the node should be copied. I am not sure what would be the most expressive.

dace/sdfg/construction_utils.py
alexnick83: Technically, they belong to a scope. I would go with a complete description for a start, e.g., "data defined in the current SDFGState/CFG's outermost scope," but it may be a good idea to discuss in a meeting if we need to introduce/clarify (new) terms, put them in the documentation, etc.

dace/codegen/codegen.py
```diff
@@ -172,23 +172,38 @@ def generate_code(sdfg: SDFG, validate=True) -> List[CodeObject]:
         import filecmp
         import shutil
         import tempfile
-        with tempfile.TemporaryDirectory() as tmp_dir:
-            sdfg.save(f'{tmp_dir}/test.sdfg', hash=False)
-            sdfg2 = SDFG.from_file(f'{tmp_dir}/test.sdfg')
-            sdfg2.save(f'{tmp_dir}/test2.sdfg', hash=False)
+        import os
+        with tempfile.NamedTemporaryFile(suffix="_.sdfg", delete=False) as tmp1, \
```
phschaad, line 176: When did this fail / what's the rationale behind the change here? It looks like we're doing essentially the same thing but with different methods and a try/except?

dace/sdfg/construction_utils.py
phschaad: I like the lambda / more complex predicate function idea

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,589 @@
...
+        new_graph.add_edge(c_src, c_dst, copy.deepcopy(e.data))
+
+    sdutil.set_nested_sdfg_parent_references(new_graph.sdfg)
+
+    return node_map
+
+
+def move_branch_cfg_up_discard_conditions(if_block: ConditionalBlock, body_to_take: ControlFlowRegion):
+    """
+    Moves a branch of a conditional block up in the control flow graph (CFG),
```
phschaad, line 91: I think for these methods in general it would be beneficial to have some example in the doc comment. These are going to be central and heavily used functions if we go down the atoms route as intended, so they should be crystal clear to understand. And (maybe that's just me) I was a bit confused first reading the description here as to what CFG gets put into what other CFG from the parameters. (Description is good, but an example would help)

dace/sdfg/construction_utils.py
phschaad: Same comment about the examples then :)

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,589 @@
...
+                if isinstance(node.loop_condition, CodeBlock) else dace.dtypes.Language.Python)
+
+    if recursive:
+        for state in sdfg.all_states():
+            for node in state.nodes():
+                if isinstance(node, dace.nodes.NestedSDFG):
+                    replace_length_one_arrays_with_scalars(node.sdfg, recursive=True, transient_only=True)
+
+
+def generate_assignment_as_tasklet_in_state(state: dace.SDFGState, lhs: str, rhs: str):
```
phschaad, line 431: Needs a doc comment

dace/sdfg/construction_utils.py
```diff
@@ -0,0 +1,589 @@
...
+
+    defined_syms = parent_graph.symbols_defined_at(nsdfg_node)
+    for sym in new_symbols:
+        if str(sym) not in nsdfg_node.sdfg.symbols:
+            nsdfg_node.sdfg.add_symbol(str(sym), defined_syms[str(sym)])
+        if str(sym) not in nsdfg_node.symbol_mapping:
+            nsdfg_node.symbol_mapping[str(sym)] = str(sym)
+
+
+def replace_length_one_arrays_with_scalars(sdfg: dace.SDFG, recursive: bool = True, transient_only: bool = False):
```
phschaad, line 371: Also one here

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,949 @@
...
+import sympy
+import dace
+from typing import Dict, Tuple, Set
+from dace.properties import CodeBlock
+from enum import Enum
+import ast
+import typing
+
+
+class TaskletType(Enum):
```
phschaad, line 20: We're doing a ton of tokenization and analysis of tokens here - why are we not using the AST? What am I missing here? This seems like something that should be doable with ASTs of the respective languages that are set in a tasklet (this manual tokenization is very sensitive to the actual input language). We certainly have the ability to use a pre-existing module for Python tasklets (we already do that in many places), but I would wager that the same exists for Fortran or C/C++ tasklets?

dace/sdfg/utils.py
```diff
@@ -2548,11 +2548,22 @@ def _specialize_scalar_impl(root: 'dace.SDFG', sdfg: 'dace.SDFG', scalar_name: s
     # -> For 2: Rm. dynamic in connector, remove the edge and the node if the degree is None
     # 3. Access Node
     # -> If access node is used then e.g. [scalar] -> [tasklet]
-    # -> then [tasklet(assign const value)] -> [access node] -> [tasklet]
+    # -> then create a [tasklet] that uses the scalar_val as a constant value inside
+    import re
+
+    def _token_replace(code: str, src: str, dst: str) -> str:
```
phschaad, line 2554: Same comment about the tokenization vs. AST - why?

dace/sdfg/utils.py
```diff
@@ -2658,6 +2672,107 @@ def repl_code_block_or_str(input: Union[CodeBlock, str], src: str, dst: str):
...
+        (float, int, str,
+         sympy.Number)), f"Expected scalar value to be float, int, str, or sympy.Number, got {type(scalar_val)}"
+    if not isinstance(scalar_val, (float, int, str)):
+        if isinstance(scalar_val, sympy.Number):
+            scalar_val = _sympy_to_python_number(scalar_val)
+
     _specialize_scalar_impl(sdfg, sdfg, scalar_name, scalar_val)
+
+
+def demote_symbol_to_scalar(sdfg: 'dace.SDFG', symbol_str: str, default_type: 'dace.dtypes.typeclass' = None):
```
phschaad, line 2701: Nice addition :)

dace/transformation/interstate/branch_elimination.py
```diff
@@ -0,0 +1,1985 @@
+import ast
```
phschaad, line 1: 👀  I'm not going to say the words :D

dace/config_schema.yml
```diff
@@ -303,7 +303,7 @@ required:
                         type: str
                         title: nvcc Arguments
                         description: Compiler argument flags for CUDA
-                        default: '-Xcompiler -march=native --use_fast_math -Xcompiler -Wno-unused-parameter'
+                        default: '--expt-relaxed-constexpr -Xcompiler -march=native --use_fast_math -Xcompiler -Wno-unused-parameter'
```
phschaad, line 306: What is the issue when not passing this arg? A warning?

dace/config_schema.yml
```diff
@@ -261,7 +261,7 @@ required:
                         type: str
                         title: Arguments
                         description: Compiler argument flags
-                        default: '-std=c++14 -fPIC -Wall -Wextra -O3 -march=native -ffast-math -Wno-unused-parameter -Wno-unused-label'
+                        default: '-fopenmp -fPIC -Wall -Wextra -O3 -march=native -ffast-math -Wno-unused-parameter -Wno-unused-label'
```
phschaad, line 264: Wait, why is this needed? We already run with OMP.. right?

dace/codegen/codegen.py
ThrudPrimrose: If you run a test multi threaded and different test operate on different sdfgs that have the same sdfg name then this can crash with a key error on 'sdfg'. (Then these files with conflict with each other due to name clashes and be overwritten)

I think using tempfile for temporary files is just nicer.

dace/sdfg/construction_utils.py
ThrudPrimrose: An example is good to add, I agree.
I also think that we need a documentation page that lists all atoms and in which file they are, like FAQ

ThrudPrimrose: Like in the developer guide, "frequently used utilities". And then there we list memset path, memset tree and add mapped tasklets, etc.

dace/sdfg/construction_utils.py
ThrudPrimrose: I will not repeat, I think you are right on asking for comments, examples. Assume I approve all of them ;)

dace/sdfg/utils.py
ThrudPrimrose: So the issue is, certain expressions make sympy crash.
One example:
`(a == 1) == 0`
In sympy a relational object can't be compared against integers. There is a very wide assumption (at least in all fortran SDFGs) that boolean true is equal to one and boolean flame is equal to zero. While this holds for almost all programming languages I know, this does not hold for symbolic math and sympy. This means all the SDFGs I work with have expressions that make sympy crash and I ended up adding a token based approach as fallback.

In my opinion, an SDFG should not involve comparisons to 1s and 0s for booleans (only if we really care about integers). I would really define `true` and false `false` and use boolean support which exists both in Candy C++.

ThrudPrimrose: @acalotoiu I am pulling you in in this one, because I want to ask, the comparison against integers, is it necessary? Does python frontend generate similar stuff too?

ThrudPrimrose: Btw. In the functions I aways try something like this:
1. If Python try to parse into symbolicexpr analyze that (e.g. on interstate edges)
2. If anything fails try token based approach
3. If not python try token based approach (I really really do not want to support CPP tasklets and I would rather crash than implement support for it)

dace/transformation/interstate/branch_elimination.py
ThrudPrimrose: :)

dace/sdfg/utils.py
ThrudPrimrose: This is very needed as the passes that gave a giant headache in icon was scalar to symbol and dead Data flow. So I have demote symbol :)

dace/config_schema.yml
ThrudPrimrose: Actually DaCe needs this flag to compile, because it uses constexpr functions in functions marked as host and device. nvidia compiler is graceful enough to compile it as long we don't really call those functions (it is rare).
I think this happens because unused headers are not necessarily need to be included and compilers optimize this out.

We get a lot of warnings about this everytime we compile GPU code.

About the openmp flag, I have some configuration problem,.openmp flag shouldn't be normally needed

dace/config_schema.yml
ThrudPrimrose: Yes, my I have some config problem in my system, I forgot the flag... Consexpr flag should be default tho imho

dace/sdfg/utils.py
ThrudPrimrose: (I know sometimes dead Data flow breaks because we have an invalid sdfg that has not been caught yet, but it is also does wrong stuff when too much stuff has been symbollified, I think there is a bug in usage check on cfgs/interstate edges)

dace/sdfg/construction_utils.py
phschaad: Fully agree :-)

PR #2202, opened 2025-10-29
Fix issue in printing in type inference

ThrudPrimrose:
Explicit thread block map might result in having `int_floor` and `int_ceil` in the expression.
Which crashes the pycode printer because it is an unknown function for pycode. This enables it. I also added some unit tests.

### DIFF
--- a/tests/safe_call_test.py
+++ b/tests/safe_call_test.py
@@ -5,14 +5,24 @@ import pytest
 
 
 @dace.program
-def indirect_access(A: dace.float64[5], B: dace.float64[5], ub: dace.int64):
-    for i in range(ub):
-        A[i] = B[i] + 1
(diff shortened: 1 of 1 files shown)


tests/transformations/add_threadblock_map_test.py
```diff
@@ -0,0 +1,125 @@
+import copy
```
phschaad, line 2: Missing copyright

phschaad (changes requested):
Small comment, otherwise good

tests/transformations/add_threadblock_map_test.py
ThrudPrimrose: I always forget that... I will add.

phschaad (approved):

PR #2207, opened 2025-10-30
LoopLocalMemoryReduction Symbolic Support

Berke-Ates:
This PR extends the LoopLocalMemoryReduction pass to also support symbolic sizes, symbolic window widths and symbolic indices.

### DIFF
--- a/dace/transformation/passes/loop_local_memory_reduction.py
+++ b/dace/transformation/passes/loop_local_memory_reduction.py
@@ -8,7 +8,7 @@ from dace.sdfg.state import LoopRegion, ConditionalBlock
 from dace.data import Scalar
 from dace.transformation import transformation as xf
 from dace.transformation import pass_pipeline as ppl
-from dace.transformation.passes.analysis import loop_analysis, StateReachability, FindAccessStates
+from dace.transformation.passes.analysis import loop_analysis, StateReachability, FindAccessStates, ConditionUniqueWrites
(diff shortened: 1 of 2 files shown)


tests/passes/loop_local_memory_reduction_test.py
```diff
@@ -507,12 +594,13 @@ def tester(b: dace.float64[16, 16], c: dace.float64[16]):
             for jj in range(4):
                 a[ii, jj] = ii + jj
 
+        # In theory, in this case, LLMR could be applied
```
phschaad: I suggest adding a TODO(later) to track supporting this, or explaining why it is infeasible to support

dace/transformation/passes/loop_local_memory_reduction.py
```diff
@@ -127,21 +132,62 @@ def apply_pass(self, sdfg: sd.SDFG, pipeline_results: Dict[str, Any]) -> Optiona
         else:
             self.access_states = FindAccessStates().apply_pass(sdfg, {})
 
+        # Maps each conditional block to a set of access nodes which are not written to the same location in all paths.
```
phschaad: Hm, maybe this whole added section would be a good idea to have as a separate analysis pass? Because a) we are doing this analysis for the whole SDFG here, and b) this may be useful information for other passes (think of propagation, SSA, etc.). Thoughts?

Berke-Ates: Agreed!

PR #2210, opened 2025-11-02
Is packed Fortran / C layout and contiguous subset utility functions

ThrudPrimrose:
I have two utility functions that I commonly use.

I am opening a PR to see if it makes sense to merge them into main.

- `is_packed_fortran_strides`, `is_packed_c_strides` returns if an array has a `packed` C-Fortran layout
- `is_contiguous_subset`. Run on a subset and a descriptor (array); if the array has a packed-C or packed-Fortran layout, it returns whether the subset would be contiguously stored in 1D memory.

I use these functions a lot in the `memcpy-memset-map-to-libnode` pass to detect whether I can generate something using `memcpy`, and in my layout transformations to help verify the correctness of transformations and their preconditions.

I wrote some unit tests and then extended them into different cases using Claude (I then manually verified correctness tho)

### DIFF
--- a/dace/data.py
+++ b/dace/data.py
@@ -1694,14 +1694,26 @@ class Array(Data):
         """Compute packed strides for Fortran-style (column-major) layout."""
         # Strides increase along the leading dimensions
         if self._packed_fortran_strides is None:
-            self._packed_fortran_strides = (1, ) + tuple(np.cumprod(self.shape[:-1], dtype=int))
+            strides = [1]
+            accum = 1
+            # Iterate in reversed order except the first dimension
+            for s in self.shape[:-1]:
+                accum *= s
+                strides.append(accum)
+            self._packed_fortran_strides = tuple(strides)
         return self._packed_fortran_strides
 
     def _get_packed_c_strides(self) -> Tuple[int]:
         """Compute packed strides for C-style (row-major) layout."""
         # Strides increase along the trailing dimensions
         if self._packed_c_strides is None:
-            self._packed_c_strides = tuple(np.cumprod(self.shape[:0:-1], dtype=int)[::-1]) + (1, )
+            strides = [1]
+            accum = 1
+            # Iterate in reversed order except the first dimension
+            for s in reversed(self.shape[1:]):
+                accum *= s
+                strides.insert(0, accum)
+            self._packed_c_strides = tuple(strides)
         return self._packed_c_strides
 
     def is_packed_fortran_strides(self) -> bool:
--- a/tests/utils/array_dimension_utils.py
+++ b/tests/utils/array_dimension_utils.py
@@ -5,6 +5,8 @@ import pytest
 import tempfile
 import os
 
+N = dace.symbol("N")
+
 
 # Helper function
 def create_array(shape, strides):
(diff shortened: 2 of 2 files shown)


dace/data.py
```diff
@@ -1680,6 +1680,35 @@ def set_shape(
         self._set_shape_dependent_properties(new_shape, strides, total_size, offset)
         self.validate()
 
+    def _get_packed_fortran_strides(self) -> Tuple[int]:
+        """Compute packed strides, if the array is stored Fortran-style (column-major)."""
+        accum = 1
+        strides = []
+        for shape in self.shape:
```
alexnick83: Just noting that you can also write this with something like `return (1, ) + tuple(np.multiply.accumulate(shape[:-1]))`. It should be faster, but unsure if it would be worth the (potential?) decrease in readability.

dace/subsets.py
```diff
@@ -931,6 +931,62 @@ def intersects(self, other: 'Range'):
...
+        Args:
+            array: array descriptor to check against
+
+        Returns:
+            True if the subset is contiguous, False otherwise
+            Returns False on all arrays that are not have a packed layout,
+            meaning that the complete array is contiguously stored in 1D memory.
+        """
+        # Any step size != 1 -> not contiguous
+        for (_, _, s) in self:
```
alexnick83:
```
if any(s != 1 for _, _, s in self):
    return False
```

dace/data.py
alexnick83: I am also wondering whether we should cache those strides instead of recomputing them.

dace/subsets.py
```diff
@@ -931,6 +931,62 @@ def intersects(self, other: 'Range'):
...
+        else:
+            return False
+
+        # Check contiguity: once we find a partial dimension, all remaining must be length 1
+        for i, (expr_len, dim) in enumerate(zip(expr_lens, shape_dims)):
+            try:
+                # Check if this dimension is partial (less than full shape)
+                if expr_len < dim:
+                    # This dimension is partial - all remaining dimensions must be length 1
+                    for j in range(i + 1, len(expr_lens)):
```
alexnick83: You can also write with `if any(...)`

dace/subsets.py
```diff
@@ -931,6 +931,62 @@ def intersects(self, other: 'Range'):
...
+            expr_lens = list(reversed([((e + 1) - b) for (b, e, s) in self]))
+            shape_dims = list(reversed(array.shape))
+        else:
+            return False
+
+        # Check contiguity: once we find a partial dimension, all remaining must be length 1
+        for i, (expr_len, dim) in enumerate(zip(expr_lens, shape_dims)):
+            try:
+                # Check if this dimension is partial (less than full shape)
+                if expr_len < dim:
```
alexnick83: Maybe you can use `expr_len != dim`, which **I think** should always return True or False. That way, you could avoid catching a TypeError and duplicating the remaining-dimensions check.

alexnick83 (approved):
LGTM. Some pedantic comments.

dace/subsets.py
ThrudPrimrose: Your comments makes sense, I will improve the readability by incorporating your suggestions if it doesn't get merged before Im back

ThrudPrimrose: After I'm back*

ThrudPrimrose: I think using any makes the imemntations more readable too.

dace/data.py
ThrudPrimrose: I added cache features. Added tests to check that the cahced values are set correctly when set shape is called and I also added test for the cases where cached values need to be updated.

dace/subsets.py
ThrudPrimrose: Yeah, any is nicer, I will use that.

dace/subsets.py
ThrudPrimrose: I will also directly access the arary elements.

PR #2219, opened 2025-11-05
Remove debug saving of SDFG in split_tasklets_test

affifboudaoud:
Removed saving of SDFG during tests. Or maybe this is necessary, and we should use a tmp file?


ThrudPrimrose:
> Removed saving of SDFG during tests. Or maybe this is necessary, and we should use a tmp file?

I forgot them in code. They should be removed.

ThrudPrimrose (approved):

ThrudPrimrose:
Phil please approve and merge this PR too.

PR #2222, opened 2025-11-06
[CI] Moving GPU tests to CSCS Alps

phschaad:
This PR moves the GPU tests to CSCS infrastructure in an effort to provide more future proofing w.r.t. scalability in the long term, and Pauli congestion mitigation in the short term.

The PR provides a workflow that is automatically triggered on pushes to `main` and PR's with `main` as the target branch. The workflow builds a Docker image, which is cached in the CSCS internal repository until either the corresponding Dockerfile or workflow definition are changed. The image is then used to launch a Slurm job on Alps, reserving a node with 4 GH200s and running 32 parallel pytest workers. This leads to an average workflow completion time of around 20 minutes at the moment.

# Current known issues / things to address:
- It is currently not possible to prevent this job from being launched with the `no-ci` label. Instead, I have set it up that PRs which are marked as draft do not launch the CI workflow, only ones that are not marked as draft. This means that if a PR should not run _any_ CI, it must have _both_ the `no-ci` label, _and_ be marked as draft. If a draft PR is marked as no longer a draft, the workflow can be triggered with a new commit, or with a comment reading just `cscs-ci run` from a trusted user on the PR. I will look into a better solution for this, but for now this has to suffice (TODO(later)).
- All 32 pytest workers appear to currently be utilizing 1 GPU. We should be able to address this with a custom fixture for pytest, but this is a future TODO

### DIFF
--- a/tests/library/gemm_test.py
+++ b/tests/library/gemm_test.py
@@ -173,11 +173,13 @@ def do_test_pure(params):
     impl = 'pure'
     _do_test_gemm(impl, params)
 
+
 @pytest.mark.gpu
 @pytest.mark.parametrize('params', _test_params)
 def do_test_cuBLAS(params):
     impl = 'cuBLAS'
     _do_test_gemm(impl, params)
+
 
 @pytest.mark.mkl
 @pytest.mark.parametrize('params', _test_params)


phschaad:
cscs-ci run GPU

phschaad:
cscs-ci run

tbennun (approved):

PR #2233, opened 2025-11-23
Fix exponential complexity in MapFusionVertical

affifboudaoud:
The `_is_data_accessed_downstream()` function was missing visited node tracking in its graph traversal. Even though the dataflow graph is acyclic, the same node can be reached via multiple paths, causing it to be processed exponentially many times. For example, in a diamond-shaped graph `A → B,C → D`, node D gets visited twice (once via B, once via C). When multiple diamonds are chained together, the visits compound multiplicatively: if D feeds into another diamond `D → E,F → G`, then G is visited 4 times (2 paths to D × 2 paths from D). With n such diamonds in sequence, the final node gets visited `2^n` times. I added a `visited` set to make sure each node is only processed once. This reduces map fusion time from 26+ minutes to seconds on a large test SDFGs.


philip-paul-mueller:
cscs-ci run

PR #2236, opened 2025-11-26
Fix `apply_transformations_once_everywhere`

iomaganaris:
I have noticed that after https://github.com/spcl/dace/pull/2223 `apply_transformations_once_everywhere` doesn't work as expected. The specified transformations are only applied until `can_be_applied` returns `False`. I guess that's not the intended behavior. I have created also a test to showcase the issue. Without commenting out the lines `249-250` in `dace/transformation/passes/pattern_matching.py` the test fails


phschaad:
cscs-ci run

PR #2237, opened 2025-11-26
[Fix | Python Parsing] Allow `MPI.COMM_WORLD` to be used for `Allreduce` replacements

FlorianDeconinck:
The way the implementation of the `Allreduce` worked we assumed that the user would take a reference to the Intracomm instead of calling directly `MPI.COMM_WORLD`. This first option was covered by technically you couldn't reach it because the argument were not marshaled properly.

We flip the reasoning here:
- `@oprepo.replaces('mpi4py.MPI.COMM_WORLD.Reduce')` is the fully defined replacement (with all guardrails)
- `@oprepo.replaces_method('Intracomm', 'Allreduce')`, `@oprepo.replaces_method('ProcessGrid', 'Allreduce')` forward to the above while sanitizing their extra argument if needed.

New unit test covers the fixed `MPI.COMM_WORLD.Allreduce(MPI.IN_PLACE, a, op=MPI.`

Other `MPI` calls might be suffering the same limitation.

### DIFF
--- a/ci/cscs_gpu.yml
+++ b/ci/cscs_gpu.yml
@@ -50,8 +50,10 @@ test_cscs_gh200:
     # running the tests. The output folder is inside the docker image
     # (${WORKDIR}/dace), so it won't take up space on $SCRATCH but in a
     # ephemeral tmpfs mount in the running node.
+    - echo $CSCS_CI_ORIG_CLONE_URL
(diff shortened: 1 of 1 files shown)


FlorianDeconinck:
I am going to go ahead and say that this PR can't reasonably triggers that much failures ?https://github.com/spcl/dace/actions/runs/19716388720/job/56489474468?pr=2237

PR #2251, opened 2025-12-11
Fix in cuda codegen: convert block_size config values to `int` values

edopao:
The values retrieved from `Config` are strings. For block_size configuration, the values need to be converted to `int` type.


ThrudPrimrose (approved):
Reasonable, LGTM.

PR #2253, opened 2025-12-11
Update C++ standard to C++20

tbennun:


ThrudPrimrose (approved):
Almost the same how I would enable C++20, lGTM

PR #2255, opened 2025-12-15
Add support for CUDA 13

zero9178:
The current version of DaCe only supports CUDA 12 since CUDA 13 has a few breaking changes.

This PR updates the runtime headers according to https://nvidia.github.io/cccl/cccl/3.0_migration_guide.html, replacing classes that moved from cub to thrust.

As I have no way of testing on CUDA 12, I have conservatively left the old code intact using preprocessor macros.

Tested on CUDA 13.1


phschaad:
Thank you for the contribution! Let's see if current GPU tests pass with this - they are running with CUDA 12.9.1

PR #2256, opened 2025-12-17
Fix handling of write-write hazard in state fusion

edopao:
This PR addresses an issue in state fusion by making the check for write-write hazard more strict. The issue is an SDFG with non-deterministic code generation, depending on the order of execution of different write edges.

The SDFG on the left is the start SDFG, provided by the new unit test, the one on the right is the result of state fusion on main:
image

The check for connecting path between two nodes, one in first state and and the other one in second state, is not checking if this is the only path writing to the second node. If other edges write to the node in the second state, state fusion cannot always rely on the sequencing assumption provided by a connecting path.

The check introduced in this PR is very conservative, and restricts state fusion to a subset of all valid cases. By only allowing a single writing edge on the second node, this PR blocks state fusion in the `durbin_kernel` SDFG, which contains two write edges to `__return` in the second state, but is indeed a valid case:
image

The condition to be checked, to be on the safe side, should be that all input write edges in the second state have a common source node which is the node on the connecting path.

### DIFF
--- a/dace/codegen/CMakeLists.txt
+++ b/dace/codegen/CMakeLists.txt
@@ -7,6 +7,7 @@ set(DACE_PROGRAM_NAME "dace_program" CACHE STRING "Name of DaCe program")
 set(DACE_SRC_DIR "" CACHE STRING "Root directory of generated code files")
 set(DACE_FILES "" CACHE STRING "List of host code files relative to the root of the source directory")
 set(DACE_LIBS "" CACHE STRING "Extra libraries")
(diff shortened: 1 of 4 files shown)


philip-paul-mueller (approved):
LGTM

edopao:
One autodiff test is failing. All this change does is to prevent applying state fusion, in some cases, because it is more conservative in checking for write-write race conditions. A possible explanation why only this test starts failing is that by skipping state fusion, we hit some issue in another transformation.

PR #2257, opened 2025-12-23
[v1 Maintenance] Back port `MapExpansion` fixes from v2

FlorianDeconinck:
Greetings from the Ol' V1 Codebase,

This PR brings the fixes to `MapExpansion` [transform done in v2](https://github.com/spcl/dace/blob/156567b1eea3b54cd3dda0b6d3f259995127be68/dace/transformation/dataflow/map_expansion.py#L135) back into v1. Fixing expansions when maps have no "true" inputs but only outputs.

Pinky promise, we are coming to v2 soon :tm:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ setup(name='dace',
       },
       include_package_data=True,
       install_requires=[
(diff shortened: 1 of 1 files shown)


FlorianDeconinck:
Networkx wasn't backported - I'll PR that too

FlorianDeconinck:
Soft poke @tbennun, this should be ready to be tucked in.

PR #2260, opened 2026-01-06
Explicit gpu global copies

ThrudPrimrose:
This pass inserts GPU global copies as tasklets for explicit scheduling later.

Since this is a pass aimed at GPU specialization, I decided not use copy nodes. (Note: Right now, copy library nodes do not exist in the main branch, but I have a separate PR upcoming for copy and memset library nodes with a pass that converts memcpy and memset kernels to use these library nodes.


dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        # Store the basic context as attributes
+        self.sdfg = sdfg
+        self.state = state
+        self.src_node = src_node
+        self.dst_node = dst_node
+        self.edge = edge
+
+        memlet = edge.data
+
+        self.copy_shape = memlet.subset.size_exact()
```
alexnick83: Extraneous due to lines 37, 39, and 42?

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        NotImplementedError
+            If a descriptor is not Scalar or Array.
+        """
+
+        # ---------------------------- helpers ----------------------------
+        def _collapse_strides(strides, subset):
+            """Remove size-1 dims; keep tile strides; default to [1] if none remain."""
+            n = len(subset)
+            collapsed = [st for st, sz in zip(strides, subset.size()) if sz != 1]
+            collapsed.extend(strides[n:])  # include tiles
```
alexnick83: What are these tile strides exactly, and why are they in `n:`? This implies that the length of `strides` may be greater than the length of `subset` (`n`), but then how would `zip` in the above line work?

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        - The copy occurs between two AccessNodes
+        - The data descriptors of source and destination are not views.
+        - The storage types of either src or dst is CPU_Pinned or GPU_Device
+        - We do not have a CPU-to-CPU copy
+        """
+        # Retrieve needed information
+        state = copy_context.state
+        src_node, dst_node = copy_context.src_node, copy_context.dst_node
+
+        # 1. Ensure copy is not occuring within a kernel
```
alexnick83: I recall that in another PR, there was a discussion about the `is_devicelevel` helper methods. If the investigated issues have been resolved there, shouldn't you be using them here?

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+
+    def applicable(self, copy_context: CopyContext) -> bool:
+        """
+        Determines whether the data movement is a host<->device memory copy.
+
+        This function returns True if:
+        - We are not currently generating kernel code
+        - The copy occurs between two AccessNodes
+        - The data descriptors of source and destination are not views.
+        - The storage types of either src or dst is CPU_Pinned or GPU_Device
```
alexnick83: GPU global?

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        num_dims = len(copy_context.copy_shape)
+        if num_dims == 1:
+            copy_call = self._generate_1d_copy(copy_context)
+
+        elif num_dims == 2:
+            copy_call = self._generate_2d_copy(copy_context)
+
+        else:
+            # sanity check
+            assert num_dims > 2, f"Expected copy shape with more than 2 dimensions, but got {num_dims}."
```
alexnick83: A bit overzealous; is this supposed to catch `num_dims == 0`?

phschaad:
cscs-ci run

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        src_location, dst_location = copy_context.get_memory_location()
+        is_contiguous_copy = (src_strides[-1] == 1) and (dst_strides[-1] == 1)
+        ctype = copy_context.get_ctype()
+        gpustream = copy_context.get_assigned_gpustream()
+
+        # ----------------- Generate backend call --------------------
+
+        if is_contiguous_copy:
+            # Memory is linear: can use {backend}MemcpyAsync
+            copysize = ' * '.join(sym2cpp(copy_shape))
```
alexnick83: Would it make sense to first compute the product symbolically or numerically, then convert it to C++? The overall expression may potentiallly be simplified but I don't know if you are trying to avoid any issues.

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+            # This allows copying a strided 1D region
+            dpitch = f'{sym2cpp(dst_strides[0])} * sizeof({ctype})'
+            spitch = f'{sym2cpp(src_strides[0])} * sizeof({ctype})'
+            width = f'sizeof({ctype})'
+            height = sym2cpp(copy_shape[0])
+            kind = f'{backend}Memcpy{src_location}To{dst_location}'
+
+            call = f'DACE_GPU_CHECK({backend}Memcpy2DAsync(_out_{dst_expr}, {dpitch}, _in_{src_expr}, {spitch}, {width}, {height}, {kind}, {gpustream}));\n'
+
+        # Potentially snychronization required if syncdebug is set to true in configurations
```
alexnick83:
```suggestion
        # Potentially synchronization required if syncdebug is set to true in configurations
```

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+            width = f'sizeof({ctype})'
+            height = sym2cpp(copy_shape[0] * copy_shape[1])
+            kind = f'{backend}Memcpy{src_location}To{dst_location}'
+
+            call = f'DACE_GPU_CHECK({backend}Memcpy2DAsync(_out_{dst_expr}, {dpitch}, _in_{src_expr}, {spitch}, {width}, {height}, {kind}, {gpustream}));\n'
+
+        else:
+            raise NotImplementedError(
+                f"Unsupported 2D memory copy: shape={copy_shape}, src_strides={src_strides}, dst_strides={dst_strides}."
+                "Please implement this case if it is valid, or raise a more descriptive error if this path should not be taken."
```
alexnick83: It is a bit strange to tell the user to raise a more descriptive error.

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+            )
+
+        # Potentially snychronization required if syncdebug is set to true in configurations
```
alexnick83:
```suggestion
        # Potentially synchronization required if syncdebug is set to true in configurations
```

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        call = call + generate_sync_debug_call()
+        return call
+
+    def _generate_nd_copy(self, copy_context: CopyContext) -> None:
+        """
+        Generates GPU code for copying N-dimensional arrays using 2D memory copies.
+
+        Uses {backend}Memcpy2DAsync for the last two dimensions, with nested loops
```
alexnick83: The description makes it sound like this does not handle all ND copies. Isn't this an issue since this is the "fallback" copy method? Maybe this is not relevant, i.e., exotic copies should be generated in the first place with a mapped tasklet?

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        # Due to applicable(), src and dst node must be AccessNodes
+        copy_shape, src_strides, dst_strides, src_expr, dst_expr = copy_context.get_accessnode_to_accessnode_copy_info()
+
+        src_location, dst_location = copy_context.get_memory_location()
+        ctype = copy_context.get_ctype()
+        gpustream = copy_context.get_assigned_gpustream()
+        num_dims = len(copy_shape)
+
+        # ----------- Guard for unsupported Pattern --------------
+        if not (src_strides[-1] == 1) and (dst_strides[-1] == 1):
```
alexnick83: So for general ND copies, NotImplementedError will be raised. Shouldn't this be checked in `OutOfKernelStrategy.applicable` so that potentially another strategy is selected instead?

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
```diff
@@ -0,0 +1,525 @@
...
+        dpitch = f'{sym2cpp(dst_strides[-2])} * sizeof({ctype})'
+        spitch = f'{sym2cpp(src_strides[-2])} * sizeof({ctype})'
+        width = f'{sym2cpp(copy_shape[-1])} * sizeof({ctype})'
+        height = sym2cpp(copy_shape[-2])
+        kind = f'{backend}Memcpy{src_location}To{dst_location}'
+
+        # Generate call and write it
+        call += f'DACE_GPU_CHECK({backend}Memcpy2DAsync(_out_{dst}, {dpitch}, _in_{src}, {spitch}, {width}, {height}, {kind}, {gpustream}));\n'
+
+        # Potentially snychronization required if syncdebug is set to true in configurations
```
alexnick83:
```suggestion
        # Potentially synchronization required if syncdebug is set to true in configurations
```

dace/transformation/passes/gpu_specialization/insert_explicit_gpu_global_memory_copies.py
```diff
@@ -0,0 +1,254 @@
...
+
+            copy_context = CopyContext(copy_sdfg, state, src_node, dst_node, edge)
+
+            # Only insert copy tasklets for GPU related copies occuring out of the
+            # kernel (i.e. a GPU_device scheduled map)
+            if not out_of_kernel_copy.applicable(copy_context):
+                continue
+
+            # If the subset has more than 2 dimensions and is not contiguous (represented as a 1D memcpy) then fallback to a copy kernel
+            if len(edge.data.subset) > 2 and not edge.data.subset.is_contiguous_subset(
```
alexnick83: Why isn't this check in `OutOfKernelCopyStrategy.applicable`?

dace/dtypes.py
```diff
@@ -71,6 +71,7 @@ class ScheduleType(AutoNumberEnum):
     GPU_ThreadBlock = ()  #: Thread-block code
     GPU_ThreadBlock_Dynamic = ()  #: Allows rescheduling work within a block
     GPU_Persistent = ()
+    GPU_Warp = ()
```
alexnick83: I believe we need this, but is it actually used anywhere in this PR?

alexnick83 (changes requested):
Very good work. Some comments and questions follow.

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: Yes, you are correct should be GPU global

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: I guess it is confusing, we try what we can using Memcpy2D, but for exotic stuff I just generate maps, I will update accordingly

dace/dtypes.py
ThrudPrimrose: It is not utilized for the copies, but it is part of the new GPU codegen, I potentially had error when porting dtypes from the GPU codegen branch to this branch.

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: True, I will update.

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: I think it was solved, I will use devicelevel gpu and re-run unit tests.

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: It is some workaround around `src_subset`, `dst_subset` and `other_susbet` shenaningans.
Techinically it should be always matching in size, I update accordingly to:
```
        def _collapse_strides(strides, subset):
            assert len(strides) == len(subset.size())
            collapsed = [st for st, sz in zip(strides, subset.size()) if sz != 1]
            return collapsed or [1]
```

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: I think any compiler can perform the same optimization, and we don't need additional changes (also considering existing bugs in Sympy printer that I can't fix because of some remnants relying on the buggy behavior)

dace/transformation/passes/gpu_specialization/helpers/copy_strategies.py
ThrudPrimrose: I updated comments.

dace/transformation/passes/gpu_specialization/insert_explicit_gpu_global_memory_copies.py
ThrudPrimrose: not applicable -> skip entirely (not a GPU out-of-kernel copy)
applicable + >2D non-contiguous -> fallback mapped tasklet
applicable + expressible as memcpy -> generate memcpy code

For the if branch at 135 we have an else 170. Both cases are applicable but require different patterns.

PR #1466, opened 2023-12-03
Schedule Trees (2/3): Tree-to-SDFG conversion

tbennun:
- [x] Keep descriptor repository in schedule tree and use in reconstruction
- [x] Find state boundaries in stree
- [x] Generic input/output memlet recovery for stree nodes
- [x] Memlet set and propagation for in/out memlet recovery for stree scopes
- [ ] Generate states or empty/happens-before memlets
- [ ] Convert control-flow blocks
- [ ] Convert leaf tree nodes
- [ ] Convert dataflow blocks
    - [ ] Create nested SDFGs as necessary (i.e., if state boundaries appear in a dataflow block, depending on behavior)

PR #1145, opened 2022-11-08
Schedule tree (1/3)

tbennun:
(re)scheduling oriented view of SDFGs.
Task list:
- [x] De-aliasing across nested SDFGs
    - [x] Arrays and memlets (use `compose`)
    - [x] Memlets and symbols in control flow blocks
    - [x] Symbol assignments
        - [x] (low-prio: think about C++ tasklets, or error out / replace if `\bNAME\b` is used?)
- [x] Normalize memlets everywhere (data / subset / other_subset)
    - [x] Copy
        - [x] If both exist, write as `dst[dst_subset] = copy src[src_subset]`
    - [x] View
- [x] Dynamic map ranges (also with aliasing)
- [x] Ensure memlet paths are not generated twice
- [x] Duplicate label removal (also consider states in different SDFGs that share the same name!)
- [x] Make many tests
    - [x] Irreducible CFGs
    - [x] Memlet path types 
        - [x] same scope
        - [x] multi-scope (3 nested maps)
        - [x] through nested SDFG
        - [x] differing src/dst subsets: same scope
        - [x] differing src/dst subsets: inside map
        - [x] offsetted
    - [x] Nesting and (de)aliasing 
        - [x] Call `nested(a[0:20], a[10:30])`
        - [x] `prog(a, b), nested(b, c)`
        - [x] Composition: `a[:, 1] -> b[-5:] -> c[2]`
        - [x] Control flow blocks with same variables in nested SDFGs
    - [x] Code->code memlets
    - [x] Dynamic map ranges
    - [x] References
    - [x] Naming clashes (`n: m, m: n` symbol mapping, different states with same name etc.)

PR #2262, opened 2026-01-09
Schedule Tree to SDFG Translation

romanc:
## Description

Schedule trees are an alternative representation of code in DaCe. It is a tree of nodes that represent the execution order of the SDFG. The tree can be used to perform schedule transformations on the SDFG, i.e., erasing an empty `if` branch, or merging two consecutive `for`-loops.

A first version of the schedule tree representation was merged with PR https://github.com/spcl/dace/pull/1145, which adds support for creating schedule trees from SDFGs. This PR builds on https://github.com/spcl/dace/pull/1466 and brings a first version of the transformation from schedule tree to SDFG.


dace/transformation/passes/dead_dataflow_elimination.py
```diff
@@ -282,6 +282,13 @@ def _is_node_dead(self, node: nodes.Node, sdfg: SDFG, state: SDFGState, dead_nod
                             and any(ie.data.data == node.data for ie in state.in_edges(l.src))):
                         return False
 
+                    # If data is connected to a Tasklet
+                    if isinstance(l.src, nodes.Tasklet):
+                        # We can't inline connected data that is a pointer
+                        ctype = infer_types.infer_out_connector_type(sdfg, state, l.src, l.src_conn)
+                        if l.src.language == dtypes.Language.Python and isinstance(ctype, dtypes.pointer):
+                            return False
```
tbennun: Can you explain this?

tests/sdfg/memlet_utils_test.py
```diff
@@ -211,6 +210,62 @@ def test_non_lin_delin_8():
     _perform_non_lin_delin_test(sdfg, e)
 
 
+def test_MemletSet() -> None:
```
tbennun: name capitalization

tests/sdfg/memlet_utils_test.py
```diff
@@ -211,6 +210,62 @@ def test_non_lin_delin_8():
...
+
+    union = empty_set.union(dace.Memlet("A[0:3]"), dace.Memlet("A[2:10]"))
+    assert dace.Memlet("A[5:7]") not in empty_set
+    assert dace.Memlet("A[5:7]") in union
+    assert len(union.internal_set["A"]) == 1
+    internal_memlet = list(union.internal_set["A"])[0]
+    assert internal_memlet.subset == dace.subsets.Range.from_string("0:10")
+
+
+def test_MemletDict() -> None:
```
tbennun: same
```suggestion
def test_memlet_dict() -> None:
```

dace/sdfg/analysis/schedule_tree/treenodes.py
```diff
@@ -427,3 +956,98 @@ def validate_has_no_other_node_types(stree: ScheduleTreeScope) -> None:
...
+    of type `ScheduleTreeRoot`.
+
+    :param stree: Schedule tree scope to be analyzed
+    :param root: If true, we raise if the top-most scope isn't of type `ScheduleTreeRoot`.
+    """
+    if root and not isinstance(stree, ScheduleTreeRoot):
+        raise RuntimeError("Expected schedule tree root.")
+
+    for child in stree.children:
+        if id(child.parent) != id(stree):
```
tbennun: should use `child.parent is not stree` - better and has similar effect

dace/sdfg/analysis/schedule_tree/treenodes.py
```diff
@@ -19,6 +23,49 @@ class UnsupportedScopeException(Exception):
...
+    def __exit__(
+        self,
+        exc_type: Optional[type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        cache_key = (self._state, id(self._node))
+        assert cache_key in self._ctx.access_cache
+
+        self._ctx.current_scope = self._parent_scope
```
tbennun: what are these classes? They are missing docstrings and their purpose

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        # state management
+        self._state_stack: List[SDFGState] = []
+
+        # dataflow scopes
+        # List[ (MapEntryNode, ToConnect) | (SDFG, {"inputs": set(), "outputs": set()}) ]
+        self._dataflow_stack: List[Tuple[nodes.EntryNode, Dict[str, Tuple[nodes.AccessNode, Memlet]]]
+                                   | Tuple[SDFG, Dict[str, Set[str]]]] = []
+
+    def _apply_nview_array_override(self, array_name: str, sdfg: SDFG) -> bool:
+        """Apply an NView override if applicable. Returns true if the NView was applied."""
```
tbennun: can you explain what a nested sdfg view override is? also in the docstring

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                return True
+
+        return False
+
+    def _parent_sdfg_with_array(self, name: str, sdfg: SDFG) -> SDFG:
+        """Find the closest parent SDFG containing an array with the given name."""
+        parent_sdfg = sdfg.parent.parent
+        sdfg_counter = 1
+        while name not in parent_sdfg.arrays and sdfg_counter < MAX_NESTED_SDFGS:
+            parent_sdfg = parent_sdfg.parent.parent
```
tbennun: for control flow blocks, don't you want to use `.parent_sdfg`, `.sdfg` or `.parent_graph`? I feel like .parent.parent is unsafe here.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+                # This NView is in use now, remove it from the free NViews.
+                del self._nviews_free[length - index]
+                return True
+
+        return False
+
+    def _parent_sdfg_with_array(self, name: str, sdfg: SDFG) -> SDFG:
+        """Find the closest parent SDFG containing an array with the given name."""
+        parent_sdfg = sdfg.parent.parent
```
tbennun: see below

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        self._ctx = tn.Context(root=node, access_cache={}, current_scope=None)
+        with node.scope(self._current_state, self._ctx):
+            self.visit(node.children, sdfg=sdfg)
+
+        assert not self._state_stack, "Expected empty state stack."
+        assert not self._dataflow_stack, "Expected empty dataflow stack."
+        assert not self._interstate_symbols, "Expected empty list of symbols to add."
+
+    def visit_GBlock(self, node: tn.GBlock, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 178: Shouldn't this be a no-op in SDFG? You could add a NamedRegion if you want, or simply nothing. Just a no-op.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def visit_StateLabel(self, node: tn.StateLabel, sdfg: SDFG) -> None:
```
tbennun, line 180: shouldn't this have already been removed/addressed in the preprocessing passes that make inter-state edges? Leave a comment if so (and especially if not!)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_GotoNode(self, node: tn.GotoNode, sdfg: SDFG) -> None:
```
tbennun, line 183: this should become a simple inter-state edge

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        raise NotImplementedError(f"Support for{type(node)} not yet implemented.")
+
+    def visit_AssignNode(self, node: tn.AssignNode, sdfg: SDFG) -> None:
+        # We just collect them here. They'll be added when state boundaries are added,
+        # see visitors below.
+        self._interstate_symbols.append(node)
+
+        # If AssignNode depends on arrays, e.g. `my_sym = my_array[__k] > 0`, make sure array accesses can be resolved.
+        input_memlets = node.input_memlets()
```
tbennun, line 192: does `input_memlets()` do something proactively? Otherwise the comment above makes no sense

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                            sdfg.add_datadesc(memlet.data, parent_sdfg.arrays[memlet.data].clone())
+
+                            # Transients passed into a nested SDFG become non-transient inside that nested SDFG
+                            if parent_sdfg.arrays[memlet.data].transient:
+                                sdfg.arrays[memlet.data].transient = False
+                                # TODO
+                                # ... unless they are only ever used inside the nested SDFG, in which case
+                                # we should delete them from the parent SDFG's array list.
+                                # NOTE This can probably be done automatically by a cleanup pass in the end.
+                                #      Something like DDE should be able to do this.
```
tbennun: This should already be done by RefineNestedAccess. Assume this is done somewhere else in dace and feel free to remove this comment whenever it appears.

dace/serialize.py
```diff
@@ -211,7 +211,7 @@ def load(*args, context=None, **kwargs):
 
 
 def dump(*args, **kwargs):
-    return json.dump(*args, default=to_json, indent=2, **kwargs)
+    return json.dump(*args, default=to_json, indent=None, separators=(',', ':'), **kwargs)
```
tbennun, line 218: would rather if there was a flag to make readable SDFGs

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+    STATE_TRANSITION = auto()  #: Creates multiple states with a state transition
+    EMPTY_MEMLET = auto()  #: Happens-before empty memlet edges in the same state
+
+
+PREFIX_PASSTHROUGH_IN: Final[str] = "IN_"
+PREFIX_PASSTHROUGH_OUT: Final[str] = "OUT_"
+MAX_NESTED_SDFGS: Final[int] = 1000
+
+
+class StreeToSDFG(tn.ScheduleNodeVisitor):
```
tbennun: 1. no docstring
2. is this intended as public API? if not, maybe call it `_STreeToSDFG`

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        self._current_state = loop_state
+        self.visit(node.children, sdfg=sdfg)
+
+        after_state = _insert_and_split_assignments(loop_region, label="loop_after")
+        self._current_state = after_state
+
+    def visit_DoWhileScope(self, node: tn.DoWhileScope, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_LoopScope(self, node: tn.LoopScope, sdfg: SDFG) -> None:
```
tbennun, line 269: what is this corresponding to?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        self._current_state = conditional_block
+
+        # add merge_state
+        merge_state = _insert_and_split_assignments(
+            conditional_block,
+            label="merge_state",
+            assignments=self._pending_interstate_assignments(),
+        )
+
+        # Check if there's an `ElseScope` following this node (in the parent's children).
```
tbennun, line 303: shouldn't this be a preprocessing nodetransformer pass? it might be unsafe to do it while converting

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+        if has_else_branch:
+            # push merge_state on the stack for later usage in `visit_ElseScope`
+            self._state_stack.append(merge_state)
+            # push condition_block on the stack for later usage in `visit_ElseScope`
+            self._state_stack.append(conditional_block)
+        else:
+            self._current_state = merge_state
+
+    def visit_StateIfScope(self, node: tn.StateIfScope, sdfg: SDFG) -> None:
```
tbennun, line 317: isn't this a simple state transition condition?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_BreakNode(self, node: tn.BreakNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_ContinueNode(self, node: tn.ContinueNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_ElifScope(self, node: tn.ElifScope, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 327: see comment above about preprocessing pass - a pass can transform the node structure so you have easier-to-work-with structures

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def visit_ElseScope(self, node: tn.ElseScope, sdfg: SDFG) -> None:
+        # get ConditionalBlock form stack
```
tbennun:
```suggestion
        # get ConditionalBlock from stack
```

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
tbennun: looks like this implementation should be almost an exact copy of the Else scope below, but with a condition. Maybe reuse the code here.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+        # visit children inside the else branch
+        self.visit(node.children, sdfg=sdfg)
+
+        # merge false-branch into merge_state
+        merge_state = self._pop_state("merge_state")
+        self._current_state = merge_state
+
+        if self._pending_interstate_assignments():
+            raise NotImplementedError("TODO: update edge with new assignments")
```
tbennun, line 347: Is this difficult? You have both states here and can reuse methods

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def _insert_nestedSDFG(self, node: tn.MapScope, sdfg: SDFG) -> None:
```
tbennun: it's not just a nested SDFG, it's an SDFG nested in a map scope. Rename method to reflect this please

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        inner_sdfg = SDFG("nested_sdfg", parent=self._current_state)
+        start_state = inner_sdfg.add_state("nested_root", is_start_block=True)
+
+        # update stacks and current state
+        old_state_label = self._current_state.label
+        self._state_stack.append(self._current_state)
+        self._dataflow_stack.append((inner_sdfg, {"inputs": set(), "outputs": set()}))
+        self._nviews_bound_per_scope[id(inner_sdfg)] = []
+        self._nviews_deferred_removal[id(inner_sdfg)] = []
+        self._current_nestedSDFG = id(inner_sdfg)
```
tbennun, line 364: I don't know if I like dictionaries with id() keys. I use it for duplicate references but I am not sure if you should not just use the index in sdfg_list here instead? but then you did not add it yet to the current state, so you cannot use that. Is there any other way?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+from typing import Dict, Final, List, Optional, Set, Tuple
+
+
+class StateBoundaryBehavior(Enum):
+    STATE_TRANSITION = auto()  #: Creates multiple states with a state transition
+    EMPTY_MEMLET = auto()  #: Happens-before empty memlet edges in the same state
+
+
+PREFIX_PASSTHROUGH_IN: Final[str] = "IN_"
+PREFIX_PASSTHROUGH_OUT: Final[str] = "OUT_"
```
tbennun, line 25: why?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+MAX_NESTED_SDFGS: Final[int] = 1000
```
tbennun: why impose a limitation at all?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        nsdfg = self._current_state.add_nested_sdfg(sdfg=inner_sdfg,
+                                                    inputs=connectors["inputs"],
+                                                    outputs=connectors["outputs"])
+        # connect nested SDFG to surrounding map scope
+        assert self._dataflow_stack
+        map_entry, to_connect = self._dataflow_stack[-1]
+
+        # connect nsdfg input memlets (to be propagated upon completion of the SDFG)
+        for name in nsdfg.in_connectors:
+            out_connector = f"{PREFIX_PASSTHROUGH_OUT}{name}"
```
tbennun, line 389: this is longer than writing `OUT_` :)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            new_in_connector = map_entry.add_in_connector(f"{PREFIX_PASSTHROUGH_IN}{name}")
```
tbennun, line 390: same

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            new_out_connector = map_entry.add_out_connector(out_connector)
+            assert new_in_connector == True
+            assert new_in_connector == new_out_connector
+
+            # Add Memlet for NView node (if applicable)
+            edge_added = False
+            for nview in self._nviews_bound_per_scope[id(inner_sdfg)]:
```
tbennun, line 397: following the above id() comment, is it really necessary to keep a dictionary here rather than a stack? You only ever look at the nviews of the current SDFG w.r.t. its immediate parent

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            self._current_state = merge_state
+
+    def visit_StateIfScope(self, node: tn.StateIfScope, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_BreakNode(self, node: tn.BreakNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_ContinueNode(self, node: tn.ContinueNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 324: super easy to implement these two nodes: just create the right state/control flow block.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                    self._current_state.add_edge(map_entry, out_connector, nsdfg, name,
+                                                 Memlet.from_memlet(nview.memlet))
+                    edge_added = True
+                    break
+
+            if not edge_added:
+                self._current_state.add_edge(map_entry, out_connector, nsdfg, name,
+                                             Memlet.from_array(name, nsdfg.sdfg.arrays[name]))
+
+        # Add empty memlet if we didn't add any in the loop above
```
tbennun, line 408: you need to do the same if there are no outputs too (empty memlet from nsdfg to map exit).
Same applies when you add tasklets

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        self._current_state.add_node(map_entry)
+        self._dataflow_stack.append((map_entry, dict()))
+
+        # visit children inside the map
+        type_of_children = [type(child) for child in node.children]
+        last_child_is_MapScope = type_of_children[-1] == tn.MapScope
+        all_others_are_Boundaries = type_of_children.count(tn.StateBoundaryNode) == len(type_of_children) - 1
+        if last_child_is_MapScope and all_others_are_Boundaries:
+            # skip weirdly added StateBoundaryNode
+            # tmp: use this - for now - to "backprop-insert" extra state boundaries for nested SDFGs
```
tbennun, line 456: what does this mean?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        map_entry = nodes.MapEntry(node.node.map)
+        self._current_state.add_node(map_entry)
+        self._dataflow_stack.append((map_entry, dict()))
+
+        # visit children inside the map
+        type_of_children = [type(child) for child in node.children]
+        last_child_is_MapScope = type_of_children[-1] == tn.MapScope
+        all_others_are_Boundaries = type_of_children.count(tn.StateBoundaryNode) == len(type_of_children) - 1
+        if last_child_is_MapScope and all_others_are_Boundaries:
+            # skip weirdly added StateBoundaryNode
```
tbennun, line 455: add a very simple transformer pass to eliminate those (if `map { state boundary; map {}}`). I don't like this special case

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        # Restore current nested SDFG
+        self._current_nestedSDFG = outer_nestedSDFG
+
+    def visit_MapScope(self, node: tn.MapScope, sdfg: SDFG) -> None:
+        dataflow_stack_size = len(self._dataflow_stack)
+        cache_state = self._current_state
+
+        # map entry
+        # ---------
+        map_entry = nodes.MapEntry(node.node.map)
```
tbennun, line 446: so you don't want to use `add_map()` because the `Map` object already exists?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                            sdfg.add_datadesc(memlet_data, parent_sdfg.arrays[memlet_data].clone())
+
+                            # Transients passed into a nested SDFG become non-transient inside that nested SDFG
+                            if parent_sdfg.arrays[memlet_data].transient:
+                                sdfg.arrays[memlet_data].transient = False
+                                # TODO
+                                # ... unless they are only ever used inside the nested SDFG, in which case
+                                # we should delete them from the parent SDFG's array list.
+                                # NOTE This can probably be done automatically by a cleanup pass in the end.
+                                #      Something like DDE should be able to do this.
```
tbennun: yes, this should be a post-processing pass, so the comment should be removed and a TODO on a pass needs to be added instead. (see RefinedNestedAccess)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                # cache local read access
+                assert memlet_data not in access_cache
+                access_cache[memlet_data] = self._current_state.add_read(memlet_data)
+                cached_access = access_cache[memlet_data]
+                self._current_state.add_memlet_path(cached_access,
+                                                    map_entry,
+                                                    dst_conn=connector,
+                                                    memlet=Memlet.from_array(memlet_data, sdfg.arrays[memlet_data]))
+
+        if isinstance(outer_map_entry, nodes.EntryNode) and self._current_state.out_degree(outer_map_entry) < 1:
```
tbennun, line 530: again, remember to add empty memlets in the other direction too.
A test would be nice

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            self._current_state.add_nedge(outer_map_entry, map_entry, Memlet())
+
+        # map_exit
+        # --------
+        map_exit = nodes.MapExit(node.node.map)
```
tbennun, line 535: no support for dynamic input maps? Also, a test would be nice

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+            if isinstance(outer_map_entry, nodes.EntryNode):
+                outer_to_connect[name] = (access_node, Memlet.from_array(name, sdfg.arrays[name]))
+            else:
+                assert isinstance(outer_map_entry, SDFG) or outer_map_entry is None
+
+        # TODO If nothing is connected at this point, figure out what's the last thing that
+        #      we should connect to. Then, add an empty memlet from that last thing to this
+        #      map_exit.
+        assert len(self._current_state.in_edges(map_exit)) > 0
```
tbennun, line 609: see my above comment. it's ok if every sink node adds an empty memlet.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def visit_ConsumeScope(self, node: tn.ConsumeScope, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 612: this is funny, but you can reuse the _exact_ same implementation as MapScope, just use node.consume and ConsumeEntry/ConsumeExit. Can you please refactor the other one to visit dataflow scope?
Also, a test would be nice.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            if memlet.data in cache:
+                cached_access = cache[memlet.data]
+                self._current_state.add_memlet_path(cached_access, tasklet, dst_conn=name, memlet=memlet)
+                continue
+
+            if isinstance(scope_node, nodes.MapEntry):
+                # get it from outside the map
+                connector_name = f"{PREFIX_PASSTHROUGH_OUT}{memlet.data}"
+                if connector_name not in scope_node.out_connectors:
+                    new_in_connector = scope_node.add_in_connector(f"{PREFIX_PASSTHROUGH_IN}{memlet.data}")
```
tbennun, line 637: wouldn't this connector already be added in  visit_MapScope?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                        sdfg.add_datadesc(memlet.data, parent_sdfg.arrays[memlet.data].clone())
+
+                        # Transients passed into a nested SDFG become non-transient inside that nested SDFG
+                        if parent_sdfg.arrays[memlet.data].transient:
+                            sdfg.arrays[memlet.data].transient = False
+                            # TODO
+                            # ... unless they are only ever used inside the nested SDFG, in which case
+                            # we should delete them from the parent SDFG's array list.
+                            # NOTE This can probably be done automatically by a cleanup pass in the end.
+                            #      Something like DDE should be able to do this.
```
tbennun: same comment on RefineNestedAccess

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            # A[1] = tasklet(A[1])
+            if memlet.data not in cache or self._current_state.out_degree(cache[memlet.data]) > 0:
+                # cache write access node
+                write_access_node = self._current_state.add_write(memlet.data)
+                cache[memlet.data] = write_access_node
+
+            access_node = cache[memlet.data]
+            self._current_state.add_memlet_path(tasklet, access_node, src_conn=name, memlet=memlet)
+
+            if isinstance(scope_node, nodes.MapEntry):
```
tbennun, line 687: a lot of this looks like a copy of visit_MapScope. What is going on here?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+                # Add out_connector in any case if not yet present, e.g. write after read
+                # Dev note: memlet.data and nview.target are identical
+                to_connect["outputs"].add(memlet.data)
+
+            else:
+                assert scope_node is None
+
+    def visit_LibraryCall(self, node: tn.LibraryCall, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 713: same as TaskletNode, but with a LibraryNode instead. Could you please refactor the method and add a test please? Thank you!

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+            self._ctx.access_cache[cache_key] = {}
+        access_cache = self._ctx.access_cache[cache_key]
+
+        # assumption source access may or may not yet exist (in this state)
+        src_name = node.memlet.data
+        source = access_cache[src_name] if src_name in access_cache else self._current_state.add_read(src_name)
+
+        # assumption: target access node doesn't exist yet
+        assert node.target not in access_cache
+        target = self._current_state.add_write(node.target)
```
tbennun, line 728: I am not sure that is a valid assumption. Try it in the cache first.
I am thinking of a case where there are two copies to the same array but different memlets. Please make a simple test (two lines)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+        self._current_state.add_memlet_path(source, target, memlet=node.memlet)
+
+    def visit_DynScopeCopyNode(self, node: tn.DynScopeCopyNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 733: This always precedes a dataflow scope directly. You can keep a list of those you found and apply them in visit_MapScope above. Should be simple 🤞

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    def visit_ViewNode(self, node: tn.ViewNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 736: shouldn't this just modify the access cache? simply add the target to the cache.
Not sure how this works with reads and writes. Maybe we can add two tests and skip them as necessary for another PR.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        # Basic working principle:
+        #
+        # - NView and (artificial) NViewEnd nodes are added in parallel to mark the region where the view applies.
+        # - Keep a stack of NView nodes (per name) that is pushed/popped when NView and NViewEnd nodes are visited.
+        # - In between, when going "down into" a NestedSDFG, use the current NView (if it applies)
+        # - In between, when "coming back up" from a NestedSDFG, pop the NView from the stack.
+        # - AccessNodes will automatically pick up the right name (from the NestedSDFG's array list)
+        self._nviews_free.append(node)
+
+    def visit_NViewEnd(self, node: tn.NViewEnd, sdfg: SDFG) -> None:
```
tbennun, line 748: What is NViewEnd? I don't see any pass inserting the end scope of an nview.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        for index, nview in enumerate(reversed(self._nviews_free), start=1):
+            if node.target == nview.target:
+                # Stack semantics: remove from the back of the list
+                del self._nviews_free[length - index]
+                return
+
+        raise RuntimeError(f"No matching NView found for target {node.target} in {self._nviews_free}.")
+
+    def visit_RefSetNode(self, node: tn.RefSetNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 768: same as ViewNode, add a `x->(set)y` graph and modify access cache. Here though you don't need to modify the access cache for writes, as opposed to views, so it is simpler to implement. I am hoping that the state boundary node addition takes care of writes. A simple test should clarify everything.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+        _insert_and_split_assignments(current_state, loop_region)
+
+        self._current_state = loop_state
+        self.visit(node.children, sdfg=sdfg)
+
+        after_state = _insert_and_split_assignments(loop_region, label="loop_after")
+        self._current_state = after_state
+
+    def visit_DoWhileScope(self, node: tn.DoWhileScope, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
```
tbennun, line 267: I am not sure why you have 3 different loop scopes here, since the ControlFlowRegion of a loop scope has `inverted` etc. to capture `do-while`, `while`, and `for`

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                del self._nviews_free[length - index]
+                return
+
+        raise RuntimeError(f"No matching NView found for target {node.target} in {self._nviews_free}.")
+
+    def visit_RefSetNode(self, node: tn.RefSetNode, sdfg: SDFG) -> None:
+        raise NotImplementedError(f"Support for {type(node)} not yet implemented.")
+
+    def visit_StateBoundaryNode(self, node: tn.StateBoundaryNode, sdfg: SDFG) -> None:
+        # When creating a state boundary, include all inter-state assignments that precede it.
```
tbennun, line 771: looks like we actually have no way to control default state boundary behavior.
I would add some field to the init ctor (`self.state_boundary_behavior`) with a default to STATE_TRANSITION. The other one is not implemented so an exception can be raised.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+    # Setup SDFG descriptor repository
+    result = SDFG(stree.name, propagate=False)
+    result.arg_names = copy.deepcopy(stree.arg_names)
+    for key, container in stree.containers.items():
+        result._arrays[key] = copy.deepcopy(container)
+    result.constants_prop = copy.deepcopy(stree.constants)
+    result.symbols = copy.deepcopy(stree.symbols)
+
+    # Insert artificial state boundaries after WAW, before label, etc.
+    stree = insert_state_boundaries_to_tree(stree)
```
tbennun: see above comments on potentially cleaning up branches

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    # Traverse tree and incrementally build SDFG, finally propagate memlets
+    StreeToSDFG().visit(stree, sdfg=result)
```
tbennun:
```suggestion
    StreeToSDFG(state_boundary_behavior).visit(stree, sdfg=result)
```

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+    propagation.propagate_memlets_sdfg(result)
```
tbennun, line 819: see above comments on post-processing passes (simplify already takes care of making non-transient data containers into transient if not used outside. We are not calling simplify by default so no need to do anything here and no need for any TODO comments).

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
tbennun: I'm wondering if there is anything else that could be made simpler with preproc passes (e.g., consolidating all dynscope assignments with their map nodes; "inlining" GBlocks, which have no SDFG equivalent; consolidating the nview mappings so you don't need to track dicts with id() keys)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+                node_index = _list_index(scope.parent.children, scope)
+
+                # If there's already a state boundary before the map, don't add another one
+                if node_index > 0 and isinstance(scope.parent.children[node_index - 1], tn.StateBoundaryNode):
+                    return visited
+
+                return [tn.StateBoundaryNode(), visited]
+            return visited
+
+    stree = NestedSDFGStateBoundaryInserter().visit(stree)
```
tbennun, line 893: oh dear, needs a RefSet state boundary inserter

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+    """
+    Creates a boundary between two states
+
+    :param boundary_node: The state boundary node to generate.
+    :param state: The last state prior to this boundary.
+    :param behavior: The state boundary behavior with which to create the boundary.
+    :return: The newly created state.
+    """
+    if behavior != StateBoundaryBehavior.STATE_TRANSITION:
+        raise NotImplementedError("Only STATE_TRANSITION is supported as StateBoundaryBehavior in this prototype.")
```
tbennun:
```suggestion
        raise NotImplementedError("Only STATE_TRANSITION is currently supported as StateBoundaryBehavior.")
```

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    # TODO: Some boundaries (control flow, state labels with goto) could not be fulfilled with every
```
tbennun:
```suggestion
    # TODO: Some boundaries (control flow, state labels with goto, pending assignments) could not be fulfilled with every
```

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+    assignments = assignments if assignments is not None else {}
+    return _insert_and_split_assignments(state, label=label, assignments=assignments)
+
+
+def _insert_and_split_assignments(
+    before_state: ControlFlowBlock,
+    after_state: Optional[ControlFlowBlock] = None,
+    *,
+    label: Optional[str] = None,
+    assignments: Optional[Dict] = None,
```
tbennun: type hint incomplete (key, value?)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1007 @@
...
+
+    DaCe validation (currently) won't let us add multiple assignment with read after
+    write pattern on the same edge. We thus split them over multiple state transitions
+    (inserting empty states in between) to be safe.
+
+    NOTE (later) This should be double-checked since python dictionaries preserve
+                 insertion order since python 3.7 (which we rely on in this function
+                 too). Depending on code generation it could(TM) be that we can
+                 weaken (best case remove) the corresponding check from the sdfg
+                 validator.
```
tbennun: The check on the validator is on purpose (and we have seen cases where transformations trigger race conditions, e.g., in special cases where constant propagation modifies something). The semantics of the SDFG should be clear in that you should not make assignments that can cause a cycle or assume any order in the application of inter-state edge assignments. The only order is that conditions precede assignments.

dace/serialize.py
tbennun: rebase on main please

tbennun (changes requested):
Looking great! I would like to see it more feature complete (as it is very close to being!) and I left some comments on the code. Maybe more preprocessing passes can make the implementations simpler

dace/transformation/passes/dead_dataflow_elimination.py
romanc: Good find, that shouldn't be anymore - bad merge. The issue got resolved in mainline with https://github.com/spcl/dace/pull/2098/.

dace/serialize.py
romanc: rebased

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: `NViewEnd` nodes are inserted in `sdfg_to_tree` when creating the tree. I found that when a nested SDFG with `NView` nodes is inlined into the stree, there's no way of knowing where the inlined nested SDFG ends and thus where the effect `NView` ends. I thus oped to explicitly add these nodes, see

https://github.com/spcl/dace/pull/2262/changes#diff-6454a828c0b26fcb8b1d616db3ee75ff0c61f6fdb45f0a8b2a4aaffad3b9ca5dR603-R608.

Also see `visit_NView` above for how I keep track of nested SDFG view nodes.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: I think it's good practice. Allows me to use the in/out passthrough prefix without having to look up if it's one or two underscores and/or if there's an underscore at the beginning or not.

Regarding the length of `"IN_"` and `"PREFIX_PASSTHROUGH_IN"`, I think we all have big enough screens and aren't limited to 70 (or so) characters per line anymore.

I can change if you insist, but I'd rather not.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: I think it's good practice and it dramatically helps debugging. Realistically, there's a limitation at some point because of limited resources of any computer parsing programs. If you think 1000 is too low, I'm happy to increase the number.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: `input_memlets()` returns memlets associated with this `AssignNode`. In case of no input memlets we are done here. If there are input memlets, we make sure that the respective data descriptor is available or - if not  - gets added to the nested SDFG. (see the rest of the function)

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: I am aware that there is only one general loop region in CFGs. We would like to keep at last `ForScope` and `WhileScope` to keep as much context easily available in the schedule tree representation.

Our use-case for specialized loop nodes is optimization. We have schedule tree level optimizations that look for "for loops", which are easily recognizable as `ForScope` nodes in this proposal. Of course, this could also work with a general `LoopScope` because from the CFG `LoopScope` object we can derive a unique "loop variant" (see `treenodes.py` / `loop_variant()`). However, figuring out the loop variant is non-trivial and we think it would be a nice perk of the schedule tree to do that out of the box. I've added OOTB-support for `ForScope` and `WhileScope` because these are the only ones that we support in user-code (and this PR is really just a "let's start by building exactly what we need and nothing more").

I think having different loop scopes as first citizen will strengthen the schedule tree.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: This visitor corresponds to any loop scope that isn't specialized. In the current proposal, `LoopScope` variants `'do-for-uncond-increment'` and `'do-for'` will end up here because only `'for'`, `'while'` and `'do-while'` are generated as special scopes (see changes to the creation of `LoopRegion`s in sdfg_to_stree.py and the specialized visitors above).

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: This was built before complete transition to CFGs and it shows now. I should be able to if/elif/else in something like  a `ConditionalScope` containing the whole `ConditionalBlock` with all branches. Then we have all the information needed in one place and don't need to "look forward" to see if there are more branches coming.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: No, not difficult, just a todo left from moving fast and coding only codepaths that were explicitly needed.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: let's talk at the top where the constants are defined.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: let's talk at the top where the constants are defined

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: What is it you don't like about dictionaries with `id()` keys? Honest question - there might be something else we can do. This is just the first thing that solved my problem at the time. Things might change around here anyway

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: what are dynamic input maps?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
tbennun: Maybe make the number configurable in that case

dace/sdfg/analysis/schedule_tree/treenodes.py
romanc: They are only used in the stree -> sdfg transformation and manage some context information. I've moved them to `tree_to_sdfg.py` and added some comments.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: See `visit_NView()` for how we keep track of nested SDFG view nodes. I've left a comment here.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: This is from back when I initially wrote the bridge and we didn't yet use control flow graphs.

Please elaborate on how `.parent_sdfg` are `.parent_graph` are supposed to be used and how they differ. I am missing a mental picture of how to use them. `.parent_graph` has no docstring and I wasn't able to find anything in the docs.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: removed

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: Is there an easy way to get all the sink nodes inside the map? Like, how can I say "give me all the nodes between MapEntry / MapExit with out degree = 0"?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: No. I'm adding accesses (and thus scope node connectors) from the inside to the outside. Connectors are added on demand when accesses inside the map need them.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: The code structure for a mapped tasklet basically is

```
// add MapEntry to state

// visit all children of the map
    // add Tasklet to the state

// add MapExit to the state
```

Because of this structure, I can directly connect _reads_ to the `MapEntry` node, but I can't yet connect _writes_. I thus keep a list, `to_connect`, with things to connect once I'm adding the `MapExit` node.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: The assumption is guarded. We'll run into the `assert` once that case arises.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: I see. I've reformulated the comment to reflect that.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: What do you mean by that? Can we do this later?

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: done

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: done

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
tbennun: As far as I knowknow, parent_graph can refer to a control flow block that the node is in, whereas parent_sdfg skips all the control flow levels until the next sdfg

tbennun:
cscs-ci run

tbennun:
cscs-ci run

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1052 @@
...
+
+                # This NView is in use now, remove it from the free NViews.
+                del self._nviews_free[length - index]
+                return True
+
+        return False
+
+    def _parent_sdfg_with_array(self, name: str, sdfg: SDFG) -> SDFG:
+        """Find the closest parent SDFG containing an array with the given name."""
+        parent_sdfg = sdfg.parent.parent
```
tbennun:
```suggestion
        parent_sdfg = sdfg.parent.parent_sdfg
```

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -0,0 +1,1052 @@
...
+        sdfg_counter = 1
+        while name not in parent_sdfg.arrays and sdfg_counter < self._max_nested_sdfg:
+            parent_sdfg = parent_sdfg.parent.parent
```
tbennun:
```suggestion
            parent_sdfg = parent_sdfg.parent.parent_sdfg
```

tbennun (changes requested):

tbennun (approved):
LGTM for now and we will keep the PR comments alive to fix them

tbennun:
cscs-ci run

PR #2269, opened 2026-01-12
Add `#pragma unroll <X>` to `LoopRegion` generated code

iomaganaris:
Allow the compiler to unroll itself the `LoopRegion`s instead of just having a manual unroll feature with the `LoopUnroll` pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ of performance optimization, regardless of the application or the target process
 DaCe generates high-performance programs for:
  * Multi-core CPUs (tested on Intel, IBM POWER9, and ARM with SVE)
  * NVIDIA GPUs and AMD GPUs (with HIP)
- * Xilinx and Intel FPGAs
+ * [Xilinx and Intel FPGAs](https://github.com/spcl/dace-fpga)
 
 DaCe can be written inline in Python and transformed in the command-line/Jupyter
 Notebooks or SDFGs can be interactively modified using our [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=phschaad.sdfv).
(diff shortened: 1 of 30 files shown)


dace/sdfg/state.py
```diff
@@ -3180,6 +3180,12 @@ class LoopRegion(ControlFlowRegion):
                                        'do-while style into a while(true) with a break before the update (at the end ' +
                                        'of an iteration) if the condition no longer holds.')
     loop_variable = Property(dtype=str, default='', desc='The loop variable, if given')
+    unroll = Property(dtype=bool,
+                      default=False,
+                      desc='If True, indicates that this loop should be unrolled during code generation.')
+    unroll_factor = Property(dtype=int,
+                             default=1,
+                             desc='If unrolling is enabled, the factor by which to unroll the loop.')
```
tbennun, line 3188: I would make the default 0, so that it can fully unroll a loop by default

dace/sdfg/state.py
```diff
@@ -3189,7 +3195,9 @@ def __init__(self,
                  update_expr: Optional[Union[str, CodeBlock]] = None,
                  inverted: bool = False,
                  sdfg: Optional['SDFG'] = None,
-                 update_before_condition=True):
+                 update_before_condition=True,
+                 unroll: bool = False,
+                 unroll_factor: int = 1):
```
tbennun:
```suggestion
                 unroll_factor: int = 0):
```

dace/sdfg/state.py
iomaganaris: Good point 👍

PR #2270, opened 2026-01-12
Add HIP-codegen for stream-order memory allocation

edopao:
Although the GPU codegen is being reworked, this small change would be useful for the time being. It a an extension to the HIP codegen of the current solution for stream-order memory allocation. It has been tested in GT4Py.


edopao:
cscs-ci run

PR #2280, opened 2026-01-28
Add Tasklet Classification

ThrudPrimrose:
The analysis returns a dictionary containing the left- and right-hand-side operands of a given tasklet. It returns the operator (or function), the constants involved, and whether they are data (array or scalar) or constants (hardcoded constants or symbols), etc.

I have implemented the analysis for the auto-vectorization to determine whether to use vector-vector intrinsics or vector-scalar intrinsics, etc. It only accepts tasklets with at most 2 rhs operands. SplitTasklets that I have merged before can be used to ensure that all supported operators and a subset of functions (such as min, sin, cos, etc.) are used. Python tasklets have 2 rhs operands.

The analysis was also useful for a student trying to implement blocked-FP numbers in DaCe, so I think it will have further use beyond auto-vectorization.

### DIFF
--- a/dace/sdfg/tasklet_utils.py
+++ b/dace/sdfg/tasklet_utils.py
@@ -1,0 +1,963 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
+"""
+Tasklet Classification Utilities
+
+This module provides utilities for analyzing and classifying DaCe tasklets based on their
+computational patterns. It parses tasklet code to determine the types of operations, operands,
+and constants involved. It also provides utilities furhter manipulate and analyze tasklets.
+"""
+
+import re
+import sympy
+import dace
+from typing import Dict, Tuple, Set
+from dace.properties import CodeBlock
+from enum import Enum
+import ast
+import typing
+
+
+class TaskletType(Enum):
+    """
+    Enumeration of supported tasklet computational patterns.
+
+    Each pattern represents a specific combination of input types (arrays, scalars, symbols)
+    and operation types (assignment, binary operation, unary operation).
+
+    Note: inside a tasklet you always have scalars, it is about he connector types
+    Assignment Operations:
+        ARRAY_ARRAY_ASSIGNMENT: Direct array-to-array copy (e.g., a = b)
+        ARRAY_SYMBOL_ASSIGNMENT: Symbol/constant assignment to array (e.g., a = sym)
+        ARRAY_SCALAR_ASSIGNMENT: Scalar variable assignment to array (e.g., a = scl)
+        SCALAR_ARRAY_ASSIGNMENT: Array assignment to scalar variable (e.g., scl = a)
+        SCALAR_SCALAR_ASSIGNMENT: Scalar assignment to scalar variable (e.g., scl = scl)
+
+    Binary Operations with Arrays:
+        ARRAY_SYMBOL: Array with symbol/constant (e.g., out = arr + 5, out = arr * N)
+        ARRAY_SCALAR: Array with scalar variable (e.g., out = arr + scl)
+        ARRAY_ARRAY: Two arrays (e.g., out = arr1 + arr2)
+
+    Binary Operations with Scalars/Symbols:
+        SCALAR_SYMBOL: Scalar with symbol/constant (e.g., out = scl + 5)
+        SCALAR_SCALAR: Two scalars (e.g., out = scl1 + scl2)
+        SYMBOL_SYMBOL: Two symbols (e.g., out = sym1 + sym2)
+
+    Unary Operations:
+        UNARY_ARRAY: Single array operand (e.g., out = abs(arr), out = arr * arr)
+        UNARY_SCALAR: Single scalar operand (e.g., out = abs(scl), out = scl * scl)
+        UNARY_SYMBOL: Single symbol operand (e.g., out = abs(sym), out = sym * sym)
+    """
+    ARRAY_ARRAY_ASSIGNMENT = "array_array_assignment"
+    ARRAY_SYMBOL_ASSIGNMENT = "array_symbol_assignment"
+    ARRAY_SCALAR_ASSIGNMENT = "array_scalar_assignment"
+    SCALAR_ARRAY_ASSIGNMENT = "scalar_array_assignment"
+    SCALAR_SCALAR_ASSIGNMENT = "scalar_scalar_assignment"
+    SCALAR_SYMBOL_ASSIGNMENT = "scalar_symbol_assignment"
+    SCALAR_SYMBOL = "scalar_symbol"
+    ARRAY_SYMBOL = "array_symbol"
+    ARRAY_SCALAR = "array_scalar"
+    SCALAR_ARRAY = "scalar_array"
+    ARRAY_ARRAY = "array_array"
+    UNARY_ARRAY = "unary_array"
+    UNARY_SYMBOL = "unary_symbol"
+    UNARY_SCALAR = "unary_scalar"
+    SCALAR_SCALAR = "scalar_scalar"
+    SYMBOL_SYMBOL = "symbol_symbol"
+
+
+def _token_split(string_to_check: str) -> Set[str]:
+    """
+    Splits a string into a set of tokens, keeping delimiters, and returns all tokens.
+    The input string is split on empty space and brackets (` `, `(`, `)`, `[`, `]`).
+
+    Parameters
+    ----------
+    string_to_check : str
+        The string to split into tokens.
+    pattern_str : str
+        (Unused in this function, kept for consistency with token_match)
+
+    Returns
+    -------
+    Set[str]
+        The set of tokens extracted from the string.
+    """
+    # Split while keeping delimiters
+    tokens = re.split(r'(\s+|[()\[\]])', string_to_check)
+
+    # Replace tokens that exactly match src
+    tokens = {token.strip() for token in tokens}
+
+    return tokens
+
+
+def replace_code(code_str: str,
+                 code_lang: dace.dtypes.Language,
+                 repldict: Dict[str, str],
+                 py_only: bool = False,
+                 use_sym_expr: bool = True) -> str:
+    """
+    Replaces variables in a code string according to a replacement dictionary.
+    Supports Python symbolic substitution and fallback string-based replacement.
+
+    Parameters
+    ----------
+    code_str : str
+        The code string to modify.
+    code_lang : dace.dtypes.Language
+        The programming language of the code.
+    repldict : Dict[str, str]
+        Mapping from variable names to their replacements.
+    py_only: bool
+        Replace only if python tasklet
+    Returns
+    -------
+    str
+        The modified code string with replacements applied.
+    """
+
+    def _str_replace(lhs: str, rhs: str) -> str:
+        code_str = token_replace_dict(rhs, repldict)
+        return f"{lhs.strip()} = {code_str.strip()}"
+
+    def _str_replace_nsplit(code_str: str) -> str:
+        new_code_str = token_replace_dict(code_str, repldict)
+        return new_code_str
+
+    if code_lang == dace.dtypes.Language.Python and use_sym_expr:
+        try:
+            lhs, rhs = code_str.split(" = ")
+            lhs = lhs.strip()
+            rhs = rhs.strip()
+        except Exception as e:
+            try:
+                new_rhs_sym_expr = dace.symbolic.SymExpr(code_str).subs(repldict)
+                cleaned_expr = sympy.pycode(new_rhs_sym_expr, allow_unknown_functions=True).strip()
+                return f"{cleaned_expr}"
+            except Exception as e:
+                return code_str
+        try:
+            new_rhs_sym_expr = dace.symbolic.SymExpr(rhs).subs(repldict)
+            cleaned_expr = sympy.pycode(new_rhs_sym_expr, allow_unknown_functions=True).strip()
+            return f"{lhs.strip()} = {cleaned_expr}"
+        except Exception as e:
+            return _str_replace(lhs, rhs)
+    else:
+        if not py_only:
+            try:
+                lhs, rhs = code_str.split(" = ")
+                lhs = lhs.strip()
+                rhs = rhs.strip()
+                return _str_replace(lhs, rhs)
+            except Exception as e:
+                try:
+                    return _str_replace_nsplit(code_str)
+                except Exception as e:
+                    return code_str
+        else:
+            assert False
+
+
+def tasklet_replace_code(tasklet: dace.nodes.Tasklet,
+                         repldict: Dict[str, str],
+                         py_only: bool = True,
+                         use_sym_expr: bool = True):
+    """
+    Replaces symbols in a tasklet's code according to a replacement dictionary.
+    Updates the tasklet's code in place.
+
+    Parameters
+    ----------
+    tasklet : dace.nodes.Tasklet
+        The tasklet whose code to modify.
+    repldict : Dict[str, str]
+        Mapping from variable names to their replacements.
+
+    Returns
+    -------
+    None
+    """
+    new_code = replace_code(tasklet.code.as_string,
+                            tasklet.code.language,
+                            repldict,
+                            py_only=py_only,
+                            use_sym_expr=use_sym_expr)
+    tasklet.code = CodeBlock(code=new_code, language=tasklet.code.language)
+
+
+def extract_bracket_tokens(s: str) -> list[tuple[str, list[str]]]:
+    """
+    Extracts all contents inside [...] along with the token before the '[' as the name.
+
+    Args:
+        s (str): Input string.
+
+    Returns:
+        List of tuples: [(name_token, string inside brackes)]
+    """
+    results = []
+
+    # Pattern to match <name>[content_inside]
+    pattern = re.compile(r'(\b\w+)\[([^\]]*?)\]')
+
+    for match in pattern.finditer(s):
+        name = match.group(1)  # token before '['
+        content = match.group(2).split()  # split content inside brackets into tokens
+
+        results.append((name, " ".join(content)))
+
+    return {k: v for (k, v) in results}
+
+
+def remove_bracket_tokens(s: str) -> str:
+    """
+    Removes all [...] patterns from the string.
+
+    Args:
+        s (str): Input string.
+
+    Returns:
+        str: String with all [...] removed.
+    """
+    return re.sub(r'\[.*?\]', '', s)
+
+
+def _extract_constant_from_ast_str(src: str) -> str:
+    """
+    Extract a numeric constant from a Python code string using AST parsing.
+
+    Supports both direct constants (e.g., 42, 3.14) and unary operations on constants
+    (e.g., -5, +3.14). The function walks the AST tree to find constant nodes.
+
+    Args:
+        src: Python code string containing a constant (e.g., "x + 3.14" or "y - (-5)")
+
+    Returns:
+        String representation of the constant value
+
+    Raises:
+        ValueError: If no constant is found in the source string
+
+    Examples:
+        >>> _extract_constant_from_ast_str("x + 3.14")
+        '3.14'
+        >>> _extract_constant_from_ast_str("y + (-5)")
+        '-5'
+    """
+    tree = ast.parse(src)
+
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Constant):
+            return str(node.value)
+        elif isinstance(node, ast.UnaryOp) and isinstance(node.operand, ast.Constant):
+            if isinstance(node.op, ast.USub):
+                return f"-{node.operand.value}"
+            elif isinstance(node.op, ast.UAdd):
+                return str(node.operand.value)
+
+    raise ValueError("No constant found")
+
+
+def _split_code_on_assignment(code_str: str) -> Tuple[str, str]:
+    """
+    Returns the LHS and RHS of the first assignment in a Python tasklet.
+
+    Args:
+        node: A Python tasklet node.
+
+    Returns:
+        A tuple (lhs_str, rhs_str) where both are strings representing
+        the left-hand side and right-hand side of the first assignment.
+    """
+    # Parse the tasklet code into an AST
+    code_ast = ast.parse(code_str)
+
+    # Find the first assignment statement
+    assign_node = next((n for n in code_ast.body if isinstance(n, ast.Assign)), None)
+    if assign_node is None:
+        raise ValueError("No assignment found in tasklet code.")
+
+    # Convert LHS to string
+    lhs_node = assign_node.targets[0]  # handle simple assignments only
+    lhs_str = ast.unparse(lhs_node).strip()
+
+    # Convert RHS to string
+    rhs_node = assign_node.value
+    rhs_str = ast.unparse(rhs_node).strip()
+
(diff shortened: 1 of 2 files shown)


dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,963 @@
+# Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,963 @@
...
+    py_only: bool
+        Replace only if python tasklet
+    Returns
+    -------
+    str
+        The modified code string with replacements applied.
+    """
+
+    def _str_replace(lhs: str, rhs: str) -> str:
+        code_str = token_replace_dict(rhs, repldict)
```
phschaad: where does this function come from? I can't find a definition or import?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,963 @@
...
+                else:
+                    c1 = bound_syms.pop()
+                    ttype = TaskletType.UNARY_SYMBOL
+                    info_dict.update({"type": ttype, "constant1": c1, "constant2": None, "op": op})
+                    return info_dict
+
+    raise NotImplementedError("Unhandled case in detect tasklet type")
+
+
+import ast
```
phschaad: already imported at the top, no?

phschaad (changes requested):
Some questions and minor corrections - after that will take closer look at remaining parts

dace/sdfg/tasklet_utils.py
ThrudPrimrose: True, I will fix

dace/sdfg/tasklet_utils.py
ThrudPrimrose: I must have accidentally not removed it. Tasklet replace code and replace code should not have been part of this PR. I will fix.

dace/sdfg/tasklet_utils.py
ThrudPrimrose: Fixed

dace/sdfg/tasklet_utils.py
ThrudPrimrose: I removed the function that should not be part of this PR

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+import re
+import dace
+from typing import Dict, Tuple, Set
+from dace.properties import CodeBlock
+from enum import Enum
+import ast
+import typing
+
+
+class TaskletType(Enum):
```
tbennun, line 19: str enum?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    """
+    Enumeration of supported tasklet computational patterns.
+
+    Each pattern represents a specific combination of input types (arrays, scalars, symbols)
+    and operation types (assignment, binary operation, unary operation).
+
+    Note: inside a tasklet you always have scalars, it is about he connector types
```
alexnick83, line 26:
```suggestion
    Note: inside a tasklet you always have scalars, it is about the connector types
```

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    Assignment Operations:
+        ARRAY_ARRAY_ASSIGNMENT: Direct array-to-array copy (e.g., a = b)
+        ARRAY_SYMBOL_ASSIGNMENT: Symbol/constant assignment to array (e.g., a = sym)
+        ARRAY_SCALAR_ASSIGNMENT: Scalar variable assignment to array (e.g., a = scl)
+        SCALAR_ARRAY_ASSIGNMENT: Array assignment to scalar variable (e.g., scl = a)
```
alexnick83, line 31: Is this a valid case? Even when, e.g., you assign A[i] to a scalar b, isn't the source connector a scalar?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+def _token_split(string_to_check: str) -> Set[str]:
+    """
+    Splits a string into a set of tokens, keeping delimiters, and returns all tokens.
+    The input string is split on empty space and brackets (` `, `(`, `)`, `[`, `]`).
+
+    Parameters
+    ----------
+    string_to_check : str
+        The string to split into tokens.
+    pattern_str : str
```
alexnick83, line 76: But the parameter is not actually kept in the signature

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+
+    Returns
+    -------
+    Set[str]
+        The set of tokens extracted from the string.
+    """
+    # Split while keeping delimiters
+    tokens = re.split(r'(\s+|[()\[\]])', string_to_check)
+
+    # Replace tokens that exactly match src
```
alexnick83, line 87: Comment doesn't make sense (copied from `token_match`?)

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        >>> _extract_constant_from_ast_str("y + (-5)")
+        '-5'
+    """
+    tree = ast.parse(src)
+
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Constant):
+            return str(node.value)
+        elif isinstance(node, ast.UnaryOp) and isinstance(node.operand, ast.Constant):
+            if isinstance(node.op, ast.USub):
```
alexnick83, line 121: Do we need the bitwise inversion? In that case, I would evaluate the result and convert it to a string.

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    ARRAY_ARRAY = "array_array"
+    UNARY_ARRAY = "unary_array"
+    UNARY_SYMBOL = "unary_symbol"
+    UNARY_SCALAR = "unary_scalar"
+    SCALAR_SCALAR = "scalar_scalar"
+    SYMBOL_SYMBOL = "symbol_symbol"
+
+
+def _token_split(string_to_check: str) -> Set[str]:
+    """
```
tbennun, line 68: not the correct (Sphinx) docstring format we use in the rest of dace (apply everywhere)

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    Splits a string into a set of tokens, keeping delimiters, and returns all tokens.
+    The input string is split on empty space and brackets (` `, `(`, `)`, `[`, `]`).
```
tbennun, line 70: double backticks (``) for code comments

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    tree = ast.parse(src)
+
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Constant):
+            return str(node.value)
+        elif isinstance(node, ast.UnaryOp) and isinstance(node.operand, ast.Constant):
+            if isinstance(node.op, ast.USub):
+                return f"-{node.operand.value}"
+            elif isinstance(node.op, ast.UAdd):
+                return str(node.operand.value)
```
tbennun, line 124: why not `+value`?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    rhs_node = assign_node.value
+    rhs_str = ast.unparse(rhs_node).strip()
+
+    assert isinstance(lhs_str, str)
+    assert isinstance(rhs_str, str)
+
+    return lhs_str, rhs_str
+
+
+def _extract_non_connector_syms_from_tasklet(node: dace.nodes.Tasklet, state) -> typing.Set[str]:
```
tbennun, line 162: missing type hints and direct inclusion of typing.
Since dace is Python 3.10+, use `set[str]` instead of typing

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    """
+    Identify free symbols in tasklet code that are not input/output connectors.
+
+    This function extracts all symbolic variables from the right-hand side of the tasklet's
+    code expression and filters out those that correspond to input/output connectors,
+    leaving only the actual free symbols (e.g., SDFG symbols or constants).
+
+    Args:
```
tbennun, line 170: first parameters, now args. Please harmonize docstrings

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+
+        # For compatibility with Python <3.8
+        def visit_Num(self, n):  # type: ignore
+            constants.append(n.n)
+
+        def visit_Str(self, n):  # type: ignore
+            constants.append(n.s)
+
+        def visit_NameConstant(self, n):  # type: ignore
+            constants.append(n.value)
```
tbennun, line 226: no need

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        for s in dace.symbolic.symbols_in_code(code_rhs,
+                                               potential_symbols={str(s)
+                                                                  for s in state.symbols_defined_at(node)})
+    }
+    real_free_syms = all_syms - connectors
+    free_non_connector_syms = {str(s) for s in real_free_syms}
+    return free_non_connector_syms
+
+
+def _extract_non_connector_bound_syms_from_tasklet(code_str: str) -> typing.Set[str]:
```
tbennun, line 199: isn't this the same as the other function?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    from a Python AST node.
+
+    Args:
+        node (ast.AST): The AST node or subtree to traverse.
+
+    Returns:
+        List of constants (int, float, str, bool, None, etc.)
+    """
+    constants = []
+    node = ast.parse(code_str, mode="exec")
```
tbennun, line 211: please explain why mode=exec

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        op = None
+
+        if isinstance(node, ast.BinOp):
+            op = _BINOP_SYMBOLS.get(type(node.op), None)
+        elif isinstance(node, ast.UnaryOp):
+            op = _UNARY_SYMBOLS.get(type(node.op), None)
+        elif isinstance(node, ast.BoolOp):
+            op = _BINOP_SYMBOLS.get(type(node.op), None)
+        elif isinstance(node, ast.Compare):
+            assert len(node.ops) == 1
```
tbennun, line 337: these should really not be assertions (same comment applies to an assertion that a tasklet language is python above)
1. they are not executed in PYTHONOPTIMIZE=1
2. codes with `a < b < c` exist and these cases should be handled gracefully (not classified...)

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+            print(f"Found unsupported op {op} in {src}")
+
+        if found is not None:
+            raise ValueError(
+                f"More than one supported operation found, first found: {found}, second: {op} for tasklets {node} ({ast.unparse(node)})"
+            )
+
+        found = op
+
+    code_rhs = src.split(" = ")[-1].strip()
```
tbennun, line 358: why the whitespace around `=`. no need

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    try:
```
tbennun, line 359: implicit (wrong) assumption: `a = b = c` never exists. Check len(split).

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        tree = ast.parse(code_rhs, mode="eval")
```
tbennun, line 360: again, please explain

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    # Parse the tasklet code into an AST
+    code_ast = ast.parse(code_str)
+
+    # Find the first assignment statement
+    assign_node = next((n for n in code_ast.body if isinstance(n, ast.Assign)), None)
+    if assign_node is None:
+        raise ValueError("No assignment found in tasklet code.")
+
+    # Convert LHS to string
+    lhs_node = assign_node.targets[0]  # handle simple assignments only
```
tbennun, line 149: need to address the case of len(targets) > 1 by raising a valueerror

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+            constants.append(n.s)
+
+        def visit_NameConstant(self, n):  # type: ignore
+            constants.append(n.value)
+
+    ConstantExtractor().visit(node)
+    return {str(c) for c in constants}
+
+
+_BINOP_SYMBOLS = {
```
alexnick83, line 232: Are the `_X_SYMBOLS` sets supposed to be exhaustive? You are missing, e.g., bitwise inversion (~)

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        state: The SDFG state containing the tasklet
+        tasklet: The tasklet node whose connectors to map
+
+    Returns:
+        Dictionary mapping connector names (str) to data descriptors (dace.data.Data)
+
+    """
+    tdict = dict()
+    for ie in state.in_edges(tasklet):
+        if ie.data is not None:
```
tbennun, line 394: ie.data.is_empty() is also important

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+            tdict[ie.dst_conn] = state.sdfg.arrays[ie.data.data]
```
tbennun, line 395: what if there is no destination connector (like dependency edges?) check.

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        rhs2: Second operand name
+
+    Returns:
+        Tuple of (left_operand, right_operand) in the order they appear in the code
+
+    Note:
+        For function calls, uses AST parsing to extract arguments in order.
+        For operators, splits the code by the operator symbol.
+    """
+    code_rhs = code_str.split(" = ")[-1].strip()
```
tbennun, line 434: again, no whitespace please and also no assumptions on number of = signs

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    if op not in _SUPPORTED_OPS:
+        try:
+            tree = ast.parse(code_rhs, mode="eval")
+            call_node = tree.body
+            if not isinstance(call_node, ast.Call):
+                raise ValueError(f"Expected a function call in expression: {code_rhs}")
+
+            args = [ast.get_source_segment(code_rhs, arg).strip() for arg in call_node.args]
+            left_string, right_string = args[0:2]
+            assert len(args) == 2
```
tbennun, line 444: same issue as the rest of the code

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    """
+    tree = ast.parse(expr, mode="eval")
+    count = 0
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Name) and node.id == name:
+            count += 1
+    return count
+
+
+def classify_tasklet(state: dace.SDFGState, node: dace.nodes.Tasklet) -> Dict:
```
tbennun, line 511: bad output type hint. Please make a TypedDict for this, or a dataclass.

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    out_conns = list(node.out_connectors.keys())
+    n_in = len(in_conns)
+    n_out = len(out_conns)
+
+    assert n_out <= 1, "Only support tasklets with at most 1 output in this pass"
+    lhs = next(iter(node.out_connectors.keys())) if n_out == 1 else None
+
+    assert isinstance(node, dace.nodes.Tasklet)
+    code: CodeBlock = node.code
+    assert code.language == dace.dtypes.Language.Python
```
tbennun, line 589: again, no assertions please

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+            if isinstance(node.func, ast.Name):
+                op = node.func.id
+            elif isinstance(node.func, ast.Attribute):
+                op = node.func.attr
+
+        if op is None:
+            continue
+
+        if op not in _SUPPORTED:
+            print(f"Found unsupported op {op} in {src}")
```
alexnick83, line 349: Shouldn't an error be raised here?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        - Tasklet must have exactly 1 output connector
+        - Tasklet must use Python language
+        - Code must contain at most one operation (See SplitTasklets pass to enforce this easily)
+    """
+    in_conns = list(node.in_connectors.keys())
+    out_conns = list(node.out_connectors.keys())
+    n_in = len(in_conns)
+    n_out = len(out_conns)
+
+    assert n_out <= 1, "Only support tasklets with at most 1 output in this pass"
```
tbennun, line 584: same

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    lhs = next(iter(node.out_connectors.keys())) if n_out == 1 else None
+
+    assert isinstance(node, dace.nodes.Tasklet)
+    code: CodeBlock = node.code
+    assert code.language == dace.dtypes.Language.Python
+    code_str: str = code.as_string
```
tbennun, line 590: it's inefficient that you already have an AST object and you convert it to a string and then reparse it. Explain your reasoning please.

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        assert len(in_edges) == 1, f"expected 1 in-edge for connector {rhs}, found {len(in_edges)}"
+        rhs_data_name = in_edges.pop().data.data
+        rhs_data = state.sdfg.arrays[rhs_data_name]
+        out_edges = {oe for oe in state.out_edges_by_connector(node, lhs)}
+        assert len(out_edges) == 1, f"expected 1 out-edge for connector {lhs}, found {len(out_edges)}"
+        lhs_data_name = out_edges.pop().data.data
+        lhs_data = state.sdfg.arrays[lhs_data_name]
+
+        # Assignment operators it will return op <- `=` and always populate `rhs1`
+        if code_str == f"{lhs} = {rhs}" or code_str == f"{lhs} = {rhs};":
```
tbennun, line 608: I would not trust an expression like this (e.g., what about `a = (b)`?). Why not use isinstance(ast.Assign/AnnAssign)?

dace/sdfg/tasklet_utils.py
alexnick83, line 358: Wouldn't it be better to convert to AST, find the assignment, and then take the value?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+                            raise ValueError(f"Unsupported Assignment Type {lhs_datadesc} <- {c1}")
+                        info_dict.update({"type": ttype, "op": "=", "constant1": c1})
+                        return info_dict
+                else:
+                    c1 = bound_syms.pop()
+                    ttype = TaskletType.UNARY_SYMBOL
+                    info_dict.update({"type": ttype, "constant1": c1, "constant2": None, "op": op})
+                    return info_dict
+
+    raise NotImplementedError("Unhandled case in detect tasklet type")
```
tbennun, line 793: Shouldn't this just return a cannot-classify-tasklet or "other" TaskletType result?

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+            if fname.lower() == "or" and len(node.args) == 2:
+                return ast.BoolOp(op=ast.Or(), values=node.args)
+            # AND
+            if fname.lower() == "and" and len(node.args) == 2:
+                return ast.BoolOp(op=ast.And(), values=node.args)
+
+        return node
+
+
+def rewrite_boolean_functions_to_boolean_ops(src: str) -> str:
```
tbennun, line 823: where is this used?

tbennun (changes requested):

dace/sdfg/tasklet_utils.py
alexnick83, line 394: Maybe worth testing what happens when the input comes from another tasklet ... we used to have some "hidden" scalar descriptors for that, but I am not up to date on that.

dace/sdfg/tasklet_utils.py
alexnick83, line 434: Same as above.

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+        - Tasklet must use Python language
+        - Code must contain at most one operation (See SplitTasklets pass to enforce this easily)
+    """
+    in_conns = list(node.in_connectors.keys())
+    out_conns = list(node.out_connectors.keys())
+    n_in = len(in_conns)
+    n_out = len(out_conns)
+
+    assert n_out <= 1, "Only support tasklets with at most 1 output in this pass"
+    lhs = next(iter(node.out_connectors.keys())) if n_out == 1 else None
```
alexnick83, line 585: Very nitpicky, but why not just `lhs = out_conns[0] if n_out == 1 else None`?

dace/sdfg/tasklet_utils.py
alexnick83: I see now, way below (line 600+), that you are looking at the data type, not the connector type. Maybe this should be clarified in the note above (line 26).

dace/sdfg/tasklet_utils.py
```diff
@@ -0,0 +1,828 @@
...
+    assert isinstance(node, dace.nodes.Tasklet)
+    code: CodeBlock = node.code
+    assert code.language == dace.dtypes.Language.Python
+    code_str: str = code.as_string
+
+    info_dict = {"type": None, "lhs": lhs, "rhs1": None, "rhs2": None, "constant1": None, "constant2": None, "op": None}
+
+    assert n_out == 1
+
+    if n_in == 1:
```
alexnick83, line 596: Given the many restrictions on supported tasklets, the following feels somewhat too long. If I understand correctly, you accept codes with a single expression (assignment). The RHS may be either an operand or a unary/binary operator. The operands can only be connector names, symbols, or other expressions that evaluate to a constant. You also have a special case where the binary operation is applied to the same connector and interpreted as a unary op. Are you going to support more generalized tasklets in the future?

alexnick83 (commented):
Very well written, but I feel that it is a bit overengineered considering the currently limited supported cases. Please find some comments and questions below, but I will probably need to have another look.

PR #2285, opened 2026-01-30
MapFusion: only warn when class is instantiated

romanc:
`MapFusion` is deprecated and `MapFusionVertical` should be used instead. `MapFusion` remains in the cobase as a compatibility layer for backwards compatibility.

This PR changes the way how `MapFusion` is deprecated. Previously, a warning would be emitted whenever the class was loaded (e.g. when access from `dace/transformation/dataflow/__init__.py`. Depending on how verbose pytest is configured, one would thus see a deprecation warning even if `MapFusion` was never actually used/instantiated. The PR suggests to only emit a warning upon class init.

The unrelated change in `tests/buffer_tiling_test.py` (unused import) is because I used test in there to make sure the warning disappears.

/cc @philip-paul-mueller FYI

### DIFF
--- a/dace/transformation/dataflow/map_fusion.py
+++ b/dace/transformation/dataflow/map_fusion.py
@@ -1,24 +1,21 @@
 # Copyright 2019-2025 ETH Zurich and the DaCe authors. All rights reserved.
 
-import dace
 import warnings
-from dace.transformation import dataflow as dftrans
+from typing import Any
 
-try:
-    from warnings import deprecated
-except ImportError:
-
-    def deprecated(msg):
-        warnings.warn(msg, DeprecationWarning, stacklevel=2)
-        return lambda cls: cls
+from dace.transformation.dataflow import MapFusionVertical
 
 
-@deprecated('MapFusion is deprecated please use MapFusionVertical instead.')
-class MapFusion(dftrans.MapFusionVertical):
+class MapFusion(MapFusionVertical):
     """Compatibility layer for deprecated `MapFusion` name.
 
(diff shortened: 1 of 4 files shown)


dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,24 +1,34 @@
...
+    def __init__(
+        self,
+        only_inner_maps: Optional[bool] = None,
+        only_toplevel_maps: Optional[bool] = None,
+        strict_dataflow: Optional[bool] = None,
+        assume_always_shared: Optional[bool] = None,
+        require_exclusive_intermediates: Optional[bool] = None,
+        require_all_intermediates: Optional[bool] = None,
+        consolidate_edges_only_if_not_extending: Optional[bool] = None,
+        never_consolidate_edges: Optional[bool] = None,
```
philip-paul-mueller:
```suggestion
        *args,
```

dace/transformation/dataflow/map_fusion.py
```diff
@@ -1,24 +1,34 @@
...
+        **kwargs: Any,
+    ) -> None:
+        """Deprecated: use `MapFusionVertical` instead."""
+        super().__init__(only_inner_maps, only_toplevel_maps, strict_dataflow, assume_always_shared,
+                         require_exclusive_intermediates, require_all_intermediates,
+                         consolidate_edges_only_if_not_extending, never_consolidate_edges, **kwargs)
```
philip-paul-mueller:
```suggestion
        super().__init__(*args, **kwargs)
```

tbennun (approved):

PR #2286, opened 2026-01-30
Drop support for python versions < 3.10

romanc:
In particular, this get rid of a bunch of if/else blocks with breaking changes in python 3.8 and 3.9.


tbennun (approved):
Nice work!

PR #2288, opened 2026-02-01
Work-Depth: support interstate edges, better tasklet depth

tbennun:
The changes include reimplementing `DepthCounter` class for AST traversal, updating tasklet analysis functions to support both work and depth metrics, extending the SDFG analysis to include interstate edge work/depth, and adding tests to validate the new logic.

### DIFF
--- a/dace/sdfg/performance_evaluation/work_depth.py
+++ b/dace/sdfg/performance_evaluation/work_depth.py
@@ -345,14 +345,6 @@ class DepthCounter(ast.NodeVisitor):
             depths.append(self.visit(node.step))
         return max(depths) if depths else 0
 
-    def visit_Module(self, node):
-        # For a module, process statements sequentially to track dependencies
-        max_depth = 0
-        for stmt in node.body:
-            stmt_depth = self.visit(stmt)
-            max_depth = max(max_depth, stmt_depth)
-        return max_depth
-
     def generic_visit(self, node):
         # For unhandled nodes, try to get max depth from children
         max_depth = 0


dace/sdfg/performance_evaluation/work_depth.py
```diff
@@ -188,58 +188,214 @@ def count_arithmetic_ops_code(code):
...
-def tasklet_work(tasklet_node, state):
+def tasklet_work(tasklet_node: nd.Tasklet, state: SDFGState):
     if tasklet_node.code.language == dtypes.Language.CPP:
         # simplified work analysis for CPP tasklets.
+        total_accesses = 0
         for oedge in state.out_edges(tasklet_node):
-            return oedge.data.num_accesses
+            if not oedge.data.is_empty():
+                total_accesses += oedge.data.num_accesses
+        return total_accesses if total_accesses > 0 else 1
```
ThrudPrimrose: If I remember and understand correctly, this is a simplifying assumption made by Cliff Hodel.
Setting the depth equal to the output volume does not make sense. E.g., if you have a reduction from N elements to a single output, output accesses should be 1, but depth is log(N).

Given that we have C++ code, the analyzability is very hard, and I guess we can keep the assumption as is. With C++ code, I would probably have returned 1, though. Since C++ tasklets are not really analyzable (with realistic effort), I would even propose a warning here as well.

dace/sdfg/performance_evaluation/work_depth.py
```diff
@@ -188,58 +188,214 @@ def count_arithmetic_ops_code(code):
...
+        depths = []
+        if node.lower:
+            depths.append(self.visit(node.lower))
+        if node.upper:
+            depths.append(self.visit(node.upper))
+        if node.step:
+            depths.append(self.visit(node.step))
+        return max(depths) if depths else 0
+
+    def visit_Module(self, node):
```
ThrudPrimrose: Can this ever happen in python-ast of an SDFG tasklet?

ThrudPrimrose: If not, we can remove this to decrease total LoC that needs to be read in the future.

PR #2289, opened 2026-02-02
Check for invalid accesses in explicit dataflow tasklets

tbennun:
This includes memlets that target symbols, symbol assignment, and reading/writing arrays or scalars without an appropriate memlet.

### DIFF
--- a/tests/polybench/doitgen.py
+++ b/tests/polybench/doitgen.py
@@ -58,8 +58,8 @@ def doitgen(A, C4):
         def compute_sum(p: _[0:NP], s: _[0:NP]):
             inA << A[r, q, s]
             inC4 << C4[s, p]
-            s >> sum(1, lambda a, b: a + b, 0)[p]
-            s = inA * inC4
+            outs >> sum(1, lambda a, b: a + b, 0)[p]
+            outs = inA * inC4
 
         @dace.map
         def compute_A(p: _[0:NP]):


tests/intarg_test.py
```diff
@@ -11,8 +11,9 @@ def intarg(A, B, integer):
     @dp.map(_[0:W])
     def compute(i):
         a << A[i]
+        iarg << integer
         b >> B[i]
-        b = a * integer
+        b = a * iarg
```
phschaad, line 16: is `integer` a predefined symbol, or is this just an aesthetic change?

tests/nested_sdfg_test.py
```diff
@@ -17,16 +17,18 @@ def sdfg_internal(input: dp.float32, output: dp.float32[1]):
 
         @dp.tasklet
         def init():
+            inp << input
             out >> output
-            out = input
```
phschaad, line 21: ah, I suppose they are functions in the AST context?

tests/intarg_test.py
tbennun: integer is a scalar that was not accessed through a memlet

tests/nested_sdfg_test.py
tbennun: `input` is a scalar that was not accessed through a memlet

PR #2304, opened 2026-02-11
Add Clang CI

iomaganaris:
- Use LLVM 21 compiler for python3.10 and simplify autoopt


tbennun:
Please use a better way to set up llvm (the llvm apt repository) and perhaps instead of sequentializing it (since this will make all the tests take longer) make it part of the configuration matrix that gets skipped most of the time

.github/workflows/general-ci.yml
```diff
@@ -29,6 +34,15 @@ jobs:
       uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Setup LLVM
+      if: matrix.compiler == 'clang'
+      run: |
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 21
```
iomaganaris, line 42:
```suggestion
        sudo ./llvm.sh
```
Not sure if we should pinpoint this to a specific version or keep this open to get the last one always

dace/config_schema.yml
```diff
@@ -268,7 +268,7 @@ required:
                         type: str
                         title: Arguments
                         description: Compiler argument flags
-                        default: '-fPIC -Wall -Wextra -O3 -march=native -ffast-math -Wno-unused-parameter -Wno-unused-label'
```
iomaganaris, line 293: This needs to be reverted however it didn't fix all the errors in the Clang CI

PR #2307, opened 2026-02-13
Fixed `remove_edge_and_dangling_path()`

philip-paul-mueller:
Fixed `remove_edge_and_dangling_path()` such that it also handles the case where the edge is _not_ located inside a Map.


dace/sdfg/nodes.py
```diff
@@ -192,11 +192,17 @@ def remove_in_connector(self, connector_name: str):
             :param connector_name: The name of the connector to remove.
             :return: True if the operation was successful.
         """
+        if not connector_name:
+            warnings.warn(f'Tried to remove the {connector_name} from the in-connectors of node {str(self)}',
```
tbennun:
```suggestion
            warnings.warn(f'Tried to remove `{connector_name}` from the in-connectors of node {str(self)}',
```

dace/sdfg/nodes.py
```diff
@@ -205,11 +211,17 @@ def remove_out_connector(self, connector_name: str):
             :param connector_name: The name of the connector to remove.
             :return: True if the operation was successful.
         """
+        if not connector_name:
+            warnings.warn(f'Tried to remove the {connector_name} from the out-connectors of node {str(self)}',
```
tbennun:
```suggestion
            warnings.warn(f'Tried to remove `{connector_name}` from the out-connectors of node {str(self)}',
```

dace/sdfg/utils.py
```diff
@@ -857,51 +857,74 @@ def set_outer_subset(e: MultiConnectorEdge[dace.Memlet], new_subset: sbs.Subset)
...
+    nb_removed_edges = 0
+    while curr_tree is not None:
+        curr_edge = curr_tree.edge
+        assert not curr_edge.data.is_empty()
+        state.remove_edge(curr_edge)
+        nb_removed_edges += 1
+
+        if curr_tree.downwards:
+            if state.degree(curr_edge.dst) == 0:
+                # If the edge is isolated we can remove it.
```
tbennun:
```suggestion
                # If the target node is isolated we can remove it.
```

dace/sdfg/utils.py
```diff
@@ -857,51 +857,74 @@ def set_outer_subset(e: MultiConnectorEdge[dace.Memlet], new_subset: sbs.Subset)
...
+            if state.degree(curr_edge.src) == 0:
+                state.remove_node(curr_edge.src)
+            else:
+                if isinstance(curr_edge.src, nd.ExitNode) and curr_edge.src_conn.startswith("OUT_"):
+                    curr_edge.src.remove_in_connector("IN_" + curr_edge.src_conn[4:])
+                if curr_edge.src_conn and len(list(state.out_edges_by_connector(curr_edge.src,
+                                                                                curr_edge.src_conn))) == 0:
+                    curr_edge.src.remove_out_connector(curr_edge.src_conn)
+
+            # The connector might be collecting.
```
tbennun, line 914: what case is this? I think it is generally disallowed in SDFG syntax

tbennun (changes requested):
Generally looks good, minor comments

dace/sdfg/utils.py
philip-paul-mueller: I actually wish that it would be disallowed but it is permitted and as far as I can tell parts of the toolchain (I suspect `ConsolidateEdge`) actually create such cases.

The case essentially represents the following:
```python=
A: np.ndarray[N, 40] = ...
for i in dace.map[0:N]:
    A[i, 0] = some_awesome_boundary_value()
    A[i, 39] = another_awesome_boundary_value()
```
So `MapExit` will have two incoming edges.
However, they do not have to go to the distinct (in-)connectors of the `MapExit` node. 
In case they go to distinct connectors then there are also two outgoing edges one with subset `[0:N, 0]` and the other with `[0:N, 39]`.

However, it is also allowed that they go to the same (in-)connector of `MapExit`.
In that case `MapExit` will only have one outgoing edge, and it will have a subset of `[0:N, 0:40]`, which grossly overestimate the range that is written.

This is actually currently happening in GT4Py and it also blocks some optimization.

tbennun: Yes, there is an exception for passthrough (`IN_*/OUT_*`) connectors. I was only thinking about tasklets/library nodes/nested SDFGs. Thank you for the clarification

tbennun (approved):
LGTM

PR #1303, opened 2023-07-10
Add a library node for np.transpose for >2 dims and a cuTENSOR implementation

lamyiowce:
Adds 
- a library node for np.transpose for >2 dims
- a cuTENSOR implementation
- pure implementation (copy-pasted what was in replacements.py before)
- cutensor environment

What I don't know:
- whether the file structure is correct
- how to make DaCe find cuTensor in a pretty way. It has to be separately downloaded (it's not in cuda) and I don't know how that's handled in DaCe.

PR #2317, opened 2026-03-08
TTranspose node with cuTensor v2

ThrudPrimrose:
I have added cutensor tensor transpose expansion for cutensor v2 and ported tensor contraction to cutensor v2 too.

After this we should also close this old PR:
https://github.com/spcl/dace/pull/1303.


dace/libraries/linalg/environments/cutensor.py
tbennun, line 1: I don't think cutensor should live in the standard library. Why did you move it here?

dace/libraries/standard/nodes/ttranspose.py
```diff
@@ -61,6 +62,8 @@ class ExpandHPTT(ExpandTransformation):
 
     @staticmethod
     def expansion(node, parent_state, parent_sdfg):
+        from dace.libraries.linalg import blas_helpers
```
tbennun: any inline import requires a comment as to why (especially since the environments and implementation used to live in linalg)

dace/libraries/linalg/nodes/ttranspose.py
```diff
@@ -84,11 +87,133 @@ def expansion(node, parent_state, parent_sdfg):
...
+        modes_a = list(range(ndim))
+        # Output modes: the permutation   [axes[0], axes[1], ...]
+        modes_c = list(node.axes)
+
+        modes_a_str = ', '.join(str(m) for m in modes_a)
+        modes_c_str = ', '.join(str(m) for m in modes_c)
+        extent_a_str = ', '.join(symstr(s) for s in inp_tensor.shape)
+        extent_c_str = ', '.join(symstr(s) for s in out_tensor.shape)
+        stride_a_str = ', '.join(symstr(s) for s in inp_tensor.strides)
+        stride_c_str = ', '.join(symstr(s) for s in out_tensor.strides)
```
tbennun: for C++ code, you need to use sym2cpp, not symstr: https://spcldace.readthedocs.io/en/latest/extensions/symbolic.html

tbennun (changes requested):
Everything looks great. I am not sure about cutensor being in the standard environment, especially as it relates to einsums and contractions rather than just transposes. Sounds like a linalg library to me but I am open for discussion.

dace/libraries/linalg/environments/cutensor.py
ThrudPrimrose: The tensor transpose library node was in standard before so I added it there. Imho, matrix transpose and tensor transpose should be in linalg. Should I move them there?

dace/libraries/linalg/nodes/tensordot.py
```diff
@@ -318,58 +324,65 @@ def expansion(node, parent_state, parent_sdfg):
...
         workspace = """
-            cutensorContractionFind_t find;
-            dace::linalg::CheckCuTensorError(cutensorInitContractionFind(&__dace_cutensor_handle, &find, CUTENSOR_ALGO_DEFAULT));
-            size_t worksize = 0;
-            dace::linalg::CheckCuTensorError(cutensorContractionGetWorkspace(
-                &__dace_cutensor_handle, &desc, &find, CUTENSOR_WORKSPACE_RECOMMENDED, &worksize));
+            cutensorPlanPreference_t planPref;
+            dace::linalg::CheckCuTensorError(cutensorCreatePlanPreference(
+                __dace_cutensor_handle, &planPref,
+                CUTENSOR_ALGO_DEFAULT, CUTENSOR_JIT_MODE_NONE));
```
tbennun: Don’t we want jit for performance?

dace/libraries/linalg/nodes/ttranspose.py
```diff
@@ -84,11 +87,135 @@ def expansion(node, parent_state, parent_sdfg):
...
+        __dace_cutensor_handle, &opDesc,
+        descA, modesA, CUTENSOR_OP_IDENTITY,
+        descC, modesC,
+        {compute_desc}));
+
+    // plan preference & plan
+    cutensorPlanPreference_t planPref;
+    dace::linalg::CheckCuTensorError(cutensorCreatePlanPreference(
+        __dace_cutensor_handle, &planPref,
+        CUTENSOR_ALGO_DEFAULT, CUTENSOR_JIT_MODE_NONE));
```
tbennun: Same jit question but here it is perfectly acceptable not to do that

tbennun (changes requested):
Looks great, one comment and ci failed (potentially unrelated)

dace/libraries/linalg/nodes/tensordot.py
ThrudPrimrose: I thought of using no-JIT for more predictable runtimes, but I think using JIT makes sense to create the plans.

ThrudPrimrose:
I think the ML autodiff tests have some tests that occasionally/spuriously fail numerical validation.

tbennun (approved):

PR #2318, opened 2026-03-09
ci: find and automatically remove unused imports

romanc:
Unused imports add a performance overhead at runtime, and risk creating import cycles. To automatically detect and remove them in the future, this PR suggests to add [ruff](https://docs.astral.sh/ruff/) as a `pre-commit` hook to detect unused imports. I've added an exception for `__init__.py` files to allow re-exports without adding an explicit `__all__` list or adding extra annotations.

The PR grew quite big. However, non-automatic changes are only in the following seven files

- `.pre-commit-config.yaml`: configure `pre-commit` to run the `ruff` linter
- `ruff.toml`: configure the `ruff` linter to only search for unused imports (rule `F401`)
- `dace/autodiff/library/library.py`: make sure we keep the `ParameterArray` import for backwards compatibility
- `dace/frontend/python/replacements/operators.py`: make sure we keep the `dace` import for evaluation of data types
- `tests/library/include_test.py`: make sure we keep the necessary import in the middle of the test
- `dace/sdfg/analysis/schedule_tree/treenodes.py`: manually remove the now trivial `if TYPE_CHECKING` branch

PR #2321, opened 2026-03-12
Add a config option for storing lineinfo into the SDFG

romanc:
When we build an SDFG, there's the option to store `DebugInfo` with some SDFG nodes. For example, this `DebugInfo` can be used to store file & line information of parsed code when building an SDFG. When using the SDFG API, the default is to inspect the python stack and extract file & line information from there. These calls to `inspect` can/will be expensive, especially for bigger graphs.

This PR proposes to add a configuration option, `compiler.lineinfo`, to drive this behavior from a single place. The defaults are kept as is, i.e. we keep inspecting the stack by default. However, the config option allows a central pace to turn `DebugInfo` off, which could be configured in production scenarios.

### DIFF
--- a/dace/config_schema.yml
+++ b/dace/config_schema.yml
@@ -212,7 +212,7 @@ required:
                 description: >
                     Specify the default data types to use in generating code.
                     If "Python", Python's semantics will be followed (i.e., `float`  and `int`
-                    are represented using 64 bits). If the property is set to "C", C's semantcs will be
+                    are represented using 64 bits). If the property is set to "C", C's semantics will be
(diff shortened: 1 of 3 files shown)


tbennun (approved):

tbennun:
cscs-ci run

romanc:
rebased to fix a conflict in the imports of `state.py`, which were also touched by PR https://github.com/spcl/dace/pull/2318. No other changes.

tbennun (approved):

tbennun:
cscs-ci run

PR #2324, opened 2026-03-14
Do not convert floats to integers

ThrudPrimrose:
The symstr might convert `0.0` to `0`. E.g. the iedge below is generated from `lc = max(-zdqs[jl], 0.0)`, can be converted to use `0`.

I had an interesting butterfly effect. Sympy reordered arguments because the max function is commutative. But not with implicit promotions. If 0 is an integer, and appears first, then the double variable is promoted to int, thus the result will be different from when the double appears first.

I thought it could be related to `sympy_numeric_fix`, but I think the fix is in improving the print Float function `symstr`. I propose to change the `_print_Float` of `symstr` to clean trailing zeroes as before, but ensuring that there is at least one and not converting floats to integers implicitly.

With the new print I avoid implicit casts to integer in CloudSC SDFG. (I add some tests too).

See my comment below for a better explanation.


tbennun (approved):
It passes all tests for now, but let's write the follow up PR that would fix the issue entirely

PR #2331, opened 2026-04-01
Reduced Size of SDFG Folder

philip-paul-mueller:
The initial idea was to reduce the size of the cache folder, by only storing the components that are needed.
To this end the code generator was modified such that different versions of build folder could be generated.
Currently there are only two versions:
- `development`: Which is the old full version, i.e. everything in a single folder.
- `production`: This is a reduced folder and only contains the libraries (stub and program library) as well the version.

The implementation has two parts. First `generate_program_folder()` was modified such that it only generates the parts that are absolutely needed, such as source files and anything else would not be generated in the first place.
Then `compile_and_configure()` was modified such that it would remove the parts that are no longer needed (an example would be the source files which are needed for compilation but not afterwards).

The changes are backwards compatible. Thus, caches that were generated _before_ this PR will still work, but they should be phased out.
The changes are also done in a way that it should be simple to add new modes later, if their need arises.

In Addition:
- Also fixes that `sdfg.view()` fails if there are external SDFGs ([727cfb1](https://github.com/spcl/dace/pull/2331/commits/727cfb1190fe801da286031ca251a1743d7d31dd))
- `sdfg.generate_code()` no longer generates the source maps as a side effect and writes them to disc. Instead they are generated by the `generate_program_folder()` ([5e1694b](https://github.com/spcl/dace/pull/2331/commits/5e1694b09b831e68c10c81c8a51e6708bd9bc81a)).
- Dumping of the configuration in the build folder ([eb54062](https://github.com/spcl/dace/pull/2331/commits/eb540625fe7ed01699bf73458caeaf1e53feb625)).
- Miscellaneous refactoring in the `ReloadableDLL`, that changed its interface.

### DIFF
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -1682,9 +1682,8 @@ def distributed_compile(sdfg: SDFG, comm, *, validate: bool = True) -> csdfg.Com
     :param validate: If True, validates the SDFG prior to generating code.
     :return: Compiled SDFG.
     :note: This method can be used only if the module mpi4py is installed.
+    :todo: Relocate this function to `dace.codegen.compiler`.
     """
-
-    # TODO: Think if it should be relocated to `codegen`
 
     rank = comm.Get_rank()
     func = None
--- a/dace/sdfg/utils.py
+++ b/dace/sdfg/utils.py
@@ -1699,7 +1699,7 @@ def distributed_compile(sdfg: SDFG, comm, *, validate: bool = True) -> csdfg.Com
 
     # Loads compiled SDFG.
     if rank > 0:
-        func = load_precompiled_sdfg(folder)
+        func = sdfg_compiler.load_precompiled_sdfg(folder)
 
     comm.Barrier()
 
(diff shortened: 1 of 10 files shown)


philip-paul-mueller:
A point of discussion is if the `CACHEDIR.TAG` should also be generated in production mode.

acalotoiu (approved):
This is a good change, fur future we should also consider compressing the sdfg in all cases.
Thank you!

dace/sdfg/utils.py
```diff
@@ -1697,6 +1682,7 @@ def distributed_compile(sdfg: SDFG, comm, *, validate: bool = True) -> csdfg.Com
     :param validate: If True, validates the SDFG prior to generating code.
     :return: Compiled SDFG.
     :note: This method can be used only if the module mpi4py is installed.
+    :todo: Relocate this function to `dace.codegen.compiler`.
```
tbennun, line 1685: Why is this a TODO? This could be moved directly.

philip-paul-mueller: While I changed some function they had an "internal smell", i.e. those brave enough to use them know that it is dangerous and accept that.
However, this function is much more public than the others, thus I decided to keep it where it is, because I am also not fully sure if it should be moved at all.

PR #2335, opened 2026-04-15
Remove Views Pass

ThrudPrimrose:
This PR introduces a pass that removes views from the SDFG by folding them into the access patterns of the arrays they view.

Context: Reduced SDFG IR
This is the first PR toward a Reduced IR. A simplified subset of the full SDFG IR that limits the number of features transformations that need support. The Reduced IR eliminates views, WCR edges, other_subset on edges, dynamic thread-block maps, and persistent device maps by lowering each to a smaller set of primitives (e.g., WCR becomes a reduction library node, other_subset requiring copies becomes explicit copy nodes).

The goal of this PR is to be part of many to provide a lower pipeline from the full IR to the Reduced IR, removing as many of these features as possible. Alongside this pass, I have also extended the Copy library node and implemented a pass that inserts explicit copies where needed. (Separate PR)

What this pass does
The pass tries to remove as many views as possible by absorbing them into surrounding memlets. For example, a view that selects a strided column from a row-major array:

image

is removed with the following subset:

image

(It supports many other patterns as well).


tbennun:
What cases does your pass remove that the existing view removal transformations in array elimination doesn’t cover?

ThrudPrimrose:
> What cases does your pass remove that the existing view removal transformations in array elimination doesn’t cover?

I compared `ArrayElimination` against `RemoveViews`. Remove views handles interstate edges, and some other complicated patterns.

| test | before | after AE Only | after RV Only | AE correct | notes |
|---|---:|---:|---:|:---:|---|
| test_view_array_array | 1 | 1 | 0 | skip | |
| test_view_slice_detect_simple | 1 | 0 | 0 | OK | |
| test_read_slice | 3 | 0 | 0 | OK | |
| test_simple_slice_view | 1 | 0 | 0 | OK | |
| test_reshape_view | 2 | 1 | 0 | OK | |
| test_squeeze_view | 1 | 0 | 0 | OK | |
| test_view_chain | 2 | 0 | 0 | OK | |
| test_noop_no_views | 0 | 0 | 0 | OK | |
| test_unsqueeze_view | 1 | 0 | 0 | OK | |
| test_multiple_views_same_state | 2 | 0 | 0 | OK | |
| test_write_view | 1 | 0 | 0 | OK | |
| test_column_view | 1 | 0 | 0 | OK | |
| test_column_view_w_offset | 1 | 0 | 0 | OK | |
| test_strided_column_view | 1 | 1 | 0 | OK | |
| test_flatten_view | 1 | 1 | 0 | OK | |
| test_view_in_interstate_edge | 1 | 1 | 0 | OK | |

dace/transformation/passes/remove_views.py
```diff
@@ -0,0 +1,989 @@
...
+
+A View node V between a viewed array A and a map
+``phi : index_space(V) -> index_space(A)`` encoded by the
+view edge memlet. We can rewrite memlets m
+that references V in terms of A by composing ``m' = phi.m``.  After
+rewriting, V has no remaining uses and can be removed.
+
+Two strategies are used depending on the view kind:
+
+Four strategies are tried in order until one succeeds:
```
tbennun, line 14: is it two or four :)

dace/transformation/passes/remove_views.py
```diff
@@ -206,9 +201,6 @@ def _delinearize_flat(flat, astrides: List[int], array_shape: List[int]):
     return [(flat // astr) % shp for astr, shp in zip(astrides, array_shape)]
```
tbennun, line 201: can you explain how this works, especially with complex permutations?

tbennun (changes requested):
Please add to simplification pipeline

dace/transformation/passes/remove_views.py
```diff
@@ -0,0 +1,989 @@
...
+import operator
+from typing import Any, Dict, List, Optional, Set, Tuple
+
+from dace import SDFG, SDFGState, config, data as dt, dtypes, properties, subsets, symbolic
+from dace.frontend.python import astutils
+from dace.properties import CodeBlock
+from dace.sdfg import nodes as nd, utils as sdutil, graph as gr
+from dace.transformation import pass_pipeline as ppl, transformation
+
+_PASS = 'RemoveViews'
```
tbennun, line 53: unnecessary

dace/transformation/passes/remove_views.py
```diff
@@ -0,0 +1,989 @@
...
+_DEBUGPRINT = config.Config.get('debugprint') in (True, '1', 'true', 'yes')
```
tbennun, line 54: this does not work. Also this does not conform to our API.
1. It will only read it on import, so someone who sets the config temporarily will not observe this change. 
2. Call `config.Config.get_bool('debugprint')`

So just call this function in the only place it is used.

tbennun (changes requested):

PR #1957, opened 2025-02-28
Enforce consistent formatting on CI

romanc:
## Description

As discussed in the weekly DaCe meeting today, we'd like to go ahead with enforcing consistent formatting as part of CI. As a reminder, this started with a discussion[^1] in December and there was no pushback on the discussion page and also not in the meeting today.

We decided to proceed in the following way.

1. @tbennun will run `pre-commit` / `yapf`  locally on the codebase and push that, see https://github.com/spcl/dace/pull/1966
2. This PR wil then be merged afterwards and just contain the changes needed to add the new CI job.

[^1]: https://github.com/spcl/dace/discussions/1804

PR #2337, opened 2026-04-17
ci: cleanup linting GitHub Action

romanc:
When we started to enforce consistent formatting on the CI (PR https://github.com/spcl/dace/pull/1957), an important discussion point was for developers to see changes that needed to be applied. For lack of better knowledge, I've added a small script to show the `git diff` output in case of failure. I recently learned that `pre-commit` has a built-in `--show-diff-on-failure` option

image

which enables exactly this behavior out of the box ([link to failing workflow run](https://github.com/spcl/dace/actions/runs/24556723300/job/71794961944?pr=2337)). It even comes with a colored output ... :star_struck: I suggest, we drop the custom script since this option is much simpler.


tbennun (approved):
Much better, thank you

tbennun:
cscs-ci run

PR #2339, opened 2026-04-20
Fix Map Fission

ThrudPrimrose:
This fixes 2 bugs in map fission.

1) Loop-dependent views are incorrectly hoisted. For example:

The read `__tmp_726_11_r` depends on `A[i]` but it is considered as the subset if not present within the NSDFG.
Pasted image

After it becomes wrongly hoisted and the SDFG does not compile anymore:

Pasted image (2)

2) Second, if the step size is not equal to 1, then the array dimensions between the maps are generated according to the map trip count, but the access expression is generated only via the map iterator. I fix this by generating the access expression for intermediate arrays via `(i - begin)/step` such that the generated access expression is always correct , regardless of the map's shape.


phschaad (approved):
Looks good to me, thank you for the fix

PR #2344, opened 2026-04-21
refactor: set default lineinfo on state/block

romanc:
## Description

Follow-up from PR https://github.com/spcl/dace/pull/2321, we shouldn't set `debuginfo` anymore to a `DebugInfo` object when adding things to a `State` (or CFG Block). Instead, we set the `default_lineinfo` on that `State`.

### DIFF
--- a/dace/frontend/fortran/ast_utils.py
+++ b/dace/frontend/fortran/ast_utils.py
@@ -15,6 +15,8 @@ from numpy import float64 as fl
 from dace.frontend.fortran import ast_internal_classes
 from typing import List, Set
 
+from dace.sdfg.state import default_line_info
+
 fortrantypes2dacetypes = {
     "DOUBLE": dtypes.float64,
     "REAL": dtypes.float32,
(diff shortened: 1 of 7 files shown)


tbennun (changes requested):
I think the current diff in newast is overkill and we can just wrap the parts that change pv.current_lineinfo with the with statement instead. Any reason not to?

romanc:
I agree, it doesn't look pretty now.  I don't have a mental picture of `newast.py` that is good enough to propose an automatically sync of `.current_lineinfo` on the `ProgramVisitor` with `._default_line_info` on the respective `State`.

I guess one could automatically set `state._default_line_info` to `pv.current_lineinfo` for any state that is added to the SDFG and then potentially update once `pv.current_lineinfo` changes. Is that what you are suggesting? That would mean tracking all new state creation and setting `_default_line_info` accordingly. 

I start to wonder why we changed the signature of `state.add_read()` in the first place. PR https://github.com/spcl/dace/pull/2345 explores partially reverting https://github.com/spcl/dace/pull/2321and avoiding the deprecation warnings all together.

romanc:
@tbennun I think you should look at https://github.com/spcl/dace/pull/2345 as a replacement for this PR. https://github.com/spcl/dace/pull/2345 is a much smaller change and gives the same results (all warnings gone).

PR #2345, opened 2026-04-22
Avoid DebugInfo warnings (take two)

romanc:
## Description

In PR https://github.com/spcl/dace/pull/2321 we introduced a way to turn of automatic stack inspection when adding a node to a state without explicitly stating debug info (such as file name, line info, ...). Since merging this PR, we get a ton of `DeprecationWarning`s from code added in `_get_debug_info()`.

In a first attempt (see PR https://github.com/spcl/dace/pull/2344), we tried to work around the issue by setting `._default_line_info` on the state. This method proved to be very verbose and unnecessarily complicates parsing, where we have a legitimate use-case for explicitly passing along an explicit `DebugInfo`.

This PR suggests to keep the API of `state.add_*()` functions as before, partially reverting changes from PR https://github.com/spcl/dace/pull/2321. Doing so, we can drop the deprecation warnings because nothing changes.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -9,8 +9,8 @@ import inspect
 import itertools
 import warnings
 import sympy
-from typing import (TYPE_CHECKING, Any, AnyStr, Callable, Dict, Iterable, Iterator, List, Literal, Optional, Set, Tuple,
-                    Union, overload)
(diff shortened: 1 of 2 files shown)


tbennun (approved):

tbennun:
cscs-ci run

PR #2349, opened 2026-04-22
Fix Loop2Map.

ThrudPrimrose:
If Loop2Map references arrays only on IfBlock or InterstateEdge, then the symbols necessary for stride/offset/shape are not added properly to the Nested SDFG, causing an invalid SDFG. There is also a new test case now.


phschaad (approved):
LGTM

phschaad:
@ThrudPrimrose check the conflicts (after your vacation) please

PR #2361, opened 2026-05-08
Distributed Grid Data Descriptors

tbennun:
Convert process grids, subarrays, and redistribution arrays to special data descriptors. Removes specialized SDFG fields and streamlines code generation.


dace/data/distributed.py
```diff
@@ -38,8 +73,7 @@ class ProcessGrid(object):
     """
 
     name = Property(dtype=str, desc="The process-grid's name.")
```
tbennun, line 75: what is the significance of `name`?

dace/data/distributed.py
```diff
@@ -38,8 +73,7 @@ class ProcessGrid(object):
...
-    is_subgrid = Property(dtype=bool, default=False, desc="If true, spanws sub-grids out of the parent process-grid.")
-    shape = ShapeProperty(default=[], desc="The process-grid's shape.")
+    is_subgrid = Property(dtype=bool, default=False, desc="If true, spawns sub-grids out of the parent process-grid.")
     parent_grid = Property(dtype=str,
```
tbennun, line 77: @alexnick83 should this be a DataProperty too, or are there cases where the user provides a communicator via string?

dace/sdfg/sdfg.py
```diff
@@ -712,7 +691,12 @@ def from_json(cls, json_obj, context=None):
         if _type != cls.__name__:
             raise TypeError("Class type mismatch")
 
-        attrs = json_obj['attributes']
+        attrs = dict(json_obj['attributes'])
+        legacy_pgrids = attrs.pop('_pgrids', None)
```
tbennun: a lot of the changes here and below are kept for backwards compatibility of the SDFG format but are not necessary. Feel free to tell me to remove them

tests/library/mpi/distributed_descriptor_frontend_test.py
```diff
@@ -0,0 +1,86 @@
+import dace
```
tbennun, line 2:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
import dace
```
oops, sorry!

tests/library/mpi/mpi4py_test.py
```diff
@@ -320,8 +320,8 @@ def test_send_recv():
 
     @dace.program
     def mpi4py_send_recv(rank: dace.int32, size: dace.int32):
-        src = np.full([1], (rank - 1) % size, dtype=np.int32)
-        dst = np.full([1], (rank + 1) % size, dtype=np.int32)
+        src = (rank - 1) % size
+        dst = (rank + 1) % size
```
tbennun, line 324: I'm not sure I like this change, but I think the new code makes more sense to test. Thoughts?

tests/sdfg/distributed_descriptors_test.py
```diff
@@ -0,0 +1,102 @@
+import copy
```
tbennun:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
import copy
```
Again, sorry

tests/sdfg/distributed_descriptors_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+    assert isinstance(loaded.arrays[pgrid.name], ProcessGrid)
+    assert isinstance(loaded.arrays[subarray.name], SubArray)
+    assert isinstance(loaded.arrays[redistr.name], RedistrArray)
+    assert loaded.process_grids[pgrid.name] is loaded.arrays[pgrid.name]
+    assert loaded.subarrays[subarray.name] is loaded.arrays[subarray.name]
+    assert loaded.rdistrarrays[redistr.name] is loaded.arrays[redistr.name]
+    assert not hasattr(loaded, '_pgrids')
+    assert not hasattr(loaded, '_subarrays')
+    assert not hasattr(loaded, '_rdistrarrays')
```
tbennun:
```suggestion
    assert not hasattr(loaded, '_rdistrarrays')


if __name__ == '__main__':
    test_distributed_descriptors_are_data_descriptors()
    test_sdfg_distributed_descriptor_storage_and_compatibility_views()
    test_distributed_descriptor_json_roundtrip()
    test_legacy_distributed_side_store_json_migrates_to_arrays()
```

tests/sdfg/distributed_descriptors_test.py
tbennun, line 1: FYI, I am totally fine with removing this file.

tbennun (commented):
Some commentary for the review @alexnick83

dace/data/distributed.py
alexnick83: It is used in code generation. For example, you can see how the initialization is generated [here](https://github.com/spcl/dace/blob/3f9b6099a78ac15f69801978e4a23c04186239a0/dace/distr_types.py#L117). It is also used as the dictionary key in the current implementation (see [here](https://github.com/spcl/dace/blob/3f9b6099a78ac15f69801978e4a23c04186239a0/dace/sdfg/sdfg.py#L2288)). Canonically, the name is `__pgrid` and uses the `SDFG._find_new_name` functionality (see [here](https://github.com/spcl/dace/blob/3f9b6099a78ac15f69801978e4a23c04186239a0/dace/sdfg/sdfg.py#L2283)).

dace/data/distributed.py
alexnick83: `parent_grid` should be the name of an existing `ProcessGrid`, so I guess it can be a DataProperty only. However, minor adjustments may be required to ensure this works properly across all current uses/samples.

tests/library/mpi/mpi4py_test.py
alexnick83: I don't recall why it was like that in the first place, but there must have been a reason. It is likely related to the MPI API using 32-bit integers by default, but I do not remember whether this was just overzealous on my part or if there were issues with passing Python integers.

tests/sdfg/distributed_descriptors_test.py
alexnick83: It depends on whether this tests something useful, i.e., generation of (sub)grids with non-trivially derived attributes, but I don't recall right now.

dace/codegen/targets/cpp.py
```diff
@@ -247,6 +247,14 @@ def ptr(name: str, desc: data.Data, sdfg: SDFG = None, framecode: 'DaCeCodeGener
         if root in sdfg.arrays and isinstance(sdfg.arrays[root], data.Structure):
             name = name.replace('.', '->')
 
+    if isinstance(desc, data.DistributedDescriptor):  # Skip CFG ID mangling for distributed descriptors
```
alexnick83, line 250: I just started reading the PR, so I am not sure how init/deinit is handled, but in the original implementation, all ProcessGrids are included in the program's state (it would be difficult otherwise; I guess now that they are data, they can be passed around in an easier manner). However, there is a specific design decision: the init/deinit occurs in the program's init/deinit (so only once, regardless of how many times the SDFG is called). This also made sense from a performance perspective; you want to perform these expensive inits only once and amortize their cost over the program's lifetime.

alexnick83: The "useful" comment here is to discuss whether it should always be `f'__state->{name}'` or do we actually ever want to have ProcessGrids that are not program-global.

dace/codegen/targets/framecode.py
```diff
@@ -592,6 +592,11 @@ def determine_allocation_lifetime(self, top_sdfg: SDFG):
             access_instances[sdfg.cfg_id] = instances
 
         for sdfg, name, desc in top_sdfg.arrays_recursive(include_nested_data=True):
+            if isinstance(desc, data.DistributedDescriptor):
```
alexnick83, line 595: So I guess these descriptors are anyway force-allocated on the top-level SDFG, which makes sense.

dace/data/distributed.py
```diff
@@ -68,14 +102,21 @@ def __init__(self,
         self.name = name
         self.is_subgrid = is_subgrid
         if is_subgrid:
-            self.parent_grid = parent_grid.name
+            self.parent_grid = parent_grid.name if isinstance(parent_grid, ProcessGrid) else parent_grid
```
alexnick83, line 105: Regarding your question above, I suppose we could set this to the parent grid object instead of the name (with minor changes later). Keeping the name probably makes serialization simpler, though.

dace/data/distributed.py
```diff
@@ -68,14 +102,21 @@ def __init__(self,
...
             self.color = color
             self.exact_grid = exact_grid
-            self.shape = [parent_grid.shape[i] for i, remain in enumerate(color) if remain]
+            if isinstance(parent_grid, ProcessGrid):
+                shape = [parent_grid.shape[i] for i, remain in enumerate(color) if remain]
+            else:
```
alexnick83, line 110: We can throw an exception here. This should never happen.

dace/data/distributed.py
```diff
@@ -110,14 +147,17 @@ def from_json(cls, json_obj, context=None):
...
+        ret.lifetime = dtypes.AllocationLifetime.Persistent
         # Check validity now
         ret.validate()
         return ret
 
     def init_code(self):
         """ Outputs MPI allocation/initialization code for the process-grid.
             It is assumed that the following variables exist in the SDFG program's state:
-            - MPI_Comm {self.name}_comm
+            - MPI_Comm {self.name}
```
alexnick83, line 160: No strong opinions here. I just found it clearer, since there are so many related (C/C++) variables required, to just add the `_comm` postfix to the actual `MPI_Comm` object.

dace/frontend/python/replacements/mpi.py
```diff
@@ -25,12 +25,13 @@ def _cart_create(pv: ProgramVisitor, sdfg: SDFG, state: SDFGState, dims: ShapeTy
...
     """
-    pgrid_name = sdfg.add_pgrid(dims)
+    name = pv.get_target_name()
+    pgrid_name = sdfg.add_pgrid(dims, name=name)
 
     # Dummy tasklet adds MPI variables to the program's state.
     from dace.libraries.mpi import Dummy
     tasklet = Dummy(pgrid_name, [
-        f'MPI_Comm {pgrid_name}_comm;',
+        f'MPI_Comm {pgrid_name};',
```
alexnick83, line 34: Since the grids are now descriptors, is there now perhaps a better way to add these variables to the state dict?

dace/sdfg/sdfg.py
alexnick83: What is the use case here? Old SDFGs that use grids being compatible with a newer DaCe version? I guess this is nice, but maybe it needs a test? I don't know if this use case realistically exists, though; I would just reparse the programs. Also, would this even work, considering that the (old) SDFG will not have the related access nodes connected to the MPI library nodes?

alexnick83 (approved):
It looks fine to me in principle. From what I see, init/deinit still occurs in the program's init/deinit, so no functionality is altered.

dace/data/distributed.py
tbennun: Since that’s no longer the case (we store the name in the data container entry), should I try removing this property?

dace/data/distributed.py
tbennun: I’ll try

dace/sdfg/sdfg.py
tbennun: You’re right. I’ll remove that

tests/library/mpi/mpi4py_test.py
tbennun: Whatever the situation was, it works now. Thanks!

dace/data/distributed.py
tbennun: I’m suggesting a DataProperty, which is still a string but constrained to the available data descriptors in sdfg.arrays

dace/frontend/python/replacements/mpi.py
tbennun: These might no longer be used, but the dummy node still creates the grids by calling the mpi functions in the init/exit code

PR #2364, opened 2026-05-11
Do not merge: working PR for NASA crew

romanc:
## Description

`dace/math.h` re-exposes a couple of math functions from `cmath`, e.g. `std::sin(x)`, which has overloads for e.g. `float` and `double`. Some functions, e.g. `asin(x)`, were not exposed in this way. This leads to precision issues when `asin(x)` is called with a `float` because in the generated code, `asin` will be mapped to [the C version](https://en.cppreference.com/c/numeric/math/asin), which is only defined for doubles. There is thus an implicit cast happening of the argument and the computation is done in double precision. Worse, the return type is always a `double`, which will force the rest of the calculation to be up-casted to double precision if `asin()` is used in an expression.

#### TODO list for later

- [x] make test case for math functions (e.g. compare against `std::log10(x)` in a cpp-tasklet)
- [ ] make test case for symbol replacement (double-replace with strings, replace symbol with non-default dtype)
- [ ] make test case for [fix: stree -> sdfg, data descriptors of nested maps](https://github.com/spcl/dace/pull/2364/commits/bec508d401a5df004d0955267a056cfb5a92c86e)
- [x] make test cases for `memlets_in_ast()` (subscript access, scalar access, indirect array access)
- [ ] add support for indirect array access in `memlets_in_ast()`
- [ ] once `memlets_in_ast()` is reliable, we can inline conditionals and loop condition/update (in stree as well as in gt4py bridge)
- [ ] refactor `tn.IfScope` / `tn.ElifScope` / `tn.ElseScope` to `tn.ConditionalScope`
- [ ] validation succeeds if conditional uses a scalar that isn't defined (anymore); see image below; might be fixed by considering scalar-memlets (to be re-tested)
- [ ] validation succeeds if mapped tasklets aren't fully connected; see image below

image

image

### DIFF
--- a/dace/config.py
+++ b/dace/config.py
@@ -402,7 +402,7 @@ class Config(object):
     @staticmethod
     def get_default(*key_hierarchy):
         """ Returns the default value of a given configuration entry.
-            Takes into accound current operating system.
+            Takes into account current operating system.
 
             :param key_hierarchy: A tuple of strings leading to the
                                   configuration entry.
--- a/dace/frontend/python/parser.py
+++ b/dace/frontend/python/parser.py
@@ -16,6 +16,9 @@ from dace.frontend.python import (newast, common as pycommon, cached_program, pr
 from dace.sdfg import SDFG, utils as sdutils
 from dace.data import create_datadescriptor, Data
 
+if TYPE_CHECKING:
+    from dace.codegen.compiled_sdfg import CompiledSDFG
+
 try:
     import mpi4py
     from dace.sdfg.utils import distributed_compile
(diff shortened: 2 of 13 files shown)


alexnick83 (changes requested):
This looks fine. I would add a test that demonstrates the issue described in the PR (up/downcasting of data) and confirms that it is now resolved. I guess just one test for, e.g., asin would be enough.

romanc:
Make sense, I'll add something tomorrow. I think I can build upon [`compare_numpy_output()`](https://github.com/spcl/dace/blob/main/tests/numpy/common.py#L17).

romanc:
putting this back into a draft state because it is now our de-facto working branch and I've started to push more unrelated changes. I'll pull this apart later and make nice PRs with unrelated changes and tests.

romanc:
We are back on mainline. Closing this PR as it is not needed anymore.

PR #2368, opened 2026-05-18
Bugfix/mapfission nested parent isolated nodes

ThrudPrimrose:
## Fix two breaking patterns in Map Fission

This PR fixes two SDFG patterns where `MapFission` produced an invalid graph
("Leftover nodes in queue" from `scope_dict`).

### Pattern 1 — map nested inside another map

Breaking pattern 1

Before fix 1:
Before fix 1

After fix 1:
After fix 1

### Pattern 2 — conditional inside the fissioned component

Breaking pattern 2

Before fix 2:
Before fix 2

After fix 2 (left untouched — refused, not corrupted):
After fix 2

### Follow-up

Supporting if/else fission requires sequentializing and splitting the branch
conditions and replicating them per fissioned map. I plan a Map Fission
extension (also for Map Fusion, vertical and horizontal, to recombine
branches), tracked separately and out of scope for this bug-fix PR.


acalotoiu (approved):
LGTM. Thank you!

PR #2369, opened 2026-05-18
StateFusionExtended: refuse unsafe write-after-read fusion

ThrudPrimrose:
## Refuse state fusion that breaks happens-before ordering

After `StateFusion` runs, some graphs become incorrect because a
happens-before dependency before the fused states is lost.

Pattern this fixes:
Pattern

Buggy result after fusion:
Buggy after fusion

One possible fix is to insert explicit dependency edges, but that bloats the
graph and is not, in my view, a good approach:
Dependency-edge alternative

**Current approach:** refuse the fusion in this case, leaving the graph
correct rather than producing an invalid one.


acalotoiu (approved):
LGTM! Thanks!

PR #2370, opened 2026-05-18
Default-typed parsing for serialized constants

ThrudPrimrose:
Adds an opt-in `default_constants` flag to `deserialize_symbolic`: untyped literals become default-typed `TypedConstant` (`5`->`int64`, `5.0`->`float64`, matching `dtypes.typeclass`), and C++ ctype casts (`double(5)`, `int(5)`) parse back into a `TypedConstant`. Default is off, so existing callers are unchanged. Also formats serialized/codegen floats in shortest round-trip form (`5.0f64` instead of `5.00000000000000f64`).

A `TypedConstant` missing `_prec` (crashes float constants inside `sympy.Min`/`Max`) is documented as a strict xfail.

### DIFF
--- a/dace/symbolic.py
+++ b/dace/symbolic.py
@@ -495,8 +495,22 @@ def _typed_constant_suffix(dtype: dtypes.typeclass) -> str:
         raise TypeError(f'Unsupported typed constant dtype {dtype.to_string()}') from ex
 
 
+def _format_float(value: float) -> str:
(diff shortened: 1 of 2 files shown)


ThrudPrimrose:
Superseded by #2371 (branch renamed to `symbolic-serialization-extensions`; the rename closed this PR since its head ref no longer exists).

PR #2371, opened 2026-05-18
Default-typed parsing for serialized constants

ThrudPrimrose:
Adds an opt-in `default_constants` flag to `deserialize_symbolic`: untyped literals become default-typed `TypedConstant` (`5`->`int64`, `5.0`->`float64`, matching `dtypes.typeclass`), and C++ ctype casts (`double(5)`, `int(5)`) parse back into a `TypedConstant`. Default is off, so existing callers are unchanged. Also formats serialized/codegen floats in shortest round-trip form (`5.0f64` instead of `5.00000000000000f64`).

A `TypedConstant` missing `_prec` (crashes float constants inside `sympy.Min`/`Max`) is documented as a strict xfail.

### DIFF
--- a/dace/symbolic.py
+++ b/dace/symbolic.py
@@ -14,6 +14,7 @@ import sympy.printing.str
 import packaging.version as packaging_version
 
 from dace import dtypes
+from dace.config import Config
 
 DEFAULT_SYMBOL_TYPE = dtypes.int32
 _NAME_TOKENS = re.compile(r'[a-zA-Z_][a-zA-Z_0-9]*')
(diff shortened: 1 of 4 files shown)


dace/symbolic.py
```diff
@@ -1414,6 +1428,18 @@ def visit_Attribute(self, node):
         return ast.copy_location(new_node, node)
 
 
+# Default typeclasses for untyped literals (matches dtypes.typeclass(int|float)).
+_DEFAULT_INTEGER_TYPECLASS = dtypes.int64
```
tbennun: This depends on the c types config entry

dace/symbolic.py
```diff
@@ -1414,6 +1428,18 @@ def visit_Attribute(self, node):
...
+_DEFAULT_FLOAT_TYPECLASS = dtypes.float64
+
+# C++ ctype spelling (e.g. ``double(5)``) -> typeclass, to parse cast fallbacks.
+_CPP_CTYPE_TO_TYPECLASS = {
```
tbennun: Seems like something we already have (or should have) in dtypes.py

tbennun (changes requested):
Minor comments. I’d use dtypes.typeclass(int) etc

dace/symbolic.py
ThrudPrimrose: Yes, I will update.

dace/symbolic.py
ThrudPrimrose: I checked the dtypes, we have the inverse of this dictionary, so I use that to generate this dictionary, can you check it again?

tbennun (approved):

PR #2372, opened 2026-05-19
Maps must have a positive step: reject descending maps, assert symbolic

ThrudPrimrose:
## Maps must have a positive step

A DaCe `Map` is an unordered, ascending iteration domain, and the CPU backend
emits an ascending loop only (`for (i = begin; i < end + 1; i += step)`). A
negative/descending step silently miscompiles. As we would need to check for `>`.
I think Loop2Map already normalizes the loops, but nothing prevents the user
from writing `dace.map[10:0:-1]` (validation passes) which I did and noticed this bug :). a

### Changes
- **`Map.validate`**: reject a *provably* descending map.
- **CPU codegen**: for a symbolic step whose sign can't be decided
  statically, emit a debug-only `assert((step) > 0)` before the OpenMP pragma.
It should not have a performance impact, as we compile with `-O3` by def.

No existing test uses a negative-step `dace.map`; positive-step maps are unchanged.

### DIFF
--- a/tests/map_positive_step_validation_test.py
+++ b/tests/map_positive_step_validation_test.py
@@ -1,0 +1,115 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
+""" Map iteration must be ascending.
+
+    A ``Map`` with a negative step fails SDFG validation. An empty
+    ``end < begin`` positive-step range is valid (it iterates zero times).
+    A symbolic step is checked at runtime by an ``assert(step > 0)`` the
+    CPU codegen emits, active only in Debug builds.
+
+    The symbolic-step assertion runs in a subprocess: a failing C
+    ``assert`` raises ``SIGABRT`` rather than a Python exception, so the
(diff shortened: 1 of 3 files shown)


tests/map_positive_step_validation_test.py
```diff
@@ -0,0 +1,115 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
+""" Map iteration must be ascending.
+
+    A ``Map`` with a negative step fails SDFG validation. An empty
+    ``end < begin`` positive-step range is valid (it iterates zero times).
+    A symbolic step is checked at runtime by an ``assert(step > 0)`` the
+    CPU codegen emits, active only in Debug builds.
+
+    The symbolic-step assertion runs in a subprocess: a failing C
+    ``assert`` raises ``SIGABRT`` rather than a Python exception, so the
```
tbennun, line 10: Sad. Need to use nanobind to propagate exceptions upwards and use `throw` in the code

tbennun (approved):

PR #2375, opened 2026-05-21
fix(frontend): give each slice memlet its own subset object

ThrudPrimrose:
## Problem

`ProgramVisitor.make_slice` builds the slice-read memlet with `Memlet.simple(array, rng, ...)`. `Memlet.simple` stores a passed-in `Subset` **by reference** (it does not copy), and `rng` is frequently the cached `Range` object handed back by the `accesses` cache on a repeated read of the same array slice.

So two sibling reads of the same slice (e.g. `arr[i, k]` in two loop bodies under a map) produce two distinct edges whose memlets **share one subset object**. That violates the invariant that each memlet owns its subset: any later in-place subset rewrite — loop-iterator renaming, symbol replacement, offsetting — on one edge silently corrupts the other.

### Reproducer

```python
N = dace.symbol('N')

@dace.program
def two_sibling_slice_reads(out: dace.float64[N, N], arr: dace.float64[N, N]):
    for i in dace.map[0:N]:
        for k in range(N):
            out[i, k] = arr[i, k]
        for k in range(N):
            out[i, k] += arr[i, k]
```

In the parsed SDFG, the two `arr[i, k]` read edges share a single `Range` subset object. A pass that rewrites a subset in place on one of them (the new test originally surfaced this through a loop-iterator rename) then corrupts the sibling.

## Fix

Deepcopy `rng` for the slice memlet's subset in `make_slice`, mirroring the `other_subset` deepcopy two lines above. This is a pure object-identity fix — subset values and the generated SDFG are unchanged.

## Test

`tests/python_frontend/slice_subset_aliasing_test.py` asserts that no two memlets in the parsed SDFG share a subset object, plus an end-to-end value check. Fails before the fix, passes after.

### DIFF
--- a/dace/autodiff/backward_pass_generator.py
+++ b/dace/autodiff/backward_pass_generator.py
@@ -740,11 +740,11 @@ class BackwardPassGenerator:
             # Create a new memlet that copies what memlet is writing to to the tmp
             new_memlet_subset = memlet.subset if memlet.data == forward_node.data else memlet.other_subset
             original_to_tmp_memlet = dace.Memlet(data=backward_node.data,
-                                                 subset=new_memlet_subset,
-                                                 other_subset=new_memlet_subset)
+                                                 subset=copy.deepcopy(new_memlet_subset),
+                                                 other_subset=copy.deepcopy(new_memlet_subset))
 
             # Remove the src_subset of the new memlet and replace the memlet in the edge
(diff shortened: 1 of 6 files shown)


tbennun (approved):
Good catch! I’m wondering why sdfg.validate didn’t catch this with the reference checker

tbennun (approved):

PR #2376, opened 2026-05-21
Use seeded test input for determinism

affifboudaoud:
Small fix for sporadic test failure caused by low floating point tolerance and unseeded random input.


tbennun (approved):
That is going to be a productivity booster, thanks for making CI more reliable!

PR #2380, opened 2026-05-22
Explicit copy and fill nodes

ThrudPrimrose:
The pass implements 0-memset and copy nodes to make copies explicit as a part of pre-codegen passes.
The auto-lowering tries to find the most suitable implementations. It also extends the patterns DaCe can correctly lower.


ThrudPrimrose:
cscs ci-run

dace/codegen/targets/cuda.py
```diff
@@ -1624,7 +1627,13 @@ def generate_devicelevel_state(self, sdfg: SDFG, cfg: ControlFlowRegion, state:
...
                 callsite_stream.write("}  // subgraph end", cfg, state.block_id)
 
-            callsite_stream.write('__gbar.Sync();', cfg, state.block_id)
+            # The grid barrier separates the states of the KERNEL's own state machine, where every
+            # thread arrives. A deeper nested SDFG runs inside one of the components above -- under
+            # a `blockIdx.x == 0 && threadIdx.x == 0` guard, or per-thread inside a map -- so only
+            # part of the grid would reach the barrier and it would never release. The enclosing
+            # state emits its own barrier after the component, which is what orders these writes.
+            if not self._below_toplevel_sdfg:
+                callsite_stream.write('__gbar.Sync();', cfg, state.block_id)
```
tbennun: not necessarily? You can just have a kernel map wrapping around a nested SDFG with states. This feels wrong.

dace/libraries/standard/nodes/copy_node.py
```diff
@@ -0,0 +1,892 @@
...
+                      parent_state: dace.SDFGState) -> bool:
+    """True if src/dst cross the CPU/GPU boundary. ``Register`` follows scope: GPU scope -> GPU,
+    else CPU."""
+    in_gpu = is_devicelevel_gpu(parent_state.sdfg, parent_state, copy_node)
+
+    src_gpu = (src_storage in GPU_RESIDENT_STORAGES) or (src_storage == dtypes.StorageType.Register and in_gpu)
+    dst_gpu = (dst_storage in GPU_RESIDENT_STORAGES) or (dst_storage == dtypes.StorageType.Register and in_gpu)
+
+    src_cpu = (src_storage in CPU_RESIDENT_STORAGES) or (src_storage == dtypes.StorageType.Register and not in_gpu)
+    dst_cpu = (dst_storage in CPU_RESIDENT_STORAGES) or (dst_storage == dtypes.StorageType.Register and not in_gpu)
```
tbennun: can you not use dtypes.GPU_STORAGE?

dace/libraries/standard/nodes/copy_node.py
tbennun, line 1: This file is quite lengthy and contains some of the implementations. Maybe worth factoring out some of the methods.

dace/codegen/targets/cuda.py
ThrudPrimrose: As discussed, the existing approach has certain cases where it will deadlock (e.g. states within a nsdfg that has conditions), I tried to match better to copy node -> by doing this: if there is a nsdfg that has only 1 state, I don't emit the gbar sync. Right now glb -> register copy + gbar.sync without the check results in in problems.

dace/libraries/standard/nodes/copy/expansions/shmem_collective.py
```diff
@@ -0,0 +1,52 @@
...
+if TYPE_CHECKING:
+    pass
+
+
+@library.expansion
+class ExpandSharedMemoryCollective(ExpandTransformation):
+    """Block-collective Shared <-> Shared/Global copy: a single Tasklet emitting
+    ``dace::CopyND<...>::Copy + __syncthreads()``, with ``_in``/``_out`` connectors matching the
+    libnode's directly (no NSDFG wrapper -- the parent kernel's ``__shared__`` array binds
+    straight in, no scope-id name mangling).
```
tbennun: This comment and the line before indicate some changes that were made during development, which are now irrelevant

dace/libraries/standard/nodes/copy/common.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+def _is_cross_cpu_gpu(src_storage: dtypes.StorageType, dst_storage: dtypes.StorageType, copy_node: "CopyLibraryNode",
+                      parent_state: dace.SDFGState) -> bool:
+    """True if src/dst cross the CPU/GPU boundary. ``Register`` follows scope: GPU scope -> GPU,
+    else CPU."""
+    in_gpu = is_devicelevel_gpu(parent_state.sdfg, parent_state, copy_node)
+
+    src_gpu = (src_storage in GPU_RESIDENT_STORAGES) or (src_storage == dtypes.StorageType.Register and in_gpu)
+    dst_gpu = (dst_storage in GPU_RESIDENT_STORAGES) or (dst_storage == dtypes.StorageType.Register and in_gpu)
```
tbennun, line 48: Why not use dtypes.GPU_STORAGES?

dace/libraries/standard/nodes/copy/common.py
```diff
@@ -0,0 +1,399 @@
...
+
+    return (src_cpu and dst_gpu) or (src_gpu and dst_cpu)
+
+
+def _copy_waives_volume_check(copy_node: "CopyLibraryNode", parent_state: dace.SDFGState) -> bool:
+    """True if a wired memlet carries ``allow_oob``, i.e. the author waived the src/dst volume check.
+
+    ``SDFGState.validate`` (``validation.py``) skips its own equal-volume check on such an edge, and
+    plain copy-edge codegen sizes the transfer from the source subset. A lifted ``CopyLibraryNode``
+    must honour the same waiver, else an SDFG that is legal as a direct copy edge stops expanding.
```
tbennun: American English spelling please

dace/libraries/standard/nodes/copy/common.py
```diff
@@ -0,0 +1,399 @@
...
+            stride_args.append(shape_strs[d])
+        if not src_static or dst_static:
+            stride_args.append(src_stride_strs[d])
+        if not dst_static:
+            stride_args.append(dst_stride_strs[d])
+
+    all_args = [in_conn, out_conn] + stride_args
+    # ``CopyND`` makes every thread copy the whole region, so a shared SOURCE is read across thread
+    # boundaries and needs the block's writes to have landed first.
+    preamble = "__syncthreads();\n" if inp.storage == dtypes.StorageType.GPU_Shared else ""
```
tbennun: You don’t always want to sync

dace/libraries/standard/nodes/copy/node.py
```diff
@@ -0,0 +1,125 @@
...
+    ``MemcpyCUDA1D``/``2D`` (``cudaMemcpyAsync``/``cudaMemcpy2DAsync``), ``MemcpyCUDANDStrided``
+    (Sequential map of ``cudaMemcpyAsync``), ``SharedMemoryCollective`` (``dace::CopyND`` +
+    ``__syncthreads()``).
+
+    Does NOT accept dynamic (Scalar) input connectors -- subset expressions must use symbols
+    already in scope at construction time, so the auto selector reasons purely from static
+    memlet subsets.
+    """
+
+    implementations = {
```
tbennun: Deprecated syntax, especially since it’s already nicely separated to files. Use the register expansion decorator on each expansion class instead

dace/libraries/standard/nodes/fill/node.py
```diff
@@ -0,0 +1,65 @@
...
+class FillLibraryNode(nodes.LibraryNode):
+    """Library node writing a constant over a contiguous output subset.
+
+    Does NOT accept dynamic (Scalar) input connectors: subset expressions must use symbols
+    already in scope, so the auto selector reasons purely from the static memlet subset.
+    """
+
+    implementations = {
+        "Auto": ExpandAuto,
+        "pure": ExpandPure,
```
tbennun: I don’t like the name pure and prefer “native”. However, other library nodes already use “pure” so I’m not sure. Thoughts?

dace/libraries/standard/nodes/fill/node.py
```diff
@@ -0,0 +1,65 @@
...
+        "CUDA": ExpandCUDA,
+        "CPU": ExpandCPU,
+        "tasklet": ExpandTasklet
+    }
+    default_implementation = 'Auto'
+
+    OUTPUT_CONNECTOR_NAME = OUTPUT_CONNECTOR_NAME
+
+    # dtype=None takes any Python constant; numpy scalars are normalized by Property.__set__.
+    value = properties.Property(dtype=None, default=0, desc='The constant written over the subset.')
```
tbennun, line 27: So value can only be a compile time constant or a symbol? No optional scalar in connector option?

dace/sdfg/infer_types.py
```diff
@@ -261,6 +261,12 @@ def _determine_schedule_from_storage(state: SDFGState, node: nodes.Node) -> Opti
             continue
         constraints.add(sched)
 
+    # Copy/Memset library nodes legitimately bridge storages; schedule on the GPU if involved.
+    from dace.libraries.standard.nodes.copy import CopyLibraryNode
```
tbennun, line 265: Why do you need this corner case in type inference? Shouldn’t this already be covered by the surrounding access nodes?

dace/transformation/passes/insert_explicit_copies.py
```diff
@@ -0,0 +1,314 @@
...
+    """Replaces implicit copy patterns with ``CopyLibraryNode`` instances: direct
+    ``AccessNode -> AccessNode`` edges (View endpoints included), and stage-in/stage-out through
+    chained ``MapEntry``/``MapExit`` (libnode placed inside the map scope)."""
+
+    # Other storages (TensorCore_*, FPGA_*, Snitch_*) use their own codegen ``copy_memory`` hook.
+    _STANDARD_STORAGES = (CPU_RESIDENT_STORAGES | GPU_RESIDENT_STORAGES
+                          | {dtypes.StorageType.Default, dtypes.StorageType.Register})
+
+    def modifies(self) -> ppl.Modifies:
+        return ppl.Modifies.States | ppl.Modifies.Nodes | ppl.Modifies.Edges
```
tbennun, line 89: Does it really modify states?

tbennun (changes requested):
Looks very good

dace/libraries/standard/nodes/copy_node.py
ThrudPrimrose: Because, GPU_STORAGES is this, and it is imho misleading. But changing it breaks 10 other places and I would prefer to not do that as part of this:
```
# A subset of on-GPU storage types
GPU_STORAGES = [
    StorageType.GPU_Shared,
]
```

ThrudPrimrose: (Taken from dace/dtypes.py)

dace/libraries/standard/nodes/copy_node.py
ThrudPrimrose: I have addressed that and expansions are in separate folders, it is ok now, right?

dace/libraries/standard/nodes/copy/common.py
ThrudPrimrose: You can check the answer above,`dtypes.GPU_STORAGES` only has `GPU_SHARED` in it, and it is not useful here.

dace/libraries/standard/nodes/copy/common.py
ThrudPrimrose: True, but I couldn't find a better solution, because if I don't sync as part of the copy, the underlying codegen doesn't do it (it just doesnt do it properly, and in existing passes it was done by adding a .sync property to loops if I remember correctly)

ThrudPrimrose: New codegen does a better job syncing once for multiple consecutive copies.

dace/libraries/standard/nodes/copy/expansions/shmem_collective.py
ThrudPrimrose: True, it was using copynd instead of global to shared for some time. I will update

dace/libraries/standard/nodes/copy/node.py
ThrudPrimrose: Okay, I'll update.

dace/libraries/standard/nodes/fill/node.py
ThrudPrimrose: Auto-opt hardcodes the name "pure". If we don't provide "pure" expansions, all auto-opt SDFGs that have copy/fill nodes might fail. I don't like the name pure much either, I would propose a small style-only PR later that replaces "pure" with a different name also we should provide these magic-words as constants (can't provide real constants in Python, but we can provide the all-capital variables in dtypes, so that we don't have hardcoded strings in optimization pipelines)

E.g. see auto_optimize.py
```
    # general nodes
    for node, _ in sdfg.all_nodes_recursive():
        if isinstance(node, nodes.LibraryNode):
            # NOTE: LibraryNodes with sequential schedule on GPU must be expanded to CUDA kernel-compatible code.
            # NOTE: Pure implementations are a safe choice for now but this should be revisited in the future.
            if device == dtypes.DeviceType.GPU and node.schedule == dtypes.ScheduleType.Sequential:
                node.implementation = "pure"
                continue
            for impl in implementation_prio:
                if impl in node.implementations:
                    if isinstance(
                            node,
                            dace.libraries.standard.nodes.reduce.Reduce) and node.implementation == 'CUDA (block)':
                        continue
                    node.implementation = impl
                    break
```

ThrudPrimrose: I think "native" makes more sense than "pure".

dace/sdfg/infer_types.py
ThrudPrimrose: This was added manually after things failed (this time not Claude overengineering). The type inference fails when a libnode touches CPU<->GPU together. It puts them to CPU schedule. I find this way to be the least intrusive way to instruct the inference.

dace/transformation/passes/insert_explicit_copies.py
ThrudPrimrose: I thought it would be safe to add States because we will eventually expand to nested SDFGs, and will add states inside the nsdfg?

dace/libraries/standard/nodes/fill/node.py
ThrudPrimrose: Good remark, I forgot about this because I have not really seen a broadcast from a scalar in the examples I had on Fortran SDFGs, this is an easy addition I will add.

dace/libraries/standard/nodes/copy/common.py
tbennun: we somehow need a TODO or some property to avoid this extraneous sync, especially in new codegen.

dace/libraries/standard/nodes/fill/node.py
tbennun: issue / PR for later ("good first issue" even)

dace/sdfg/infer_types.py
tbennun: still ugly but ok

tbennun (approved):
LGTM!

Issue #2382, opened 2026-05-27
Influence of Unused Transient on the Signature


philip-paul-mueller:
As far as I can tell the recent [fix to `LoopToMap`](https://github.com/spcl/dace/pull/2349) introduced a change in behaviour.
The PR also modified how `SDFG::_used_symbols_internal()` works and in my opinion changed it in a non desirable way.

Consider the following script:
```python=
import dace
import pytest

def _make_sdfg_with_unused_transient(create_unused_transient: bool) -> dace.SDFG:
    sdfg = dace.SDFG("unused_transient")
    state = sdfg.add_state()

    for name in "ab":
        sdfg.add_array(
                name,
                shape=(10,),
                dtype=dace.float64,
                transient=False,
        )

    sdfg.add_symbol("x_shape", dace.int32)
    if create_unused_transient:
        sdfg.add_array(
                "x",
                shape=("x_shape",),
                dtype=dace.float32,
                transient=True,
        )

    state.add_mapped_tasklet(
            "map",
            map_ranges={"__i": "0:10"},
            inputs={"__in": dace.Memlet("a[__i]")},
            code="__out = __in + 1.90",
            outputs={"__out": dace.Memlet("b[__i]")},
            external_edges=True,
    )

    return sdfg

if __name__ == "__main__":
    create_unused_transient = True
    sdfg: dace.SDFG = _make_sdfg_with_unused_transient(create_unused_transient)

    print(f"Free Symbols: {sdfg.free_symbols}")
    print(f"arglist(): {sdfg.arglist()}")
    print(f"signature_arglist(): {sdfg.signature_arglist()}")
    print(f"signature_init_arglist(): {sdfg.init_signature()}")
```

As you can see depending on `create_unused_transient` the SDFG may or may not contain data descriptor with name `x`, that when it is present is not used.
Furthermore, the array has a symbolic size of `x_shape`, that symbol, however, is always present.
If you run that script using DaCe `2.0.0a3`, then `x_shape` will _not_ be part of the signature.

But if you use current `main` (`a6717b608f497190205d`) then `x_shape` will part of the signature if `create_unused_transient` is `True`.
However, if `create_unused_transient` is `False` then it will not be present.
It is important to notice that even if `x_shape` is present in the signature it is not used, i.e. the array is not allocated.

My question now is what is the correct behaviour?
In my opinion `x_shape` should not be part of the signature and this is an unintended side effect.

PR #2385, opened 2026-05-31
Transformation bugfixes

ThrudPrimrose:
# Bugfixes

While working on the canonicalization pipeline and ensuring that we can automatically parallelize CloudSC, I have encountered many bugs and fixed many of them.

---

### `symbols_defined_at` does not include enclosing LoopRegion vars
- **Where:** `dace/sdfg/state.py::ControlFlowBlock.symbols_defined_at`.
- **How to reproduce:** Query `symbols_defined_at` on any node inside a `LoopRegion` body.
- **Pre-fix:** The enclosing loop variable (e.g. `i`) is missing from the returned set.
- **Post-fix:** The loop variable is reported as defined.
- **Downstream effect:** Memlet propagation's widening fallback consults `symbols_defined_at`; with `i` absent it treats per-iteration subsets like `arr[*, jk-1, *]` as un-parameterised and widens them to whole-array unions, which then blocks `LoopToMap`.
- **Fix:** Walk up considers LoopRegions' smybols too.
- **Tests:** `tests/sdfg/symbols_defined_at_test.py`.

### Output-array arg dropped when outgoing memlet from MapExit is source-relative
- **Where:** `dace/sdfg/state.py::DataflowGraphView.unordered_arglist`, the `AccessNode/CodeNode -> ExitNode` branch.
- **How to reproduce:** Build a map whose outgoing edge from `MapExit` to an outer AccessNode (`D`) carries a memlet whose `data` field still names an inner transient (`tmp`) rather than `D`. (This can appear because of `oterh_subset` copies etc.).
- **Pre-fix:** `D` is missing from the arglist. The code took `oedge.data.data == 'tmp'`; `tmp` is already in `descs` (the AccessNode `tmp` is in-scope), so nothing reached `additional_descs`.
- **Post-fix:** `D` (and its stride/shape symbols) is in the arglist.
- **Fix:** Resolve the destination from the graph -- `memlet_path(oedge)[-1].dst` (equivalently `memlet_tree(oedge).root().edge.dst`) -- and use that node's `.data`, with `oedge.data.data` as fallback when the path does not terminate at an `AccessNode`.
- **Tests:** New focused unit test `tests/sdfg/exit_arglist_test.py::test_arglist_resolves_outer_destination_from_source_relative_outgoing_memlet`. End-to-end GPU compile of the same shape: `tests/codegen/argument_signature_test.py::test_argument_signature_test`.

### Veclen lookup crashed on non-AccessNode endpoints
- **Where:** `dace/sdfg/validation.py::validate_state`, the edge dimensionality check + the two View-exception branches.
- **How to reproduce:** Construct any edge with `other_subset` set where one endpoint is a scope node (`NestedSDFG` / `MapEntry` / `MapExit` / `ConsumeEntry` / `ConsumeExit`), then call `sdfg.validate()`.
- **Pre-fix:** `AttributeError: 'NestedSDFG' object has no attribute 'data'` raised from `sdfg.arrays[src_node.data].veclen`. The pre-fix code assumed both endpoints were `AccessNode`s; scope nodes route data through connectors and have no `.data` field.
- **Post-fix:** Validation reaches a verdict.
- **Fix:** Read each side's `veclen` only when its endpoint is an `AccessNode`; default to `1` for scope nodes. The same guard applied to the two View-exception branches downstream.
- **Tests:** `tests/sdfg/validation/subset_size_test.py::test_veclen_lookup_guarded_on_non_accessnode_endpoint`.

### Unused transient's shape symbol leaked into signature (#2382)
- **Where:** `dace/sdfg/sdfg.py::SDFG._used_symbols_internal`.
- **How to reproduce:** Build two identical SDFGs differing only by one declaring an extra unused transient whose shape uses a symbol (e.g. `x_shape`). Compare `used_symbols(all_symbols=False)` and `arglist()`.
- **Pre-fix:** The SDFG with the unused transient gets `x_shape` added to both its used-symbols set and its arglist -- merely declaring an unused array changes the SDFG's signature.
- **Post-fix:** Both SDFGs report identical `used_symbols` and `arglist`.
- **Fix:** Gate the extent-symbol expansion on `read_and_write_sets()` rather than on every declared array.
- **Tests:** `tests/sdfg/free_symbols_test.py::test_unused_array_does_not_leak_shape_symbol`.

---

### `sympy_numeric_fix` demoted integer-valued floats to int

The previous merge to main fixed the error on roundtrip but, still it appeared on CloudSC coming from Python frontend. 

- **Where:** `dace/symbolic.py::sympy_numeric_fix` (plus a consolidation of `_print_Float` / `_format_float`).
- **How to reproduce:** Call `sympy_numeric_fix` on a finite Python / numpy / sympy float like `1.0`; then construct a `sympy.Min` of that value and a symbol.
- **Pre-fix:** `1.0` is returned as Python `int(1)`; `Min(zanew, 1)` is a type-mismatched int-Min that SymPy re-canonicalises through the integer branch and silently truncates downstream.
- **Post-fix:** `1.0` is returned as `sympy.Float(1.0)`; `Min(zanew, 1.0)` preserves float typing end-to-end.
- **Fix:** Preserve any finite Python / numpy / sympy float as `sympy.Float`. Also fixes `inf.0` and 17-digit Float print bugs surfaced by the same path.
- **Tests:** `tests/symbolic_print_test.py` + `tests/symbolic_roundtrip_test.py`.

### `_print_floor` emitted a literal `floor(...)` instead of integer division in cpp_mode

I guess users should never use // and use `int_floor`, but this fix should help an unaware user from getting incorrect CPP without any warning or error.

- **Where:** `dace/symbolic.py::DaceSympyPrinter._print_floor`.
- **How to reproduce:** Form `(LEN - 1) // 8` on an integer symbol `LEN` (sympy normalises this to `floor(LEN/8 - 1/8)`), then print with `DaceSympyPrinter(cpp_mode=True)`.
- **Pre-fix:** The printer emits the literal `floor(LEN/8 - 1/8)`. C++ integer-divides `1 / 8` to `0`, so the bound effectively collapses to `LEN / 8` and the loop overshoots by one.
- **Post-fix:** The printer emits `((LEN - 1) / (8))` -- a single integer division.
- **Fix:** Detect the common-denominator fraction-difference via `arg.together().as_numer_denom()` and emit a single integer-division pair.
- **Tests:** `tests/symbolic_print_test.py::test_cpp_floor_of_fraction_difference_recombines_to_integer_division` -- asserts the cpp output contains no `floor(` and no `1/8`.

---

### TrivialTaskletElimination misoriented the surviving memlet when source was a MapEntry
- **Where:** `dace/transformation/dataflow/trivial_tasklet_elimination.py`, the `expr_index=1` (MapEntry-source) branch.
- **What the pass does:** Eliminates a passthrough copy tasklet by folding `read --in_edge--> tasklet --out_edge--> write` into a single replacement edge from `read` to `write`. The replacement memlet must describe the side that owns the connector the edge actually leaves through.
- **How to reproduce:** Build a passthrough copy tasklet whose source is a `MapEntry` reading an offset slice (e.g. `a[i + 1]`) and whose destination is a transient scalar (`a_idx[0]`). Apply `TrivialTaskletElimination`; inspect the surviving edge's `.data` and `.subset`.
- **Pre-fix:** Surviving edge has `data='a_idx', subset='0', other_subset='i + 1'` -- the pass reused the write-side memlet and stuffed the read offset into `other_subset`. The SDFG validates and runs correctly because the dataflow is still consistent across the two subsets.
- **Post-fix:** Surviving edge has `data='a', subset='i + 1', other_subset='0'`.
- **Fix:** When `expr_index=1` (MapEntry as per the transformation) keep `in_edge.data` (the read-side memlet) as the surviving memlet, with the write subset moved into `other_subset`.
- **Tests:** `tests/transformations/trivial_tasklet_elimination_test.py::test_trivial_tasklet_map_source_preserves_offset_subset` -- asserts the surviving edge's `.data == 'a'` and `.subset == 'i + 1'`. Pre-fix the test fails with `AssertionError: surviving edge must describe the read data 'a', got 'a_idx'`.

### MapFusionVertical InOut split + matching ConditionFusion parent-fixup
- **Where:** `dace/transformation/dataflow/map_fusion_vertical.py` (`_split_inout_for_intermediate`) + `dace/transformation/interstate/condition_fusion.py` (`fuse_consecutive_conditions`).
- **Symptom:** Fusing maps that share a `NestedSDFG` `InOut` connector for the intermediate name merged the two sides into one connector and clobbered the upstream feedthrough; the matching parent-block bookkeeping in `ConditionFusion.fuse_consecutive_conditions` had the analogous clobber.
- **Fix:** When an intermediate's data name matches an `InOut` connector of a `NestedSDFG` in the producer's body, split the connector inside the NestedSDFG (rename the inner read-side accesses to a fresh array bound to a new input connector) so the standard rename machinery rewires the output-only connector without a mismatched-InOut validation error. Apply the same split-aware bookkeeping in `ConditionFusion`'s parent fixup.
- **Tests:** `tests/transformations/map_fusion_vertical_test.py`.

### ConditionFusion: use `isinstance(ControlFlowBlock)` instead of `hasattr(node, 'sdfg')`
- **Where:** `dace/transformation/interstate/condition_fusion.py`, the post-rewrite parent-fixup walk in `fuse_nested_conditions`.
- **Symptom (not directly probed; analysed from the code):** The walk does `if hasattr(node, 'sdfg'): node.sdfg = parent.sdfg` over `all_nodes_recursive()`. `hasattr(node, 'sdfg')` is `True` for `NestedSDFG` nodes too -- but `NestedSDFG.sdfg` is the **inner** SDFG. The assignment overwrites the inner-SDFG slot with the outer container, producing a graph cycle that infinite-recurses `all_nodes_recursive` on the next walk.
- **Fix:** `isinstance(node, ControlFlowBlock)` matches `SDFGState` / `ControlFlowRegion` / `ConditionalBlock` (whose `.sdfg` legitimately names the containing SDFG) and skips `NestedSDFG`. Defensive type tightening matching the same pattern already corrected in `fuse_consecutive_conditions`.
- **Tests:** `tests/transformations/interstate/condition_fusion_test.py` (existing suite; the specific NestedSDFG-inside-condition shape is not directly exercised here -- this is a defensive same-family tightening rather than a regression-catcher fix).

### RedundantArrayCopyingIn folded chains it should have refused
- **How to reproduce (partial copy):** Build a four-node chain `A -> B -> C -> D` where one of the copies covers only part of the array (e.g. `[0:2]` of a size-4 array). Run the pass.
- **Pre-fix:** The pass applies and the output is corrupted: the redirected writers' subsets are preserved, but the partial copy that originally restricted them is gone, so positions the chain never wrote get overwritten.
- **How to reproduce (extra consumer):** Build the chain `A -> B -> C -> D` plus a second consumer `C -> E`. Run the pass.
- **Pre-fix:** `C` is removed; `E` is left isolated; validation fails with `InvalidSDFGNodeError: Isolated node E`.
- **Post-fix (both cases):** The pass refuses (`applied == 0`) and the SDFG stays correct.
- **Fix:** Two guards in `can_be_applied`: `out_degree(med_array) == 1` (no second consumer) AND `_is_full_copy` on both `in -> med` and `med -> out` (each side's subset equals `Range.from_array(<that side's desc>)`). Either failure refuses the fold.
- **Tests:** `tests/transformations/redundant_copy_test.py::test_in_failure_partial_copy` and `::test_in_failure_extra_consumer` -- both fail pre-fix with the symptoms above, pass post-fix.

### AugAssignToWCR did not detect the copy-wrapped read-modify-write shape
- **Symptom:** The frontends lower an in-place update like `arr[S] = arr[S] + x` as a 4-node chain `arr[S] -> copy_in -> tasklet -> copy_out -> arr[S]` -- the accumulator slice is materialised into a scalar transient before the combining tasklet and copied back after it. The existing `AugAssignToWCR` matcher only recognised the direct shape and missed the copy-wrapped one, so loop-carried reductions stayed sequential and could not parallelize via `LoopToMap`.
- **Fix:** A new `expr_index=2` matches the chain and rewrites it to a WCR write on `arr[S]`. Supports `+`, `*`, `min`, `max`, left-sub.
- **Tests:** `tests/transformations/wcr_conversion_test.py`.

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -1427,21 +1427,47 @@ class SDFG(ControlFlowRegion):
                                                 free_syms=free_syms,
                                                 used_before_assignment=used_before_assignment,
                                                 with_contents=with_contents)
-        # Expand array-descriptor stride/shape/offset symbols into the free
-        # set. Without this, a ``ConditionalBlock`` guard or memlet subset
-        # referencing ``A[i, j]`` leaves the symbols used in ``A`` 's strides
-        # out of the computed free-symbol set, causing
-        # ``generate_nsdfg_header`` to emit a nested function signature
-        # missing those symbols, ceating an invalid SDFG.
+        # Expand the stride/shape/offset symbols of arrays that are referenced
+        # in a control-flow code block (a ``ConditionalBlock`` guard or a
+        # ``LoopRegion`` condition/range) into the free set. Such an array's
+        # extent symbols are otherwise dropped, because the code block records
+        # only the array name -- not the symbols needed to compute its layout --
+        # which makes ``generate_nsdfg_header`` emit a nested function signature
+        # missing those symbols, creating an invalid SDFG. Memlet- and
+        # access-node-referenced arrays already contribute their extent symbols
+        # through the contents analysis above, and interstate-edge-referenced
+        # arrays are expanded by the parent ``ControlFlowRegion`` analysis; only
+        # code-block references need this extra step.
         res_free, res_defined, res_before = result
         if with_contents:
-            for desc in self.arrays.values():
-                res_free |= {str(s) for s in desc.used_symbols(all_symbols)}
+            for name in self._arrays_used_in_code_blocks(all_symbols):
+                res_free |= {str(s) for s in self.arrays[name].used_symbols(all_symbols)}
             # Don't drag in symbols that are genuinely defined inside this
             # SDFG (e.g., LoopRegion loop variables); keep only the ones
             # outside ``defined_syms``.
             res_free -= res_defined
         return res_free, res_defined, res_before
+
+    def _arrays_used_in_code_blocks(self, all_symbols: bool) -> Set[str]:
+        """
+        Returns the names of data descriptors that are referenced inside the
+        control-flow code blocks of this SDFG (conditional-block guards and
+        loop conditions/ranges), as opposed to merely being declared.
+
+        An array is considered referenced when its name appears as a free
+        symbol of the code block once it has been parsed into a symbolic
+        expression. Nested SDFGs are not traversed: each computes the extent
+        symbols of its own referenced arrays.
+
+        :param all_symbols: If False, only symbols that are needed as arguments
+                            (used in the generated code) are considered.
+        :returns: A set of data-descriptor names referenced in code blocks.
+        """
+        array_names = self.arrays.keys()
+        used = set()
+        for cfr in self.all_control_flow_regions(recursive=False):
+            used |= cfr.used_symbols(all_symbols=all_symbols, with_contents=False) & array_names
+        return used
 
     def get_all_toplevel_symbols(self) -> Set[str]:
         """
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -1427,47 +1427,18 @@ class SDFG(ControlFlowRegion):
                                                 free_syms=free_syms,
                                                 used_before_assignment=used_before_assignment,
                                                 with_contents=with_contents)
-        # Expand the stride/shape/offset symbols of arrays that are referenced
-        # in a control-flow code block (a ``ConditionalBlock`` guard or a
-        # ``LoopRegion`` condition/range) into the free set. Such an array's
-        # extent symbols are otherwise dropped, because the code block records
-        # only the array name -- not the symbols needed to compute its layout --
-        # which makes ``generate_nsdfg_header`` emit a nested function signature
-        # missing those symbols, creating an invalid SDFG. Memlet- and
-        # access-node-referenced arrays already contribute their extent symbols
-        # through the contents analysis above, and interstate-edge-referenced
-        # arrays are expanded by the parent ``ControlFlowRegion`` analysis; only
-        # code-block references need this extra step.
+        # A used array needs its stride/shape/offset symbols in the free set, but a
+        # merely-declared one must not leak its shape symbol into the signature
+        # (issue #2382). ``read_and_write_sets`` already reports exactly the arrays
+        # that are used -- read or written, including those referenced only by a
+        # code-block guard/condition -- so expand the extent symbols of those alone.
         res_free, res_defined, res_before = result
         if with_contents:
-            for name in self._arrays_used_in_code_blocks(all_symbols):
+            read_set, write_set = self.read_and_write_sets()
+            for name in (read_set | write_set) & self.arrays.keys():
                 res_free |= {str(s) for s in self.arrays[name].used_symbols(all_symbols)}
-            # Don't drag in symbols that are genuinely defined inside this
-            # SDFG (e.g., LoopRegion loop variables); keep only the ones
-            # outside ``defined_syms``.
-            res_free -= res_defined
+            res_free -= res_defined  # drop symbols defined inside (e.g. loop vars)
         return res_free, res_defined, res_before
-
-    def _arrays_used_in_code_blocks(self, all_symbols: bool) -> Set[str]:
-        """
-        Returns the names of data descriptors that are referenced inside the
-        control-flow code blocks of this SDFG (conditional-block guards and
-        loop conditions/ranges), as opposed to merely being declared.
-
-        An array is considered referenced when its name appears as a free
-        symbol of the code block once it has been parsed into a symbolic
-        expression. Nested SDFGs are not traversed: each computes the extent
-        symbols of its own referenced arrays.
-
-        :param all_symbols: If False, only symbols that are needed as arguments
-                            (used in the generated code) are considered.
-        :returns: A set of data-descriptor names referenced in code blocks.
-        """
-        array_names = self.arrays.keys()
-        used = set()
-        for cfr in self.all_control_flow_regions(recursive=False):
-            used |= cfr.used_symbols(all_symbols=all_symbols, with_contents=False) & array_names
-        return used
 
     def get_all_toplevel_symbols(self) -> Set[str]:
         """
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -1643,6 +1643,20 @@ class SDFGState(OrderedMultiDiConnectorGraph[nd.Node, mm.Memlet], ControlFlowBlo
             for e in sdfg.edges():
                 symbols.update(e.data.new_symbols(sdfg, symbols))
 
+        # Add the loop variables of the control-flow loops enclosing this state,
+        # outermost first. Without this only global, inter-state-edge and dataflow-scope
+        # (map) symbols are seen; a node inside a LoopRegion must also see the loop
+        # variable as defined -- e.g. so memlet propagation keeps a ``jk``-indexed
+        # nested-SDFG access parametric instead of widening it to the whole array.
+        enclosing_loops = []
+        cfg = self.parent_graph
+        while cfg is not None:
+            if isinstance(cfg, LoopRegion) and cfg.loop_variable:
+                enclosing_loops.append(cfg)
+            cfg = cfg.parent_graph
+        for loop in reversed(enclosing_loops):
+            symbols.update(loop.new_symbols(symbols))
+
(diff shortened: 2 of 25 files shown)


ThrudPrimrose:
## SymbolPropagation -- what was fixed

Six correctness fixes in `dace/transformation/passes/symbol_propagation.py` that together make the pass safe to run inside `FixedPointPipeline` and correct on CloudSC / TSVC patterns.

1. **Same-edge race guard + cross-CFG assert + termination cap.** A carried value referencing a symbol the current edge also reassigns is no longer propagated.

2. **Simultaneous-RHS resolve (cyclic-swap).** Multiple assignments on one edge (e.g. `{a: b, b: a}`) are now resolved against the pre-edge table simultaneously. 

3. **Return propagated set (or `None`) so `FixedPointPipeline` converges.** Pre-fix `apply_pass` returned `set()` on a no-op, which the driver reads as "modified" and re-runs forever.

4. **Dead-iedge sweep after substitution.** Once `k_plus_1 = klev + 1` is substituted into every use, the defining assignment is dead but was left behind; now cleaned.

5. **Substitute into array descriptors before the dead-iedge sweep.** A propagated symbol must reach `Array.shape` / `strides` BEFORE its iedge assignment is dropped, else the descriptor references an undefined symbol and the SDFG fails validation.

6. **Invalidate carried value whose RHS reads a symbol the current edge reassigns.** `k = j + 1` carried in from the predecessor is stale for the block past an edge that reassigns `j` (e.g. `j = k + 1`); pre-fix produced off-by-two indices (TSVC s128 / CloudSC).


## LoopToMap -- what was fixed

Four fixes in `dace/transformation/interstate/loop_to_map.py`, each addressing a distinct false-refusal or false-acceptance in `can_be_applied`.

1. **Iteration-independence through a `NestedSDFG` body.** A loop body that is a `NestedSDFG` propagates a whole-array external write memlet (`b[0:N]`) if the write is dynamic. This can happen if the array is written to inside an if block, which hides the per-iteration inner write. New `_nested_writes_iter_indexed` walks past the connector, rewrites inner write subsets through the node's `symbol_mapping` into the outer iteration symbol, and applies the same `a*i + b` check (recursive across nested NestedSDFGs, conservative). Effect: `LoopToMap -> MapToForLoop -> LoopToMap` recovers the map.

### What the fix does

When the standard `_check_range(dst_subset, a, itersym, b, step)` fails and the source of the edge is a `NestedSDFG`, the fix calls a new helper `_nested_writes_iter_indexed(nsdfg, conn, itersym, a, b, step)` which looks PAST the connector instead of trusting the external memlet:

  * Walk into `nsdfg.sdfg` and find every write edge to an AccessNode whose `.data` equals the outer connector name (`"b"`). Those are the inner per-iteration writes the external memlet was a union over.

  * For each such inner write edge, take its `dst_subset` (e.g. `b[i]` inside the NestedSDFG, where `i` is the INNER symbol name).

  * Translate that subset into the OUTER iteration symbol via `nsdfg.symbol_mapping`. The mapping looks like `{i_inner: i_outer, N: N}`; substituting it into `b[i_inner]` yields `b[i_outer]`.

  * Apply the same `_check_range` against the translated subset. If it matches `a*i + b`, that inner write is per-iteration unique.

  * If a deeper `NestedSDFG` sits inside, recurse, composing the `symbol_mapping`s on the way down.

2. **Refuse when the loop range reads a body-assigned symbol.** A loop whose `start` / `end` / `step` references a symbol the body reassigns would, if parallelised, see iterations execute with each other's bounds.

3. **Shared affine iteration-var dimension is cross-iteration disjoint.** Two writes that index a dimension by the same injective `a*i + b` (`a != 0`) collide only when their iterations coincide -- they cannot race across iterations. Parallelises scatter writes like `A[0, idx, i] = ...; A[idx, 0, i] = ...` despite the opaque `idx`.

4. **Peeled fixed-read pattern.** A loop-INVARIANT read (no iteration symbol) is no longer an automatic conflict in `test_read_memlet`; it's a conflict only when it actually overlaps a write. `a[0]` vs `a[0:N]` stays refused; `a[0]` vs `a[1:N]` is now correctly parallelisable (Which was refused before).

PR #2388, opened 2026-06-03
Make Symbol Serialisation Thread Safe

philip-paul-mueller:
[PR#2366](https://github.com/spcl/dace/pull/2366) introduced a new global that was used during serialization. However, this fails when multiple threads are active, which is not the default mode of operation (except in GT4Py and the [DaCe tests themselves](https://github.com/spcl/dace/pull/2126)).
This PR made the serialization thread safe.


ThrudPrimrose (approved):
LGTM

Issue #2391, opened 2026-06-04
PR#2366 (Typesafe Symbols) Breaks Roundtrip


philip-paul-mueller:
PR#2366, that introduced typesafe symbol serialization breaks roundtrip, i.e. an SDFG after serialization and deserialization is no longer the same.
The attached reproducer constructs a copy between two AccessNodes.
Before serialization `memlet_copy_to_absolute_strides()` reports that the `copy_shape` has length 1, i.e. it means that it can be expressed as a 1d `memcpy()`.
However, after it is serialized and restored `copy_shape` has length 3, which means that it can not be expressed as a 1d copy anymore.

I run the reproducer using different versions of DaCe, which confirms that it is indeed PR#2366:
- `main` (`a6717b608f4`): Fails
- PR#2366 (`17be66ec4a`): Fails
- PR#2378 (`4200db54`, i.e. `PR#2366~1`): Passes

```python=
import sys

import dace
from dace.sdfg import nodes as dace_nodes
from dace.sdfg import graph as dace_graph
from dace.symbolic import pystr_to_symbolic as s2s
from dace.codegen.targets import cpp as dace_cpp


def _make_sdfg() -> tuple[dace.SDFG, dace.SDFGState, dace_nodes.AccessNode, dace_graph.MultiConnectorEdge[dace.Memlet], dace_nodes.AccessNode]:
    sdfg = dace.SDFG("copy_sdfg")
    state = sdfg.add_state()

    for sname in ["__out_IDim_range_0", "__out_IDim_range_1", "__out_IDim_stride", "__out_JDim_range_0", "__out_JDim_range_1", "__out_JDim_stride", "__out_KDim_range_0", "__out_KDim_range_1", "__out_KDim_stride"]:
        sdfg.add_symbol(sname, dace.int32)

    sdfg.add_array(
            "a",
            dtype=dace.float64,
            transient=False,
            shape=(
                s2s("max(0, -__out_IDim_range_0 + __out_IDim_range_1)"),
                s2s("max(0, -__out_JDim_range_0 + __out_JDim_range_1)"),
                s2s("max(0, -max(0, __out_KDim_range_0) + __out_KDim_range_1)"),
            ),
            strides=(
                1,
                s2s("max(0, -__out_IDim_range_0 + __out_IDim_range_1)"),
                s2s("max(0, -__out_IDim_range_0 + __out_IDim_range_1)*max(0, -__out_JDim_range_0 + __out_JDim_range_1)"),
            ),
    )
[...]

PR #2392, opened 2026-06-04
Fix Issue #2391

ThrudPrimrose:
pystrings that have Abs, Min and Max (capital start) are considered user functions until deserialized, creating issues. (see: #2391).

This Pr fixes that.


tests/symbolic_serialization_test.py
```diff
@@ -376,7 +373,7 @@ def test_minmax_with_float_literal_roundtrip(op, literal):
 
 def test_minmax_with_ctype_cast_int_literal_normalizes_then_stable():
     first = symbolic.serialize_symbolic(symbolic.deserialize_symbolic('Max(double(7), 5.0)'))
-    assert first == 'Max(7f64, 5.0)'
+    assert first == '7f64'
```
tbennun, line 387: This seems to indicate some simplification happens in serialization. Why

tests/symbolic_serialization_test.py
```diff
@@ -364,9 +364,6 @@ def test_cpp_ctype_cast_parses_to_typed_constant(text, ctype):
     '5.0f64',
     '5f32',
     '7f64',
-    '(3.0 + 4.0j)c128',
-    '(3.0 - 4.0j)c128',
-    '(3.0 + 4.0j)c64',
```
tbennun, line 369: Why are these removed

dace/symbolic.py
```diff
@@ -129,9 +129,9 @@ def __new__(cls, name=None, dtype=None, **assumptions):
         if 'integer' in assumptions or not numpy.any(is_integer):
             # Using __xnew__ as the regular __new__ is cached, which leads
             # to modifying different references of symbols with the same name.
-            self = sympy.Symbol.__xnew__(cls, name, **assumptions)
+            self = sympy.Symbol.__xnew__(cls, name, **{**assumptions, 'commutative': True})
```
tbennun: No documentation as to why this assumption is added. Please explain.

tbennun (changes requested):

tests/symbolic_serialization_test.py
ThrudPrimrose: Sympy by default tries to simplify Max and Min (as they are not user functions)

tests/symbolic_serialization_test.py
ThrudPrimrose: Min and Max are not defined for complex numbers, it was working because previously Min and Max were user functions and were treated differently than min and max. Writing min/max in AST was treated as min and max. Min and Max were treated as user functions.

ThrudPrimrose: With the fix for Min/Max these crash

dace/symbolic.py
ThrudPrimrose: I added it because, without that, the serialization was changing in roundtrip, but I think that might be another issue.

Also, adding ceiling alongside ceil also changes the behaviour when roundtripping because sympy simplifies it.  The simplification is correct when I check the CI output.

ThrudPrimrose: I have not found a good solution to fix the current issue, because imho, ceiling/ceil/Min/Max/max/min should all be detected by sympy.

ThrudPrimrose:
I think the issue was missing sorts. I guess the non-determinism was coincidentally hidden, and adding "Min", "Max" and "ceiling" as first-class Sympy functions triggered it.

ThrudPrimrose:
Somehow, somewhere `commutative` assumption is explicitly stored. I can't find where...

``` 
FAILED tests/symbolic_roundtrip_test.py::test_symbolic_expression_serialization_preserves_integerness - AssertionError: assert 'Max(Integer(...teger=True)))' == 'Max(Integer(.....
>       assert sympy.srepr(original) == sympy.srepr(restored)
E       AssertionError: assert 'Max(Integer(...teger=True)))' == 'Max(Integer(...teger=True)))'
E         
E         - Max(Integer(0), Add(Mul(Integer(-1), symbol('__out_IDim_range_0', integer=True)), symbol('__out_IDim_range_1', integer=True)))
E         + Max(Integer(0), Add(Mul(Integer(-1), symbol('__out_IDim_range_0', commutative=True, integer=True)), symbol('__out_IDim_range_1', commutative=True, integer=True)))
E         ?                                                                   ++++++++++++++++++                                             ++++++++++++++++++
```

ThrudPrimrose:
> Somehow, somewhere `commutative` assumption is explicitly stored. I can't find where...
> 
> ```
> FAILED tests/symbolic_roundtrip_test.py::test_symbolic_expression_serialization_preserves_integerness - AssertionError: assert 'Max(Integer(...teger=True)))' == 'Max(Integer(.....
> >       assert sympy.srepr(original) == sympy.srepr(restored)
> E       AssertionError: assert 'Max(Integer(...teger=True)))' == 'Max(Integer(...teger=True)))'
> E         
> E         - Max(Integer(0), Add(Mul(Integer(-1), symbol('__out_IDim_range_0', integer=True)), symbol('__out_IDim_range_1', integer=True)))
> E         + Max(Integer(0), Add(Mul(Integer(-1), symbol('__out_IDim_range_0', commutative=True, integer=True)), symbol('__out_IDim_range_1', commutative=True, integer=True)))
> E         ?                                                                   ++++++++++++++++++                                             ++++++++++++++++++
> ```

I dont know how to fix other than explicitly adding the commutativity assumption...

tbennun (approved):

PR #2404, opened 2026-06-14
Fix `lru_cache` type-conflation in `dace.symbolic.simplify`

Spielopoly:
## Summary

`dace.symbolic.simplify` is memoized with `functools.lru_cache`. `lru_cache`
keys its entries by `hash`/`==`, and in Python booleans *are* integers:

```python
>>> hash(True) == hash(1)
True
>>> True == sympy.Integer(1)
True
```

As a result the cache cannot tell `simplify(True)` apart from
`simplify(sympy.Integer(1))` — they collapse onto a single entry. Whichever is
evaluated first wins, and the other caller gets the wrong-typed result:

```python
>>> from dace.symbolic import simplify
>>> simplify(True)                 # caches sympy.true under a key equal to 1
True
>>> simplify(sympy.Integer(1))     # cache hit -> returns the *boolean*
True                               # expected: 1  (a sympy.Integer)
>>> type(simplify(sympy.Integer(1)))
<class 'sympy.logic.boolalg.BooleanTrue'>
```

The same conflation applies to `False` / `Integer(0)`.

## Impact

The corruption is **process-global and order dependent**: any code path that
feeds a plain Python `bool` to `simplify` (for example a comparison like
`simplify(s == 1)`, which evaluates to `True`/`False` before reaching
`simplify`) poisons the cache for every later `simplify(sympy.Integer(1))` /
`simplify(sympy.Integer(0))`.

Downstream, this surfaces as a hard crash in memlet-volume propagation. For a
single-iteration map the propagated volume is `sympy.Integer(1)`, so
`propagate_subset` (`dace/sdfg/propagation.py`) calls `simplify(Integer(1))`,
gets back `BooleanTrue`, and the `Memlet.volume` setter rejects it:

```
TypeError: Property volume must be a literal or symbolic expression,
got: <class 'sympy.logic.boolalg.BooleanTrue'>
```

Because it depends on evaluation order across the whole process, the failure is
intermittent and hard to attribute — an unrelated SDFG breaks long after the
`bool` was simplified.

## Fix

Construct the cache with `typed=True` so arguments of different types are cached
under distinct entries, even when they hash and compare equal:

```diff
-@lru_cache(maxsize=2048)
+@lru_cache(maxsize=2048, typed=True)
 def simplify(expr: SymbolicType) -> SymbolicType:
     return sympy.simplify(expr)
```

With `typed=True`, `simplify(True)` (a `bool`) and `simplify(sympy.Integer(1))`
(a `sympy.Integer`) occupy separate cache slots, so neither can return the
other's result.

## Testing

New regression test: `tests/simplify_cache_typed_test.py`. It covers:

- `simplify(True)` then `simplify(Integer(1))` (and the reverse) stay correctly
  typed — both directions, for `True`/`1` and `False`/`0`.
- Normal symbolic expressions still simplify and are still served from the cache
  (`cache_info().hits`).
- End-to-end: after a `bool` is passed to `simplify`, memlet-volume propagation
  over a single-iteration map still succeeds (this reproduced the original
  `TypeError` verbatim before the fix).


tests/simplify_cache_typed_test.py
```diff
@@ -0,0 +1,133 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```
ThrudPrimrose, line 1: This comment text has too much AI-fluff, can you shorten it?

ThrudPrimrose: `Regressions tests to ensure LRU cache is type-aware` could be enough.

ThrudPrimrose (changes requested):
Other than the overly verbose comments on the tests it looks fine to me.

ThrudPrimrose (approved):

tbennun (approved):

edopao:
cscs-ci run

PR #2410, opened 2026-06-22
Subset 1D Slicing Contiguous 

philip-paul-mueller:
This PR adds a special case for 1D slices that are contiguous in that dimension, but are not packed.


philip-paul-mueller:
While it fixes some issues in ICON4Py it does not fixes all, i.e. ICON4Py still has problems and the errors looks like they are related.

ThrudPrimrose (approved):
LGTM.

PR #2272, opened 2026-01-15
Remove the deprecated `NestedSDFG.schedule` property

tbennun:

PR #1701, opened 2024-10-22
Check host_maps and host_data in the GPU transformations

ThrudPrimrose:

PR #2413, opened 2026-06-23
fix: gpu transform node selection

romanc:
## Description

This part of the gpu transforms decides which nodes should move to the GPU. The original selection was enhanced/changed with the following two PRs

- https://github.com/spcl/dace/pull/1701
- https://github.com/spcl/dace/pull/2272

In combination, I think we ended in an inconsistent state where any `NestedSDFG` node would end up in the list of "gpu nodes". This in turn would force all inputs/outputs of that nested SDFG to GPU-bound memory.

The other issue that I could see is the following: in the current version, if a library node has an input/output that is marked to be forced on the host, the library node will anyway be exectured on the device (and the input/output data will be moved to the device).

### DIFF
--- a/dace/transformation/interstate/gpu_transform_sdfg.py
+++ b/dace/transformation/interstate/gpu_transform_sdfg.py
@@ -361,10 +361,9 @@ class GPUTransformSDFG(transformation.MultiStateTransformation):
             sdict = state.scope_dict()
             for node in state.nodes():
                 if sdict[node] is None:
-                    if isinstance(node, (nodes.LibraryNode, nodes.NestedSDFG)):
-                        if node.guid:
-                            if isinstance(node, nodes.LibraryNode):
-                                node.schedule = dtypes.ScheduleType.GPU_Device
+                    if isinstance(node, (nodes.LibraryNode)):
+                        if not self._output_or_input_is_marked_host(state, node):
+                            node.schedule = dtypes.ScheduleType.GPU_Device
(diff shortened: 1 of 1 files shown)


romanc:
@ThrudPrimrose / @tbennun what do you think? Does this look reasonable?

ThrudPrimrose:
> @ThrudPrimrose / @tbennun what do you think? Does this look reasonable?

The change makes sense, the Host-Data functionality was not handling library nodes correctly. LGTM

ThrudPrimrose (approved):

tbennun:
cscs-ci run

tbennun:
> @ThrudPrimrose / @tbennun what do you think? Does this look reasonable?

The change seems reasonable but the question is what about the contents of the nested SDFG? Are they traversed and mapped to the GPU appropriately?

romanc:
> > @ThrudPrimrose / @tbennun what do you think? Does this look reasonable?
> 
> The change seems reasonable but the question is what about the contents of the nested SDFG? Are they traversed and mapped to the GPU appropriately?

I'm honestly not sure. It might be more complicated than the proposed changes. Feel free to take this PR and develop further. I'm in the middle of a deadline ... I can come back to it first half of July.

tbennun:
Note for later: despite all tests passing, one of the cases we should explicitly test for is a map within which there is a loop (like the old double buffered matrix multiplication).

PR #2418, opened 2026-07-02
Expose additional math functions in `dace/math.h`

romanc:
## Description

`dace/math.h` re-exposes a couple of math functions from `cmath`, e.g. `std::sin(x)`, which has overloads for e.g. `float` and `double`. Some functions, e.g. `asin(x)`, were not exposed in this way. This leads to precision issues when `asin(x)` is called with a `float` because in the generated code, `asin` will be mapped to [the C version](https://en.cppreference.com/c/numeric/math/asin), which is only defined for doubles. There is thus an implicit cast happening of the argument and the computation is done in double precision. Worse, the return type is always a `double`, which will force the rest of the calculation to be up-casted to double precision if `asin()` is used in an expression.

This is a re-hash (including the missing test) of https://github.com/spcl/dace/pull/2364, which grew way too big and is now plucked apart into smaller, more meaningful PRs.

@alexnick83 would you mind giving this another look? I've added a simple test for `asin()` as you asked in https://github.com/spcl/dace/pull/2364#pullrequestreview-4264812947.

### DIFF
--- a/dace/config.py
+++ b/dace/config.py
@@ -402,7 +402,7 @@ class Config(object):
     @staticmethod
     def get_default(*key_hierarchy):
         """ Returns the default value of a given configuration entry.
-            Takes into accound current operating system.
+            Takes into account current operating system.
 
             :param key_hierarchy: A tuple of strings leading to the
                                   configuration entry.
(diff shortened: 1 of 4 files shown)


romanc:

tbennun (approved):

tbennun:
cscs-ci run

romanc:
test failure seems unrelated?

PR #2419, opened 2026-07-02
stree: consider read-after-write when calculating input_memlets of tree nodes

romanc:
## Description

Every tree node has functions to calculate input and output memlets of that node. Tree scopes have a default implementation to gather all inputs/outputs of their children. That default implementation for scopes didn't consider read after write (within the same scope). This caused "too many inputs" to be returned, which - in turn - caused the dependency analysis of the state boundary inserter to generate wrong results, which - in the end - generated wrong SDFGs from a given schedule tree.

This PR changes the default implementation of `_gather_memlets_in_scope()` to account for read-after-write memlets and adds a test.

This PR is part of a series of commits to bring back the fixes from the June push (i.e. https://github.com/spcl/dace/pull/2364) into `main` as clean commits with test cases.

### DIFF
--- a/dace/sdfg/analysis/schedule_tree/treenodes.py
+++ b/dace/sdfg/analysis/schedule_tree/treenodes.py
@@ -108,7 +108,25 @@ class ScheduleTreeScope(ScheduleTreeNode):
 
         # Fast path, no propagation necessary
         if keep_locals:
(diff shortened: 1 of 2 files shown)


romanc:

tbennun (approved):

tbennun:
cscs-ci run

PR #2420, opened 2026-07-02
[stree] stree to SDFG fixes

romanc:
## Description

This PR brings back a couple of stree -> SDFG fixes from the NASA working branch https://github.com/spcl/dace/pull/2364, which accumulated a bunch of changes last month. In particular, this PR contains

1. consider scalar memlets e.g. when collecting memlets of conditions. our use-case is a conditional inside a triple-nested loop where the conditional depends on a scalar condition that is calculated outside the loop (see test case).
2. connect all source / sink nodes inside a map with empty memlets to the corresponding entry / exit node of the containing map (e.g. see image below of what was wrong before)
3. proper support for (self-assigning) copy nodes
4. ensure unique names of added loop regions (this is enforced by validation)

Support for collecting scalar memlets bleeds out into `sdfg.py` and `state.py`. All changes there (outside the schedule tree world) are fully backwards compatible.

---

problematic stree -> sdfg transformation before fix number 2

image

### DIFF
--- a/dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
+++ b/dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
@@ -253,7 +253,7 @@ class _StreeToSDFG(tn.ScheduleNodeVisitor):
         memlets = loop_region.get_meta_read_memlets(self._ctx.root.containers, include_scalars=True)
         self._ensure_data_descriptors(memlets, sdfg)
 
-        cf_region.add_node(loop_region)
+        cf_region.add_node(loop_region, ensure_unique_name=True)
         prefix = self._loop_state_name_prefix(node)
         loop_state = loop_region.add_state(f"{prefix}_loop_state_{id(node)}", is_start_block=True)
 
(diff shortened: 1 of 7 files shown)


romanc:

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -537,6 +578,13 @@ def visit_MapScope(self, node: tn.MapScope, sdfg: SDFG) -> None:
 
         # connect writes to map_exit node
         for name in to_connect:
+            # Special case:
+            # This tasklet is a sink node and just needs an (empty) memlet connection to the MapExit node.
+            if name.startswith(PREFIX_SINK_TASKLET):
```
tbennun: I don't like this name convention approach. Not saying we cannot use it, but what is the use case that necessitates this and can we mark it separately without the special case?

tbennun (changes requested):
I have one comment and I am not sure we are testing for this case.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
romanc: You are right, we don't need the name convention. I can just see if the `Memlet` is empty instead. The newly added test cases

- `test_create_map_scope_read()`: read - no write
- `test_create_map_scope_write_in_two_tasklets()`: write - no read
- `test_create_map_scope_hello_world()`: neither read nor write

are covering all possibilities, no?

tbennun (approved):

tbennun:
cscs-ci run

PR #2421, opened 2026-07-02
Miscellaneous type hint fixes and typos

romanc:
## Description

This is the last commit in cleaning up "the NASA working branch", i.e. https://github.com/spcl/dace/pull/2364. It contains a couple of smaller unrelated (type) fixes and typos. It is assumed that all of these changes are covered (in one way or another) by existing test cases.

### DIFF
--- a/dace/frontend/python/parser.py
+++ b/dace/frontend/python/parser.py
@@ -6,16 +6,18 @@ import copy
 import os
 import sympy
 import sys
-from typing import Any, Callable, Dict, List, Optional, Set, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Sequence, Tuple, Union, TYPE_CHECKING
 from typing import get_origin, get_args
 import warnings
 
 from dace import data, dtypes, hooks, symbolic
-from dace.codegen.compiled_sdfg import CompiledSDFG
 from dace.config import Config
 from dace.frontend.python import (newast, common as pycommon, cached_program, preprocessing)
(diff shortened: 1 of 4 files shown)


romanc:

dace/frontend/python/parser.py
```diff
@@ -11,6 +11,7 @@
 import warnings
 
 from dace import data, dtypes, hooks, symbolic
+from dace.codegen.compiled_sdfg import CompiledSDFG
```
tbennun: potentially dangerous with import loop. Can you add it under a TYPE_CHECKING branch?

tbennun (changes requested):
One tiny change

dace/frontend/python/parser.py
romanc: For function definitions, python requires that type definitions are available at runtime. Changing this to


```suggestion
if TYPE_CHECKING:
    from dace.codegen.compiled_sdfg import CompiledSDFG
```

will result in a `NameError: CompiledSDFG` not defined so we can't do that. If you consider the import an issue, we could

1. not have the type hint all all (i.e. removing the wrong `None` type)
2. make `CompiledSDFG` a string type

tbennun: when adding TYPE_CHECKING, you have to convert your TYPE_CHECKING-guarded type hints to strings. This is how we do it in the rest of the codebase. IDEs and type checkers follow this convention well.

tbennun (approved):

tbennun:
cscs-ci run

PR #2424, opened 2026-07-03
Made it possible to disable GPU detection

philip-paul-mueller:
Currently DaCe will _always_ compile and run a test program to determine the CUDA/HIP architecture.
This is problematic because this will create a CUDA-Context and prevents parallel compilations.

This PR changes this by changing the meaning of `compiler.cuda.{hip,cuda}_archs`.
Before they were the additional architectures to compile for, since they carried a value, in the case for CUDA the value was `60`, we would _always_ compile for two architectures the one that was detected and for `60`.
The new meaning of these keys are "the architectures we want to compile for" and if they are empty, the current default, the architecture is auto detected.

Also the auto detection changed.
Before a test program was compiled to probe for the architecture.
The auto detection for CUDA was switched to [`native`](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html), i.e. the compiler does it itself.
However, this does not work for HIP thus the program is retained.


dace/config_schema.yml
```diff
@@ -345,19 +345,28 @@ required:
...
+                            See also 'no_gpu_arch_detection'.
+                        default: ''
+
+                    no_gpu_arch_detection:
+                        type: bool
+                        title: Disable GPU architecture detection
+                        description: >
+                          If set to 'true' then there will be no GPU detection set before compilation,
+                          which prevents the creation of a CUDA/HIP context. Instead the value
+                          specified in 'hip_arc' and 'cuda_arch' will be used, respectvely.
```
ThrudPrimrose: Typo

dace/config_schema.yml
```diff
@@ -345,19 +345,28 @@ required:
...
                         title: Additional CUDA architectures
                         description: >
-                            Additional CUDA architectures (separated by commas)
-                            to compile GPU code for, excluding the current
-                            architecture on the compiling machine.
-                        default: '60'
+                            Additional CUDA architectures (separated by commas) to compile GPU
+                            code for, excluding the current architecture on the compiling machine.
+                            See also 'no_gpu_arch_detection'.
+                        default: ''
```
ThrudPrimrose, line 395: `CUDA_ARCHITECTURES` already support a list, and Nvidia also supports `arch=native` to auto-detect.
Do we even need `get_cuda_arch.cpp`  anymoer btw?

ThrudPrimrose: If we use the flag native then we should not need to run these scripts, and then we would also not need the `np_gpu_arch_detection` (as a list would be passed for cross-compiling), do I understand correctly?

ThrudPrimrose: Instead of the new flag, imo, the better option is:
- Remove compile+execution based detection
- Use `arch=native` for that by default
- Still allow passing a list or an explicit arch for cross-compiling if GPU is not available. Update CMake accordingly?

ThrudPrimrose: We can bump cmake dependency to be >3.24. 
CMake does not have an equivalent for HIP, so leaving it completely empty or defaulting to native might be a problem. (Even though amdclang has --offload-arch=native)

ThrudPrimrose (changes requested):

dace/config_schema.yml
philip-paul-mueller: Flag was removed.

dace/config_schema.yml
philip-paul-mueller: I removed the auto detection for cuda, instead when nothing is specified it is now the same as passing `native` to it.
The HIP is still on the compiled code so no change, however, it only does it when the architecture is empty.

philip-paul-mueller:
cscs-ci run

ThrudPrimrose (approved):
LGTM with the changes

dace/config_schema.yml
```diff
@@ -343,21 +343,18 @@ required:
...
                         type: str
-                        title: Additional CUDA architectures
+                        title: CUDA architectures to compile for
                         description: >
-                            Additional CUDA architectures (separated by commas)
-                            to compile GPU code for, excluding the current
-                            architecture on the compiling machine.
-                        default: '60'
+                            If empty the cuda archtecture is autodetected by the compiler and has
+                            the same effect as passing 'native' to it.
```
edopao:
```suggestion
                            If empty, the SDFG is compiled for the 'native' CUDA architecture.
```

dace/config_schema.yml
```diff
@@ -343,21 +343,18 @@ required:
...
                     hip_arch:
                         type: str
-                        title: Additional HIP architectures
+                        title: HIP architectures to compile for
                         description: >
-                            Additional HIP architectures (separated by commas)
-                            to compile GPU code for, excluding the current
-                            architecture on the compiling machine.
-                        default: 'gfx906'
+                            If empty then an autodetection step is performed.
```
edopao:
```suggestion
                            If empty, the SDFG build will detect the local HIP architecture.
```

edopao (commented):
Some suggestions

tbennun (changes requested):
Need to also change setup.py: https://github.com/spcl/dace/blob/main/setup.py#L35
for the minimum cmake version required
Otherwise lgtm

tbennun (approved):

ThrudPrimrose:
The frequent GPU crashes might benefit from this: https://github.com/spcl/dace/pull/2471, such that we can read the error


I this feature is creating a non deterministic issue where dace compiles to default arch (could be sm70-60) and then cublas is sm90. The 2471 is already in merge queue.

ThrudPrimrose (approved):

PR #2426, opened 2026-07-03
Relax non-negative-integer Pow to ipow (fix symbolic-exponent codegen)

ThrudPrimrose:
Here I provide a pass relaxes `pow` to `ipow` if the exponent is nonnegative. It analyzes loop ranges / map ranges to check the bound of the expression, as appearing in stockham_fft. This way, we can compile stockham_fft.


dace/frontend/python/parser.py
```diff
@@ -131,7 +131,8 @@ def infer_symbols_from_datadescriptor(sdfg: SDFG,
                     sym_dim = sym_dim.subs(repldict)
 
                 if symbolic.issymbolic(sym_dim - real_dim):
-                    equations.append(sym_dim - real_dim)
+                    # ipow is semantically Pow; restore it so the solver can invert the shape.
+                    equations.append((sym_dim - real_dim).replace(symbolic.ipow, lambda b, e: b**e))
```
tbennun: shouldn't this just be in repldict above?

dace/runtime/include/dace/math.h
```diff
@@ -546,8 +546,21 @@ namespace dace
             return result;
         }
 
-        template<typename T>
-        DACE_HDFI T ipow(const T& a, const unsigned int& b) {
+        // Scalar types seed at ``T(1)`` so ``ipow(a, 0) == 1``.
+        template<typename T,
+                 typename std::enable_if<std::is_constructible<T, int>::value>::type* = nullptr>
```
tbennun, line 562: careful with `int` and negative numbers

dace/runtime/include/dace/math.h
```diff
@@ -546,8 +546,21 @@ namespace dace
...
+            T result = T(1);
+            for (unsigned int i = 0; i < b; ++i)
+                result *= a;
+            return result;
+        }
+
+        // Vector types have no scalar constructor, so seed at ``a``. Only ever reached with a
+        // compile-time exponent >= 1 (the constant-power path emits a literal 1 for exponent 0).
+        template<typename T,
+                 typename std::enable_if<!std::is_constructible<T, int>::value>::type* = nullptr>
```
tbennun, line 573: same

tbennun (changes requested):
minor comments about the pervasiveness of the pass

dace/runtime/include/dace/math.h
ThrudPrimrose: The power operand is an unsigned integer, so a negative value should not be passed. 

If the user explicitly passes a negative number, then there is not much we can do. 

It is also the reason why I run the pass just before codegen, as relaxed `pow` ->`ipow` needs to ensure the loop structure has not changed and the symbol assumptions etc are not changed.

dace/frontend/python/parser.py
ThrudPrimrose: I tried running `replace_dict` but it calls `.subs` and that doesn't replace pow. So I used this instead, an example:
```
import sympy
from dace.symbolic import pystr_to_symbolic, ipow

e = pystr_to_symbolic('pow(x, 2)')  
print('parsed        :', e, '   head:', type(e).__name__)
print('replace->ipow :', e.replace(sympy.Pow, lambda b, n: ipow(b, n))) 
print('subs no-op    :', e.subs({'pow': 'ipow'}))
```

Output:
```
parsed        : x**2    head: Pow
replace->ipow : ipow(x, 2)
subs no-op    : x**2
```

PR #2427, opened 2026-07-07
`nanobind` Based Interface

philip-paul-mueller:
Adds a Python <-> C interface that uses `nanobind` instead of `ctypes`.
The `NanobindCompiledSDFG` wrapper does not replace the original `ctypes` based `CompiledSDFG` class; both coexist.
Its advantage is that calling it is much faster, as the argument processing is moved into C++.

The nanobind interface is deliberately NOT feature complete.
Its scope is: primitive data (scalars of integers, floats, complex, bool, plus string scalars), arrays of primitive data (including `float16`, nullable, GPU and vector arrays), anything nanobind handles natively (in particular `pyobject` scalars, arrays and returns), and array return values.
This covers the vast majority of use cases (and all of GT4Py's) while keeping the interface simple; everything else stays on the `ctypes` interface.
In the absence of hooks, `NanobindCompiledSDFG.__call__()` is essentially `return self._handle(*args, **kwargs)`.

### Interface Selection: `compiler.interface=auto` (new default)
`compiler.interface` now accepts `auto`, `ctypes` and `nanobind`, with `auto` as the default.
Under `auto`, each SDFG is inspected at `compile()` time and the nanobind interface is selected iff the nanobind package is available and the interface supports all of the SDFG's arguments and return values; otherwise the `ctypes` interface is used.
A missing nanobind package therefore degrades `auto` to ctypes-only with a one-time warning (nanobind is a declared dependency, so its absence usually indicates a broken installation), while explicit `nanobind` keeps failing hard in that situation.
The decision lives in one external function, `dace.codegen.compiler.resolve_compiler_interface()`; `SDFG.compile()` calls it once and passes the decided mode down as function arguments (no attribute is attached to the SDFG, and code generation never re-reads the configuration).
The supportability check is the same predicate the bindings generator's refusals use (`nanobind_bindings.argument_unsupported` / `unsupported_reason`), so the detector and the codegen cannot diverge.
Explicit `nanobind` remains a demand: an out-of-scope SDFG is then refused fail-fast at code generation with an error pointing at `ctypes`, never silently redirected.
Loading is unaffected: `load_precompiled_sdfg()` dispatches on the folder's `INTERFACE` marker only, so folders of both interfaces coexist transparently.

Out of scope (routed to `ctypes` under `auto`, refused under explicit `nanobind`): `Structure` arguments, `ContainerArray`s, record-dtype (`dtypes.struct`) arrays, callbacks, the ml_dtypes-backed low-precision types (`bfloat16`, `float8_*`) in every form, `float16` and vector *scalars*, string arrays, non-array returns and returns with a nonzero offset.
If a particular feature is needed later, it can be added then.

### Most Important Changes
Here is a short summary of the most important changes a user should be aware of.
- Before, `CompiledSDFG` made sure that for every instance there was a separate `.so` file and performed a copy if needed.
    This is no longer possible as the name of the file now has a significance for Python.
    Instead now multiple `NanobindCompiledSDFG` instances can use the same `.so` file.
- The above change implies that the `.so` file can not have a state.
    All state information is contained in the handle, which is (still) managed by the compiled SDFG object.
- Before it was possible to unload a `.so` file, which was done when the `CompiledSDFG` was garbage collected or `finalize()` was called.
    This is no longer possible as Python does not allow to unload modules.
    If an SDFG is compiled while its identity is already loaded (see below) it will be silently renamed.
- The generated `.so` files are now Python C extensions that are "normal" modules that can be imported.
    They are imported under the `dace.generated` package, as `dace.generated.<magic>.<name>`, where `<magic>` is a hash of the resolved build folder.
    A module's identity is thus the pair (build folder, SDFG name): SDFGs with the same name that build into different folders coexist in one process.
    Loading the same pair again returns the already imported module; the loader verifies that it refers to the same file and fails loudly otherwise.
    The generated C++ types additionally carry a content hash in their namespace, so two different programs sharing an SDFG name can not be dispatched into each other (nanobind's type registry is process wide and keyed by type name).
- `SDFG.is_loaded()` treats an identity as loaded when it is resident under EITHER interface (the nanobind module registry or the ctypes DLL bookkeeping).
    Both interfaces deliberately share the `lib<name>.so` file name, and `dlopen` short-circuits on an exact pathname match against the link map: rebuilding a path whose previous artifact is still mapped under the other interface would hand the loader the stale mapping.
    Reporting such an identity as loaded drives the rename loop instead; this matters routinely under `auto`, where ctypes-pinned and auto compiles of the same name interleave.
- Recompiling an UNCHANGED SDFG whose identity is already loaded reuses the loaded module: the module bakes the source SDFG's pre-codegen content hash, `compile()` compares against it and, on a match, returns a fresh handle - no rename, no rebuild.
    A changed SDFG still renames and rebuilds (a module cannot be reloaded), and instrumentation changes count as changes (they alter the generated code even though the semantic SDFG hash ignores them).
    Hash instability can only cause a redundant rebuild, never a wrong reuse.
    SDFGs carrying unresolved external nested SDFGs participate as well: the reuse key is computed on a resolved deepcopy (via the new `SDFG.load_external_nsdfgs()` helper), so the caller's object is never mutated by the lazy loading.
    The reuse also re-checks the folder's `INTERFACE` marker: a `ctypes` rebuild of the same (folder, name) identity invalidates the loaded module, and the compile renames instead.
    Gated by `compiler.nanobind_reuse_loaded` (default: enabled).
- `compile()` now returns a handle whose `sdfg` attribute is an isolated deepcopy (pinned to the build folder) on EVERY return path.
    Previously the cached-binary (`use_cache`) path and the no-regenerate path passed the original object, so later mutations of the caller's SDFG leaked into the handle.
    Note that this deliberately also applies to the `ctypes` interface (the paths are shared).
- The renaming on a collision follows two regimes.
    If the build folder is derived from the configuration (`cache` set to `name`, `hash` or `unique`) the renamed program also derives its own build folder from the new name, thus it never overwrites the artifacts of the original folder.
    If the build folder was set explicitly (which essentially behaves as `cache` set to `single`) the folder is fixed and the renamed program builds in place inside it; keeping such a folder consistent is the responsibility of the user.
    Instead of renaming, an error can be requested through `compiler.nanobind_name_collision`.
- Positional call arguments are mapped to names through `sdfg.arg_names` before dispatch, exactly as under `ctypes`.
    Extra positionals beyond `arg_names` are dropped with a warning (`ctypes` truncated them silently; forwarding them raw could silently bind the wrong buffer), and an argument passed both positionally and as a keyword raises a clear `TypeError`.
    Keyword-only calls skip the mapping entirely.
- `initialize()` returns the state handle as a `ctypes` pointer on BOTH interfaces, a value the `get_exported_function` functions accept as their state argument.
    Both wrappers (and the `CompiledSDFGProtocol`) expose a public `is_initialized` property; generic code must use it instead of the ctypes-private `_initialized`.
- `NanobindCompiledSDFG` no longer stores the (processed) arguments of the previous call.
    This means that calls such as `get_workspace_sizes()` and `set_workspace()` now needs the symbols.
    However, they accept the same call arguments as `__call__()` and `initialize()`.
- Compilation is significantly longer and the resulted files are larger.
    It is thus recommended to set `compiler.build_folder_mode` to `production`.
- Scalar `pyobject` arguments are supported: the object passes through as an opaque `PyObject*` and arrives as the very same object.
    The wrapper keeps the object alive for the duration of the call; the program must not retain the pointer beyond it.
    Arrays of `pyobject`s are supported too: a numpy `dtype=object` array binds through `__array_interface__` (DLPack refuses object arrays outright), and the caller's array keeps the contained objects alive for the call - the same lifetime contract as `ctypes`, which also hands out pointers into the caller's buffer.
    `pyobject` return values are supported and decay to the single contained object, matching the `ctypes` convention (`_return_arrays[i].item()`).
- The interface is able to perform symbol inferences in C.
    This means if the symbol argument `N` is also used as shape/strides of an array and is not provided, then it is taken from there.
    If it is present then no consistency check is performed.
    Note that symbols that are listed in `arg_names` can not be inferred but can act as source for symbol inferences.
    As an example, assume you have an array with symbolic shape of `(a + b,)` and you have added `b` to `arg_names`, then `b` can be inferred.
    Since return values are allocated inside the binding after inference ran, inferred symbols may also size return arrays.
- A symbol that appears ONLY in a return value's shape or strides is not part of `arglist()`, but it still binds as a parameter (it only sizes the allocation and is never forwarded to the program call).
    Omitting it raises a clear error naming the symbol, unless it is inferable from an input array.

### Things Still done in Python
There are some things that are not ported to C and are still done in Python.
- Hooks are still managed in Python; a call only pays for them when hooks are registered.
- Array-like WRAPPER objects (NDSL's `Quantity`, GT4Py storages - anything implementing `__array_interface__` / `__cuda_array_interface__`) are accepted through a repair pass:
    when the compiled dispatcher rejects a call, the wrapper substitutes zero-copy views for such arguments and retries once (by-reference semantics preserved).
    Native argument types (numpy, and DLPack speakers such as torch and cupy) never pay for this; if nothing was unwrapped, the original dispatch error is re-raised.
- The positional-to-name mapping above runs in Python, only when positionals are present.

GPU errors are checked in the compiled binding, using the same mechanism as `ctypes`:
after each program call the binding reads (and clears) the error the generated code recorded for this SDFG (`__dace_gpu_last_error`) and raises it.
The process-global CUDA last-error slot is never consulted; it is per-host-thread and shared with every other GPU user in the process, so it can carry third-party state.
The check is only compiled in when a GPU target emitted its init/exit pair.
Translating a nonzero `__dace_exit` code in `finalize()` still goes through the Python-side GPU runtime.

Return values are allocated by the compiled binding itself (through NumPy/CuPy via the Python API, so ownership and dtype semantics are unchanged), AFTER the compiled symbol inference - so inferred symbols may size a return array.
The binding also builds and returns the result (bare array or tuple); a call with return values no longer needs any Python-side processing.
Whether a caller-provided ``__return*`` buffer is accepted is baked in at code generation from ``compiler.nanobind_allow_return_override`` (changing it requires a recompile); an accepted buffer is validated against the symbol-derived return shape.
Return arrays must have offset 0 (violations refuse at code generation), and when ``do_not_execute`` suppresses the program run the call returns ``None``.

### Build Integration
- Generated code compiles with the same flags as under `ctypes`:
    `nanobind_add_module()`'s defaults are disabled (`NOMINSIZE`, `PROTECT_STACK`), since they would apply to the whole module and thus override `compiler.build_type`'s optimization level (`-Os`) and drop the stack protector on the kernels.
- The runtime ships `dace/runtime/include/dace/nanobind.h`, the umbrella of everything the generated bindings TU needs; the TU includes only that.
    When the nanobind package is importable, the header joins the precompiled binary header (`dace_prewarm.h`, `compiler.precompiled_header`), keyed on the nanobind and Python identity so a different environment can never reuse a stale one.
    Without the package everything is byte-identical to before, and a declined binary header always falls back to textual inclusion - the mechanism stays advisory.
- nanobind's helper library archive is cached across build folders (keyed by interpreter and nanobind version), and recorded builds replay without a CMake configure; see the build-cache machinery for details.
- The generated bindings use only public nanobind and CPython API (no `nb::detail` internals), so a nanobind major upgrade cannot silently break generated code again (the 2.x -> 3.x bump did, once, through an internal helper).

### Good to Know
To achieve best performance the following rules should be followed:
- Only pass arrays of primitive types, as numpy arrays or DLPack-compatible objects (array-like wrappers work, but pay the repair pass on every call).
- Prefer keyword arguments; positionals go through the Python-side mapping.
- `NanobindCompiledSDFG.gpu_error_check` (also a constructor argument) can disable the per-call GPU error check, but it is now a plain in-library read of the SDFG's own error record, so disabling it buys next to nothing.
- It is no longer possible to copy and rename the generated file.
- The original `ctypes` based `CompiledSDFG` class was renamed to `CtypesCompiledSDFG` and put into a different file.
    `compiled_sdfg.py` now contains the definition of a protocol but also exports `CompiledSDFG` as a deprecated alias.
- The `fast_call()` interface is no longer supported, a call can only be made using `__call__()`.
- An SDFG with external (workspace) memory refuses to run before `set_workspace()` was called, with a clear error naming the storage type.
    Under `ctypes` the same call silently dereferences a null workspace pointer (usually a segfault).
    The requirement is per state: `finalize()` drops the association, so a re-initialized handle must be given its workspace again.

### Not Supported
Everything listed as out of scope above stays on the `ctypes` interface; `auto` routes such SDFGs there automatically.
In addition:
- Return values are arrays, or a `pyobject` (which decays to the single contained object, as under `ctypes`).
    Python scalar returns are unsupported on BOTH interfaces - `SDFG.validate` rejects them, so this is not a compatibility gap.
- `SDFG.safe_call()` is refused (it hides the compiled object, whose collision rename would make post-call queries on the original SDFG unsound, e.g. `get_latest_report()`).
    Use `compile()` and `CompiledSDFG.safe_call()` instead, and query instrumentation reports via the compiled object's `sdfg`.
- The same rename unsoundness affects data-instrumentation report lookups:
    a same-name recompile (e.g. re-instrumenting a program that is already loaded) renames into its own build folder, so an implicit "latest report" lookup on the new compiled object finds nothing.
    Pass the report explicitly (`dace.instrument_data(..., restore_from=dreport)`); `SDFG.call_with_instrumented_data` always receives it explicitly and works unchanged.
    Programs sharing one name also share report folders under `ctypes` already - give programs distinct names when their reports must stay apart.
- An already-imported module cannot be reloaded after a same-path recompile (CPython cannot unload extension modules; `ctypes`' `ReloadableDLL` covers that workflow).
    Recompilation under a taken (build folder, name) identity renames instead.
- No caching of previous call arguments (`_lastargs`): the workspace getters/setters take the symbol values they need per call.
- Thread-safety model is documented on the class and in the generated code (same initialization race as ctypes; no shared per-call state, unlike `_lastargs`).

### Behavioral Divergences the `auto` Detector Cannot See
The detector inspects argument DESCRIPTORS, not call values, so when `auto` selects nanobind the following call-time behaviors differ from `ctypes`:
- Python lists are rejected for array parameters (`ctypes` ran `np.asarray`, silently copying - and dropping any writes); pass numpy arrays.
- Objects exposing `__array_interface__` / `__cuda_array_interface__` (NDSL's `Quantity`, GT4Py storages) ARE accepted: nanobind's dispatcher rejects them (it speaks DLPack / buffer protocol), and the wrapper then repairs the call once by substituting zero-copy views before retrying - native argument types never pay for this.
- Views pass zero-copy via DLPack; `compiler.allow_view_arguments` is not consulted.
- Lossy scalar casts (e.g. `float` -> `int` parameter) are rejected instead of warn-and-truncate; numpy scalars are accepted unless `compiler.nanobind_strict_scalar_cast` is set.
- Passing `None` for an array requires the descriptor to be marked optional (e.g. an `Optional[T]` hint); `ctypes` silently passes a null pointer for ANY array.
- Extra positional arguments beyond `arg_names` are dropped with a WARNING; `ctypes` dropped them silently.

### Configuration added
The following configurations were added:
- `compiler.interface` (`auto` | `ctypes` | `nanobind`, default `auto`):
    Decides which backend is used; `auto` selects per SDFG (see "Interface Selection").
- `compiler.nanobind_strict_scalar_cast` (default off):
    Decides if (safe) widening casts are allowed (only applies to Python scalars).
- `compiler.nanobind_allow_return_override` (default off):
    Decides if it is allowed to explicitly pass the implicit `__return` arguments to `__call__()`.
    Consulted at code generation only - the decision is baked into the module, changing the option requires a recompile.
    The ML frontend (`ONNXModel`, the torch dispatchers) bakes it in at its own compile sites, since it allocates the model outputs itself.
- `compiler.nanobind_name_collision` (`rename` | `error`, default `rename`):
    Decides what happens when an SDFG is compiled while its (build folder, name) identity is already loaded: silently rename or refuse.
- `compiler.nanobind_reuse_loaded` (default on):
    Reuse an already-loaded, content-identical module on recompile instead of rename-and-recompile (see "Most Important Changes").

### Modified Tests
With `auto` as the default, suites exercising out-of-scope features (callbacks, structures, container arrays, low-precision data) need no markers: their SDFGs route to `ctypes` automatically.
Tests that assert `ctypes`-SPECIFIC behavior are pinned to the `ctypes` interface through per-file fixtures (dropping the `DACE_compiler_interface` env var first, then `set_temporary`), so the behavior runs in every CI leg instead of being skipped:

#### Python `list`s not Coerced
`ctypes`' marshaller runs `np.asarray` on every argument, so a Python list passed for an array parameter is silently converted to an array (and any writes to that array are not propagated).
The nanobind binding is `nb::ndarray<T>.noconvert()`, which accepts only objects exposing the DLPack or buffer protocol; a bare `list` has neither and is rejected at dispatch, and it has no buffer a zero-copy view could wrap.

#### Module Reload After Same-Path Recompile
These assert that recompiling an SDFG in place yields a fresh, reloaded library within the same process.
`ctypes` wraps the shared object in `ReloadableDLL` (dlclose + dlopen).
A `nanobind` module is a CPython **extension module**, which cannot be unloaded or reloaded once imported; a recompile is therefore renamed, and a same-path reimport returns the already-loaded module.
The reload semantics cannot exist under `nanobind`.

#### View Rejection not Applicable
`compiler.allow_view_arguments` is a `ctypes`-marshalling switch that, when off, rejects non-contiguous / view arrays.
`nanobind` passes any DLPack-compatible view zero-copy and never consults that config, so the "views are rejected" assertion does not apply.

#### Return-override is opt-in
`ctypes` lets a caller pass their own buffer for a `__return` value.
`nanobind` forbids this by default; it can be enabled by setting `compiler.nanobind_allow_return_override` (baked in at code generation; changing it requires a recompile).
The default-behaviour assertion is `ctypes`-pinned, the opt-in path has its own `nanobind` test.

#### Lossy Scalar Cast Rejected vs. Warned
`ctypes`' marshaller emits a `Casting` UserWarning and truncates a lossy scalar, e.g. `0.1` -> `int` parameter.
`nanobind` rejects a lossy cast outright (overflow-checked casters), so there is no warning to assert; the test is `ctypes`-pinned.

#### `SDFG.safe_call()` Refused
`SDFG.safe_call()` compiles internally and hides the compiled object; a compiled `nanobind` module cannot be reloaded, so a recompile under an already-loaded identity is silently renamed into its own build folder.
Post-call queries on the original SDFG (e.g. `get_latest_report()`) would then silently look in the wrong folder, hence the method raises on the `nanobind` interface.
The tests exercising it are `ctypes`-pinned; the `*_precompiled` variants cover the supported route (`compile()` + `CompiledSDFG.safe_call()`, reports queried via the compiled object's `sdfg`).

#### Array-Interface Objects Run on Both Interfaces
`tests/array_interface_test.py` is deliberately NOT pinned: under `ctypes` it exercises the marshaller's `__array_interface__` coercion, under nanobind the wrapper's repair pass - both by reference.

#### Weakened
- **Missing-argument error type**: a missing program argument raises `KeyError` on ctypes and `TypeError` from the nanobind dispatcher; the test accepts both.
- **Workspace symbols per call**: the external-memory API takes the symbol values it needs per call under nanobind (there is no `_lastargs` cache), so the test passes symbols to the workspace getters under nanobind and none under ctypes.


philip-paul-mueller:
cscs-ci run

philip-paul-mueller:
cscs-ci run

philip-paul-mueller:
cscs-ci run

tests/safe_call_test.py
```diff
@@ -3,6 +3,12 @@
 import numpy as np
 import pytest
 
+skip_sdfg_safe_call_on_nanobind = pytest.mark.skipif(
```
ThrudPrimrose, line 6: Does this mean with nano-bind performance reporting by collecting performance reports of a compiled SDFG won't work?

ThrudPrimrose: I think we should discuss this shortly on this week's dace meeting.

tests/sdfg/operational_intensity_test.py
```diff
@@ -156,6 +156,11 @@ def reduction_library_node(x: dace.float64[N]):
 }
 
 
+# The tiled_mmm cases take minutes of pure sympy on a slow runner with coverage enabled. CI's
```
ThrudPrimrose, line 159: I think this comment is bloated and doesn't need to be this long.
E.g.
`This test times out without setting the timeout`
Or here, just no comments at all?

tests/transformations/local_storage_test.py
```diff
@@ -244,7 +244,11 @@ def test_uneven(self):
         sdfg = arange.to_sdfg()
         sdfg.apply_transformations([MapTiling, OutLocalStorage], options=[{'tile_sizes': [5]}, {}])
         dace.propagate_memlets_sdfg(sdfg)
-        sdfg(N=16, __return=output)
+        # The caller-provided output buffer is the point of this test (the
```
ThrudPrimrose, line 247: Is there a reason to have this a config?
Why not always opt-in?
(E.g. is there a reason to avoid this such as opting in making things slower?)
If possible I would prefer the always opt-in, because it is better to keep less configuration knobs imho

tests/codegen/position_independent_code_test.py
```diff
@@ -24,8 +25,13 @@ def addone(a: dace.float64[8]):
 
     sdfg = addone.to_sdfg()
     sdfg.build_folder = str(tmp_path / 'build_folder')
-    with set_temporary('compiler', 'cpu', 'args', value=''):
-        csdfg = sdfg.compile()
+    # ``compile_commands.json`` only survives in the development folder mode -- the production mode
```
ThrudPrimrose, line 28: I dont understand this part exactly, but, from what I get, could it be a bug in DaCe folder management?

Plus, we always enforce compiling with `-fPIC` by passing it on CMake via
`set(CMAKE_POSITION_INDEPENDENT_CODE ON)`
both for CUDA and CPU.

.github/workflows/general-ci.yml
```diff
@@ -17,9 +17,16 @@ jobs:
...
+      # the others, their results are needed to tell leg-specific failures
+      # from general ones.
+      fail-fast: false
       matrix:
         python-version: ['3.10','3.14']
-        simplify: [0,1,autoopt]
+        # The nanobind leg pins `compiler.interface`; it runs without
+        # simplification because the interface does not influence how SDFGs are
+        # optimized - simplification coverage is the other legs' job.
+        simplify: [0,1,autoopt,nanobind]
```
ThrudPrimrose, line 29: Do we really need to make the (2x3) matrix into (2x4) ?
I think repeated testing of nanobind increases CI time a lot but adds not much.

I would propose having a separate run that does (3.24 x 1) + nanobind to keep CI time reasonable.

dace/codegen/targets/cuda.py
```diff
@@ -477,6 +491,32 @@ def get_generated_codeobjects(self):
     return __err;
 }}
 
+// Discard whatever the runtime's per-thread error slot already holds, and say what was discarded.
```
ThrudPrimrose, line 494: Comments too bloated low information density.
I think it is obvious what the code is doing and thus this could be shortened a lot.

tests/lib_reuse_test.py
```diff
@@ -8,6 +8,14 @@
 from dace.codegen.compiler import load_precompiled_sdfg
 import numpy as np
 
+# A nanobind extension module cannot be reloaded in-process: a recompile renames
```
ThrudPrimrose, line 11: There is a lot of nanobind-specific decorators for pytest, Maybe we should collect them and move to conftest some other file ?

dace/frontend/ml/onnx/importer.py
```diff
@@ -284,7 +284,7 @@ def __init__(self,
                  simplify: bool = False,
                  onnx_simplify: bool = True,
                  storage: Optional[dtypes.StorageType] = None,
-                 save_transients: Optional[Dict[str, torch.Tensor]] = None,
+                 save_transients: "Optional[Dict[str, torch.Tensor]]" = None,
```
ThrudPrimrose, line 287: Is the string based type declaration change necessary?

ThrudPrimrose: Why is this change needed?

dace/frontend/operations.py
```diff
@@ -84,11 +84,27 @@ def __call__(self, compiled_sdfg: 'CompiledSDFG', args: Tuple[Any, ...]):
         times = np.ndarray(self.repetitions + 1, dtype=np.float64)
         times[0] = timer()
 
-        for i in iterator:
-            # Call function
-            compiled_sdfg._cfunc(compiled_sdfg._libhandle, *args)
+        if hasattr(compiled_sdfg, '_handle'):
```
ThrudPrimrose, line 87: Why don't we have make `_handle` present on both nanobind and ctypes variants ?

tests/codegen/build_cache_test.py
```diff
@@ -151,6 +167,29 @@ def test_caches_disabled_still_builds(tmp_path):
...
+        mp.delenv('DACE_compiler_interface', raising=False)
+        with dace.config.set_temporary('compiler', 'interface', value='nanobind'):
+            with dace.config.set_temporary('compiler', 'precompiled_header', value=False):
+                cache_dir = compiler.nanobind_static_cache_dir()
+                os.makedirs(cache_dir)
+                with open(os.path.join(cache_dir, 'libnanobind-static-stale.a'), 'wb') as fh:
+                    fh.write(b'not an archive')
+                build_and_check(tmp_path, 'staletolerant')
+    published = sorted(os.path.basename(p) for p in glob.glob(os.path.join(cache_dir, 'libnanobind*.a')))
+    assert len(published) == 2 and 'libnanobind-static-stale.a' in published, \
```
ThrudPrimrose, line 189: Are you sure it should be an assert and not runtime error?
I think it should be a runtime error.

Under python -O (together wiith build_folder_mode=production scenario this PR recommends) the assert is stripped, and it can fail, I would propose a type error or something like that.

dace/codegen/targets/framecode.py
```diff
@@ -264,11 +264,29 @@ def generate_footer(self, sdfg: SDFG, global_stream: CodeIOStream, callsite_stre
...
+        # contamination can arrive between any two calls - hence also here, per invocation.
+        # Declared rather than included: this function lives in the generated .cu, exactly like
+        # __dace_init_cuda below.
+        gpu_drain_decl = ''
+        gpu_drain_call = ''
+        # getattr, not attribute access: a user-registered code generator need not define
+        # target_name (the codegen tutorial's MyCustomLoop does not), and this walks every used
+        # target. The loop below only reads target_name inside a has_initializer/has_finalizer
+        # branch, so it never reached those.
+        if any(getattr(target, 'target_name', None) == 'cuda' for target in self._dispatcher.used_targets):
```
ThrudPrimrose, line 280: Is getattr check necessary, all targets have target_name, no?

dace/codegen/nanobind_bindings.py
```diff
@@ -0,0 +1,1389 @@
...
+
+from dace import data as dt, dtypes, symbolic
+from dace.codegen.common import sym2cpp
+from dace.config import Config
+
+# ml_dtypes-backed low-precision types: numpy cannot export their arrays via
+# DLPack or the buffer protocol, so they bypass nb::ndarray entirely (see the
+# __array_interface__ branch in _argument_binding) and are excluded from
+# shape/stride inference sources and from user_args.
+_LOWP_TYPES = (dtypes.bfloat16, dtypes.float8_e4m3fn, dtypes.float8_e5m2)
```
ThrudPrimrose, line 28: Wouldn't it be better to have this in dace dtypes py?

dace/codegen/__init__.py
```diff
@@ -1,3 +1,4 @@
 # Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
 
-from dace.codegen.compiled_sdfg import CompiledSDFG
+from dace.codegen.compiled_sdfg import CompiledSDFG, CompiledSDFGProtocol
+from dace.codegen.ctypes_compiled_sdfg import CtypesCompiledSDFG
```
ThrudPrimrose, line 4: Shouldn't naming structure be 
`CompiledSDFGProtocol` -> (`NanobindCompiledSDFG`, `CtypesCompiledSDFG`)?

dace/generated/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
+"""
+This package is used by the ``nanobind`` interface to load the generated modules, it should be kept empty.
+"""
```
ThrudPrimrose, line 4: Then we need a .gitkeep file here?

dace/codegen/__init__.py
ThrudPrimrose: I see nanobind compiled SDFG is present too, shouldn't the name be exported then?

dace/codegen/compiled_sdfg.py
```diff
@@ -1,814 +1,100 @@
...
+from typing import Any, Callable, Dict, Optional, Protocol, runtime_checkable
+
+from dace import dtypes
+
+# Re-exported for backward compatibility (they used to be defined in this module).
+from dace.codegen.ctypes_compiled_sdfg import CtypesCompiledSDFG, ReloadableDLL  # noqa: F401 [unused-import]  # Compatibility reexport.
+
+
+@runtime_checkable
+class CompiledSDFGProtocol(Protocol):
```
ThrudPrimrose, line 20: Why don't we have CompiledSDFG as the base class that uses one of the 2 implementations? Then Protocol could be combined with the legacy-placeholder class to something useful?

ThrudPrimrose: CompiledSDFG then can still delegate to nanobind or ctypes depending on user setting it or according to the rules (struct -> ctypes) as discussed?

dace/codegen/nanobind_bindings.py
```diff
@@ -0,0 +1,1389 @@
...
+
+# RAII wrapper around a Python buffer view; nested inside the generated handle,
+# emitted only when the SDFG has struct arguments. PyBUF_SIMPLE yields the raw
+# contiguous bytes regardless of the compound dtype that makes nb::ndarray
+# reject record arrays; the resulting address is the struct's/array's data
+# pointer. The guard outlives call()'s nested GIL release, so PyBuffer_Release
+# always runs with the GIL held.
+_PYBUFFER_HELPER = '''    struct _DacePyBuffer {
```
ThrudPrimrose, line 36: These seem to be not f-strings such that they are replaced dynamically.
Do we really need these C++ snippets as inline-python-string? Can't we provide these as headers? Is there an issue preventing this? This stuff seems to be going to be same across multiple SDFGs, no?

dace/codegen/nanobind_bindings.py
```diff
@@ -0,0 +1,1389 @@
...
+                # and the nb::object parameter keeps it alive across the call, so the slots
+                # stay valid while the program dereferences them. This is what the ctypes
+                # marshaller does too - it hands out pointers into the same buffer.
+                setup_stmts.append(
+                    f'nb::object {name}__ai = {name}.attr("__array_interface__");\n'
+                    f'        const std::string {name}__ts = nb::cast<std::string>({name}__ai["typestr"]);\n'
+                    f'        if ({name}__ts.size() < 2 || {name}__ts[1] != \'O\')\n'
+                    f'            throw std::invalid_argument("SDFG argument error: argument \'{name}\': expected '
+                    f'an object array (numpy dtype=object), got typestr \'" + {name}__ts + "\'.");\n'
+                    f'        const std::uintptr_t {name}__ptr = '
```
ThrudPrimrose, line 339: We run generally with GCC and it does transitively include things, from my experience when compiling dace with LLVM, I sometimes have the issue that we were relying.
So if you are generating these headers we need to ensure:
```
#include <cstdint>      // std::uintptr_t
#include <stdexcept>    // std::invalid_argument
#include <string>       // std::string
#include <utility>   
```

I couldn't see it.
But I think having a nanobind header + generating as little as possible (just for array name casts etc.) makes more sense ?

dace/codegen/nanobind_bindings.py
```diff
@@ -0,0 +1,1389 @@
...
+            # Array of structures: the caller passes a numpy array of
+            # per-element pointers (e.g. ctypes.addressof), whose data pointer
+            # is forwarded cast to the element-pointer type - no device
+            # constraint, the pointer table is passed through as-is. Must
+            # precede the Array branch - ContainerArray subclasses Array.
+            params_by_name[name] = f'nb::ndarray<uint64_t> {name}'
+            call_args.append(f'reinterpret_cast<{ctype} *>({name}.data())')
+            nb_args_by_name[name] = f'nb::arg("{name}").noconvert()'
+
+        elif isinstance(desc, dt.Structure):
```
ThrudPrimrose, line 597: I thought we discussed that nanobind would not support structured (unless the named tuple thing), and use ctypes?

Does this mean structures can be supported by nanobind? Does nanobind pass the struct as a raw pointer to it?

tests/codegen/data_instrumentation_test.py
```diff
@@ -394,6 +436,9 @@ def dinstr(A: dace.float64[20]):
     # only be triggered on SDFG states.
     # TODO(later): Make it so symbols can be instrumented on any Control flow block
     sdfg = dinstr.to_sdfg(simplify=False)
+    # Unique name: same-named programs from other tests share report folders, and the
```
ThrudPrimrose, line 439: Having unique names is a good idea, the comments are kinda redundant and can be pruned imho

tests/codegen/gpu_error_drain_test.py
```diff
@@ -0,0 +1,142 @@
...
+def _sources(sdfg: dace.SDFG):
+    """The generated CUDA file and the frame (host) file, as strings."""
+    objs = sdfg.generate_code()
+    cu = next(o.clean_code for o in objs if o.language == 'cu')
+    frame = next(o.clean_code for o in objs if o.language == 'cpp' and o.name == sdfg.name)
+    return cu, frame
+
+
+def test_gpu_drain_emitted_in_init_and_per_call():
+    """The drain is defined in the .cu, called from __dace_init_cuda, and called again at
```
ThrudPrimrose, line 37: I think these tests checking if code is generated are very obvious and not helpful.
Since all GPU code has this, I think these tests can be removed as checking if codegen emitted are trivial tests.

ThrudPrimrose: If anything goes wrong there, the whole CI would turn red, and these tests dont add anything

tests/codegen/position_independent_code_test.py
ThrudPrimrose: (I actually fixed the position independent code option to be directly in cmake and enforced globally instead of a flaky CPU argument passed, so this test is kinda redundant now?)

tests/numpy/retval_test.py
```diff
@@ -1,6 +1,15 @@
 # Copyright 2019-2021 ETH Zurich and the DaCe authors. All rights reserved.
 import dace
 import numpy as np
+import pytest
+
+# The nanobind interface forbids a caller-provided `__return` buffer by default
```
ThrudPrimrose, line 6: I think this skipif definitions is repeated. We should merge them to a single file?

Also again, why not have "nanobind_allow_return_override" always on?

ThrudPrimrose:
General comment, we need `nanobind_helpers.h`, inline code (C++ as python-strings) is very bloated and makes reading things very hard.

Something like this:
```
dace/runtime/include/dace/nanobind_helpers.h
```

```
#pragma once
#include <nanobind/nanobind.h>
#include <Python.h>

namespace nb = nanobind;

namespace dace { namespace nanobind_detail {
    struct DacePyBuffer { /* ... _PYBUFFER_HELPER body ... */ };
}}

namespace nanobind { namespace detail {
    template <> struct dtype_traits<dace::float16> { /* _FLOAT16_TRAITS */ };
    template <> struct type_caster<dace::bool_t>   { /* _DACE_BOOL_CASTER */ };
}}
```

```
_PYBUFFER_HELPER (RAII Py_buffer guard)	nanobind_bindings.py L2605	-> header
_FLOAT16_TRAITS (dtype_traits<half> spec)	-> header
_DACE_BOOL_CASTER (type_caster<dace_bool>)	-> header
#include block + namespace nb = nanobind;
```

ThrudPrimrose:
Also, fix the conflicts please

ThrudPrimrose:
All SDFGs assume input arguments are non-aliasing, when passing views as argument this assumption will be broken.
Do you keep the warnings alive (if numpy-view is passed there is currently a warning describing that a view argument is passed)?

ThrudPrimrose:
About this:

> SDFG.safe_call() Refused (skip_sdfg_safe_call_on_nanobind)
> 
> SDFG.safe_call() compiles internally and hides the compiled object; a compiled nanobind module cannot be reloaded, so a recompile under an already-loaded identity is silently renamed into its own build folder.
> Post-call queries on the original SDFG (e.g. get_latest_report()) would then silently look in the wrong folder, hence the method raises on the nanobind interface.
> The four tests exercising it are skipped; the *_precompiled variants cover the supported route (compile() + CompiledSDFG.safe_call(), reports queried via the compiled object's sdfg).

Would safe call exposing the compiled SDFG object (and not hiding) , be better to avoid the workarounds? @tbennun

tests/safe_call_test.py
ThrudPrimrose: (Unless I understood correctly)

dace/codegen/nanobind_bindings.py
ThrudPrimrose: They are named "ml_dtypes" by the numpy extension, we can follow similar naming.
This would also make things visible when somebody adds a new ml dtype (e.g. mxfp4, fp4, int4)

dace/generated/__init__.py
ThrudPrimrose: First, I don't see an update to gitignore to avoid this folder; second, I think it is not good to generate artifacts in the source folders. (Previous project did this and people repeatedly tried PRs where artifacts were in it, I would really try to do this somewhere in the build cache)

dace/codegen/nanobind_bindings.py
ThrudPrimrose: (Check next comment on nanobind helper header)

tests/transformations/local_storage_test.py
philip-paul-mueller: It is mostly if a user is using passing `__return` directly then the user is doing something terrible wrong.
Either accept that it is allocated by the wrapper or use a normal `out/inout` argument for that.

So the opt in behaviour is my way of saying: Stop doing what you are doing.

dace/codegen/targets/framecode.py
philip-paul-mueller: See line 276 which say: "yes we need it"

dace/codegen/compiled_sdfg.py
philip-paul-mueller: I think this would be a huge mess, since there is so little they actually share.

dace/codegen/nanobind_bindings.py
philip-paul-mueller: They can be at least to some degree.
But I am actually not sure if we should kick them out.
However, then we would be even further away from ctypes compatibility.

dace/frontend/ml/onnx/importer.py
philip-paul-mueller: If torch is not present, then `troch == None`, see beginning of this file, this means that `torch.Tensor` will error.
So it is a fix.

dace/frontend/operations.py
philip-paul-mueller: Because in `nanobind` `_handle` is a super cool class that is wrapped and accessible to Python, while for `ctypes` it is a pointer that is cast to an integer.
Thus they are very different and need different handling.

dace/generated/__init__.py
philip-paul-mueller: First of all `__init__.py` is your `.gitkeep` file.
Second you do not generate them here, you import them, this folder will remain empty, it is only that in the Python interpreter it will _look_ as it is here.
It is important here, i.e. under `dace.generated`, because you need to import it somewhere.
You can not do that under the global namespace because you can not guarantee that some will create a Python package with the name `dace_extremly_mangled_name`.
So you must import it somewhere under `dace.` and import it as `dace.nanobind_mangled_name` is also strange, so the only option left was creating a dedicated namespace and announce it to the world, that the name is taken, which this folder does.

dace/codegen/__init__.py
philip-paul-mueller: I agree the nanobiind one should be imported to.

dace/codegen/nanobind_bindings.py
```diff
@@ -0,0 +1,1389 @@
...
+        _DacePyBuffer(const _DacePyBuffer &) = delete;
+        _DacePyBuffer &operator=(const _DacePyBuffer &) = delete;
+        ~_DacePyBuffer() { if (ok) PyBuffer_Release(&view); }
+        void *data() const { return view.buf; }
+    };
+'''
+
+# dtype_traits specialization advertising dace::float16 (= dace::half) as a
+# 16-bit DLPack float, so nb::ndarray<dace::float16, ...> accepts a numpy/cupy
+# float16 array. nanobind's own detection uses std::is_floating_point, false for
```
philip-paul-mueller, line 53: Add explanation why not in external header:
- Precompiled header, would not work with precompiled header.

PR #2430, opened 2026-07-09
Fixed Multiple Return Values and Refined Validations

philip-paul-mueller:
Before the return values were processed by using `sorted()`, which sorted the tuple return values, i.e. `__return_{id}`, alphabetically. This has the effect that `__return_11` is ordered before `__return_2`. This PR changes this and ensures that they are in the correct order. Furthermore it speeds up `CompiledSDFG._initialize_return_values()` by not iterating over all arrays in teh SDFG but only through the actual return values.

Furthermore, this PR refines the type rules in `sdfg.validation`. Essentially it restricts return values to `Arrray`s. This is done because their allocations is done NumPy it is limited by what NumPy can allocate.


dace/codegen/compiled_sdfg.py
```diff
@@ -253,9 +253,17 @@ def __init__(self, sdfg, lib: ReloadableDLL, argnames: Optional[List[str]] = Non
         #  ``False`` either means that a tuple is returned or there are no return values.
         # NOTE: Necessary to handle the case of a tuple with one element.
         self._is_single_value_ret: bool = False
+        self._return_args: Tuple[str] = ()
         if '__return' in self._sdfg.arrays:
```
ThrudPrimrose, line 257: What happens if we have `__return` and `__return_0` at the same time? Then we will return `__return` only?

ThrudPrimrose: Ah you added the validation check, great.

dace/sdfg/validation.py
```diff
@@ -275,6 +275,33 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG', references: Set[int] = None, **context
...
                                   f'expected to find "{const_type}" but found "{sdfg.symbols[const_name]}".')
 
+        # Test the return value.
+        tuple_return_args = {n for n in sdfg._arrays if n.startswith('__return_')}
+        if '__return' in sdfg._arrays and tuple_return_args:
+            raise InvalidSDFGError(
+                'Ambiguous return values: an SDFG cannot have both a `__return` (single value) '
+                'and `__return_<i>` (tuple) data descriptor.', sdfg, None)
+        elif '__return' in sdfg._arrays:
+            tuple_return_args = {'__return'}  # This is abuse
```
ThrudPrimrose: Unnecesary comment.

ThrudPrimrose (changes requested):
Other than the unnecessary comment I mentioned, looks good.

ThrudPrimrose (approved):

PR #2438, opened 2026-07-15
Aligned Allocations

philip-paul-mueller:
For allocations `codegen` (`dace/codegen/targets/cpu.py:492` and `:541`) generated the following pattern:
```c++
    mid_deg = new double DACE_ALIGN(64)[2];
```
where `DACE_ALIGN(64)` would expand to `__attribute__((aligned(N)))`, with the intent of allocating it as 64 bytes.
However, this had no effect as it was not applied to the allocation itself and was essentially ignored.
Starting with GCC16 this was turned into an error (in case of known sizes with small sizes).

This PR fixes two things:
- The bug mentioned above.
- Make the code generator respect the `Array.alignment` property (which was ignored before); at least on CPU.

On CPU aligned allocation is performed using the aligned allocation function (`new (std::align_val_t)`) if supported, i.e. we are on C++17 or newer. Otherwise normal allocation, i.e. not aligned allocation (which is the current behaviour), is performed.

The value for alignment is taken from the `Array.alignment` property. If that value is 0, then the default, 64 byte, is used. If it is negative then un-alignment allocation is performed.

Note, alignment is only respected on the CPU backend. On GPU it has no effect, the main reason for it is that the CUDA API does not expose a suitable function for that.
Furthermore, the `alignment` property remains on the `Array` descriptor as moving it to `Data` does not make sense since for most data descriptor types it would be ignored anyway.


acalotoiu (approved):
LGTM.

tbennun (changes requested):
While we’re here: the alignment is always frozen at 64. However, every data descriptor has an alignment field. If 1, allocation should be unaligned. 0 is the default and will result in this 64. Other numbers should change the alignment appropriately? Can we check if alignment is used?

ThrudPrimrose:
> While we’re here: the alignment is always frozen at 64. However, every data descriptor has an alignment field. If 1, allocation should be unaligned. 0 is the default and will result in this 64. Other numbers should change the alignment appropriately? Can we check if alignment is used?

To my knowledge, the alignment is hardcoded to 64, so the alignment property is possible ignored. I think it is a good chance to fix that behavior too.

philip-paul-mueller:
Yes 64 is/was hardcoded.

However it is a bit different, `Data` does not have an `alignment` filed only `Array`s have one.
I think what you meant @tbennun (`1 => unaligned` and `0 => aligned`) was probably the `alignment` argument of [`Data.strides_from_layout()`](https://github.com/spcl/dace/blob/bcb0257f89b0c110330da43eee9e3ed77163b92a/dace/data/core.py#L164).

According to the (property) description `0` means "compiler default", it is unclear if this means "how `malloc()` feels" or `64`.
Furthermore, according to the class' docstring, `alignment` is the alignment of the first element, but it is unclear what "first element" is if `offset != 0`.

However, a quick `grep` shows that it is not used (no idea what `alignment_offset` is):
```
> pygrep -P '\.alignment' -r  
dace/frontend/python/replacements/array_manipulation.py:        narr.alignment_offset = 0
dace/frontend/python/replacements/array_manipulation.py:                                    alignment=desc.alignment,
dace/frontend/python/replacements/array_manipulation.py:                              alignment=desc.alignment,
dace/frontend/python/replacements/array_manipulation.py:                                   alignment=desc.alignment,
dace/frontend/python/replacements/array_manipulation.py:                                        alignment=desc.alignment,
dace/data/core.py:        self.alignment = alignment
dace/data/core.py:                          self.strides, self.offset, self.may_alias, self.lifetime, self.alignment, self.debuginfo,
dace/data/core.py:                                   alignment=viewed_container.alignment,
dace/data/core.py:                               alignment=viewed_container.alignment,
dace/data/core.py:                                    alignment=viewed_container.alignment,
dace/transformation/interstate/sdfg_nesting.py:                                       alignment=desc.alignment,
dace/transformation/dataflow/redundant_array.py:                                                    in_desc.alignment, in_desc.debuginfo, in_desc.total_size)
dace/transformation/dataflow/redundant_array.py:                                                         dtypes.AllocationLifetime.Scope, out_desc.alignment,
```

Given the fact that it is effectively not used we can make bigger changes.
I thus propose the following:
- `alignment` is for the case of `offset == 0`.
- A value of `0` means an alignment of `64` on CPU and whatever the `cudaMalloc` thinks on GPU.
- Otherwise we pass the value as alignment to the allocator.

What are you thought @ThrudPrimrose @tbennun ?

ThrudPrimrose:
> Yes 64 is/was hardcoded.
> 
> However it is a bit different, `Data` does not have an `alignment` filed only `Array`s have one. I think what you meant @tbennun (`1 => unaligned` and `0 => aligned`) was probably the `alignment` argument of [`Data.strides_from_layout()`](https://github.com/spcl/dace/blob/bcb0257f89b0c110330da43eee9e3ed77163b92a/dace/data/core.py#L164).
> 
> According to the (property) description `0` means "compiler default", it is unclear if this means "how `malloc()` feels" or `64`. Furthermore, according to the class' docstring, `alignment` is the alignment of the first element, but it is unclear what "first element" is if `offset != 0`.
> 
> However, a quick `grep` shows that it is not used (no idea what `alignment_offset` is):
> 
> ```
> > pygrep -P '\.alignment' -r  
> dace/frontend/python/replacements/array_manipulation.py:        narr.alignment_offset = 0
> dace/frontend/python/replacements/array_manipulation.py:                                    alignment=desc.alignment,
> dace/frontend/python/replacements/array_manipulation.py:                              alignment=desc.alignment,
> dace/frontend/python/replacements/array_manipulation.py:                                   alignment=desc.alignment,
> dace/frontend/python/replacements/array_manipulation.py:                                        alignment=desc.alignment,
> dace/data/core.py:        self.alignment = alignment
> dace/data/core.py:                          self.strides, self.offset, self.may_alias, self.lifetime, self.alignment, self.debuginfo,
> dace/data/core.py:                                   alignment=viewed_container.alignment,
> dace/data/core.py:                               alignment=viewed_container.alignment,
> dace/data/core.py:                                    alignment=viewed_container.alignment,
> dace/transformation/interstate/sdfg_nesting.py:                                       alignment=desc.alignment,
> dace/transformation/dataflow/redundant_array.py:                                                    in_desc.alignment, in_desc.debuginfo, in_desc.total_size)
> dace/transformation/dataflow/redundant_array.py:                                                         dtypes.AllocationLifetime.Scope, out_desc.alignment,
> ```
> 
> Given the fact that it is effectively not used we can make bigger changes. I thus propose the following:
> 
>     * `alignment` is for the case of `offset == 0`.
> 
>     * A value of `0` means an alignment of `64` on CPU and whatever the `cudaMalloc` thinks on GPU.
> 
>     * Otherwise we pass the value as alignment to the allocator.
> 
> 
> What are you thought @ThrudPrimrose @tbennun ?

I think best way to handle this imho is to have alignment as an integer property.
About, the offset, I am not sure and I don't exactly know how it is implement. From my understanding offset should impact only the logical indices. E.g. If array is of size `(N, N)` we have offset `(1, 1)` then depending on the implementation we might have (i) the size `(N, N)` in allocation but the coordinates are not from `(0, 0)`... to `(N-1, N-1)` but `(1, 1)`... to `(N, N)`, or (ii) allocation size of `(N+1, N+1)`.
Imho option (i) feels correct, in that case offset should not matter for alignment, and since alignment is a porperty of the allocation, alignment would be the same but  offset would be just a mapping between memory and logical coordinate space? I think Tal needs to clarify this more

-1 -> default (no alignment hint)
0 -> no alignment (compiler default) == what malloc and cudamemcpy feels like
\>0 -> alignment is overriden

tbennun:
@ThrudPrimrose Agreed. If necessary, make alignment a Data property (but scalar or stream alignment doesn’t make much sense to me)

ThrudPrimrose:
> @ThrudPrimrose Agreed. If necessary, make alignment a Data property (but scalar or stream alignment doesn’t make much sense to me)

I think the codegen can just ignore the field's value on streams. (I think this is nicer than making a special attribute only for array, or should we just make an attribute available only for arrays?)

philip-paul-mueller:
Sorry I meant `start_offset` which is manual alignment.

I think the table by @ThrudPrimrose make sense.
The last question is where we add the alignment property.
I also agree that it does not make sense to add them to scalar or stream, structs might make sense?
However, I would say that the correct place is add it to `Array` because it is the most logical place to put it (the downside is that also `ArrayView` will have it which does not make sense).
Otherwise we can add it to `Data` but then someone could write something like `scalar_that_i_want_to_align.alignment = 128` and then in a Tasklet `assert(&scalar_that_i_want_to_align % 128 == 0)` which might fail.

I think that all options are wrong, but adding it to `Array` is in my opinion better, but if you (@ThrudPrimrose @tbennun) think that adding it to `Data` is the right call I will go with that.

ThrudPrimrose:
> Sorry I meant `start_offset` which is manual alignment.
> 
> I think the table by @ThrudPrimrose make sense. The last question is where we add the alignment property. I also agree that it does not make sense to add them to scalar or stream, structs might make sense? However, I would say that the correct place is add it to `Array` because it is the most logical place to put it (the downside is that also `ArrayView` will have it which does not make sense). Otherwise we can add it to `Data` but then someone could write something like `scalar_that_i_want_to_align.alignment = 128` and then in a Tasklet `assert(&scalar_that_i_want_to_align % 128 == 0)` which might fail.
> 
> I think that all options are wrong, but adding it to `Array` is in my opinion better, but if you (@ThrudPrimrose @tbennun) think that adding it to `Data` is the right call I will go with that.

Regardless of the options we have, the codegen will either:
- Has to ignore the field on array views, scalars and streams
- Or crashes if it tries to access alignment on anything other than arrays or array views, and ignore when it is an array view

A hard crash will be easier to detect and fix, so technically, putting it a special property of an array might make more sense.

I am fine with both options.

dace/OUT
ThrudPrimrose, line 1: Accidentally added?

philip-paul-mueller: Thanks no idea why It was here.

tbennun (approved):

PR #2439, opened 2026-07-16
Fixed `CompiledSDFGProfiler`.

philip-paul-mueller:
`CompiledSDFGProfiler` sets `CompiledSDFG.do_not_execute` to `False` and then `yield`, in case of throwing it is never restored.
This PR fixes this.


ThrudPrimrose (approved):

acalotoiu (approved):
LGTM

PR #2444, opened 2026-07-19
Deduplicate is_access_contiguous onto Subset.is_contiguous_subset

ThrudPrimrose:
`dace.libraries.mpi.utils.is_access_contiguous` does the same thing as `is_contiguous_subset` I have implemented. Furthermore this function assumed C-packing, abd gave wrong answers for Fortran-packed / permuted layouts, and ignored non-unit steps etc..

I propose we use that function instead.


acalotoiu (approved):
Good simplification, great that you added tests too!

PR #2446, opened 2026-07-20
Fixed `half` CPU Mode

philip-paul-mueller:
The DaCe runtime header provides a CPU implementation for half precision, `half`, which is only active when such code is generated for CPU. This PR fixes two errors related to that class:
- The class could not be default initialized, which lead to errors in some cases.
- The conversion to `float` was not correct in cases if the value has left the normal range, which includes `0.0`.


ThrudPrimrose:
Seems implementing the same thing is: https://github.com/spcl/dace/pull/2395

philip-paul-mueller:
Should be closed in favour of https://github.com/spcl/dace/pull/2395

PR #2449, opened 2026-07-21
Replace getattr/hasattr with static type checks

ThrudPrimrose:
Removes 75 of 393 `getattr`/`hasattr` call sites in `dace/`, replacing them with `isinstance` checks, direct attribute access, or class-level defaults. Reflection with no static equivalent is kept (serialization, `visit_<TypeName>` dispatch, runtime-string lookup, sympy/numpy/ast duck-typing, ONNX schema attributes, runtime-attached markers).

Three defects found while auditing:

- `condition_fusion` assigned the containing SDFG into `NestedSDFG.sdfg`, overwriting the nested graph.
- `validate_state` tested a `**kwargs` dict with `hasattr`, always False, so the `in_gpu` value threaded down from `validate_sdfg` was discarded and recomputed per state.
- `_layernorm_axis` guarded `node.axis` with `hasattr` and dereferenced it in the else branch anyway.

Fixing the latent bug in state fusion and state fusion extended, another bug was exposed. GPU instrumentation's data dump does not insert a synchronization correctly and can thus dump stale data, where GPU tests randomly fail. I have also fixed that.

### DIFF
--- a/dace/frontend/python/parser.py
+++ b/dace/frontend/python/parser.py
@@ -105,9 +105,11 @@ def infer_symbols_from_datadescriptor(sdfg: SDFG,
     for arg_name, arg_val in args.items():
         if arg_name in sdfg.arrays:
             desc = sdfg.arrays[arg_name]
-            if not hasattr(desc, 'shape') or not hasattr(arg_val, 'shape'):
+            if not hasattr(arg_val, 'shape'):
                 continue
-            symbolic_values = list(desc.shape) + list(getattr(desc, 'strides', [])) + list(getattr(desc, 'offset', []))
+            # Distributed descriptors (process grids, subarrays) have no strides
+            desc_strides = [] if isinstance(desc, data.DistributedDescriptor) else desc.strides
+            symbolic_values = list(desc.shape) + list(desc_strides) + list(desc.offset)
             given_values = list(arg_val.shape)
             given_strides = []
             if hasattr(arg_val, 'strides'):
--- a/dace/frontend/python/parser.py
+++ b/dace/frontend/python/parser.py
@@ -107,9 +107,7 @@ def infer_symbols_from_datadescriptor(sdfg: SDFG,
             desc = sdfg.arrays[arg_name]
             if not hasattr(arg_val, 'shape'):
                 continue
-            # Distributed descriptors (process grids, subarrays) have no strides
-            desc_strides = [] if isinstance(desc, data.DistributedDescriptor) else desc.strides
-            symbolic_values = list(desc.shape) + list(desc_strides) + list(desc.offset)
+            symbolic_values = list(desc.shape) + list(desc.strides) + list(desc.offset)
             given_values = list(arg_val.shape)
             given_strides = []
             if hasattr(arg_val, 'strides'):
--- a/dace/frontend/python/preprocessing.py
+++ b/dace/frontend/python/preprocessing.py
@@ -1324,7 +1324,7 @@ class CallTreeResolver(ast.NodeVisitor):
         if not isinstance(function, DaceProgram):
             # Make set of parameters from __sdfg_signature__
             parameters = [(name, inspect.Parameter.POSITIONAL_OR_KEYWORD) for name in function.__sdfg_signature__()[0]]
-            objname = None
+            objname = False
         else:
             parameters = [(aname, arg.kind) for aname, arg in function.signature.parameters.items()]
             objname = function.objname
--- a/dace/frontend/python/preprocessing.py
+++ b/dace/frontend/python/preprocessing.py
@@ -1320,14 +1320,14 @@ class CallTreeResolver(ast.NodeVisitor):
         kwargs = [kwarg.arg for kwarg in node.keywords]
         result = set()
 
+        # "objname" holds the name of the "self" argument, if any
         if not isinstance(function, DaceProgram):
             # Make set of parameters from __sdfg_signature__
             parameters = [(name, inspect.Parameter.POSITIONAL_OR_KEYWORD) for name in function.__sdfg_signature__()[0]]
+            objname = None
         else:
             parameters = [(aname, arg.kind) for aname, arg in function.signature.parameters.items()]
-
-        # Handle "self" argument
-        objname = getattr(function, 'objname', False)
+            objname = function.objname
 
         nargs = len(posargs)
         arg_ind = 0
--- a/dace/libraries/sparse/nodes/csrmm.py
+++ b/dace/libraries/sparse/nodes/csrmm.py
@@ -12,7 +12,7 @@ import numpy as np
 
 
 def _is_complex(dtype):
-    if hasattr(dtype, "is_complex") and callable(dtype.is_complex):
+    if isinstance(dtype, dtypes.typeclass):
         return dtype.is_complex()
     else:
         return dtype in [np.complex64, np.complex128]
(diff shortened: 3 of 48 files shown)


tbennun:
cscs-ci run

dace/frontend/python/parser.py
```diff
@@ -105,9 +105,11 @@ def infer_symbols_from_datadescriptor(sdfg: SDFG,
     for arg_name, arg_val in args.items():
         if arg_name in sdfg.arrays:
             desc = sdfg.arrays[arg_name]
-            if not hasattr(desc, 'shape') or not hasattr(arg_val, 'shape'):
+            if not hasattr(arg_val, 'shape'):
                 continue
-            symbolic_values = list(desc.shape) + list(getattr(desc, 'strides', [])) + list(getattr(desc, 'offset', []))
+            # Distributed descriptors (process grids, subarrays) have no strides
+            desc_strides = [] if isinstance(desc, data.DistributedDescriptor) else desc.strides
```
tbennun: I would just add an empty strides `@property` to DistributedDescriptors instead.

dace/frontend/python/preprocessing.py
```diff
@@ -1320,14 +1320,14 @@ def _get_given_args(self, node: ast.Call, function: 'DaceProgram') -> Set[str]:
         kwargs = [kwarg.arg for kwarg in node.keywords]
         result = set()
 
+        # "objname" holds the name of the "self" argument, if any
         if not isinstance(function, DaceProgram):
             # Make set of parameters from __sdfg_signature__
             parameters = [(name, inspect.Parameter.POSITIONAL_OR_KEYWORD) for name in function.__sdfg_signature__()[0]]
+            objname = None
```
tbennun: before this PR, objname was set to False rather than None. Not sure what the significance of that is.

dace/libraries/sparse/nodes/csrmm.py
```diff
@@ -12,7 +12,7 @@
 
 
 def _is_complex(dtype):
-    if hasattr(dtype, "is_complex") and callable(dtype.is_complex):
+    if isinstance(dtype, dtypes.typeclass):
```
tbennun, line 15: do all typeclasses have is_complex?

dace/sdfg/replace.py
```diff
@@ -199,7 +199,7 @@ def replace_properties_dict(node: Any,
             # Don't replace variables that appear as an input or an output
             # connector, as this should shadow the outer declaration.
             reduced_repl = set(repl.keys())
-            if hasattr(node, 'in_connectors'):
+            if isinstance(node, nodes.Node):
```
tbennun, line 202: wouldn't that always be true?

dace/sdfg/validation.py
```diff
@@ -300,19 +300,21 @@ def validate_sdfg(sdfg: 'dace.sdfg.SDFG', references: Set[int] = None, **context
                             f'Transient data container "{name}" contains undefined symbol in dimension {i}, '
                             f'which is required for memory allocation', sdfg, None)
 
-                # Check strides if array
-                if hasattr(desc, 'strides'):
+                # Only descriptors that are really allocated have strides/total_size to check. A
+                #  DistributedDescriptor (ProcessGrid/SubArray/RedistrArray) describes an MPI
+                #  communicator and has neither, so test for the allocated kinds positively -- an
+                #  unknown Data subclass is then skipped rather than crashing here.
+                if isinstance(desc, (dt.Array, dt.Scalar, dt.Stream, dt.Structure)):
```
tbennun: again, adding empty properties would work well

dace/transformation/interstate/state_fusion.py
```diff
@@ -309,13 +318,16 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
-            if len(resulting_ccs) > 1:
-                # Side-effect tasklets cannot be fused if could lead to data races
-                for node in first_state.nodes():
-                    if isinstance(node, nodes.CodeNode) and getattr(node, 'side_effects', False):
+            # A side-effect node (fusion barrier, I/O, side-effecting library node) must never be fused:
+            # fusion preserves only dataflow order. Exception: __pystate-ordered callbacks keep their order
+            # via that data dependence and are gated by frontend.dont_fuse_callbacks above.
+            for state in (first_state, second_state):
+                for node in state.nodes():
+                    if isinstance(node, nodes.Tasklet) and node.has_side_effects(sdfg):
```
tbennun: CodeNode became Tasklet here! should modify the check and add an overridden method to CodeNode

dace/sdfg/replace.py
tbennun: unless dealing with a State, so keep this check please

tbennun (changes requested):
Overall looks good but can be made even faster :)

dace/frontend/python/parser.py
ThrudPrimrose: I think it is a good change. I will add `strides` and `total_size`

dace/frontend/python/preprocessing.py
ThrudPrimrose: I am not sure, but to be safe I will set to default-False.

dace/libraries/sparse/nodes/csrmm.py
ThrudPrimrose: Yes, it is a function of the typelcass class

dace/sdfg/replace.py
ThrudPrimrose: The replace calls non-node instances, so yes, node.Node should stay here

dace/transformation/interstate/state_fusion.py
ThrudPrimrose: So, `side_effects` property is Tasklet only. LibraryNode also has side effects -> but it is a property, and Tasklet one is a function. I think we should unify the design of has_side_effects to be a property and call for both.

dace/transformation/interstate/state_fusion.py
```diff
@@ -240,23 +231,12 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
-                pass
-            try:
-                next(node for node in second_state.nodes()
-                     if (isinstance(node, nodes.LibraryNode) and type(node).__name__ == 'Waitall')
-                     or node.label == '_Waitall_')
-                return False
-            except StopIteration:
-                pass
+            # Library nodes and nested SDFGs carry dependencies fusion cannot see.
+            # Tasklet callbacks are governed by dont_fuse_callbacks above.
```
tbennun, line 235: I don't mind having that other check here and just asking isinstance(node, nodes.CodeNode)

dace/transformation/interstate/state_fusion.py
```diff
@@ -318,17 +298,12 @@ def can_be_applied(self, graph, expr_index, sdfg, permissive=False):
...
-                        if StateFusion.is_pystate_ordered(state, node):
-                            continue
-                        return False
-                    if isinstance(node, nodes.LibraryNode) and node.has_side_effects:
-                        return False
+            if len(resulting_ccs) > 1:
+                # Declared side effects would race across parallel components.
+                for state in (first_state, second_state):
+                    for node in state.nodes():
+                        if isinstance(node, nodes.Tasklet) and node.side_effects:
```
tbennun, line 305: tasklet or codenode?

tbennun (approved):

tbennun (approved):

PR #2450, opened 2026-07-21
Use int_floor, not `//`, for the symbolic divisions in `view`

ThrudPrimrose:
`expr // d` on a symbolic stride builds sympy `floor(expr / d)`. sympy distributes the division over the sum inside the floor, and `sym2cpp` prints the argument **without** the floor, so each term is left to truncate on its own.

`view` divided `desc.strides`, `newshape[contigdim]` and `desc.total_size` with `//`. For a contiguous array the divisibility check on the last axis makes the division fold away, so this was latent there. A non-contiguous symbolic stride keeps it: strides `[2 * N + 1, 1]` viewed `int16 -> int32` left `floor(N*(2*N + 1)/2)` as the view's `total_size`.

Three call sites, plus one test that pins the view descriptors to `int_floor`. The test fails on the previous code with that bare `floor(...)` in the expression list.

Note: `tests/numpy/reshape_test.py` has pre-existing compilation failures in my environment on an unmodified `main` checkout (`test_reinterpret_smaller` and others), so I verified the new test in isolation and confirmed the baseline failures are not caused by this change.


ThrudPrimrose:
Folded into #2451, which covers the `view` sites plus the rest of the symbolic `//` -> `int_floor`/`int_ceil` conversions in one change.

PR #2452, opened 2026-07-21
Fix LoopUnroll on descending loops

ThrudPrimrose:
`get_loop_end` reports the **inclusive** last value of the iterate, so `range()`'s exclusive bound has to sit one unit past it *in the direction of travel*: above it while ascending, below it while descending. The bound was hardcoded one unit above, which for a negative stride lands on the wrong side and silently drops the last `ceil(2 / |stride|)` iterations.

```python
# i = 9; i > -1; i -= 1   ->   loop_diff = 0 - 9 + 1 = -8   ->   range(0, -8, -1)
# 8 iterations emitted, 10 expected; indices 0 and 1 never written
```

The result is a valid SDFG that quietly computes the wrong thing — no validation error, no warning.

Two further defects sat behind that one, both reachable only once descending loops run all their iterations:

- The empty predecessor a start-block loop needs was guarded on `loop_diff > 0`, false for every descending loop regardless of how many iterations it runs. With the predecessor always present, the start-block re-designation that followed removing the loop is unreachable and goes away.
- Iteration regions are named after the iterate value, and a negative int renders a bare `-`, which is not a legal identifier — a loop crossing zero produced names like `l_i-1_body` and failed validation outright. The fallback is the enumeration position, always non-negative. `LoopPeeling` passes its own loop counter, already non-negative, for the same purpose.

Also: reject a zero stride up front (it never advances the iterate, so there is no finite unrolling), and substitute the iterate through `symstr` rather than `str`, which misrenders an operator-function as its sympy class name.

### Test change

`test_unroll_loop_with_negative_iterate_values` asserted 6 nodes — a number the first two bugs happened to produce together: 3 truncated iterations x 2 states each, and no predecessor. The same loop now yields its 5 iterations plus the predecessor, so the expectation becomes 11, with the iteration count asserted directly rather than inferred from a node total.

`test_unroll_covers_every_iteration` is new and parametrized over ascending/descending, strides 1 and 3, and `<`/`<=`/`>`/`>=` conditions, checking which array elements actually get written.

### Verification

128 passed across `tests/transformations/interstate/` and `tests/transformations/loop_manipulation_test.py`.

### DIFF
--- a/dace/transformation/interstate/loop_peeling.py
+++ b/dace/transformation/interstate/loop_peeling.py
@@ -69,17 +69,14 @@ class LoopPeeling(LoopUnroll):
         start = loop_analysis.get_init_assignment(self.loop)
         end = loop_analysis.get_loop_end(self.loop)
         stride = loop_analysis.get_loop_stride(self.loop)
-        is_symbolic = any([symbolic.issymbolic(r) for r in (start, end)])
 
         if self.begin:
             # Create states for loop subgraph
             peeled_iterations: List[ControlFlowRegion] = []
             for i in range(self.count):
-                # Instantiate loop contents as a new control flow region with iterate value.
+                # `i` is non-negative here, so it doubles as the label-safety fallback index.
                 current_index = start + (i * stride)
-                is_symbolic |= symbolic.issymbolic(current_index)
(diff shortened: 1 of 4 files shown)


acalotoiu (approved):
Good catch! Thanks!

PR #2453, opened 2026-07-21
Cache CMake configure results and precompile the DaCe runtime header

ThrudPrimrose:
Precompiling the DaCe runtime header and caching CMake configure + build results lets repeated builds skip work that can't change between two SDFGs compiled the same way. **10s → 0.5s** per repeat on my machine. It also fixes several compiler-flag bugs surfaced along the way.

## Changes

1. Every TU re-parses `<dace/dace.h>` (most of a small kernel's compile) -> precompiled header, force-included into host sources only.
2. Every SDFG re-runs `project()` detection + `find_package` -> configure cache: transplant `CMakeCache.txt`.
3. CMake re-derives identical compile/link commands per SDFG -> `ninja -t compdb` record -> replay (no CMake/Ninja); Ninja default, jobs = `build_jobs`.
4. Default `-ffast-math` unsafe. It breaks `isnan/isinf`, NPBench LU ⇒ granular safe subset `-fno-math-errno -fno-trapping-math -fno-signed-zeros -freciprocal-math`.
5. Configured build type silently overridden (`-O3` args beat CMake's `-O2`) -> drop `-O`, default build type comes from the cmake config.
6. Flags handed to wrong compiler -> empty `cpu.executable` → CMake's `$CXX` is used, which can be different than `c++` if  `$CXX` set to something else -> always pass DaCe's compiler as `-DCMAKE_CXX_COMPILER` (never fall back to `$CXX`), pin to CMake's id.
7. PAPI/LIKWID append flags to global config -> leak per-SDFG (`-lpapi` everywhere) -> scoped library environments (just like other proper libraries).
8. PAPI vec-report flag hardcoded to GCC spelling (others would fail compiling) -> chosen in CMake per `COMPILER_ID` as a `papi_vectorization.cmake` CMake module.
9. Enforce no cmake extensions other than the flag provided (avoids defaulting to gnu++20 on GCC compilers and accidentally generating gnu-only code AFAIK)

## Safety
Since I could not test on Windows, the PCH and ninja is not enabled there.


dace/codegen/CMakeLists.txt
```diff
@@ -264,6 +276,7 @@ target_link_libraries(${DACE_PROGRAM_NAME} PUBLIC ${DACE_LIBS})
 
 # Set C++ standard to C++20 (or the configured standard)
 set_property(TARGET ${DACE_PROGRAM_NAME} PROPERTY CXX_STANDARD ${DACE_CPP_STANDARD})
+set_property(TARGET ${DACE_PROGRAM_NAME} PROPERTY CXX_EXTENSIONS OFF)
```
tbennun, line 292: what's that?

dace/codegen/compiler.py
```diff
@@ -162,6 +166,110 @@ def generate_program_folder(
...
+    try:
+        os.makedirs(os.path.join(staging, 'CMakeFiles'), exist_ok=True)
+        shutil.copy2(os.path.join(build_folder, 'CMakeCache.txt'), staging)
+        shutil.copytree(versions[0], os.path.join(staging, 'CMakeFiles', os.path.basename(versions[0])))
+        os.rename(staging, entry)  # atomic, and loses harmlessly to a concurrent publisher
+    except OSError:
+        shutil.rmtree(staging, ignore_errors=True)
+
+
+def _dace_pch_dir(targets) -> Optional[str]:
```
tbennun: function name is misleading

dace/codegen/CMakeLists.txt
ThrudPrimrose: We enforce C++ standard and ensure it is not GNU++20

ThrudPrimrose: I do it out of habit so that we don't accidentally produce a program that works with GCC where LLVM refuses

dace/codegen/compiler.py
ThrudPrimrose: Maybe we should name it to prepare pch or prepare precompiled_headers?

Also, as an addition: from testing on AULT and Daint, since their file system is super slow, putting it to memory backed storage is important. That is why there is the directory argument so that it can point to `/dev/shm` or such.

Also, since I do not test any of these on Windows, I should disable this feature on windows to be save?

dace/codegen/instrumentation/likwid.py
```diff
@@ -67,6 +67,42 @@ def is_installed():
         return len(LIKWID.cmake_libraries()) > 0
 
 
+class LIKWIDMarkers:
```
tbennun, line 70: Nice find!

tests/codegen/map_launch_test.py
```diff
@@ -16,7 +16,9 @@ def _make_sdfg(
     ub: Union[str, int],
     on_gpu: bool,
 ) -> tuple[dace.SDFG, dace.nodes.MapEntry]:
-    sdfg = dace.SDFG("map_test_" + ("gpu" if on_gpu else "cpu") + f"_{int(time.time())}")
+    # Nanoseconds, not seconds: the four CPU tests here share this name, and at second resolution two
```
tbennun:
```suggestion
    # The four CPU tests here share this name, and at second resolution two
```

tbennun (approved):

PR #2456, opened 2026-07-23
Fix GPU callback synchronization race (test_gpu_callback)

ThrudPrimrose:
A host callback that touches GPU data runs its device work (e.g. cupy) on a stream dace cannot see, so its data movement raced the surrounding async copies and gave wrong results under load (the flaky test_gpu_callback CI failure). This forces a GPU-touching callback that is not stream-aware (does not use `__dace_current_stream`) onto the default stream so ordering holds, and warns that a non-default stream needs a `dace.current_stream` argument.


acalotoiu (approved):
LGTM

PR #2458, opened 2026-07-23
Fix boolean-literal comparison guards dropped on SDFG serialization

ThrudPrimrose:
An interstate guard comparing a data access to a boolean literal (e.g. `A[i] != True`) collapsed to `True` on SDFG serialization/re-parse because SymPy folds `Ne(<non-Symbol>, True)` to a constant ("only Booleans can equal Booleans"), silently dropping the condition and miscompiling the graph on reload; the parser now rewrites such comparisons to boolean logic (`X == True`/`X != False` → `X`, `X != True`/`X == False` → `Not(X)`) so they survive round-tripping.


acalotoiu (approved):
LGTM

PR #2460, opened 2026-07-23
Optimize LiftTrivialIf

ThrudPrimrose:
Ports the optimized `LiftTrivialIf` (range-aware boundary-guard folding, cartesian-product/unsatisfiable branch pruning, deterministic relabeling, whole-SDFG bookkeeping walks gated on an actual removal so it is no longer quadratic in block count) and narrows `should_reapply` to `Modifies.States` so the pass re-runs only when a conditional block is added or removed, not on every interstate-edge change.


acalotoiu (approved):
LGTM

PR #2461, opened 2026-07-24
External Memory Types not Correctly Detected

philip-paul-mueller:
The storage types of the external memory is computed in the same loop that scans the data descriptor for GPU memory, which is needed to set `CompiledSDFG.has_gpu_code`, once such a descriptor is found, the loop exits.
This means if a GPU data descriptor is found before any data descriptor with external memory external memory is not found.
This PR moves the computation of the external memory storage types into its separate loops.
Furthermore, it makes the computation of the `has_gpu_code` property a bit more elegant by using `any()` and a less nested `if` construct.

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -265,20 +265,20 @@ class CompiledSDFG(object):
             warnings.warn('You passed `None` as `argnames` to `CompiledSDFG`, but the SDFG you passed has positional'
                           ' arguments. This is allowed but deprecated.')
 
-        self.has_gpu_code = False
-        self.external_memory_types = set()
-        for _, _, aval in self._sdfg.arrays_recursive():
-            if aval.storage in dtypes.GPU_STORAGES:
-                self.has_gpu_code = True
-                break
-            if aval.lifetime == dtypes.AllocationLifetime.External:
-                self.external_memory_types.add(aval.storage)
-        if not self.has_gpu_code:
-            for node, _ in self._sdfg.all_nodes_recursive():
-                if (isinstance(node, (nodes.EntryNode, nodes.ExitNode, nodes.LibraryNode))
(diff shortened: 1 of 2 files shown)


dace/codegen/compiled_sdfg.py
```diff
@@ -265,20 +265,20 @@ def __init__(self, sdfg, lib: ReloadableDLL, argnames: Optional[List[str]] = Non
...
-                break
-            if aval.lifetime == dtypes.AllocationLifetime.External:
-                self.external_memory_types.add(aval.storage)
-        if not self.has_gpu_code:
-            for node, _ in self._sdfg.all_nodes_recursive():
-                if (isinstance(node, (nodes.EntryNode, nodes.ExitNode, nodes.LibraryNode))
-                        and node.schedule in dtypes.GPU_SCHEDULES):
-                    self.has_gpu_code = True
-                    break
+        if any(aval.storage in dtypes.GPU_STORAGES for _, _, aval in self._sdfg.arrays_recursive()):
```
ThrudPrimrose: A question dtypes.GPU_STORAGES resolve to shared memory only 
```
# A subset of on-GPU storage types
GPU_STORAGES = [
    StorageType.GPU_Shared,
]
```

IMHO this is incorrect but most of old codegen relies on this so for the new codegen I have new variables defined.

Regardless, shared memory can't be external as shared memory is allocated per kernel. So my question is - shared + external should not be even valid? 

The other changes seem an improvement for the better.

philip-paul-mueller: Done

philip-paul-mueller:
cscs-ci run

ThrudPrimrose (approved):

acalotoiu (changes requested):
LGTM - one of the tests seems to have failed though, I would propose adding a test to check for the issue in order for it no get reverted in future.

acalotoiu (approved):
Thank you

PR #2462, opened 2026-07-25
Fix stale block-label cache handing out names already in use

ThrudPrimrose:
## Problem

`AbstractControlFlowRegion._ensure_unique_block_name` decides whether its label cache is stale by comparing it against the **node count**:

```python
if self._labels is None or len(self._labels) != self.number_of_nodes():
    self._labels = set(s.label for s in self.nodes())
return dt.find_new_name(proposed or 'block', self._labels)
```

Blocks are renamed **in place** in several places, and a rename leaves the node count unchanged:

- region inlining prefixes every child (`node.label = self.label + '_' + node.label`, `state.py:2800` and `:3346`)
- `ConditionFusion` relabels regions and their nodes (`condition_fusion.py:190-192`, `:277-279`)
- `DoubleBuffering` relabels a state (`double_buffering.py:185`)

After such a rename the cache is still considered fresh while holding names no block carries any more, and missing the ones they now carry. The next request is then handed a name that is already taken.

Every uniqueness path goes through this helper — `add_state`, `add_return`, `add_node(ensure_unique_name=True)` — so a rename anywhere upstream can surface as a duplicate anywhere downstream, and the SDFG fails validation with `Found multiple blocks with the same name in <region>`.

## Reproducer

On unmodified `main`:

```python
sdfg = dace.SDFG('repro')
a = sdfg.add_state('s', is_start_block=True)
b = sdfg.add_state('s')          # -> 's_0'; cache and node count agree
a.label = 'renamed'              # in-place rename; count unchanged
c = sdfg.add_state('renamed')    # cache still believed fresh
```

```
labels: ['renamed', 's_0', 'renamed'] -> duplicate: True
validate RAISES: Found multiple blocks with the same name in repro
```

Found in the wild on a large Fortran-derived pipeline, where loop unrolling inlines its iteration regions and a later block collides with a renamed one; the graph then fails validation several passes after the rename that caused it.

## Fix

Derive the label set from the live blocks on every call. A node count cannot detect a rename, and there is no cheaper check that can.

## Test

`tests/sdfg/control_flow_inline_test.py::test_unique_block_name_after_in_place_rename` — fails on `main` with `stale label cache reissued a taken name: ['renamed', 's_0', 'renamed']`, passes with the fix.

`tests/sdfg/control_flow_inline_test.py`, `loop_region_test.py`, `nested_control_flow_regions_test.py`, `conditional_region_test.py`: 25 passed.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -2881,8 +2881,9 @@ class AbstractControlFlowRegion(OrderedDiGraph[ControlFlowBlock, 'dace.sdfg.Inte
         return super().add_edge(src, dst, data)
 
     def _ensure_unique_block_name(self, proposed: Optional[str] = None) -> str:
-        if self._labels is None or len(self._labels) != self.number_of_nodes():
-            self._labels = set(s.label for s in self.nodes())
+        # Ledger of issued names, refreshed every call: in-place renames keep the node count, so a
+        # count guard rots.
+        self._labels = (self._labels or set()) | {s.label for s in self.nodes()}
         return dt.find_new_name(proposed or 'block', self._labels)
 
     def add_node(self,
--- a/tests/sdfg/control_flow_inline_test.py
+++ b/tests/sdfg/control_flow_inline_test.py
@@ -308,6 +308,22 @@ def test_loop_inlining_invalid_update_statement():
     assert len(nodes) == 3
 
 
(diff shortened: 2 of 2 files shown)


acalotoiu (approved):
LGTM, thanks!

PR #2464, opened 2026-07-27
Rewrite a dot to an arrow only between names, not inside an index expression

ThrudPrimrose:
The struct-member dot-to-arrow rewrite in `emit_memlet_reference` ran over the whole pointer expression, so a decimal literal in an index offset became `0->5` and the generated C++ did not compile.

Reproducer (`tests/codegen/cpp_test.py::test_pointer_argument_keeps_a_decimal_literal`, fails before this change): a nested SDFG whose pointer argument carries a subset with a decimal coefficient emits

```
nested_0_0_3(__state, &A[(0->5 * j)], M, N);
```

A blanket `replace('.', '->')` cannot serve both cases: `&s.f[(0.5 * i)]` needs the first dot converted and the second left alone. A dot between names is the struct access; a dot followed by a digit is part of a literal.

### DIFF
--- a/dace/codegen/targets/cpp.py
+++ b/dace/codegen/targets/cpp.py
@@ -9,6 +9,7 @@ import functools
 import itertools
 import math
 import numbers
+import re
 import warnings
 
 import sympy as sp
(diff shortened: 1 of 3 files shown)


acalotoiu (approved):
LGTM

PR #2465, opened 2026-07-28
`build_folder` is serialized

philip-paul-mueller:
`SDFG.build_folder` was not  serialized, which was not a problem if it was not set explicitly.
If set explicitly it was also not serialized, thus the following code will fail:
```python
sdfg = make_an_awesome_sdfg()
sdfg.build_folder = "/my/special/build_folder"
dolly = dace.SDFG.from_json(sdfg.to_json())
assert dolly.build_folder == sdfg.build_folder
```
This PR serialized `build_folder` is set otherwise it is not included, ensures compatibility with older versions.


dace/sdfg/sdfg.py
```diff
@@ -683,6 +684,16 @@ def _strip_transformation_history(json_obj: Any):
             _strip_transformation_history(tmp)
 
         tmp['attributes']['name'] = self.name
+        # An explicitly-set build folder is part of the user's contract and
```
ThrudPrimrose: If build folder is an attribute that needs serialization, why don't we make it an attribute with default None -> resolved to hash_sdfg?

Otherwise the PR makes sense

philip-paul-mueller: Initially I was thinking that it was not possible because Claude told me that serialization would serialize `sdfg.build_folder`, which resolves to the default.
I did a small check but I thought "yes".
But during writing this answer I found out that it is possible to do it, although at the price of making the getter of `build_folder` a free function.
While I initially did not like the idea, I think it is cleaner this way.
But we can also back to the other special case solution.

tbennun (approved):

PR #2467, opened 2026-07-29
[stree] Fix propagation of memlets with undefined symbols

mohsinm-dev:
## Description

Normalize explicitly undefined symbols before memlet propagation and route subsets containing them through the conservative fallback. This prevents scope-local symbols from leaking into outer schedule-tree memlet summaries while preserving precise propagation for symbols that remain defined.

This also re-enables the symbol-assignment and dynamic-range schedule-tree regression tests, and adds coverage for affine expressions with string- and symbol-valued undefined-variable sets.

Closes #2293.

## Testing

- `pre-commit run --all-files`
- Relevant memlet propagation and schedule-tree suite: 118 passed, 2 xfailed
- Focused regressions: 4 passed

### DIFF
--- a/dace/sdfg/propagation.py
+++ b/dace/sdfg/propagation.py
@@ -1686,10 +1686,8 @@ def propagate_subset(memlets: List[Memlet],
     else:
         defined_variables = set(defined_variables)
 
-    if undefined_variables is not None:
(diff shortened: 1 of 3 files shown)


tbennun (approved):
LGTM

PR #2468, opened 2026-07-30
Fixed Instrumentation

philip-paul-mueller:
This PR fixes two small and related bugs.
The first is `SDFG.is_instrumented()`, which uses a very clever way to check if `SDFGState`s and `AccessNode`s are instrumented. However, these two entities uses different enumeration types to and so a simple `node.instrument != {Data}InstrumentationType.No_Instrumentation` is not enough.

The second one is related to the `perf/` folder, in which the instrumentation reports are stored. If the folder is not created it will fail (because the runtime uses a bare `std::ofstream`). Before the folder was only generated in `development` mode. Now it is generated also if it is needed.

### DIFF
--- a/dace/codegen/compiler.py
+++ b/dace/codegen/compiler.py
@@ -152,8 +152,18 @@ def generate_program_folder(
 
     # Generate the parts of the folder that are exclusive to the development folder mode.
     if folder_mode in ["development"]:
-        # Copy a full snapshot of configuration script
+        # NOTE: There is a bug here, as this only saves they keys inside the configuration
+        #   `dict`. It ignores the configuration values set through environment variables.
+        #   instead it will store the ones in the `dict`.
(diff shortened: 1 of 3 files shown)


ThrudPrimrose (approved):
The reason of the fix to repeat is the following:
DaCe has 2 instrumentation types: InstrumentationType for performance counter, which is an attribute of state, CFGs and SDFGs; and DataInstrumentationType, an attribute of access nodes.
But attributes are named `.instrument` but have different types, which means all types of danger.
Certain enum values of these types are also shared thus: `type(n.instrument).No_Instrumentation` is a valid fix. Another cleaner way to fix would be to globally ensure the attributes (properties) of 2 types never share a name but that would require changing many lines of code. Thus I consider this fix to be OK.

acalotoiu (approved):
LGTM

PR #2470, opened 2026-07-30
Add CloudSC as the simplification-pipeline integration test

ThrudPrimrose:
CloudSC comes in as the integration case for the simplification pipeline. The kernel and its data
generator live in `tests/corpus/cloudsc`; the parse costs minutes, so it is memoized per process and
handed out as a copy.

- `tests/passes/constant_propagation_on_cloudsc_test.py` compiles and runs the propagated SDFG against
  the un-propagated one under IEEE flags, so a schedule change that altered a value cannot pass.
- `tests/corpus/cloudsc_regression_test.py` holds a wall-clock budget: five copies, one simplify each,
  median under 140s.

These tests carry the `long` marker that the General Tests matrix excludes, so they get their own workflow.

Split from the ConstantPropagation scheduling change, which is now #2476 and should merge first — this
branch was rebuilt on current `main` and holds only the CloudSC corpus, its tests and the CI job.


acalotoiu (approved):
LGTM! THanks for adding it!

PR #2471, opened 2026-07-31
Report GPU errors from the SDFG's own record, and stop segfaulting after reporting one

ThrudPrimrose:
`fast_call`'s `do_gpu_check` read `cudaGetLastError()`. That slot is per-host-thread and shared with every GPU user in the process, so an error left by one SDFG is charged to whichever SDFG is checked next (any call using the default `do_gpu_check=False`, or any PyTorch/CuPy activity in the same process, will do it), and because reading it also clears it, DaCe goes blind whenever another consumer reads first — even though the generated code's own `lasterror` still holds the failure. The same failure also gets reported twice, once from `fast_call` and again from `finalize()`, the second time inside `__del__` as an unraisable exception.

This exports `__dace_gpu_last_error` from the CUDA target and has `fast_call` read that instead, gated on the symbol being present rather than on the `has_gpu_code` heuristic. The accessor clears `lasterror` on read, so a failure is delivered exactly once.

Second, coupled defect: `fast_call`'s error handler `dlclose`s the library but left `_initialized = True`, so the later `finalize()`/`__del__` called through dangling function pointers and segfaulted. That is why a reported GPU error currently does not reach the user — the process dies first, and it kills whatever test runs next on the same worker. `finalize()` is now guarded on `_initialized`, which is cleared on unload.

Tests fail on `main` (4 failed, one of them by crashing the worker) and pass with the fix. The provoked error is a grid-Y-over-65535 launch, chosen because it is non-sticky and so does not poison the context for later tests.

### DIFF
--- a/dace/codegen/compiled_sdfg.py
+++ b/dace/codegen/compiled_sdfg.py
@@ -235,6 +235,9 @@ class CompiledSDFG(object):
         self._exit = lib.get_symbol('__dace_exit_{}'.format(sdfg.name))
         self._exit.restype = ctypes.c_int
         self._cfunc = lib.get_symbol('__program_{}'.format(sdfg.name))
+        # Present exactly when a GPU target emitted its init/exit pair, which is a sharper test than
+        # the ``has_gpu_code`` heuristic below.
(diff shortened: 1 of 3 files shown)


acalotoiu (approved):
LGTM

PR #2472, opened 2026-08-02
Add a `@dace.always_inline` decorator

tbennun:
With a robust Python frontend, some functions cannot be trusted to be side-effect free. This decorator can be placed on functions to ensure that dace understands that they will in fact be side-effect free. Avoids the need to wrap the function with `dace.inline(func_call(...))` and setting the global dace.program `constant_functions` kwarg, which would be dangerously assuming all callbacks are inlineable.

### DIFF
--- a/dace/frontend/python/preprocessing.py
+++ b/dace/frontend/python/preprocessing.py
@@ -751,12 +751,14 @@ class GlobalResolver(astutils.ExtNodeTransformer, astutils.ASTHelperMixin):
         return self._visit_potential_constant(node, True)
 
     def visit_Call(self, node: ast.Call) -> Any:
-        from dace.frontend.python.interface import in_program, inline  # Avoid import loop
+        from dace.frontend.python.interface import in_program, inline, is_always_inline  # Avoid import loop
 
         if hasattr(node.func, 'value') and isinstance(node.func.value, SDFGConvertible):
             # Skip already-parsed calls
             return self.generic_visit(node)
 
+        # Callees marked with ``@dace.always_inline`` are folded regardless of the ``resolve_functions`` setting
+        always_inline = False
         try:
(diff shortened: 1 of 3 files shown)


dace/frontend/python/preprocessing.py
```diff
@@ -767,15 +769,21 @@ def visit_Call(self, node: ast.Call) -> Any:
...
+            if self.resolve_functions or always_inline:
                 global_val = astutils.evalnode(node, self.globals)
             else:
                 global_val = node
         except SyntaxError:
+            if always_inline:
+                raise DaceSyntaxError(
+                    None, node, f'Cannot evaluate the call to "{astutils.unparse(node.func)}" at compile time, '
+                    'even though it is annotated with @dace.always_inline. Make sure all of its arguments are '
+                    'compile-time constants.')
```
ThrudPrimrose, line 782: Wouldn't be technically possible for a fuction having pure read-only inputs (constant inputs but not compile time constants) + a return statement be side-effect free? (Most CloudSC and GT4py helper functions are like that, reads some stuff, returns a scalar)

ThrudPrimrose (approved):
LGTM.  (Check the minor comment)

dace/frontend/python/preprocessing.py
ThrudPrimrose: Since I don't think the setattr/getattr would become a bottleneck here, I guess this is not that important but: Would setting `f.always_inline = False` to globally False, and later not needing setattr/getattr improve the time?

tbennun: 1. Inlining here refers to compile-time inline (of expressions as the test shows)
2. I thought of that, but then I thought that we want to modify external objects as rarely as possible

ThrudPrimrose (approved):

PR #2473, opened 2026-08-02
Fix unsound lifting of loops whose guard block contains dataflow

philip-paul-mueller:
## Fix unsound lifting of loops whose guard block contains dataflow

`LoopLifting` miscompiled loops whose guard block carries dataflow (expr. index 0/1 with a non-empty guard).
Two defects:

1. **Uninitialized condition read.**
    The guard block runs before every condition check, so it runs once more than the loop body - and the condition may depend on data the guard computes.
    The lifted, pre-checked loop region evaluated the condition *before* the guard's dataflow ever ran.
    If the condition reads a transient the guard writes, that first check reads uninitialized memory: the generated code contained `int64_t cont;` followed by `for (; (!cont); )`.
    Whether this misbehaved depended on stack garbage, which made it surface as a rare, scheduling-dependent wrong result (`tests/passes/constant_propagation_test.py::test_dependency_change_same_edge[True]`, which intermittently computed 0 instead of 540 in CI and resisted local reproduction for a long time).
2. **Dead "loop not executed" branch.**
    The construction that should run the guard once when the loop body never executes negated the condition *after* the edge-rewrite loop had already replaced it with `1`, producing the never-taken branch condition `(not 1)` - visible as `if (false)` in the generated code.

### Before / after

The miscompiled pattern (from `test_lift_while_loop_with_data_dependent_condition`, distilled): the guard block computes the transient `cont`, and the loop continues while `not cont`.
The original state machine, written as Python:

```python
i61, i17 = 1, 0
while True:
    a[0] += i17          # guard block dataflow...
    cont = (i61 == 10)   # ...computes the condition input
    if cont:             # the condition is checked AFTER the block ran
        break
    i61, i17 = i61 + 1, i61 * 12    # latch/extra assignments
```

Generated code **before** (defects marked):

```c
int64_t cont;                       /* never written before the loop check */
i61 = 1; i17 = 0;
if (false) {                        /* (2) dead branch: negate of the        */
    a[0] += i17;                    /*     already-overwritten condition '1' */
    cont = (i61 == 10);
} else {
    for (; (!cont); ) {             /* (1) reads uninitialized cont (UB)     */
        a[0] += i17;
        cont = (i61 == 10);
        i60 = i61; i61 = i60 + 1; i17 = i60 * 12;
    }
}
```

Whether this misbehaves depends on the garbage in `cont`: a zeroed stack slot enters the loop and computes the right result, a non-zero one skips the loop entirely.

Generated code **after** (the rotation):

```c
int64_t cont;
i61 = 1; i17 = 0;
a[0] += i17;                        /* unconditional copy of the guard block */
cont = (i61 == 10);                 /* -> cont is defined before the check   */
for (; (!cont); ) {
    i60 = i61; i61 = i60 + 1; i17 = i60 * 12;
    a[0] += i17;                    /* original guard block, moved to the    */
    cont = (i61 == 10);             /* END of the iteration                  */
}
```

Every condition check now happens immediately after an execution of the guard block, exactly as in the state machine.

The same rotation also fixes the execution count (from `test_lift_for_loop_with_dataflow_guard`): a guard that writes `A[i]` in a 10-iteration loop runs **11** times in the state machine (it also runs on the final, failing check), which the rotated form preserves (`1` pre-loop + `10` in-loop) - the old pre-checked form ran it only 10 times.

### The fix

Guard loops with dataflow are now lifted by **rotation** instead of the conditional construction:
an unconditional copy of the guard executes before the loop region, and the original guard moves to the end of each iteration.
Every condition check therefore happens immediately after a guard execution, exactly as in the original state machine, and the guard's extra final execution is preserved.
The rotated region deliberately carries no iteration-variable metadata; initialization stays on the init edge and the increment on the former back edge.

For self loops (expr. index 4) the conditional construction is kept, but its branch condition is now derived from the condition captured *before* the rewrite (fixing defect 2 there); a `FIXME` documents that the pre-checked region still drops the block's final execution, since the sound do-while form is not yet expected by downstream transformations such as `LoopToMap`.

Loops with empty guards - the common case - are unaffected.

### Tests

- `test_lift_for_loop_with_dataflow_guard`: deterministic; the guard writes `A[i]` and must run 11 times for a 10-iteration loop.
- `test_lift_while_loop_with_data_dependent_condition[*]`: the miscompiled pattern from the constant-propagation test; compiled with `-ftrivial-auto-var-init=pattern` so an uninitialized condition read fails deterministically instead of depending on stack garbage (skips if the compiler lacks the flag).
    Also asserts that no lifted conditional branch carries a constant-false condition.

Note: running the whole test suite with `-ftrivial-auto-var-init=pattern` exposes an unrelated pre-existing issue: `test_inverted_loop_with_additional_increment_assignment` constructs an SDFG that reads its transient scalar `i` uninitialized and crashes with an out-of-bounds write under that flag.


ThrudPrimrose:
Can you write a C-code or Python-code based before and after sample to the docs & explanation such that easier to visually map when reviewing?

philip-paul-mueller:
@ThrudPrimrose A friend of mine updated the PR description.

ThrudPrimrose:
Code sample and change makes sense in the first look.
The changes are a little hard to understand due to it is being form a part I'm not too familiar with. I need some more time for a final review.

ThrudPrimrose:
I added 1 more test, as I was not really sure if the detection would work in this weird case of nested while-true loops with many breaks. Seems to work fine.
```
NI, NJ = 3, 4
A = np.zeros((NI + 1,), np.int32)      # outer-guard execution counter
B = np.zeros((NI, NJ + 1), np.int32)   # inner-guard execution counter
C = np.zeros((NI, NJ), np.int32)       # body execution counter

i = 0
while True:
    A[i] += 1                 # OUTER guard dataflow - runs BEFORE the check
    if i >= NI:               # ...so it also runs on the final, failing check
        break
    j = 0
    while True:
        B[i, j] += 1          # INNER guard dataflow - same, one per check
        if j >= NJ:
            i = i + 1         # outer increment sits on the inner exit edge
            break
        C[i, j] += 1          # body
        j = j + 1

assert (A == 1).all()   # outer guard ran NI+1 = 4 times, not NI
assert (B == 1).all()   # inner guard ran NJ+1 = 5 times per outer iteration
assert (C == 1).all()
```

ThrudPrimrose (approved):
I never dealt with the lifting of unstructured control flow to loop regions, so I'm not sure if there is an edge case we are missing, well, there were more edge cases missing before. I do not see anything that could introduce a new problems. The add-node does not check for start-block explicitly, but it was also not being done before, so I assume it was safe before as well.

I think the two of tests are really complicated making this PR safe.

Wrt. code-qualtiy the implementation looks alright. I don't see any code smells or anti-patterns, should be fine.

acalotoiu (approved):
LGTM

philip-paul-mueller:
cscs-ci run

PR #2478, opened 2026-08-03
Forward replace_keys when replace_dict recurses into a nested control flow region

ThrudPrimrose:
`ControlGraphView.replace_dict` dropped `replace_keys` when recursing into child blocks and `LoopRegion`/`ConditionalBlock` dropped it again, so each fell back to its own conflicting default and a rename could rewrite an assignment's uses without its target, silently splitting one symbol into two.

### DIFF
--- a/tests/sdfg/replace_keys_test.py
+++ b/tests/sdfg/replace_keys_test.py
@@ -1,0 +1,30 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
+""" Tests that ``replace_keys`` reaches interstate edges inside nested control flow regions. """
+
+import dace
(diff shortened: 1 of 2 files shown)


tests/sdfg/replace_keys_test.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2019-2024 ETH Zurich and the DaCe authors. All rights reserved.
```
phschaad:
```suggestion
# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
```

phschaad (changes requested):
LGTM % one small change

phschaad (approved):

PR #2479, opened 2026-08-05
Fix `get_binary_name()`

philip-paul-mueller:
`get_binary_name()` did not respect the `FOLDER_MODE` file automatically. The new behaviour is to probe the mode, using `get_folder_mode()` and only if that fails fall back to the configuration key.
This is the same behaviour as `load_precompiled_sdfg()` has.


tests/compile_folder_mode_test.py
```diff
@@ -234,7 +311,12 @@ def test_already_loaded_and_comple_again():
 
 
 if __name__ == '__main__':
+    import tempfile
```
ThrudPrimrose: We should make this a fixture? And enable the 2 tests via passing some kind of pytest lists?

ThrudPrimrose: We should make this a fixture? And enable the 2 tests via passing some kind of pytest lists?

ThrudPrimrose (changes requested):

ThrudPrimrose (approved):

acalotoiu (approved):
LGTM

PR #2480, opened 2026-08-05
Fix three GPU codegen bugs that silently produce a wrong or uncompilable program

ThrudPrimrose:
Fixes three GPU code generation bugs that each yield a silently wrong or non-compiling program: unchecked CUB calls in the device reduction let a null workspace return all zeros, the legacy default stream that a stream-unaware callback is pinned to was never synchronized at the end of a state, and an array reached only indirectly through a tiled map's exits was dropped from the kernel signature while its body still referenced it. Position-independent code is also enforced through `CMAKE_POSITION_INDEPENDENT_CODE` rather than a configurable compiler argument, and every change carries tests that fail on current main.


acalotoiu (approved):
Excellent! LGTM

PR #2482, opened 2026-08-06
Keep Range bounds symbolic when they are assigned, not only when constructed

ThrudPrimrose:
`Range.__init__` coerces every bound through `pystr_to_symbolic` and `ndrange()` is annotated SymbolicType, but `__setitem__` stored the tuple verbatim. So `r[i] = (0, n - 1, 1)` puts plain Python ints into a container whose contract says symbolic.

Callers trust that contract: `LoopToMap._check_range` calls `.match()` on a bound, others call `.subs()` or read `.free_symbols`. An int bound therefore fails far from the assignment that caused it, as `AttributeError: 'int' object has no attribute 'match'`. LoopToMap catches matcher exceptions and downgrades them to a warning, so the loop silently stays sequential with no reason reported — one corpus sweep hit this 1028 times.

Coercion moves into a helper shared by the constructor and the setter, so the two paths cannot drift again, and a malformed range now raises where it is written.

Test fails on the current main with `2 is a raw int`.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -412,9 +412,11 @@ def add_indirection_subgraph(sdfg: SDFG,
                         toreplace = 'index_' + fname + '_' + str(len(accesses[fname]) - 1)
 
                     if direct_assignment:
-                        # newsubset[dimidx] = newsubset[dimidx].subs(expr, toreplace)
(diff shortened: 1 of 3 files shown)


acalotoiu (approved):
LGTM

PR #2483, opened 2026-08-06
Register a control-flow region in the CFG list when it is added

ThrudPrimrose:
An unregistered region does not fail loudly. `cfg_id` is `cfg_list.index(self)`, and every region is constructed holding a list containing only itself, so a region that was never registered answers **0** — the same id as the root SDFG and as every other unregistered region. Then things break.

### Also fixed

`add_branch` never propagated the SDFG into the branch it was given. A branch is reached only through `branches`, never through `nodes()`, so the bookkeeping in `add_node` never runs for it; its blocks kept `sdfg is None` and `to_json` failed on a fresh conditional with `AttributeError: 'NoneType' object has no attribute 'symbols'`.

### Tests

`tests/sdfg/cfg_list_registration_test.py` covers the four construction paths, a region carrying a subtree, and a serialization round trip. Both tests fail on current main (`regions missing from cfg_list: ['myloop']` / `['loop', 'cond', 'br']`).

### Note

Reset cfg list is not too slow, but we could update lazily by setting an "invalid cfg list" flag and resetting only an API call needs to latest one.

### DIFF
--- a/dace/sdfg/state.py
+++ b/dace/sdfg/state.py
@@ -2914,7 +2914,12 @@ class AbstractControlFlowRegion(OrderedDiGraph[ControlFlowBlock, 'dace.sdfg.Inte
         node.sdfg = sdfg
         if isinstance(node, AbstractControlFlowRegion):
             for n in node.all_control_flow_blocks():
-                n.sdfg = self.sdfg
+                n.sdfg = sdfg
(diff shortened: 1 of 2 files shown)


acalotoiu (approved):
LGTM

phschaad (approved):
LGTM

PR #2484, opened 2026-08-06
Rank-aware build cache: one folder per rank under cache_distaware

ThrudPrimrose:
Ranks of one job derive the same build folder. Ranks that each compile therefore build on top of
each other, and a rank can load a library another rank is still writing. Eight processes running one
GPU test out of one folder failed six times with `Could not load library ...so`; with a folder each,
none.

### What this does

A new config entry, `cache_distaware` (`DACE_cache_distaware=1`), names the build cache *root* after
the rank the launcher advertises: `.dacecache_rank3/`. Splitting the root rather than the SDFG name
means every `cache` mode gets the separation, not only the ones that mangle the name.

It is off by default, because sharing one build is a valid setup too: `distributed_compile` has rank
0 build and every other rank load its folder. That path now pins the broadcast folder on the ranks
that hold the SDFG, so no rank goes looking in a folder nobody built. Ranks that only load may keep
passing `None`, as `tests/library/mpi` does.

### Tests

`tests/custom_build_folder_test.py` covers every launcher variable, all four cache modes, the
untouched default, a process no launcher started, and `distributed_compile` sending every rank to
rank 0's folder.

### DIFF
--- a/dace/config_schema.yml
+++ b/dace/config_schema.yml
@@ -738,6 +738,16 @@ required:
             potentially build time, but disallows executing SDFGs in parallel
             and caching of more than one simultaneous SDFG.
 
+    cache_distaware:
+        type: bool
+        default: false
+        title: Distribution-aware build cache
+        description: >
+            Give every rank of a job its own build folder, named after the rank its launcher
+            (MPI, Flux, Slurm) advertises. Without it, ranks that each compile share one folder
+            and can load a library another rank is still writing. Leave it off when only one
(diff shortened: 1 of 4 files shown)


tbennun:
cscs-ci run

tbennun (approved):

romanc:
> It is off by default, because sharing one build is a valid setup too

Quick side-note in case anyone else lands here searching for what has changed: Contrary to what is claimed in the description, the new behavior is __on by default__ (as can be seen from the default config, which is set to `true`).

PR #2487, opened 2026-08-10
[fix] Parse `array is None` if CFG conditionals when quering for memlets

FlorianDeconinck:
We encounter a case of CFG within a NestedSDFG where the condition is of the form 
```python
array is None
```
and where `array` is a legit optional parameter. The SDFG -> STREE (and other) system relies on parsing the conditional AST to extract the arrays and propagate the memlets properly. The code ends up in  `dace.sdfg.memlets_in_ast` where it is stated that the limitations are:
- no indirect arrays (expecting to be able to use symbolic expression)
- scalars

We propose here to add a narrow parse of `array is None` to strengthen the list. Implementation restrict the parsing to a single form to evade run away checks.

### DIFF
--- a/dace/sdfg/sdfg.py
+++ b/dace/sdfg/sdfg.py
@@ -123,6 +123,7 @@ def memlets_in_ast(node: ast.AST, arrays: Dict[str, dt.Data], *, include_scalars
     """
     Generates a list of memlets from each of the subscripts that appear in the Python AST.
     Assumes the subscript slice can be coerced to a symbolic expression (e.g., no indirect access).
+    Can also parse a None check of the form `array is [not] None`.
 
     :param node: The AST node to find memlets in.
     :param arrays: A dictionary mapping array names to their data descriptors (a-la ``sdfg.arrays``)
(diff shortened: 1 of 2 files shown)


dace/sdfg/sdfg.py
```diff
@@ -133,10 +134,16 @@ def memlets_in_ast(node: ast.AST, arrays: Dict[str, dt.Data], *, include_scalars
 
     for subnode in ast.walk(node):
         if isinstance(subnode, ast.Subscript):
-            data = astutils.rname(subnode.value)
             data, slc = astutils.subscript_to_slice(subnode, arrays)
             subset = sbs.Range(slc)
             result.append(mm.Memlet(data=data, subset=subset))
+        elif (isinstance(subnode, ast.Compare) and len(subnode.ops) == 1
+              and isinstance(subnode.ops[0], ast.Is | ast.IsNot) and len(subnode.comparators) == 1
```
tbennun: Please use tuples for more than one type

tbennun (changes requested):
Very minor change, otherwise lgtm

acalotoiu (approved):
LGTM

tbennun (approved):

PR #2490, opened 2026-08-11
Emit host-side synchronization for GPU-to-host copies

ThrudPrimrose:
A device-to-host copy is issued as an asynchronous `MemcpyAsync` on a stream, but the synchronization site after it in the CUDA code generator was a bare `pass` deferring to the destination, and the destination-side helper returns early on any node that stream assignment stamped with a `_cuda_stream` -- so no host synchronization was emitted anywhere. The shape that bites is trisolv: a device scalar is copied into host memory and the next kernel launch takes it by value, which the host reads while packing launch arguments; pageable host memory hides this because the driver blocks on it anyway, while pinned memory, HIP and graph capture do not.

This emits the `StreamSynchronize` where the copy is issued, forces it for any host-located destination regardless of the stream stamp, and narrows the surrounding predicate so a same-stream consumer -- which the stream already orders -- no longer requests one. `tests/codegen/gpu_d2h_host_sync_test.py` asserts on the emitted code that the synchronization sits between the device-to-host copy and the kernel launch that reads it and names the copy's own stream, and that a device-to-device copy still gets no host wait.


phschaad (approved):
LGTM

PR #2492, opened 2026-08-12
Do not charge a foreign GPU error to the next DaCe program

philip-paul-mueller:
## The problem

The CUDA runtime's last-error slot is **per-host-thread and shared with every other GPU user in
the process** — CuPy, another SDFG, any library.
A value left pending in it is returned by the next runtime call DaCe makes, so the first
`DACE_GPU_CHECK`-wrapped call in a generated module reports someone else's failure as its own.

That is misleading twice over: it blames an innocent call, and it hides whoever actually failed.

DaCe already knows this slot is unreliable. `__dace_gpu_last_error` deliberately reads the SDFG's
own record instead, with this comment in `dace/codegen/targets/cuda.py`:

> The runtime's own last-error slot is per-host-thread and shared with every other GPU user in the
> process, so it is not a reliable carrier for this SDFG's failures.

The init path simply had no equivalent protection.

## How it was found

Intermittent `invalid device ordinal` failures in a GPU CI job running 32 pytest-xdist workers
over 4 GPUs.
The failures landed on a **different test each run** —
`reduce_test.py::test_multidim_gpu`, `wcr_cudatest.py::test`,
`block_allreduce_cudatest.py::test_blockallreduce` — and occurred under **both** compiled-SDFG
interfaces, which ruled out anything interface-specific.

`DACE_GPU_CHECK` prints `__FILE__:__LINE__` before recording, and pytest shows captured stdout on
failure. Two runs reported the identical generated line, in different SDFGs.
Since code generation needs no GPU, the failing translation unit can be regenerated on a
developer machine, which identifies it exactly:

```c
DACE_GPU_CHECK(cub::DeviceReduce::Reduce(nullptr, __cub_ssize_test_sdfg_0_2, ...));
```

That is the CUB temp-storage size query emitted into `__dace_init_cuda` by
`dace/libraries/standard/nodes/reduce.py`.
It passes `d_temp_storage = nullptr` and launches nothing — but it consults the runtime, and CUB
checks the error state internally.
It is the usual victim simply because it is often the **first checked call** in a generated
module. All three affected tests are CUB reductions.

### 1. Drain the slot, on every call

A new exported `__dace_gpu_drain_error` in the generated `.cu` discards whatever is pending and
prints what it discarded. It is called from two places:

- the top of `__dace_init_cuda`, before any checked call;
- the top of **every** `__program_<name>` invocation — initialization runs once per state, while
  contamination can arrive between any two calls.

The frame code declares the function and emits the call only when the cuda target is in use,
exactly as it already declares `__dace_init_<target>`.
The generated `.cpp` cannot include the CUDA headers, so the call has to cross into the `.cu`.

Design points:

- **Clearing is required, not incidental.** `cudaPeekAtLastError()` would observe without
  clearing, but the stale value would still be pending when the next checked call runs — the
  problem would be diagnosed and not fixed.
- **Only non-sticky errors can be discarded.** A sticky one (illegal access, corrupted context)
  survives `gpuGetLastError()` and is reported through the normal path by the next real call.
- **It reports rather than throws.** A pending error is by definition not this SDFG's, and failing
  there would penalize the very SDFG this protects.

### 2. Check the memory-pool setup

`cudaDeviceGetDefaultMemPool` and `cudaMemPoolSetAttribute` were unchecked.
An unchecked call that can fail is precisely what leaves an error pending for the next checked
call to inherit, so they are now wrapped in `DACE_GPU_CHECK`.

Other unchecked backend calls were left alone deliberately: `GetDeviceCount` has its own explicit
error handling, `DeviceSynchronize` in `__dace_exit_cuda` has its result returned, and
`LaunchKernel` is followed by `DACE_KERNEL_LAUNCH_CHECK`.

### 3. Two fixes needed to make that wrapping safe

These were not part of the original intent but are required for correctness:

- The pool setup was emitted **before** `__state->gpu_context` is constructed, and
  `DACE_GPU_CHECK` records into that context.
  Wrapping the calls as written would have turned a mempool failure into a dereference of an
  unset pointer.
  The pool setup now runs after the context exists.

- `gpu_context` was a raw pointer with **no initializer** in a state struct allocated via `new T`
  (default-initialized, so the member was indeterminate), while the runtime warm-up allocation in
  `__dace_init_cuda` is already checked before the context is assigned.
  It now has a `= nullptr` initializer, and `DACE_GPU_CHECK` guards the recording on a non-null
  context while printing either way.
  **This is a pre-existing latent bug on the init error path**, independent of everything else
  here.

## Verification, and its limits

The GPU code is verified by **regenerating it, not by compiling or running it** — there is no CUDA
toolchain on the machine this was written on. Stated plainly so reviewers can weight it.

Confirmed on a real GPU SDFG (`tests/wcr_cudatest.py`):

- the drain is defined in the `.cu` and called from `__dace_init_cuda`;
- it is declared before use in the `.cpp` and called at the top of `__program_<name>`;
- a pooled SDFG emits both mempool calls wrapped, with the context constructed first;
- a CPU-only SDFG emits no drain at all.

The CPU path is exercised for real: **137 tests pass across `tests/codegen`**.
The frame code change touches every generated program, which is why that coverage matters.

**This is not claimed as a proven fix** for the CI flakiness that prompted it.
The evidence there is failures before and green runs after; the warning was never observed firing,
because pytest suppresses captured stdout for *passing* tests (confirming it on a green run would
need `-rP` or `--capture=no`).
The change stands on its own as correct error attribution regardless.

A separate root cause in pytest's own test distribution was identified independently by a DaCe
developer and is not addressed here.

## Review questions

1. **Should the warning be unconditional?** If some workflow routinely leaves errors pending, the
   `printf` fires on every call. Gating it behind a debug config, or routing it through DaCe's
   logging rather than `printf`, would both be reasonable.
2. **Is `__program_<name>` the right hook** for the per-call drain, or would a target-provided
   per-invocation preamble be preferable to the frame code knowing about the cuda target by name?
3. **Should the drain also run at program exit**, so an error DaCe leaves pending does not
   contaminate the next user of the slot?

## Not included

`dace/codegen/targets/cuda.py` still hardcodes the device ordinal in
`cudaDeviceGetDefaultMemPool(&mempool, 0)`.
That call is now checked, but whether ordinal `0` is right in a multi-device or
`CUDA_VISIBLE_DEVICES`-restricted context is a separate question, left alone here.

### DIFF
--- a/dace/codegen/targets/cuda.py
+++ b/dace/codegen/targets/cuda.py
@@ -258,7 +258,11 @@ class CUDACodeGen(TargetCodeGenerator):
         self._compute_pool_release(sdfg)
 
         # Write GPU context to state structure
-        self._frame.statestruct.append('dace::cuda::Context *gpu_context;')
+        # Initialized, because the state struct is allocated with `new T` (default-initialized, so
+        # a raw pointer member would be indeterminate) and DACE_GPU_CHECK reads this field. The
+        # warm-up allocation in __dace_init_cuda is checked before the context is constructed, so
+        # on that error path the field is read before it is assigned.
+        self._frame.statestruct.append('dace::cuda::Context *gpu_context = nullptr;')
 
         # Collect all defined symbols and argument lists with one traversal
         shared_transients = {}
@@ -399,9 +403,9 @@ class CUDACodeGen(TargetCodeGenerator):
             poolcfg = int(Config.get('compiler', 'cuda', 'mempool_release_threshold'))
             pool_header = """
     {backend}MemPool_t mempool;
-    {backend}DeviceGetDefaultMemPool(&mempool, 0);
+    DACE_GPU_CHECK({backend}DeviceGetDefaultMemPool(&mempool, 0));
     uint64_t threshold = {poolcfg_threshold};
-    {backend}MemPoolSetAttribute(mempool, {backend}MemPoolAttrReleaseThreshold, &threshold);
+    DACE_GPU_CHECK({backend}MemPoolSetAttribute(mempool, {backend}MemPoolAttrReleaseThreshold, &threshold));
 """.format(backend=self.backend, poolcfg_threshold=('UINT64_MAX' if poolcfg == -1 else poolcfg))
 
         self._codeobject.code = """
@@ -413,6 +417,7 @@ class CUDACodeGen(TargetCodeGenerator):
 DACE_EXPORTED int __dace_init_cuda({sdfg_state_name} *__state{params});
 DACE_EXPORTED int __dace_exit_cuda({sdfg_state_name} *__state);
 DACE_EXPORTED int __dace_gpu_last_error({sdfg_state_name} *__state);
+DACE_EXPORTED void __dace_gpu_drain_error({sdfg_state_name} *__state);
 DACE_EXPORTED bool __dace_gpu_set_stream({sdfg_state_name} *__state, int streamid, gpuStream_t stream);
 DACE_EXPORTED void __dace_gpu_set_all_streams({sdfg_state_name} *__state, gpuStream_t stream);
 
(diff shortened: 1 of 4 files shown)


ThrudPrimrose:
I kind of think that, the tests, which are only testing whether the code is emitted, are unnecessary.
Technically speaking, writing a test that emits and error to further drain is far fetched, and I think for this PR, it could make sense to skip tests? what do you think?

ThrudPrimrose:
Closed because the order GPU flake PR has these featuers implemented too.

PR #2494, opened 2026-08-12
[opt] cache sympy pattern for matching in `simplify_ext(expr)`

romanc:
## Description

`dace/symbolic.py` exposes a function `simplify_ext(expr)` that matches expressions like `min(a, b) + c` and `max(a, b) + c`. This function is known to be slow and previous attempts (e.g. PR https://github.com/spcl/dace/pull/2093) have been made to avoid the function at all cost.

In this PR, we are looking into the function, optimizing a crucial part. To find the min/max expressions, the function uses sympy pattern-matching. The patterns are re-built for every call. This is unnecessary because patterns (similar to regex) can be re-used once created. It turns out, pattern creation is a (surprisingly) expensive operation. By simply caching the patterns, we can get significant speedup for codes calling the function often (e.g. subset covers).

#### Motivation

This PR is motivated by SDFG -> schedule tree -> SDFG roundtrips. While it's fast to go from an SDFG to a schedule tree, the reverse transformation (from schedule tree to SDFG) takes a long time. This is mostly due to the fact that (re-)building an SDFG from the schedule tree, we have to make a ton of calls to figure out if one subset overlaps with another or if ti's covered.

For our testcase, we spend 6:40 in `covers()` before the optimization where 4:47 min is spent creating sympy patterns as shown from the profile below.

image

_profile before_

with the pattern caching proposed in this PR, we can remove this time entirely, reducing the runtime of `covers()` to 1:18 min and the total runtime from 7+ minutes to 2+ minutes.

image

_profile after_

### DIFF
--- a/dace/symbolic.py
+++ b/dace/symbolic.py
@@ -2,7 +2,7 @@
 import ast
 import contextlib
 from collections import Counter
-from functools import lru_cache
+from functools import lru_cache, cache
 import sympy
 import threading
 import pickle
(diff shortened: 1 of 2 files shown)


dace/symbolic.py
```diff
@@ -1445,6 +1445,23 @@ def sympy_divide_fix(expr):
     return nexpr
 
 
+@cache
```
romanc, line 1448: According to [the docs](https://docs.python.org/3/library/functools.html#functools.cache), `@cache`  is equivalent but faster than an unbound `@lru_cache`. Since the function has no arguments and thus returns/caches only one set of values, there's no chance of cache explosion and thus no need for eviction.

ThrudPrimrose (approved):
Nice, LGTM!

romanc:
thx for the approval and auto-erge - someone with access rights will need to kick off the GPU tests, otherwise, this will sit waiting for it

edopao:
cscs-ci run GPU

PR #2497, opened 2026-08-13
[opt] guard expensive pattern matching in simplify_ext(expr)

romanc:
## Description

`dace/symbolic.py` exposes a function `simplify_ext(expr)` that matches expressions like `min(a, b) + c` and `max(a, b) + c`. This function is known to be slow and previous attempts (e.g. PR https://github.com/spcl/dace/pull/2093) have been made to avoid the function at all cost.

This PR builds on PR https://github.com/spcl/dace/pull/2494 and not only caches the pattern to be matched, but guards pattern matching with a much simpler check. This - again - speeds up repeated calls to the function were we can't match (e.g. single symbols).

#### Motivation

See #2494 for the full story. Basically we start from a profile where we spend 1:18 min in `covers()` mostly due to matching `{Min,Max}(a, b) + c` 

image

_profile before this PR_

With this PR, we can reduce the time spent in `covers()` to 5 seconds!

image

_profile after this PR_

bringing down the whole test case to something just over a minute.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,5 +37,6 @@ Yihang Luo
 Alexandru Calotoiu
 Phillip Lane
 Samuel Martin
+Roman Cattaneo
 
 and other contributors listed in https://github.com/spcl/dace/graphs/contributors
--- a/dace/symbolic.py
+++ b/dace/symbolic.py
@@ -1474,13 +1474,15 @@ def simplify_ext(expr):
 
     # Push expressions into both sides of min/max.
     # Example: Min(N, 4) + 1 => Min(N + 1, 5)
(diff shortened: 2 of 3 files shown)


ThrudPrimrose (approved):
LGTM

edopao:
cscs-ci run GPU

acalotoiu (approved):
LGTM

tbennun:
@romanc  looks like match in memlet propagation could use the same treatment :)

romanc:
> @romanc looks like match in memlet propagation could use the same treatment :)

Indeed it shows up "next" in the profile. Feel free to open up a PR during your day. Otherwise, I'll see what can be done tomorrow. This is gonna be interesting. Memlet propagation is used in a bunch of places.

PR #2498, opened 2026-08-13
Fix affine memlet propagation dropping the multiplier of a zero-offset access

ThrudPrimrose:
`AffineSMemlet.propagate` has a special case for the `A[i:i+stride]` shape inside a `begin:end:stride` map range, which returns the map range verbatim:

```python
if (node_rb == result_begin and (re - rb + 1) == node_rs and rs == 1 and rt == 1):
    return (node_rb, node_re, 1, 1)
```

A **single-element** access has `re - rb + 1 == 1`, which equals `node_rs` on any unit-stride map range, and `a * i` at a zero map begin makes `result_begin == node_rb`. Both tests therefore hold for an access the case was never meant to cover, and returning the map range verbatim discards the multiplier.

```
C[2*i] over i=0:N   ->  C[0:N]        (wrong)
                        C[0:2*N-2:2]  (expected)
C[3*i] over i=0:N   ->  C[0:N]        (wrong)
C[2*i+1]            ->  C[1:2*N-1:2]  (correct -- non-zero offset takes the general path)
```


acalotoiu (approved):
LGTM

PR #2500, opened 2026-08-14
[opt] guard sympy matching in memlet propagation

romanc:
## Description

As learned in PRs #2497, pattern matching in sympy is costly and a simple up-front check can help reduce that burden.

This PR follows the motivating example in #2494 manages to shave off 10% runtime in that benchmark.

image

_initial profile_

In particular, we are attacking `sympy.match()` (light green box) showing inside `propagate_memlets_state()`. More generally, memlet propagation is used in a bunch of places, so it might also be visible in other places (given "bigger" SDFGs).

image

_profile with the changes of this PR_

/cc @tbennun quick PR as discussed. Bottleneck starts to shift towards free symbols now ... For us, this is an okay point to stop optimizing sdfg -> stree -> sdfg roundtrip times. There's still lots to do and we'll be back for sure in the future.


acalotoiu (approved):
LGTM

romanc:
Thx. Someone with avmccess rights will need to kick off GPU tests. They don't run automatically for me (must be some kind of permission issue).

tbennun:
cscs-ci run

PR #2501, opened 2026-08-17
frontend: drop the redundant process-grid merge from ProgramVisitor.defined

ThrudPrimrose:
### What

`ProgramVisitor.defined` is read while parsing every statement. Its MPI block read
`self.sdfg.process_grids` **inside** a dict comprehension over `self.variables`:

```python
result.update({
    v: self.sdfg.process_grids[v]
    for k, v in self.variables.items() if v in self.sdfg.process_grids
})
```

`process_grids` is a property that builds a fresh dict by scanning every descriptor in
`SDFG._arrays`, so the membership test re-scanned the whole descriptor repository once per
variable — `|variables| x |arrays|` work per access, on a property read once per parsed
statement.

### DIFF
--- a/dace/frontend/python/newast.py
+++ b/dace/frontend/python/newast.py
@@ -1372,11 +1372,11 @@ class ProgramVisitor(ExtNodeVisitor):
         # Add SDFG arrays, in case a replacement added a new output
         result.update(self.sdfg.arrays)
 
-        # MPI-related stuff
-        result.update({
-            v: self.sdfg.process_grids[v]
-            for k, v in self.variables.items() if v in self.sdfg.process_grids
-        })
+        # MPI-related stuff. ``process_grids`` is a PROPERTY that rebuilds a dict by scanning every
+        # descriptor in the SDFG, so reading it inside the comprehension re-scanned once per variable:
+        # |variables| x |arrays| per access, and this property is read while parsing every statement.
+        process_grids = self.sdfg.process_grids
+        result.update({v: process_grids[v] for k, v in self.variables.items() if v in process_grids})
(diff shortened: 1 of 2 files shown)


acalotoiu (approved):
LGTM

dace/frontend/python/newast.py
```diff
@@ -1372,11 +1372,11 @@ def defined(self):
...
-        # MPI-related stuff
-        result.update({
-            v: self.sdfg.process_grids[v]
-            for k, v in self.variables.items() if v in self.sdfg.process_grids
-        })
+        # MPI-related stuff. ``process_grids`` is a PROPERTY that rebuilds a dict by scanning every
+        # descriptor in the SDFG, so reading it inside the comprehension re-scanned once per variable:
+        # |variables| x |arrays| per access, and this property is read while parsing every statement.
+        process_grids = self.sdfg.process_grids
+        result.update({v: process_grids[v] for k, v in self.variables.items() if v in process_grids})
```
tbennun: @ThrudPrimrose or you could just loop over arrays and ask isinstance(distributeddescriptor) / processgrid

tbennun: `result.update({v: self.sdfg.arrays[v] for k, v in self.variables.items() if isinstance(self.sdfg.arrays[v], ProcessGrid)})`

tbennun: or, because all data descriptors are covered above (see line 1373), just do nothing, since it now includes process grids?

ThrudPrimrose: You are correct, we can even drop it (and become even faster).

PR #2502, opened 2026-08-17
LoopToMap Fixes

ThrudPrimrose:
LoopToMap crashes or refuses on loop shapes it can prove safe, and pays for the dependence proof even on loops it refuses for cheaper reasons. This collects those fixes with a test per behaviour, taking the corpus to 44 cases.

```python
# refused before; the read and the write never share an element
for i in range(2, 40, 2):
    A[i] = A[i - 1] * 2.0
```


acalotoiu (approved):
LGTM

PR #2505, opened 2026-08-17
SymbolPropagation: Fixing it Finally

ThrudPrimrose:
SymbolPropagation looked up a nested block's containers in the top-level SDFG, joined branches without a meet so one side's value escaped, refused read-only Scalars, rendered operator-backed functions as their sympy class names, and could return None after modifying the SDFG. This fixes those and adds the eight cases that cover them.

```python
# both arms bind fac; post-join uses folded to the literal 1.0 before
if cond: fac = 1.0
else:    fac = zfokoop[jl]
```

Also, adding symbol prop to simplify exposed some bugs:
- replace dicts on codeblock has issues
- pyinterop of abs has an issues on complex numbers (abs of complex type is a real type not complex)
- C++ "auto" emits different types than the symbol's dtype and produces different code than SDFG wants, so we need to ensure we collect the dtype of a symbol


acalotoiu (approved):
LGTM

acalotoiu (approved):

PR #2507, opened 2026-08-17
Fix autodiff symbol identity, subset parsing, and nested-SDFG determinism and UB in autodiff

ThrudPrimrose:
Subset-dimension strings in the data-forwarding store path are parsed with `symbolic.pystr_to_symbolic` instead of `sympy.sympify`, so `//` stays `int_floor` rather than folding to a rational that loses the loop index.
Symbolic tasklet differentiation now mints `dace.symbolic.symbol` rather than bare sympy symbols and differentiates against the instance already present in the expression, which previously returned a silent zero gradient.
Every `add_nested_sdfg` call site in `dace/autodiff` passes an explicit `symbol_mapping` resolved against the parent scope and ordered connector mappings instead of plain sets, so the missing-symbol validation can fire and connector order no longer depends on `PYTHONHASHSEED`.
An annotation naming a class that does not exist is corrected and `TYPE_CHECKING` imports are added so the `BackwardPassGenerator` forward references resolve.

For my experience missing symbols or using sympy bare symbols is a very common cause of non determinism due to sympy caches, and this might fix the non determinism in ML/autodiff.


dace/autodiff/implementations/dace_reduction_nodes.py
```diff
@@ -153,10 +154,27 @@ def _backward_reduction(forward_node: Node, context: BackwardContext, result: Ba
...
+            # counter -- the k-th reversed iteration then hands back grad/k instead of grad.
+            count_shape = out_desc.shape if len(out_desc.shape) > 0 else (1, )
+            zero_state.add_mapped_tasklet(
+                f"_zero_{type_name}_count_", {
+                    f"o{i}": f"0:{shape}"
+                    for i, shape in enumerate(count_shape)
+                }, {},
+                "__zero = 0",
+                {"__zero": Memlet.simple(count_arr_name, ",".join(f"o{i}" for i in range(len(count_shape))))},
+                external_edges=True)
```
affifboudaoud: One issue I've had with this zero initialization state technique is that when used in a NestedSDFG, it blocks it from being inlined and becomes a very annoying blocker for fusion later, so I tried to avoid it whenever possible by using setzero and being careful with allocation scopes.

dace/autodiff/implementations/dace_reduction_nodes.py
```diff
@@ -153,10 +154,27 @@ def _backward_reduction(forward_node: Node, context: BackwardContext, result: Ba
...
+            zero_state = sdfg.add_state(f"zero_{type_name}_count_{id(forward_node)}", is_start_block=True)
             count_state = sdfg.add_state(f"count_{type_name}_{id(forward_node)}")
             grad_state = sdfg.add_state(f"grad_{type_name}_{id(forward_node)}")
+            sdfg.add_edge(zero_state, count_state, dace.InterstateEdge())
             sdfg.add_edge(count_state, grad_state, dace.InterstateEdge())
 
+            # State 0: Clear the counter. It has to happen on every execution of this state, so it
+            # cannot be an AccessNode's ``setzero``: that zeroes the transient where it is ALLOCATED,
+            # i.e. once per SDFG invocation. Between calls that is enough, but under a loop the WCR
+            # below keeps accumulating across iterations, and the gradient tasklet divides by the
```
affifboudaoud: There is no single correct answer for the gradients here in the case where multiple values are the max. We divide by the counter to match how PyTorch returns the gradients.

affifboudaoud (commented):
Looks good in general. I would prefer seprate targerted PRs but this is still resonable.

dace/autodiff/implementations/dace_reduction_nodes.py
ThrudPrimrose: This code here should not create nested SDFG, but if future inline calls, fails it means the inline sdfg needs to be fixed.

ThrudPrimrose: Upon discussion I reverted back to setzero approach

affifboudaoud (approved):

PR #2508, opened 2026-08-18
Stop materializing the used-name set on every mint

ThrudPrimrose:
In HPC agentbench, parsing of Python SDFGs take a lot of time, this one PR to cache the things used when parsing.
I know new Python frontend is coming, but I still think we can have the old one faster until that one comes.

`_find_new_name` built the union of every array, constant and symbol name each time a name was minted, but `find_new_name` never iterates what it is handed -- it only asks whether a candidate is taken, so a membership view over the existing `is_name_used` answers it from three dict lookups. `find_new_name`'s annotation widens from `Sequence` to `Container` to say what it actually requires.

```python
def _find_new_name(self, name: str):
    return dt.find_new_name(name, _UsedNames(self))
```


acalotoiu (approved):
LGTM

PR #2511, opened 2026-08-20
Reduce warning frequency (Using sets as inputs is discouraged...)

edopao:
Reduce frequency of warning message by restricting to relevant case.


ThrudPrimrose (approved):

dace/sdfg/state.py
```diff
@@ -1716,8 +1716,8 @@ def add_tasklet(
         debuginfo = _get_debug_info(debuginfo or self._default_lineinfo)
 
         # Make dictionary of autodetect connector types from set
-        if isinstance(inputs, set) or isinstance(outputs, set):
```
philip-paul-mueller, line 1719: While I fully agree with the modification to the test, I think that the test itself (even before) makes the type annotation look wrong.
The annotation essentially tells you "use `set` or a `dict`", while the check tells you "do not use `set`s", which is kind of strange and indicates the the type annotation itself is wrong.
I think that the true type annotation should be `Iterator[Union[str, Tuple[str, dtypes.typeclass]]]`.
Then we can check if it is a `set` and issue an error, if it has length one (`Iterable` does not support `len()` so we would have to do it afterwards).

However, doing that would require a big rework, not only of this function but many more.
In addition I am not sure if PR is the right place to do it.

PR #2514, opened 2026-08-20
CI: uv migration

ThrudPrimrose:
Port pip + venv based implementation to faster modern implementation using uv.


.github/workflows/copilot-setup-steps.yml
```diff
@@ -28,13 +36,12 @@ jobs:
 
       - name: Install Python dependencies
         run: |
-          python -m pip install --upgrade pip
           # Install additional testing dependencies
-          python -m pip install pytest pytest-cov
-          python -m pip install flake8 pytest-xdist coverage
+          uv pip install pytest pytest-cov
+          uv pip install flake8 pytest-xdist coverage
```
tbennun: In my version of this PR I just added pytest-xdist to pyproject.toml and installing for `[testing]` is a one-liner. It's also faster

.github/workflows/general-ci.yml
```diff
@@ -37,8 +45,8 @@ jobs:
         # Install dependencies
         sudo apt-get update
         sudo apt-get install -y libyaml-dev cmake libblas-dev libopenblas-dev liblapacke-dev libpapi-dev papi-tools
-        pip install flake8 pytest-xdist coverage
-        pip install -e ".[testing]"
+        uv pip install flake8 pytest-xdist coverage
```
tbennun: same

.github/workflows/gpu-ci.yml
```diff
@@ -26,21 +26,26 @@ jobs:
...
+        enable-cache: true
+        pyproject-file: pyproject.toml
+        cache-dependency-glob: pyproject.toml
     - name: Install dependencies
       run: |
         rm -f ~/.dace.conf
-        rm -rf .dacecache tests/.dacecache
-        python -m venv ~/.venv      # create venv so we can use pip
+        rm -rf .dacecache* tests/.dacecache*
+        uv venv --clear ~/.venv      # create venv so we can use uv
```
tbennun: wouldn't it make things slower?
should we not just use uv sync or something?

.github/workflows/ml-ci.yml
```diff
@@ -34,9 +42,8 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y libyaml-dev cmake
         sudo apt-get install -y libblas-dev libopenblas-dev liblapacke-dev
-        python -m pip install --upgrade pip
-        pip install flake8 pytest-xdist coverage
-        pip install -e ".[ml-testing,ml]"
+        uv pip install flake8 pytest-xdist coverage
```
tbennun: same

tbennun (changes requested):

.github/workflows/gpu-ci.yml
ThrudPrimrose: uv sync requires as to commit uv.lock to the repo, but we can do that. Should we do that?

.github/workflows/copilot-setup-steps.yml
ThrudPrimrose: I did uv pip to not commit uv.lock to repo, I can use uv sync and no pip in CI to remove pip completely.

pyproject.toml
```diff
@@ -40,16 +40,21 @@ dependencies = [
...
     "absl-py",
     "opt_einsum",
     "pymlir",
     "click",
     "ipykernel",
     "nbconvert",
-    "pytest-timeout",
 ]
+mpi = ["mpi4py", "pytest-mpi"]
+gpu = ["cupy", "cutensor-cu12"]
```
tbennun: so as to not build from source, should we not provide something like `cupy-cuda13x`?

.github/workflows/gpu-ci.yml
tbennun: @ThrudPrimrose we did but still you run `uv venv --clear`. Why?

tbennun:
@ThrudPrimrose two comments remain. Thanks!

pyproject.toml
ThrudPrimrose: In a recent PR I fixed the isues with the cuda13 compatabilities, so we can now use cuda13x instead of 12x. I think adding these to pyproject.toml makes sense. I will perform the changes and regenerate uv.lock.

ThrudPrimrose: It makes sense to sue cuda13x in this case

.github/workflows/gpu-ci.yml
ThrudPrimrose: Looks like I just forgot to remove it, I will remove it and trigger the CI

.github/workflows/pyFV3-ci.yml
```diff
@@ -51,26 +57,32 @@ jobs:
...
 
     - name: Print versions
       run: |
         gcc --version
-        python --version
-        pip --version
-        pip list
+        uv run python --version
+        uv --version
+        uv pip list
```
romanc: While we're here, let's just delete the whole block. The log will contain this information in one way or another anyway.

pyproject.toml
```diff
@@ -40,16 +40,21 @@ dependencies = [
 ml = ["onnx", "torch", "onnxsim", "onnxscript", "onnxruntime", "protobuf", "ninja"]
 testing = [
```
romanc: If I may add a suggestion: `pyproject.toml` allows you to distinguish

- optional project dependencies that users might or might not want to install
- from optional project dependencies that only developers need (e.g. linting / testing / docs)

The former stay in `project.optional-dependencies` while the latter go into `dependency-groups`. Dependency groups can then be installed with `--group` (instead of `--extra`). As an example, in [NDSL](https://github.com/NOAA-GFDL/NDSL/blob/develop/pyproject.toml), I've three different dependency groups (linting tools, test dependencies, and things needed for docs).

By default, `uv sync` will install the `dev` group automatically (and in the above example, I've included all other dependency groups in there such that developers just can call `uv sync`  and be done with it). The nice thing about dependency groups is that you can also use `--only-group <group_name>` , which will install just that group and nothing else. In NDSL, I use that feature e.g. in [the linting workflow](https://github.com/NOAA-GFDL/NDSL/blob/develop/.github/workflows/lint.yaml) to just install the linting tools and nothing else.

Feel free to consider or discard. I'm also happy to make a follow-up PR with a concrete proposal how to restructure optional dependencies, if you like.

romanc (commented):
Very happy to see this PR 🥳

pyproject.toml
ThrudPrimrose: It makes sense, I think we can merge both together.
Maybe you can open a PR to directly merge to this one before we merge this branch to main?

romanc: https://github.com/spcl/dace/pull/2543

romanc: Thanks for merging and sorry for breaking the cscs-ci. Looks like I forgot to update `cscs_gpu.yml`. I've fixed that in https://github.com/spcl/dace/pull/2544, which avoids duplicating the testing dependencies in a testing extra and the testing group. I've ported over your minimal version requirements from the testing extra to the testing group.

There's no way for me to test this locally, so I'm afraid, the quickest way is to just merge https://github.com/spcl/dace/pull/2544 here and see if there's breakage again.

ThrudPrimrose: Your approach is better, so I have merged it.

tbennun (approved):
LGTM

PR #2518, opened 2026-08-23
Replace symbols inside a SymExpr bound

ThrudPrimrose:
Symbol replacement returns anything that is not a `sympy.Basic` unchanged, and a `SymExpr` — the (exact, over-approximation) pair strip mining produces — is exactly that, so a range whose end is strip-mined comes out half-renamed. Rewriting both halves fixes it.

```python
sdfg.apply_transformations(MapTiling, options=dict(tile_sizes=(32,)))   # inner: tile_i:Min(N - 1, tile_i + 31)
sdfg.replace("tile_i", "q")                                            # inner: q:Min(N - 1, tile_i + 31)
```

### DIFF
--- a/dace/sdfg/replace.py
+++ b/dace/sdfg/replace.py
@@ -22,6 +22,12 @@ tokenize_cpp = re.compile(r'\b\w+\b')
 
 
 def _internal_replace(sym, symrepl):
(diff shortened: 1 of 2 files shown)


acalotoiu (approved):
LGTM

PR #2519, opened 2026-08-23
[WIP] Derive a scope's symbol table once per scope

ThrudPrimrose:
`symbols_defined_at` walks every descriptor in the SDFG, and `propagate_memlet` derived it once per
memlet, so building a program cost O(scopes x descriptors) sympy work -- on a 2000-line generated
kernel, 8845 walks over 2082 descriptors, 81% of a 183 s parse. Callers propagating several memlets
through one scope node now derive it once, `Data.free_symbols` is memoized behind the property
setter that is the only way its sources can change, and the descriptor scan reads `s.name` instead
of printing the symbol.

```
tests/corpus/cloudsc, build_cloudsc_sdfg(simplify=False)   -- this branch vs main
  before  872.6 s      after  800.3 s      (6342 blocks, 2687 descriptors)

warpx_field_gather (hpcagent_bench, 2000 lines), to_sdfg(simplify=False)
  before  182.6 s      after   38.4 s      (7269 blocks, 2948 descriptors)
```

The second kernel is measured on the `extended` branch: it needs a data-dependent shape main's
frontend refuses, so main cannot parse it either way.

### DIFF
--- a/dace/data/core.py
+++ b/dace/data/core.py
@@ -129,6 +129,11 @@ class Data:
         """Returns a string for a Data-Centric Python function signature (e.g., `A: dace.int32[M]`). """
         raise NotImplementedError
 
+    #: Sources of ``free_symbols``, all tuples or sympy expressions: reassigning one is the only way
+    #: any can change, so that is what drops the memo. Structure overrides and computes fresh.
+    SYMBOL_SOURCE_ATTRIBUTES = frozenset({'_shape', '_strides', '_offset', '_total_size', '_transient', '_dtype'})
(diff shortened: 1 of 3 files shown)


ThrudPrimrose:
Holding this on the extended branch for now: the win is large on generated kernels and small on the cloudsc corpus, so it wants more evidence before landing on main.

ThrudPrimrose:
Reopened -- keeping it as a draft. The change is running on the extended branch meanwhile, which is where the generated-kernel numbers come from.

PR #2521, opened 2026-08-24
Keep both parts when promoting a real and a complex type

ThrudPrimrose:
`result_type_of` compared floating-point types by byte width alone, so complex64 tied with float64 and argument order settled it: one way the answer dropped the imaginary part, the other dropped half the real precision. Half a complex type's width is its imaginary part, so where exactly one side is complex the result is now complex and wide enough for the other side, matching numpy.

```
result_type_of(complex64, double)  before: double      after: complex128
result_type_of(double, complex64)  before: complex64   after: complex128
```


acalotoiu (approved):
lgtm

PR #2522, opened 2026-08-25
Sync a persistent kernel's stream reset against the pushes of its state

ThrudPrimrose:
A `GPU_Persistent` state resets its streams from every block and only synchronizes at the end of the state, so a block entering late rewinds the queue head under a block that already pushed and the consumer -- whose count is a separate atomic -- then indexes slots that were never written, which is the `CUDA_ERROR_ILLEGAL_ADDRESS` `tests/persistent_fusion_cudatest.py` hits on GH200 (264 blocks) and which poisons the CUDA context for every later GPU test on that xdist worker.

```c
__1_stream2.reset();
__gbar.Sync();          // new
for (int x = ...) { ... __1_stream2.push(neighbor); ... }
__gbar.Sync();
```


acalotoiu (approved):
lgtm

PR #2524, opened 2026-08-25
Keep the floor and ceiling of an expression that has no denominator

ThrudPrimrose:
`sympy_intdiv_fix` matched `floor(e/b)` with `b = 1` for any non-atomic argument, so `floor(sin(x))` was rewritten to `int_floor(sin(x), 1)` and printed as `((sin(x)) / (1))` -- the rounding silently dropped.


ThrudPrimrose:
Folded into #2525.

PR #2526, opened 2026-08-26
Use `dict` instead of `set` when adding a `NestedSDFG` in stree -> SDFG

romanc:
# Description

Inserting nested SDFGs within the stree -> SDFG translation layer was emitting the following warning:

```
Using sets as inputs is discouraged as it leads to indeterministic behavior.
```

We thus avoid the `set` and use a `dict` instead (akin to PR https://github.com/GridTools/gt4py/pull/2807 in gt4py).

Tested upstream (in `gt4py.cartesian`) and verified before/after that the warning appears/disappears.


ThrudPrimrose (approved):

acalotoiu (approved):
LGTM

romanc:
Thanks for approving and enabling auto-merge. Someone will have to kick off the gpu tests for me. I don't have permission to do so and they don't seem to run automatically on my PRs ...

tbennun:
cscs-ci run

edopao:
cscs-ci run

PR #2527, opened 2026-08-26
Add distribution of thread blocks to chiplets in old codegen

iomaganaris:
The goal of this PR is to distribute the thread blocks of the GPU backend to AMD GPU chiplets in such way that the X dimension of the grid is divided by the number of chiplets, so each one of them gets a continuous range of the domain. The domains that execute on the same chiplet can use then the same L2 cache, improving cache hits when there are data that can be reused between thread blocks in the same region or data that live in one dimension in two dimensional kernels (i.e. ICON neighbor tables and other vertical level independent fields).
This change has a ~7% performance improvement in the `icon4py` dycore.
We have already discussed about the necessity of this solution with @ThrudPrimrose I was interested to implement it in the old codegen though because we still haven't switched to the new one


ThrudPrimrose:
The assignment of thread-blocks to physical locations is a matter of codegen, yet right now the choice is too implicit. It is also not providing the user the option to choose between options. (This change would force all kernels to be generated either the old way or the new chiplet-aware way)

I think it would make more sense to have a map-level attribute (to also not break the existing assumption of "GPU_Device" is a GPU kernel and no a new schedule type to update everything)

I see that, through the tests the existence of tblock maps it should be fine.

For this pattern (and new codegen enforces it):
```
for i,j in dace.map[0:D0:16, 0:D1:16] @ GPU_Device:
  for bi, bj in dace.map[0:16, 0:16] @ GPU_ThreadBlock: 
     <body>
```

Codegen only changes the tblocks are assigned to chiplets so it seems fine to me.

ThrudPrimrose:
For test quality we should have a test where we check numerical equivalence between the two strategies (e.g. maybe a very small icon loopnest snippet written using the python frontend?). Also please add a test involving persistent schedule (so that we dont break it unknowingly)

ThrudPrimrose (commented):
All previous comments are addressed.

The codegen is opt-in and I like it. We also have enough numerical correctness checks.

Only one question? Would it make sense, and justify the effort to auto-detect chiplet count if a value like 0 is provided? 
We could have a simple dictionary between compute arch/model to chiplet count to make it simpler to run on multiple amd devises? What do you think?

iomaganaris:
> All previous comments are addressed.
> 
> The codegen is opt-in and I like it. We also have enough numerical correctness checks.
> 
> Only one question? Would it make sense, and justify the effort to auto-detect chiplet count if a value like 0 is provided? We could have a simple dictionary between compute arch/model to chiplet count to make it simpler to run on multiple amd devises? What do you think?

Thank you very much for the review.
I have some logic added [here in gt4py](https://github.com/GridTools/gt4py/pull/2832/changes) to detect the number of XCDs based on the `amdsmi` python package if it exists. I would be okay moving this logic to DaCe or use a look up table. Whatever you prefer

dace/sdfg/nodes.py
```diff
@@ -1068,6 +1062,13 @@ class Map(object):
 
     gpu_force_syncthreads = Property(dtype=bool, desc="Force a call to the __syncthreads for the map", default=False)
 
+    allow_chiplet_threadblock_distribution = Property(
+        dtype=bool,
+        default=True,
+        desc="Allow the thread-blocks of this kernel to be distributed over the chiplets of the GPU "
+        "(see the `compiler.cuda.chiplet_number` configuration entry)",
+        serialize_if=lambda m: m.schedule in dtypes.GPU_SCHEDULES)
```
tbennun: this should only be for GPU_Device and GPU_Persistent, right? not GPU_Threadblock[Dynamic]

dace/config_schema.yml
```diff
@@ -431,6 +413,19 @@ required:
                             Can be set to 'max' to maximize occupancy.
                         default: '32,1,1'
 
+                    chiplet_number:
+                        type: int
+                        title: Number of chiplets of the target GPU
+                        description: >
+                            Number of chiplets (XCDs) of the GPU the code is generated for. Set to 6 on MI300A.
```
tbennun:
```suggestion
                            Number of chiplets (XCDs) of the GPU the code is generated for. Set to 6 on MI300A,
                            or 8 on MI300X.
```

dace/codegen/targets/cuda.py
```diff
@@ -2281,6 +2306,22 @@ def get_kernel_dimensions(self, dfg_scope):
...
+                                                        extra_grid_dims)
+        if self._kernel_chiplet_count > 1:
+            # ``blockIdx.x`` carries the chiplet ID: the hardware dispatches thread-block ``f`` to
+            # chiplet ``f % gridDim.x`` in a round-robin fashion, and the flattened block index is
+            # ``blockIdx.x + blockIdx.y * gridDim.x + ...``, so making ``gridDim.x`` the number of
+            # chiplets pins each chiplet to one value of ``blockIdx.x``. Each chiplet then receives a
+            # contiguous chunk of ``ceil(grid_size[0] / chiplets)`` blocks of the first dimension,
+            # together with the full second dimension, which moves to ``blockIdx.z``.
+            original_grid_size = grid_size
+            grid_size = [self._kernel_chiplet_count, int_ceil(grid_size[0], self._kernel_chiplet_count), grid_size[1]]
```
tbennun: I am not sure why we should change the grid size; that means that the grid sizes are effectively minimized because blockdim x is the one that can be the largest of the three.
I'd consider a straightforward code generation approach (changing the blocks based on a modulo operator rather than changing the actual map dimensions.

tbennun (changes requested):
I am a big fan of the idea, but I am not sure we should change the map dimensions to provide that. It should be a purely generated-code approach as it changes the mapping of logical thread-blocks to physical thread-block numbers.

dace/codegen/targets/cuda.py
iomaganaris: I deliberately wanted to avoid the module operations and the thread ID translation because they add extra resource usage and complicates further the generated code. Unless there is actual need to support 3D maps I would keep this simpler approach

dace/sdfg/nodes.py
iomaganaris: If `GPU_Persistent` is enabled then based on the `chiplet_count` function nothing happens. I thought it would actually make sense probably to set this for the `GPU_Threadblock` schedule maybe?

dace/codegen/targets/cuda.py
```diff
@@ -2281,6 +2345,22 @@ def get_kernel_dimensions(self, dfg_scope):
...
+            warnings.warn(f'Distributing the grid of kernel "{kernelmap_entry.map.label}" over '
```
iomaganaris: This warning prints a lot of warnings like:
```
/capstor/scratch/cscs/ioannmag/cycle38/icon4py-benchmarks/venv_mi300/lib/python3.12/site-packages/dace/codegen/targets/cuda.py:2321: UserWarning: Distributing the grid of kernel "map_0_fieldop" over 6 chiplets, adjusting its size from [649, 1, 1] to [6, 109, 1].
```
when the option is enabled.
I am not really sure if this is beneficial or not

dace/codegen/targets/cuda.py
iomaganaris: Since I realized that different grid dimensions have different allowed ranges I have updated the logic to be codegen only with `%` and `/` usage

ThrudPrimrose (approved):

ThrudPrimrose:
@tbennun I think the latest stage is good can re-review?

tbennun:
cscs-ci run

tbennun:
Thanks, will review

tbennun:
@iomaganaris looks like the test is failing

iomaganaris:
cscs-ci run

iomaganaris:
@tbennun the test is now fixed. Could you please have another look?

tbennun (approved):

PR #2259, opened 2026-01-05
[Draft, for CI] New GPU Codegen Complete PR

ThrudPrimrose:
The complete PR for the NEW GPU codegen. (To be split into individual branches until everything has been merged).

PR #2528, opened 2026-08-26
[New GPU Codegen Prep] Const-qualify read-only nested-SDFG array arguments

ThrudPrimrose:
Emits an input array connector the callee never writes as `const T*`, deriving "never written" from `read_and_write_sets` and propagating non-const up a write-direction View chain. Split out of the new GPU codegen PR (#2259) as a prerequisite.


acalotoiu (approved):
LGTM

PR #2529, opened 2026-08-26
[New GPU Codegen Prep] An unused array must not leak its shape symbol into the signature

ThrudPrimrose:
Expands the extent symbols of only the arrays `read_and_write_sets` reports as used, so declaring an unused transient no longer changes `arglist()` or `init_signature()` (issue #2382). Split out of the new GPU codegen PR (#2259) as a prerequisite.


acalotoiu (approved):
LGTM

PR #2537, opened 2026-08-27
[To Discuss] Division and modulo semantics: % is C's, frontends name the others

ThrudPrimrose:
A bare `%` now means C's truncating modulo everywhere in an SDFG (tasklets, memlets, maps and inter-state edges; symbolic `CMod`), and frontends name the other semantics as `PyMod`, `FtnMod` and `FtnModulo`, which fold to `sympy.Mod` or `CMod` where the roundings agree. The Python frontend keeps Python's `//` and `%` by emitting `py_floor` and `PyMod`, the Fortran frontend emits `FtnMod`/`FtnModulo`, the runtime replaces `Mod`/`Mod_float`/`Modulo`/`Modulo_float`/`cpp_mod` with `py_mod`/`c_mod`/`ftn_mod`/`ftn_modulo`, and `doc/sdfg/ir.rst` documents the table.

### DIFF
--- a/dace/codegen/cppunparse.py
+++ b/dace/codegen/cppunparse.py
@@ -851,14 +851,24 @@ class CPPUnparser:
         "Sub": "-",
         "Mult": "*",
         "Div": "/",
-        "Mod": "%",
         "LShift": "<<",
         "RShift": ">>",
         "BitOr": "|",
         "BitXor": "^",
         "BitAnd": "&"
     }
-    funcops = {"FloorDiv": (" /", "dace::math::ifloor"), "MatMult": (",", "dace::gemm")}
+    # ``//`` and ``%`` are Python's, hence numpy's: the quotient rounds toward negative infinity and
+    # the remainder therefore takes the divisor's sign (``-32 // 7 == -5``, ``-32 % 7 == 3``). C
+    # rounds toward zero and gives the remainder the dividend's sign, so neither can be written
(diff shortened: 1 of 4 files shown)


ThrudPrimrose:
Two points for the discussion, neither of them fixed here.

**Division by zero still diverges from NumPy.** NumPy answers `0` for integers and `inf`/`nan` for floats; the emitted C traps or is undefined. Matching it puts a branch on every division, which did not seem like a trade to make silently.

**`//` and `%` are no longer infix.** They emit `py_floor(a, b)` / `py_mod(a, b)`, so anything asserting on emitted C text for these operators will see a different string.

The table in the new test is 64 operand pairs per dtype over `{int32, int64, float32, float64}` x `{host, device}`, compared against NumPy itself. The device half is not redundant with the host half: the floor correction is a branch, and that branch is what held the `std::div` call, so only running it on a GPU shows the branch survived compilation.

tbennun:
As per the discussion, let's add a few sympy functions:
* CMod
* FtnMod
* PyMod

have them lower to sympy.Mod in the cases that it can be evaluated at compile time

Also, modify the docs so that in dace, tasklet code is translated to C Modulo semantics, memlets and symbolic expressions in maps/ISedges/... follow SymPy Modulo semantics, and each respective frontend adds their own (PyMod, etc.).

@ThrudPrimrose

PR #2539, opened 2026-08-28
Compose the view subset affinely when removing a slice view

ThrudPrimrose:
`RemoveSliceView` built the composed subset from the inner memlet's offset and size, which discards its step and re-derives the end as `offset + size - 1`, so `image[0:N, 0:N][:, 0:2*H:2]` folded into `image[0:N, 0:H]` -- the first half of every row instead of its even columns. This composes the two ranges affinely instead, and adds a regression test that fails on main and passes here.


acalotoiu (approved):
LGTM

PR #2540, opened 2026-08-28
Do not treat an empty memlet as a read or a write

ThrudPrimrose:
`FindAccessNodes` classified access nodes by raw degree, so a node reached only by empty memlets -- ordering edges that transfer no data -- was reported as a write, and `ScalarFission` then renamed it twice: the edge was left naming a container that is neither of its endpoints, and the reads it dominated were versioned onto a container nothing ever writes. Classify by the memlet instead, and let the rename guard also accept a version already minted for the same container.


acalotoiu (approved):
lgtm

PR #2541, opened 2026-08-28
Fix connectors on stree

ThrudPrimrose:
A small bugfix to not forget adding a connector to sdfg on certain read cases.


tbennun (approved):

romanc:
Thanks for finding and fixing! :pray:

PR #2548, opened 2026-09-01
stree -> sdfg: add missing nested SDFG connectors

romanc:
## Description

Follow-up from PR https://github.com/spcl/dace/pull/2541: there's one more case of "read after write" where we need to add a connector to the nested SDFG to retain the connection to "outside" memory and not confuse simplify.

Founds as part of our work to port the FV3 dynamical core.

image

This is DelnFlux_2, last state. Look at `__g_self__fy2`: it's read after written and thus no input connector for `__g_self__fy2`  is added. That's not good because `__g_self__fy2` is only partially written here. Also, since there's no input connector and since the output connector will be (rightfully) pruned, `__g_self__fy2`  will be "inlined" by `DeadDataflowElimination`, which then breaks stuff down the line (e.g. latest in `D_SW`).

The PR adds a simpler reproducer (in tests) and ensures as part of the unit tests that this won't regress.

### DIFF
--- a/dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
+++ b/dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
@@ -207,7 +207,6 @@ class _StreeToSDFG(tn.ScheduleNodeVisitor):
                         use_nview = self._apply_nview_array_override(memlet.data, sdfg)
                         if not use_nview:
                             sdfg.add_datadesc(memlet.data, parent_sdfg.arrays[memlet.data].clone())
-
                             # Transients passed into a nested SDFG become non-transient inside that nested SDFG
                             if parent_sdfg.arrays[memlet.data].transient:
                                 sdfg.arrays[memlet.data].transient = False
@@ -288,7 +287,6 @@ class _StreeToSDFG(tn.ScheduleNodeVisitor):
                     use_nview = self._apply_nview_array_override(memlet.data, sdfg)
                     if not use_nview:
                         sdfg.add_datadesc(memlet.data, parent_sdfg.arrays[memlet.data].clone())
-
                         # Transients passed into a nested SDFG become non-transient inside that nested SDFG
                         if parent_sdfg.arrays[memlet.data].transient:
                             sdfg.arrays[memlet.data].transient = False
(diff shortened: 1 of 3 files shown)


romanc:
Actually, this doesn't look quite right yet ...

Notes for tomorrow:  From what I understand, the problem is "read after partial write", i.e. 

image

This is DelnFlux_2, last state. Look at `__g_self__fy2`: it's read after written and thus no input connector for `__g_self__fy2`  is added. That's not good because `__g_self__fy2` is only partially written here. Also, since there's no input connector and since the output connector will be (rightfully) pruned, `__g_self__fy2`  will be "inlined" by `DeadDataflowElimination`, which then breaks stuff down the line (e.g. latest in `D_SW`).

romanc:
ML tests fail with network issue

```none
E   urllib.error.HTTPError: HTTP Error 404: Not Found
```

can someone re-run them? (I don't have rights to do so ...)

romanc:
I don't understand how this can/should break ML tests. I'm hitting a 404 again. Seems like something isn't online (anymore)?!

ThrudPrimrose:
> I don't understand how this can/should break ML tests. I'm hitting a 404 again. Seems like something isn't online (anymore)?!

Downloading models is failing, cloudflare must be down or something
Let's try again to merge tomorrow.

ThrudPrimrose (approved):
Core fix looks good. Semantically, might be better to split the bugfix and syntactic-cleaning, regardless all good.

ThrudPrimrose:
Need to wait until cloudflare or whatever webservice that is down is back

romanc:
> Core fix looks good. Semantically, might be better to split the bugfix and syntactic-cleaning, regardless all good.

I can split stuff if you want me to. Since we anyway have to wait and everything is neatly in commits it's gonna be very easy and we won't loose time.

romanc:
> > I don't understand how this can/should break ML tests. I'm hitting a 404 again. Seems like something isn't online (anymore)?!
> 
> Downloading models is failing, cloudflare must be down or something Let's try again to merge tomorrow.

Do these models change often? If not, could we cache them or are they way too big for GHA caches?

romanc:
> Core fix looks good. Semantically, might be better to split the bugfix and syntactic-cleaning, regardless all good.

https://github.com/spcl/dace/pull/2554

romanc:
I've split out the unrelated changes. Someone with access rights will need to kick off the GPU tests ...

edopao:
cscs-ci run GPU

romanc:
back to draft state: there's more to do here ...

romanc:
This is ready for review again. I've added more book keeping to check if we are (in case of read after write) reading something that "comes from outside", i.e. has previously been access outside the current nested SDFG that we are building. If so, we need to add the input connector to make the connection.

The current fix (i.e. https://github.com/spcl/dace/pull/2541), would always add an input connector upon read. Doing so will add a connector for e.g. scalars that are really only used inside the nested SDFG. Doing so will

a. trigger sdfg validation warnings of reading unused temporaries
b. raise the scope of everything that's ever read inside a nested SDFG all the way up to the SDFG root

acalotoiu (approved):
LGTM

edopao:
cscs-ci run GPU

PR #2554, opened 2026-09-04
simplify: rename collection of recursive passes

romanc:
## Description

This PR renames `_nonrecursive_passes` to `_recursive_passes`  in `dace/transformation/passes/simplify.py` because those are the passes that run recursively.


romanc:
whoever has access rights: just restart the failing NASA test. test data download failed. we should really cache that data: it never changes and those intermittent network issues are annoying ...

ThrudPrimrose (approved):

tbennun:
cscs-ci run

PR #2556, opened 2026-09-06
Treat states after return, break, and continue as dead

### LINKED ISSUE #2051: Make DeadStateElimination eliminate after return / break / continue blocks
The `DeadStateElimination` [pass](https://github.com/spcl/dace/blob/main/dace/transformation/passes/dead_state_elimination.py) removes states that are definitely not going to be executed in SDFGs, similarly to Dead Code Elimination in control-centric compilers such as LLVM. The control flow blocks [ReturnBlock](https://github.com/spcl/dace/blob/main/dace/sdfg/state.py#L2553), [BreakBlock](https://github.com/spcl/dace/blob/main/dace/sdfg/state.py#L2539), and [ContinueBlock](https://github.com/spcl/dace/blob/main/dace/sdfg/state.py#L2525), under some conditions, will generate "dead states" after them. We would like to modify `DeadStateElimination` to consider them as dead states as well.

The PR's implementation should include amendments to the pass as well as tests for the behavior.

For more information, see the [SDFG IR documentation](https://spcldace.readthedocs.io/en/latest/sdfg/ir.html).

mohitt31:
DeadStateElimination walks the CFG from the start block and follows the edges out of ReturnBlock, BreakBlock and ContinueBlock, so whatever sits after one of them still counts as reachable. Control flow does not continue past those blocks, so those states are dead and should go.

The fix is to stop the traversal when it reaches one of the three. Anything reachable only through them then falls out of the visited set and gets removed the same way as any other unreachable block.

Tests cover a return at the top level, and a break and a continue inside a loop region. There is also a test where a state is reachable both through the return and around it, so the pass has to keep it. That one passes before the change as well, it is there to catch the traversal being cut too early.

Closes #2051.

### DIFF
--- a/dace/transformation/passes/dead_state_elimination.py
+++ b/dace/transformation/passes/dead_state_elimination.py
@@ -7,7 +7,8 @@ from typing import List, Optional, Set, Tuple, Union
 from dace import SDFG, InterstateEdge, SDFGState, symbolic, properties
 from dace.properties import CodeBlock
 from dace.sdfg.graph import Edge
-from dace.sdfg.state import ConditionalBlock, ControlFlowBlock, ControlFlowRegion
+from dace.sdfg.state import (BreakBlock, ConditionalBlock, ContinueBlock, ControlFlowBlock, ControlFlowRegion,
+                             ReturnBlock)
 from dace.sdfg.validation import InvalidSDFGInterstateEdgeError, InvalidSDFGNodeError
 from dace.transformation import pass_pipeline as ppl, transformation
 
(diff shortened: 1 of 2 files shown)


ThrudPrimrose (approved):

mohitt31:
This is approved and up to date with main, but no CI has ever run on it, the checks tab is empty. I think workflow runs still need approving for a first contribution here. Let me know if you want anything changed in the meantime.

acalotoiu (approved):
LGTM

tbennun:
cscs-ci run

PR #2557, opened 2026-09-07
AI-generated implementations for DaCe library nodes

tbennun:
This PR provides two elements:

1. A reserved `'ai'` implementation, available on every library node. Expanding with it asks a language model to write the replacement tasklet, given the node's context in the SDFG and on the local machine.
2. `dace.libraries.ai.nodes.ai_node.AINode`, a library node that carries only a natural-language description, for microkernels that have no library node of their own.
3. `ai.refine` allows AI sessions to refine the implementation (e.g., if it crashes or not fast enough) 

The model is reached through a provider selected by the `ai.provider` configuration entry. It supports the Anthropic Messages API, the OpenAI Responses API, and arbitrary AI models through a `manual` mode that lets you paste prompts and responses.


dace/libraries/ai/nodes/ai_node.py
```diff
@@ -0,0 +1,70 @@
...
+class AINode(nodes.LibraryNode):
+    """
+    A library node that carries a natural-language description instead of an implementation.
+
+    Every DaCe library node can be expanded with the reserved ``'ai'`` implementation, which asks a
+    language model to write the tasklet. This node exists for the case where there is no library
+    node to begin with -- a microkernel DaCe does not expose, a vendor intrinsic sequence, a call
+    into an external library -- so the description *is* the specification::
+
+        node = AINode('fma', 'Compute out = a * b + c elementwise over the whole tile.',
```
ThrudPrimrose, line 23: I think the pure/naive lowering of the Library can be provided to describe the semantics on a more constrained way?

tbennun: the pure/native lowering can be arbitrarily complex (with e.g., multiple maps in reduce) and so it might require a lot of prompt real estate and to actually use the dace codegen (or print out e.g. a schedule tree?)

Not sure. Potential future work?

dace/libraries/ai/nodes/ai_node.py
```diff
@@ -0,0 +1,70 @@
...
+
+    implementations = {}  # 'ai' is a reserved name, resolved by LibraryNode.expand
+    default_implementation = 'ai'
+
+    description = properties.Property(dtype=str,
+                                      default='',
+                                      desc='Natural-language description of what this node must compute. This is '
+                                      'the specification handed to the model, so state the intended semantics, '
+                                      'the expected numerical behavior, and any implementation technique that is '
+                                      'required (e.g. "use AVX2 intrinsics").')
```
ThrudPrimrose, line 40: I think DaCe could (or should?) provide a default pre- or post-amble for the description that DaCe requires input language in C++, CUDA, HIP or Python for library node implementations.

tbennun: Yes, already part of the prompt (see prompt.py)

dace/libraries/ai/nodes/ai_node.py
ThrudPrimrose: Makes sense, also to consider a library node implementation valid we should use symbolic shapes to create fuzzed correctness tests.

I think this should be a future work, I'm doing this for the ICLR submission, we can expand this after ICLR submission. Having measures against reward-hacking and vagueness is critical to avoid token wastage.

dace/libraries/ai/nodes/ai_tasklet.py
```diff
@@ -0,0 +1,72 @@
...
+    the code generator emits it identically (see the ``_generate_AITasklet`` alias in
+    :mod:`dace.codegen.targets.cpu`) -- and differs only in remembering its origin.
+    """
+
+    provenance = Property(dtype=str,
+                          default='',
+                          desc='JSON recording the session, the round, and the serialized library '
+                          'node this tasklet was generated from. Empty if unknown.')
+
+    def to_json(self, parent):
```
ThrudPrimrose, line 37: Wouldn't be simpler to load AITasklet/AINode together with tasklet to not need this?

dace/libraries/ai/providers/responses_provider.py
```diff
@@ -0,0 +1,72 @@
...
+
+import json
+from typing import Dict, List
+
+from dace.config import Config
+from dace.libraries.ai import backend
+from dace.libraries.ai.exceptions import AIExpansionError
+
+
+class ResponsesProvider:
```
ThrudPrimrose, line 12: If the "responses" API is refering to the OpenAI package -  then maybe we should name it OpeAI and Antrophic Provider respectively?

dace/libraries/ai/providers/manual_provider.py
```diff
@@ -0,0 +1,176 @@
...
+
+    {response_path}
+
+A ```json fenced block is fine; the fences are stripped. A saved answer is reused on later runs,
+so re-running this program will not ask again unless the prompt itself changes.
+================================================================================
+"""
+
+
+class ManualProvider:
```
ThrudPrimrose, line 47: I dont have anything against the implementation's code.
But I think an interactive environment to paste code is going to be useful. Considering the implementation can get long. This is my personal opinion, but I would definitely not use it.
I think it would much more convenient if use DaCe to generate the prompt into a .md file, and then provide an API to read the response impl?

ThrudPrimrose: It would also make the implementation simpler to have:

```
ai_node.generate_prompt() -> str
ai_node.register_impl(str)
```

pyproject.toml
```diff
@@ -79,6 +79,8 @@ dependencies = [
 ml = ["onnx", "torch", "onnxsim", "onnxscript", "onnxruntime", "protobuf", "ninja"]
 mpi = ["mpi4py", "pytest-mpi"]
 gpu = ["cupy-cuda13x", "cutensor-cu13"]
+ai-anthropic = ["anthropic>=1.0"]
```
ThrudPrimrose, line 82: For simplicity we can enforce installing both anthropic and open ai when ai is requested?

dace/libraries/ai/nodes/ai_tasklet.py
ThrudPrimrose: Then we dont need serialization specialization, just need to add the provenance field.

dace/libraries/ai/providers/manual_provider.py
ThrudPrimrose: We can also provide API at the SDFG level, e.g.
`sdfg.generate_ai_node_prompts()`, or `sdfg.generate_ai_prompt_folders()`
`sdfg.collect_implementations(List[str] | Path)`

ThrudPrimrose: I think following such an approach would allow trimming a lot of stuff like needing regexes below etc?

dace/libraries/ai/backend.py
```diff
@@ -0,0 +1,344 @@
...
+
+import os
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Protocol
+
+from dace.config import Config
+from dace.libraries.ai.exceptions import AIExpansionError
+
+#: Environment variable names used when ``ai.api_key_envvar`` is left at its schema default.
+DEFAULT_KEY_ENVVARS = {'anthropic': 'ANTHROPIC_API_KEY', 'responses': 'OPENAI_API_KEY'}
```
ThrudPrimrose, line 19: I think responses name is confusing seeing this again, we should definitely name it `openai` imho

dace/libraries/ai/context.py
```diff
@@ -0,0 +1,811 @@
...
+
+from dace import data as dt
+from dace import dtypes
+from dace.config import Config
+from dace.sdfg import nodes
+from dace.sdfg import SDFG, SDFGState
+from dace.sdfg.scope import devicelevel_block_size, get_node_schedule, is_devicelevel_gpu
+
+#: Storage types that live in GPU memory and are therefore unreachable from host code.
+DEVICE_ONLY_STORAGE = (dtypes.StorageType.GPU_Global, dtypes.StorageType.GPU_Shared)
```
ThrudPrimrose, line 32: I think GPU_Storages being GPU_Shared is bad.
I think we need to provide good new names that cover these locations to not repeated these 3 declarations again and again?

dace/libraries/ai/context.py
```diff
@@ -0,0 +1,811 @@
...
+    :param state: The state containing the node.
+    :param sdfg: The SDFG containing the state.
+    :return: The complete expansion context.
+    """
+    # Computed once and shared: it deep-copies the SDFG, and every part of the context that
+    # mentions a schedule or a storage type has to agree with it.
+    defaults = infer_defaults(sdfg)
+    connectors = collect_connectors(node, state, sdfg, defaults)
+    capabilities = collect_capabilities(node, state, sdfg, connectors, defaults)
+    nesting = collect_nesting(node, state, sdfg, defaults)
```
ThrudPrimrose, line 784: Is the nesting really necessary?
Shouldn't we just need `is_devicelevel_gpu` + `schedule` of the ai tasklet?

dace/libraries/ai/context.py
```diff
@@ -0,0 +1,811 @@
...
+def collect_context(node: nodes.LibraryNode, state: SDFGState, sdfg: SDFG) -> ExpansionContext:
+    """
+    Gathers everything the model needs in order to write this tasklet.
+
+    :param node: The library node being expanded.
+    :param state: The state containing the node.
+    :param sdfg: The SDFG containing the state.
+    :return: The complete expansion context.
+    """
+    # Computed once and shared: it deep-copies the SDFG, and every part of the context that
```
ThrudPrimrose, line 779: Same, do we really need the full SDFG? Isn't the concept of connector+tasklet is to have information needed for a tasklet locally and derive it just from the input output data?

dace/libraries/ai/nodes/ai_tasklet.py
tbennun: I don't think I understand this question. AINode is a library node with descriptions. AITasklet is a Tasklet expanded from the AINode during codegen. I couldn't find a simpler way to deal with build errors, which would need the expanded version to run through codegen and compilation. We should discuss

dace/libraries/ai/providers/responses_provider.py
tbennun: Good point. The code used to be Messages API and Responses API but I changed to Anthropic and Responses. This is because OpenAI has the chat API and other non-openai models support the Responses API.

dace/libraries/ai/providers/manual_provider.py
tbennun: That's a cool idea. I like the "paste it here" approach (which also accepts a json file) but the regexp can be reduced to `strip()`.

pyproject.toml
tbennun: sure, I wanted to not call the dependency `ai` though to not confuse users with DaCeML. Thoughts?

dace/libraries/ai/context.py
tbennun: hmm we can also reuse the dtypes code `is_accessible` for your current schedule type and the given storage type.

dace/libraries/ai/context.py
tbennun: I just wanted a full list of all the functions and control flow (CATS-style) so that the model knows that, e.g., it is in a time loop.

dace/libraries/ai/context.py
tbennun: The deep copy can be removed in principle. However, we will need to refactor some code to do a read-only/immutable connector type inference. It is the same code that sets the default schedule and storage types (infer_types.py) when a library node is expanded.

PR #2558, opened 2026-09-07
fix deprecation warnings in schedule tree tests

romanc:
## Description

Using a `set` as data type for inputs / outputs of `NestedSDFG`s or `Tasklet`s is deprecated because sets are not ordered and thus generated SDFGs might not be deterministic.

Replacing all inputs/outputs with `Set`  type is an ongoing effort. This PR updates what is needed to get no warnings when running all tests under `tests/schedule_tree/`.


ThrudPrimrose (approved):

tbennun:
cscs-ci run

PR #2559, opened 2026-09-07
Remove RTLTasklet class from nodes.py

tbennun:
Removed the RTLTasklet class, which is a leftover of the FPGA codegen migration.


acalotoiu (approved):
LGTM

PR #2561, opened 2026-09-07
Answer a cross-boundary copy with a Memcpy, or refuse it

ThrudPrimrose:
`CopyLibraryNode`'s Auto selection returned `MappedTasklet` for every in-kernel multi-element copy without looking at the storages, and again as the host-level fallback for anything the CUDA branch's storage set did not cover, so the expansion then rejected the copy as its own limitation and pointed at `MemcpyCUDA1D` — which device code cannot issue either. Inside a kernel a cross-boundary copy has no implementation at all and is now refused at selection with a message naming the kernel, while at host level it always resolves to a Memcpy, including for a `Register` endpoint that sits outside the storage set the branch tested.


acalotoiu (approved):
LGTM

PR #2562, opened 2026-09-08
Scale a strided tile's over-approximation by the map step

ThrudPrimrose:
`StripMining._create_strided_range` over-approximated a non-divides-evenly tile as `tile_size` indices wide while its exact bound spans `tile_size * td_step`, so `Range.size()` reported a strided map's tile as `tile_size / td_step` iterations -- tiling a step-2 map by 128 produced a thread-block map that `InferGPUGridAndBlockSize` measured at 64 threads and rejected against the kernel's declared `gpu_block_size`. The approximation now carries the step, matching the exact bound and the `divides_evenly` branch one line above, and is identical for `td_step == 1` so no existing caller changes.


acalotoiu (approved):
LGTM

PR #2565, opened 2026-09-09
fix: rescale nested SDFG subsets by the outer memlet step

ThrudPrimrose:
A map body carried into a nested SDFG got a local descriptor sized from the number of elements of the outer subset, while the inner memlet was only shifted by the outer start. For a strided outer subset such as A[15*i:15*i+28:3] this left an inner range of 0:28:3 against a shape-(10,) array, and parsing failed with "Memlet subset out-of-bounds".


acalotoiu (approved):
LGTM

PR #2566, opened 2026-09-09
Array elimination of scalars

romanc:
## Description

`ArrayElimination` removes unused descriptors, but for some reason, `data.Scalar`  are excluded. It is not mentioned why and with `FindAccessStates` tracking all data descriptors in `sdfg.arrays`, I see no reason not to also act on `data.Scalar`.


romanc:
Failing tests raised a discussion on whether a memlet between two tasklets should be able to carry data (and thus assign the "output" of the first tasklet to that data descriptor while passing it as input to the second tasklet.

To be discussed in the DaCe meeting on the 17th of September.

**Example**
The value of `lookup` is saved in `sdfg.arrays["__bla_value"]` without an explicit access node for `__bla_value`.

image

PR #2567, opened 2026-09-10
New parser test: no transient argument to dace.program

romanc:
## Description

Follow-up from #2565: no "transient" argument to `@dace.program` because

1. it doesn't make any sense
2. that was not the intention of what should be tested.


ThrudPrimrose (approved):
Thanks for fixing this! LGTM

tbennun:
cscs-ci run

PR #2568, opened 2026-09-10
Avoid GHA warnings: unexpected inputs(s) 'pyproject-file'

romanc:
The action `astral-ash/setup-uv@v7` is complaining about an unexpected input `pyproject-file`, which is true: according to their documentation, there's no such configuration option. I thus removed that option from all the places that is was used.

image

I've also removed the `cache-dependency-glob`  configuration because the default includes not only `pyproject.toml`, but also `uv.lock`, see [1] for reference. Since the `uv.lock` file can change while the `pyproject.toml` remains untouched, that glob is better because a changed `uv.lock` file will match exatly what versions need to be installed.

[1]: https://github.com/astral-sh/setup-uv/tree/v7/#inputs

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -17,8 +17,6 @@ jobs:
         with:
           python-version: '3.11'
           enable-cache: true
-          pyproject-file: pyproject.toml
-          cache-dependency-glob: pyproject.toml
 
       - name: Set up C++ compiler
         run: |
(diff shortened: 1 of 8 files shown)


romanc:
not sure what `pre-commit`  is complaining out. `yapf` is running on python files and no python files have been changed ... works okay locally. maybe just someone with permissions re-run the check.

ThrudPrimrose (approved):

tbennun:
cscs-ci run

PR #2569, opened 2026-09-11
fix: update `get_view_node()` to follow the memlet path

romanc:
Fixes `get_view_node()` on a view that is connected to another access node via a map entry/exit node by following the memlet path.

Previous implementation was failing in case of e.g. `AccessNode -> MapEntry -> View`, where `get_view_node()` would wrongly return the `MapEntry` instead of the viewed `AccessNode`.

image


romanc:
@ThrudPrimrose this is now ready for review (including the promised test case).

@tbennun FYI found this while plying with views

ThrudPrimrose (approved):

tbennun:
cscs-ci run

tbennun:
I have a similar fix

PR #2570, opened 2026-09-11
Refuse the overapproximation matches on a structural test first

ThrudPrimrose:
A small sympy performance improvement.
Checking for existence makes the sympy operations faster when locally tested.


acalotoiu (approved):
LGTM

PR #2573, opened 2026-09-12
Say that all_nodes_between discards its whole walk on a sink

ThrudPrimrose:
`all_nodes_between` documented a `None` return for a path that does not reach `end`, but it has always returned an empty set and the discard is total. One node with no out-edges throws away every node already collected, so a predicate written over the result reports "nothing found" without inspecting anything. The docstring now states that, points callers who want a scope's nodes at `scope_subgraph`, and four tests pin the behaviour including that an empty walk is indistinguishable from an empty scope.

### DIFF
--- a/tests/sdfg/all_nodes_between_sink_test.py
+++ b/tests/sdfg/all_nodes_between_sink_test.py
@@ -1,0 +1,71 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
+"""``all_nodes_between`` discards its whole walk when the region holds a sink.
+
+The docstring now says so, and these pin it. The behaviour is easy to read as "omit the dead-ending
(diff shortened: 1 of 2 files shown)


tests/sdfg/all_nodes_between_sink_test.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2019-2026 ETH Zurich and the DaCe authors. All rights reserved.
+"""``all_nodes_between`` discards its whole walk when the region holds a sink.
+
+The docstring now says so, and these pin it. The behaviour is easy to read as "omit the dead-ending
```
tbennun:
```suggestion
The docstring now says so, and these pin it. The behavior is easy to read as "omit the dead-ending
```

tbennun (approved):

PR #2574, opened 2026-09-12
Keep tile sizes and mapped symbol types in replace_dict

ThrudPrimrose:
Replacing any symbol rebuilt every map range from `(start, end, step)` only and re-parsed every nested SDFG `symbol_mapping` value from its string, which reset tile sizes to 1 and re-minted mapped symbols with the default dtype and no assumptions (e.g. `P: N` with `N` int64 nonnegative became a plain int32 `N` after `replace_dict({"M": "K"})`). `InterstateEdge.replace_dict` also compared a string with an AST and re-spelled every condition and assignment, dropping the parsed condition, and all three paths now leave values that name no key untouched.


tbennun (approved):
Great catch!

PR #2576, opened 2026-09-13
Define enclosing loop iterators for nested SDFGs and symbols_defined_at

ThrudPrimrose:
`SDFGState.defined_symbols` missed enclosing loop iterators, `symbols_defined_at` saw none, and `add_nested_sdfg` typed nested symbols from `sdfg.symbols` alone, so a nested SDFG in an inner loop body raised `Missing symbols` for the inner iterator; the new test in `tests/sdfg/loop_region_test.py` fails on main. `defined_symbols` and `symbols_defined_at` now add the iterators of the enclosing regions (map symbols stay node-scoped), `add_nested_sdfg` builds on `defined_symbols`, and `LoopRegion.new_symbols` memoizes the inferred iterator type on the header text and the types of the names the header reads, so codegen stays fast.


tbennun:
cscs-ci run

dace/sdfg/state.py
```diff
@@ -1826,9 +1847,7 @@ def add_nested_sdfg(
             # Add new global symbols to nested SDFG
             for sym, symval in s.symbol_mapping.items():
                 if sym not in sdfg.symbols:
-                    # TODO: Think of a better way to avoid calling
```
tbennun, line 1829: This was a performance related comment. Fine with removing it though

tbennun (approved):

PR #2577, opened 2026-09-13
[pre-NSDFG-fixes] Rename the nested SDFG symbol mapping key in OTFMapFusion's advanced_replace

tbennun:
advanced_replace renames a symbol inside nested SDFGs but left the enclosing node's symbol mapping keyed by the old name, so the renamed symbol was no longer passed in and the SDFG failed validation.


ThrudPrimrose (approved):
LGTM (I actually have this fix on my working branch as well)

PR #2578, opened 2026-09-13
[pre-NSDFG-fixes] Declare the grid barrier for device maps in nested SDFGs within a kernel

tbennun:
A kernel declares its grid barrier only when a device map in its own scope needs to synchronize, so consecutive device maps inside a nested SDFG in the kernel emitted __gbar.Sync() on an undeclared barrier. Check nested SDFGs too, using the same predicate the scope generator uses to emit the synchronization.


tests/codegen/gpu_nested_grid_barrier_test.py
```diff
@@ -0,0 +1,92 @@
...
+                              memlet=dace.Memlet('B[i, j]'))
+    sdfg.validate()
+    return sdfg
+
+
+def _check_barrier(nested: bool):
+    sdfg = _kernel_with_consecutive_maps(nested)
+    code = next(obj.clean_code for obj in sdfg.generate_code() if obj.language == 'cu')
+    assert '__gbar.Sync();' in code
+    assert re.search(r'cub::GridBarrier\s+__gbar|GridBarrier\s*&\s*__gbar', code)
```
ThrudPrimrose, line 79: Is `cub::GridBarrier` check even necessary when we have `GridBarrier`, The `cub` name becomes `hipcub` but ti should work on AMD too, so I guess it is ok?

dace/codegen/targets/cuda.py
```diff
@@ -1705,6 +1700,10 @@ def generate_scope(self, sdfg: SDFG, cfg: ControlFlowRegion, dfg_scope: StateSub
                 # Create grid barrier only if there is a synchronization requirement on nested GPU_Device maps
                 if any(p is not scope_entry for p in dfg_scope.predecessors(node)):
                     create_grid_barrier = True
+            elif isinstance(node, nodes.NestedSDFG) and node.sdfg is not None:
+                # Device maps in nested SDFGs within the kernel may also synchronize the grid
```
ThrudPrimrose, line 1704: Maybe my understand of the nested device maps is wrong.
From my understanding what happens is this:

```
for i in 0..N <at> GPU_Device
  for j in 0..N <at> GPU_Device
      <body1>
  for j in 0..N <at> GPU_Device
      <body2>
```

is actually emitted as:
```
for i,j in (0..N), (0..N) <at> GPU_Device
    if range1: <body1>
    grid_sync
    if range2: <body2>
```
And nested SDFG follows this.

Btw as a future todo for the new GPU codegen I have a pass that does this at codegen level.
I should merge this later so that we can simplify the codegen later. (It collapses dimensions, and then each pervious kernels body is put inside an if + grid sync state between the ifs)

ThrudPrimrose: Maybe we should make it an issue to not forget?

ThrudPrimrose (approved):

tests/codegen/gpu_nested_grid_barrier_test.py
tbennun: Yes, both platforms should work with this code. I never tested hipcub though

dace/codegen/targets/cuda.py
tbennun: The existence of nested sdfg in between the two device maps should not affect generated code. That’s part of the point of the PR.

We did have some recent discussion on the role of nested sdfgs for synchronization, but that should not affect this PR. Do you concur?

ThrudPrimrose: That is true, that case was only relevant if the node has a parent conditional region, in this case both variants generate the same code and thus correct.

PR #2580, opened 2026-09-13
[pre-NSDFG-fixes] Fix DoubleBuffering loop variable being used registered outside loop

tbennun:
Makes post-loop code invalid because the symbol is not properly defined outside


tbennun:
cscs-ci run

ThrudPrimrose (approved):
LGTM

PR #2581, opened 2026-09-13
[pre-NSDFG-fixes] Format node and edge validation errors raised on non-state blocks

tbennun:
InvalidSDFGNodeError and InvalidSDFGEdgeError index the nodes and edges of the block their id names, which fails for a control flow block that is not a state (e.g., a conditional block) and hides the actual validation message.


tbennun:
cscs-ci run

acalotoiu (approved):
LGTM

PR #2582, opened 2026-09-13
[pre-NSDFG-fixes] Detect write conflicts of WCR writes through views

tbennun:
A WCR write into a view was treated as a write into a transient container, so no atomic operation was emitted even though the view aliases a container shared across parallel iterations. Follow views to the viewed container before checking for conflicts.


tbennun:
cscs-ci run

ThrudPrimrose (approved):

PR #2583, opened 2026-09-13
[pre-NSDFG-fixes] Replace symbols in list properties and keep name-typed slots strings

tbennun:
SDFG.replace did not rename list properties such as Map.params, which is why MultiExpansion renamed map parameters by hand. Replace list elements based on the declared element type: whole identifiers in string lists, symbolic substitution in numeric and symbolic lists, and nothing otherwise (booleans, code fragments, objects). MultiExpansion's manual renames are now redundant.

A replacement may also be given as a symbolic expression, which broke dictionary keys (e.g., interstate edge assignments) and LoopRegion's string loop variable. Convert those back to strings.


tbennun:
cscs-ci run

dace/sdfg/state.py
```diff
@@ -3787,7 +3787,7 @@ def replace_dict(self,
                      replace_keys: bool = True):
         if replace_keys:
             if self.loop_variable and self.loop_variable in repl:
-                self.loop_variable = repl[self.loop_variable]
+                self.loop_variable = str(repl[self.loop_variable])
```
ThrudPrimrose, line 3790: When removing the `str()` here, some tests fail. The test makes sense, but why would `repl` change the dtype of the input?

The function signature is:
```
    def replace_dict(self,
                     repl: Dict[str, str],
                     symrepl: Optional[Dict[symbolic.SymbolicType, symbolic.SymbolicType]] = None,
                     replace_in_graph: bool = True,
                     replace_keys: bool = True):
```

So I would have expected this `str()` wrapper to be unnecessary

tbennun: The reason is that someone is not playing nice. Maybe we should track down the bad users instead.

dace/sdfg/replace.py
```diff
@@ -196,6 +196,48 @@ def replace_in_codeblock(codeblock: properties.CodeBlock,
...
+        if not isinstance(item, str) or item not in repl:
+            return item
+        new_name = str(repl[item])
+        return new_name if new_name.isidentifier() else item
+
+    is_symbolic_type = (element_type is symbolic.SymExpr
+                        or (isinstance(element_type, type) and issubclass(element_type, sp.Basic)))
+    if element_type in (int, float) or is_symbolic_type:
+        try:
+            newitem = symbolic.pystr_to_symbolic(str(item)).subs(symrepl)
```
ThrudPrimrose, line 228: Doesn't calling `str()` before `pystr_to_symbolic()` defeat the purpose?

dace/sdfg/state.py
ThrudPrimrose: `safe_replace` passes symbol so the typehint should be:
`Dict[str, str|dace.symbol]`?

ThrudPrimrose: I think this is a task that can be done by an opus session - even though the overhead of calling `str` is not massive it is not something super fast (from my memory), finding the callers might be better

ThrudPrimrose: I also fear the string roundtrips through `str` might cause the issue where 2 symbols have same string name but have mismatching assumptions?
Because, `pystr(str(x))` would drop assumptions (although in this context it should be safe)

dace/sdfg/replace.py
tbennun: Fixes some type issues - that was the old code too

ThrudPrimrose (approved):

PR #2584, opened 2026-09-13
[pre-NSDFG-fixes] Allocate views in the scope of their access node, with Scope lifetime

tbennun:
A view allocates no memory: it binds a pointer at its access node, often using the enclosing scope's parameters. The frame and SubgraphView.arglist treated a view like a container of its storage type, so a view inside a scope whose schedule cannot allocate that storage (e.g., GPU_Global inside a kernel, or a custom schedule) was bound outside the scope, where its parameters are undefined. Keep views in their scope, and allocate containers in the scope of the tutorial's custom map code generator.

View.view() and Reference.view() copied the viewed container's lifetime, but views and references only support Scope lifetime. Create them with Scope lifetime and reject other lifetimes during SDFG validation.


ThrudPrimrose (approved):
LGTM

PR #2585, opened 2026-09-13
[pre-NSDFG-fixes] Pass hoisted nested SDFG transients into the generated nested function

tbennun:
When a nested SDFG's transient cannot be allocated within the enclosing scope's schedule, the frame allocates it in an ancestor scope. The CUDA generator passes such containers into kernels, but a CPU nested SDFG function referred to them by a name that is not defined in its body. Pass them as arguments, unless the calling code generator already does.


tbennun:
cscs-ci run

ThrudPrimrose:
This codegen fix looks alright. We had the same problem on nested SDFGs within GPU device maps, our solution was a pass to move shared-memory scope allocation data from a nsdfg to parent scope and pass it from there (thus making it non transient inside).  Once that PR is merged we can expand that pass for a clearer distinction of concerns.

ThrudPrimrose (approved):

PR #2586, opened 2026-09-14
[pre-NSDFG-fixes] Inline nested SDFGs that read and write a container through one access node

tbennun:
InlineSDFG required every input connector to have a pure source access node and every output connector a pure sink, so a nested SDFG that writes and then reads the same container through a single access node was never inlined. Allow a connector that is both an input and an output when both outer edges bind the same container at the same offsets and the inner access nodes stay valid in the outer scope; its outer paths are only dropped if no outer dataflow depends on their ordering. Also detach path edges whose scope connector lost its counterpart when removing unused connector paths.


dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -96,6 +96,85 @@ def replfunc(mapping):
...
         return all(istr == ostr for istr, ostr in zip(istrides, ostrides))
 
+    @staticmethod
+    def _shared_connectors(state: SDFGState, sdfg: SDFG, nested_sdfg: nodes.NestedSDFG,
+                           candidates: Set[str]) -> Set[str]:
+        """
+        Returns the connectors that are both inputs and outputs of the nested SDFG, are bound to the same outer
+        container with the same offsets, and whose inner access nodes can be kept as they are upon inlining. Such
+        access nodes may be read and written in any order, since reads and writes of the connector resolve to the
+        same outer memory after inlining.
```
ThrudPrimrose, line 106: On a side track, does dace allow passing an array `A` to in connector `B`, while out connector `B`  goes to array `C` ?

like this `A -> (B) [ NSDFG ] (B) -> C` ?

dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -96,6 +96,85 @@ def replfunc(mapping):
...
+        :param sdfg: The SDFG containing ``state``.
+        :param nested_sdfg: The nested SDFG node.
+        :param candidates: Names of connectors that are both input and output connectors.
+        :return: The subset of ``candidates`` that fulfill the above conditions.
+        """
+        nstate: SDFGState = nested_sdfg.sdfg.nodes()[0]
+        result = set()
+        for conn in candidates:
+            desc = nested_sdfg.sdfg.arrays[conn]
+            in_edge = next(state.in_edges_by_connector(nested_sdfg, conn))
```
ThrudPrimrose, line 118: What if, the same out connector has two out edges?
```
| nested |(B) ->
| sdfg   |(B) ->  
```

This would be a valid graph still, right?

dace/transformation/interstate/sdfg_nesting.py
```diff
@@ -96,6 +96,85 @@ def replfunc(mapping):
...
+                    for neighbor in itertools.chain(nstate.predecessors(n), nstate.successors(n))):
+                continue
+            result.add(conn)
+        return result
+
+    @staticmethod
+    def _can_drop_shared_paths(state: SDFGState, nested_sdfg: nodes.NestedSDFG, conn: str, no_source: bool,
+                               no_sink: bool) -> bool:
+        """
+        Checks whether a shared input/output connector (see ``_shared_connectors``) without a source (or sink) access
```
ThrudPrimrose, line 154: How can a nested SDFG in-connector not have a source/sink node? because src/sink node should be an access node and the connector should be eventually connect to an access node?

ThrudPrimrose: (AccessNode) -> (MapEntry) -> (Nsdfg) then the access node would be the src?

dace/transformation/interstate/sdfg_nesting.py
tbennun: No, and we should have a validation test for that case specifically. Regardless, the no-view nested SDFG contract #1696 will disallow all of that

dace/transformation/interstate/sdfg_nesting.py
tbennun: Please clarify: two out edges from the same connector or one empty memlet and one edge out of a connector?

dace/transformation/interstate/sdfg_nesting.py
tbennun: I think the intent here is tasklet->nsdfg (which is allowed in the IR semantics) but I will double check

dace/transformation/interstate/sdfg_nesting.py
ThrudPrimrose: 2 out edges from the same connectors (with a non empty memlet)

dace/transformation/interstate/sdfg_nesting.py
ThrudPrimrose: We should consider if `(tasklet) -> (tasklet|nsdfg)` should be allowed, I think enforcing `(tasklet) -> access node -> (tasklet|nsdfg)` makes sense.

ThrudPrimrose: (But then that discussion is outside the scope of this PR)

dace/transformation/interstate/sdfg_nesting.py
tbennun: that is strictly disallowed in validation

ThrudPrimrose (approved):
I don't see any issues in the impl. LGTM.
(Although design-wise we should discuss and potentially disallow memlet-edges between tasklets/code nodes and require an access node, I have opened an issue for it for discussion)

PR #2587, opened 2026-09-14
[pre-NSDFG-fixes] Convert missing node types from schedule trees to SDFGs

tbennun:
StreeToSDFG raised NotImplementedError for views, dynamic scope copies, reference sets, library calls, do-while and for loops, elif branches, and break/continue. Implement them. Views are created where they are accessed, with a views edge to the viewed container that is passed through enclosing maps and nested SDFGs like any other container. Copies inside maps and nested SDFGs now use the same input/output logic as tasklets, and state boundary insertion treats accesses through a view as accesses to the viewed container.

@romanc This is most of our TODOs but not all of them. Just to be safe, please check this branch against your code too if possible.


tbennun:
@copilot resolve the merge conflicts in this pull request

tbennun:
FYI this was one of the worst ways to manage merging (took ours over main....)

tbennun:
@ThrudPrimrose @acalotoiu @romanc fixed merge now.

FlorianDeconinck:
Re: `View` node.

The implementation buries view in the maps

Roman has been working on our own implementation over at https://github.com/romanc/dace/tree/romanc/stree-view-support

The idea is that we want to have
```python
@dace.program
def tester(a: dace.float64[10, 10, 30], b: dace.float64[10, 10, 30]):
    for k in dace.map[0: 30] @ dace.ScheduleType.Sequential:
        r_I = a[:, :, k]
        for j in dace.map[0:10] @ dace.ScheduleType.Sequential:
            r_I_J = r_I[:, j]
            for i in dace.map[0:10] @ dace.ScheduleType.Sequential:
                b[i, j, k] = exp(r_I_J[i])
```
generates the nicely hoisted pointers C code
```c++
for (int i, ...)
{
    float* r_i = &a[i]
    for (int j, ...)
    {
         float* r_i_j = &r_i[j] {}
    }
}
```
but the current implementation buries the view all the way down, e.g.

```c++
...
for (auto j = 0; j < 10; j += 1) {
    {
        for (auto k = 0; k < 30; k += 1) {
            const double* r_I;
            r_I = &a[(300 * i)];
            const double* r_I_J;
            r_I_J = &r_I[(30 * j)];
```

tests/schedule_tree/roundtrip_test.py
```diff
@@ -56,6 +69,243 @@ def test_name_propagation():
...
+    sdfg = _inverted_loop_sdfg('tester', LoopRegion('loop', 'i < 0', inverted=True))
+    new_sdfg = _roundtrip(sdfg, tn.DoWhileScope, simplify)
+
+    a = np.zeros(10)
+    new_sdfg(A=a)
+    assert np.allclose(a, [1] + [0] * 9)
+
+
+@pytest.mark.parametrize('update_before_condition', (False, True))
+@pytest.mark.parametrize('simplify', (False, True))
```
ThrudPrimrose: Do we need simplify parametrization since the CI is already running all general tests using a matrix of `(simplify=0, simplify=1, auto-opt cpu)`

romanc (commented):
Thanks for pushing the schedule tree forward and for cleaning up some stuff.

I had a coarse look and this PR seems okay. No testing on NDSL generated code (so far). 

With the new `View` handling, does this mean, we can get rid of `NView` / `NViewEnd`  nodes in `treenodes.py`? If so, that would make `visit_{NView/NViewEnd}`, a couple of mappings in the tree init and also the creation of `NView{End}` nodes in [sdft_to_tree](https://github.com/spcl/dace/blob/2b56f4eca189b7de2f917ecb42c755206e2603cf/dace/sdfg/analysis/schedule_tree/sdfg_to_tree.py#L610-L613) unnecessary, right? I'm happy with a follow up PR.

tbennun:
@romanc #1696 already gets rid of those nodes, yes. No need for a follow up PR as this PR is preparation for that other PR

tests/schedule_tree/roundtrip_test.py
tbennun: you are right but the code passed simplify=False to the SDFG, so it was parameterizing simplifying post stree->sdfg conversion. @ThrudPrimrose I pushed a fix that now uses the config. Thanks!

FlorianDeconinck:
> Re: `View` node.
> 
> The implementation buries view in the maps
> 
> Roman has been working on our own implementation over at https://github.com/romanc/dace/tree/romanc/stree-view-support
> 
> The idea is that we want to have
> 
> ```python
> @dace.program
> def tester(a: dace.float64[10, 10, 30], b: dace.float64[10, 10, 30]):
>     for k in dace.map[0: 30] @ dace.ScheduleType.Sequential:
>         r_I = a[:, :, k]
>         for j in dace.map[0:10] @ dace.ScheduleType.Sequential:
>             r_I_J = r_I[:, j]
>             for i in dace.map[0:10] @ dace.ScheduleType.Sequential:
>                 b[i, j, k] = exp(r_I_J[i])
> ```
> 
> generates the nicely hoisted pointers C code
> 
> ```c++
> for (int i, ...)
> {
>     float* r_i = &a[i]
>     for (int j, ...)
>     {
>          float* r_i_j = &r_i[j] {}
>     }
> }
> ```
> 
> but the current implementation buries the view all the way down, e.g.
> 
> ```c++
> ...
> for (auto j = 0; j < 10; j += 1) {
>     {
>         for (auto k = 0; k < 30; k += 1) {
>             const double* r_I;
>             r_I = &a[(300 * i)];
>             const double* r_I_J;
>             r_I_J = &r_I[(30 * j)];
> ```

This is no longer relevant as we taking another road to address it.

dace/sdfg/analysis/schedule_tree/tree_to_sdfg.py
```diff
@@ -340,31 +350,74 @@ def visit_IfScope(self, node: tn.IfScope, sdfg: SDFG) -> None:
             assignments=self._pending_interstate_assignments(),
         )
 
-        # Check if there's an `ElseScope` following this node (in the parent's children).
+        self._continue_conditional_block(node, conditional_block, merge_state)
+
+    def _continue_conditional_block(self, node: tn.ControlFlowScope, conditional_block: ConditionalBlock,
```
ThrudPrimrose: This is a bit of a nit, but continue-conditional sounds like as if there should have been a continue statement inside conditional.
Maybe a name like "process next conditional block" be more descriptive?

ThrudPrimrose (approved):

PR #2588, opened 2026-09-14
[pre-NSDFG-fixes] Reject nested calls that pass fewer elements than the callee declares

tbennun:
The Python frontend accepted a call such as passing a[:, i] (5 elements) to a function declared with a 10-element argument, silently producing accesses past the end of the argument. Raise a DaceSyntaxError instead; arguments of a different shape but sufficient size (e.g., reshapes) remain allowed. Fix the schedule tree test that relied on the mismatch.


tests/python_frontend/argument_test.py
```diff
@@ -50,7 +50,40 @@ def tester(x: dace.float64[20]):
...
+    @dace.program
+    def caller(a: dace.float64[5, 10]):
+        for i in range(5):
+            callee(a[:, i])
+
+    with pytest.raises(dace.frontend.python.common.DaceSyntaxError, match='declared with 10 elements'):
+        caller.to_sdfg(simplify=False)
+
+
+def test_nested_call_with_reshaped_argument():
```
ThrudPrimrose, line 68: If the feature is about rejected different shapes, shouldn't the nested call test include cases where it should refuse, e.g., `callee(a.reshape(20, ))` when callee requires, `a: dace.float20[10]`?

dace/frontend/python/newast.py
```diff
@@ -4375,6 +4375,29 @@ def _parse_sdfg_call(self, funcname: str, func: Union[SDFG, SDFGConvertible], no
...
+            symbolic.pystr_to_symbolic(k): symbolic.pystr_to_symbolic(v)
+            for k, v in (mapping or {}).items() if isinstance(v, (str, int, sympy.Basic))
+        }
+        for a, m in itertools.chain(inputs.items(), outputs.items()):
+            if not isinstance(m, Memlet) or a not in sdfg.arrays or a.startswith('__return'):
+                continue
+            inner_desc = sdfg.arrays[a]
+            if not isinstance(inner_desc, data.Array) or inner_desc.transient:
+                continue
+            inner_size = sympy.Integer(1)
```
ThrudPrimrose, line 4389: Don't we already have a utility function to get the total number of elements of an SDFG array? `array.total_size`  in `dace/data/core.py`?

tests/python_frontend/argument_test.py
tbennun: I think your comment has a syntax error

dace/frontend/python/newast.py
tbennun: That’s more of an allocation property rather than a shape/reshape property

tests/python_frontend/argument_test.py
ThrudPrimrose: Maybe I wrote it confusingly, I was thinking of a test to practice the pattern where the new checks should raise like this:
```python
def test_nested_call_with_reshaped_argument():
    @dace.program
    def callee(a: dace.float64[10]):
        a += 1

    @dace.program
    def caller(a: dace.float64[5, 4]):
        callee(a.reshape((20, )))

    A = np.random.rand(5, 4)

    with pytest.raises(TypeError):
        caller(A)
```

```

dace/frontend/python/newast.py
ThrudPrimrose: Ah, ok, then all good.

I think another PR for later might introduce `total_elements` of an array as a member function, because I think this information is needed in many place (I use it in multiple places). And total_size is `total_elements * size(element_type)`

tbennun: num_elements() sounds good. total_size is numel*sizeof by default but it's modifiable

tests/python_frontend/argument_test.py
tbennun: Good idea. Done.

ThrudPrimrose (approved):

PR #2589, opened 2026-09-14
[pre-NSDFG-fixes] Only redirect gradient memlets that name a zeroed-out container

tbennun:
When the backward pass zeroes out a gradient, it redirects memlets to the transient copy that holds the values. It also redirected memlets that name the other end of the edge (e.g., a copy from a map entry into a transient scalar), giving their subsets the wrong container. Redirect only memlets that name the container the copy stands in for, using the recorded copies rather than the copy's name.


affifboudaoud (approved):
Looks good to me. Thanks! I am curious did this break an existing test or is it just a cleanup pass?

tbennun:
Yes, it breaks tests when we change nested SDFGs to not define new containers.

PR #2593, opened 2026-09-14
Memlet schedules

tbennun:
A memlet serves two purposes at the same time. It produces an address, and it moves bytes. Currently, the default is copy-on-access, recomputing the offset at every access. This PR adds `Memlet.schedule`, which allows to modify that. At code generation, the memlet schedule is lowered to LibraryNodes (just like the copy node today) and can now represent concepts such as pointer incrementation (cursors), buffer resource descriptors, TMA-generated reads, or general address-generation/iterators/accessors.


FlorianDeconinck:
Ongoing testing on GEOS' FV3 Dynamics core routine of D_SW.

Testing the `LoopCursor` schedule for memlets (via `ScheduleLoopCursors`)

In the case of single K (vertical) loop and many J I (horizontal) dual-loop where code within the kernel is heavily dependent on `if` condition to restrict computation further
- 35% speed up while keeping validation

Overall this looks like it's clearly superior for serial CPU code

FlorianDeconinck:
> Ongoing testing on GEOS' FV3 Dynamics core routine of D_SW.
> 
> Testing the `LoopCursor` schedule for memlets (via `ScheduleLoopCursors`)
> 
> In the case of single K (vertical) loop and many J I (horizontal) dual-loop where code within the kernel is heavily dependent on `if` condition to restrict computation further
> 
> * 35% speed up while keeping validation
> 
> Overall this looks like it's clearly superior for serial CPU code

After more thorough look, `ForLoop` already to direct referencing per `For` axis, which seems to lead to a similar speed up. The transform remains valid (as in it validates numerics) but it adds a large amount of pointer/integer that bloats the asm

PR #2594, opened 2026-09-15
Add num_elements() to data descriptors

### LINKED ISSUE #2592: Add total_elements to dace.data descriptors, next to total_size (including Scalar)
Data descriptors expose `total_size`, but there is no `total_elements`. `Array.total_size` is the allocated span, which can exceed the logical element count when strides or padding are given, so callers that need the number of elements recompute `prod(shape)` themselves.

Proposal: add a `total_elements` property to `dace.data` (at least `Data`, `Array`, `Scalar`, `Stream`), with `Scalar` returning 1, so that `total_elements` and `total_size` are both valid on every descriptor, scalars included.

mohitt31:
Adds a `total_elements` property to the data descriptors, as described in #2592.

It lives on `Data` and returns the product of the shape, so `Array`, `Stream`, `Structure` and the view and reference classes all get it without extra code. `Scalar` returns 1 explicitly, the same way it already does for `total_size`. The `Array` docstring now points at it from the `total_size` bullet, since that is where the difference matters.

The test covers a plain array, a scalar, a stream, a symbolic shape, and the padded example from the `Array` docstring, where `total_size` is 63 but `total_elements` is 40. It fails without the change.

I didn't touch existing callers that compute `prod(shape)` themselves. One that might be worth a look separately is `dace/libraries/onnx/op_implementations/normalization_ops.py`, which assigns `total_elements = data.total_size`, so it would count padding if the input had any.

Closes #2592


tbennun:
Not sure if it's better to call it num_elements() instead
@ThrudPrimrose Thoughts?

mohitt31:
Either works for me. I went with `total_elements` because the issue used it and it sits next to `total_size` as a property. `num_elements()` would match what subsets already have, so if you'd rather go that way I'll rename it, just let me know whether you want a method or a property.

ThrudPrimrose:
> Either works for me. I went with `total_elements` because the issue used it and it sits next to `total_size` as a property. `num_elements()` would match what subsets already have, so if you'd rather go that way I'll rename it, just let me know whether you want a method or a property.

I think `num_elements` is more descriptive, and is better than what I had written in the issue. Could you update the naming scheme? Thanks!

mohitt31:
Renamed, and rebased on main. I made it a method rather than a property, since that is how `num_elements()` already works on `Range`, `Indices` and `Memlet`. Say the word if you'd rather have it as a property next to `total_size`.

tbennun (approved):

tbennun:
cscs-ci run

PR #2596, opened 2026-09-16
perf: match two-node patterns without edges without VF2

edopao:
## Summary
- `get_transformation_metadata` used networkx VF2 for every pattern with more than one node that isn't a single edge, including two unconnected nodes (e.g. `MapFusionHorizontal`). For such patterns, VF2 runs its feasibility test on every node of the state for the second pattern node, once per candidate for the first, which dominated matching time on large states.
- Adds `_unconnected_pair_matcher`, a dedicated matcher for two-node patterns with no edge between them. It yields the same matches in the same order as `_subgraph_isomorphism_matcher`: ordered pairs of distinct type-matching nodes that are not adjacent in either direction and have no self-edge.

## Test plan
- [x] New test `unconnected_pair_matching_test.py` compares the new matcher against VF2 on random graphs and on an SDFG.

Co-authored-by: @havogt


dace/transformation/passes/pattern_matching.py
```diff
@@ -491,6 +493,31 @@ def _node_matcher(digraph, nxpattern, node_pred, edge_pred):
...
+
+    def candidates(pattern_node):
+        return [
+            nid for nid in digraph if node_pred(digraph.nodes[nid], pattern_node) and not digraph.has_edge(nid, nid)
+        ]
+
+    second_candidates = candidates(second_pattern_node)
+    for u in candidates(first_pattern_node):
+        for v in second_candidates:
+            if u != v and not digraph.has_edge(u, v) and not digraph.has_edge(v, u):
```
tbennun: I wonder if you don't want to do `u is not v` because it's cheaper

edopao: Good catch, fixed.

dace/transformation/passes/pattern_matching.py
```diff
@@ -491,6 +493,31 @@ def _node_matcher(digraph, nxpattern, node_pred, edge_pred):
...
+        feasibility test on every graph node for the second pattern node, once per candidate for
+        the first, which dominates the matching time on large states.
+    """
+    first, second = nxpattern
+    first_pattern_node = nxpattern.nodes[first]
+    second_pattern_node = nxpattern.nodes[second]
+
+    def candidates(pattern_node):
+        return [
+            nid for nid in digraph if node_pred(digraph.nodes[nid], pattern_node) and not digraph.has_edge(nid, nid)
```
tbennun: I am not sure if the check should be has_edge or has_path (more expensive...) This way two chained map-entries can match. I am not sure about the rules. Can we add a test please for some dist(i,j)>2 path and see how VF2 behaves? Thanks!

tbennun (changes requested):
One more comment about the connectivity of the two nodes. Otherwise LGTM

PR #2597, opened 2026-09-16
perf: avoid linear node-ID scans in pattern matching

edopao:
## Summary
- For every candidate match, `_try_to_match_transformation` looked up the ID of each matched node with `graph.node_id`, a Python-level scan over all nodes of the state — even though `collapse_multigraph_to_nx` numbers the nodes in the order of `graph.nodes()`, so the index in the collapsed graph already *is* the node ID.
- `_try_to_match_transformation` now uses that index directly instead of calling `graph.node_id`.
- `OrderedDiGraph.node`, used whenever a pattern node of a transformation is read, now skips to the position with `itertools.islice` instead of enumerating in Python when given an integer index.

## Test plan
- [x] New test `node_lookup_test.py` checks that node IDs are positions in the collapsed graph for states, regions and conditional blocks, and pins `node` and `node_id` against a reference scan.

Co-authored-by: @havogt


dace/sdfg/graph.py
```diff
@@ -669,7 +669,11 @@ def nx(self):
 
     def node(self, id: int) -> NodeT:
         try:
-            return next(n for i, n in enumerate(self._nodes.keys()) if i == id)
+            if isinstance(id, int) and id >= 0:
+                # Same result as the scan below, without a Python-level loop over the nodes
+                return next(itertools.islice(self._nodes.keys(), id, None))
+            else:
+                return next(n for i, n in enumerate(self._nodes.keys()) if i == id)
```
tbennun: isn't isinstance() also expensive? I think that based on the type hint you can just assume id is an int.

edopao: Good point, fixed.

tbennun (approved):
LGTM

PR #2598, opened 2026-09-16
perf: do not repeat pattern matching per transformation

edopao:
## Summary
- `PatternMatchAndApplyRepeated` with `order_by_transformation=True` (the default of `SDFG.apply_transformations_repeated()`) ran one match-and-apply loop per transformation. Those loops all enumerate the same matches: `match_patterns()` receives `self._metadata`, which holds every transformation of the pass, and ignores its `patterns` argument. So the ordered variant applied the same matches in the same order as the unordered one, and only added enumerations that apply nothing: one per remaining transformation, plus a full round of them as soon as anything applied.
- Runs a single match-and-apply loop in both the ordered and unordered case, keeping the per-transformation warning about experimental control flow block incompatibility. For a pass of two transformations with k applications, matching is now enumerated k+1 times instead of k+4.
- Should `match_patterns()` ever start honouring its `patterns` argument over the metadata, ordering by transformation becomes meaningful again and this optimization would need to be revisited, since it would then change the application order.

## Test plan
- [x] New test `order_by_transformation_test.py` checks that the ordered and unordered variants apply the same matches, in the same order, and counts the pattern-matching enumerations for both.

Co-authored-by: @havogt


dace/transformation/passes/pattern_matching.py
```diff
@@ -211,50 +211,36 @@ def _apply_pass(self, sdfg: SDFG, pipeline_results: Dict[str, Any], apply_once:
...
+        while applied:
+            applied = False
+            for match in match_patterns(sdfg,
+                                        permissive=self.permissive,
+                                        patterns=xforms,
+                                        states=self.states,
+                                        metadata=self._metadata):
+                self._apply_and_validate(match, sdfg, start, pipeline_results, applied_transformations)
+                applied = True
+                break
```
philip-paul-mueller:
```suggestion
            matched_pattern = next(match_patterns(sdfg,
                                        permissive=self.permissive,
                                        patterns=xforms,
                                        states=self.states,
                                        metadata=self._metadata), None)
             if matched_pattern is not None:
                 self._apply_and_validate(match, sdfg, start, pipeline_results, applied_transformations)
                 applied = True
```

edopao: Sorry, I don't follow. You suggested edit does not look equivalent: the transformation is not applied.

philip-paul-mueller: You are right.
I updated the suggestion.
What I do not like at the thing is that the `for` is essentially used as an `if` statement, hence the unconditional `break`.

edopao: Good point, thank you.

dace/transformation/passes/pattern_matching.py
```diff
@@ -211,58 +211,45 @@ def _apply_pass(self, sdfg: SDFG, pipeline_results: Dict[str, Any], apply_once:
...
-                applied = False
-                for match in match_patterns(sdfg,
-                                            permissive=self.permissive,
-                                            patterns=xforms,
-                                            states=self.states,
-                                            metadata=self._metadata):
-                    self._apply_and_validate(match, sdfg, start, pipeline_results, applied_transformations)
-                    applied = True
-                    break
+            # `match_patterns()` matches on `self._metadata`, which covers every transformation of
```
tbennun: something is off here. The check is only done if order_by_transformations is true, and then the order is the same no matter what?

tbennun (changes requested):
Not sure if the PR is correct as-is

PR #2599, opened 2026-09-16
fix: write the FOLDER_MODE marker atomically

edopao:
## Summary
- `generate_program_folder()` wrote `FOLDER_MODE` with `open(..., "w")`, so the file existed and was empty until the write landed. A concurrent `get_folder_mode(probe=True)` — e.g. from `get_binary_name()` in another process sharing the build folder — could read `''` and raise `ValueError: Unknown folder mode '' found.`
- `FOLDER_MODE` (and `CACHEDIR.TAG`) are now written to a uniquely named file in the same directory and moved into place with `os.replace()`, so a reader never observes a partial/empty file. The staging name is not PID-based, since build folders on a shared file system can be written by several hosts with colliding PIDs.
- Belt-and-suspenders: `get_folder_mode(probe=True)` now also treats an existing-but-empty-or-unknown marker (content that isn't `development` or `production`) as "no marker" rather than raising, since a build folder shared with an older DaCe that still writes the marker in place can still expose such a file.

## Test plan
- [x] `tests/compile_folder_mode_test.py` (8 tests) passes, including new cases added by these commits for the atomic write and for probing an empty/unknown marker.

Co-authored-by: @havogt


dace/codegen/compiler.py
```diff
@@ -654,6 +663,9 @@ def get_folder_mode(object_folder: Union[pathlib.Path, str], probe: bool = False
     if (object_folder / 'FOLDER_MODE').exists():
         with open(object_folder / 'FOLDER_MODE', 'rt') as F:
             folder_mode = F.readline().strip()
+        if probe and folder_mode not in ('development', 'production'):
+            # E.g. empty, because an older DaCe, which does not write it atomically, is still writing it.
```
philip-paul-mueller: This comment does not make much sense.

edopao:
```suggestion
            # E.g. an older DaCe version, which does not write the file atomically, might have left it empty.
```

philip-paul-mueller (commented):
LGTM

tbennun (approved):
LGTM.

PR #2600, opened 2026-09-16
fix: require string replacements in replace_dict

### LINKED ISSUE #2591: Drop str() casts in replace/replace_dict and fix callers that pass non-string replacements
`replace` / `replace_dict` are typed `Dict[str, str]`, but some callers pass other types. `symbolic.safe_replace` hands its callback sympy symbols (default `value_as_string=False`), and `LoopUnroll` passes the loop index as a number. #2583 converts these back with `str()` where names are stored (`LoopRegion.loop_variable`, interstate edge assignment keys). That also silently stores non-identifiers such as `"3"`.

Proposal: remove the `str()` casts, make `replace`/`replace_dict` reject non-string replacement values, and fix the offending callers at the source. This touches many call sites and needs a full CI run (see the discussion in #2583).

alibro005:
## Description

Fixes #2591.

Some `replace` and `replace_dict` callers were passing non-string replacement values, which were being converted to strings internally. This could hide incorrect usage and result in invalid names being stored in the SDFG.

This PR removes those implicit conversions and makes the replacement functions reject non-string keys and values. The affected callers have been updated to convert their values explicitly where needed.

I also added regression tests for invalid replacement values and updated the affected LoopUnroll tests.

### Testing

* `replace_untouched_names_test.py` passes
* Selected LoopUnroll tests pass
* `git diff --check` passes

The full test suite could not be run locally because of existing Windows/Python environment issues, and some LoopUnroll tests require a C++ toolchain that is not available locally.


ThrudPrimrose (changes requested):
The issue is not about casting expressions into symbolic via symstr.
The reason this issue exists to pinpoint which callers not adhere to the contract and fix it there.

alibro005:
Revisited the affected call sites and updated them based on the expected replacement type. I also removed the unnecessary `symstr()` conversions from the propagation paths, since those values are already handled as textual expressions.

Updated in `9a30ac1`.

tbennun:
I'm not sure that we don't want to allow replacing by integer constants.

alibro005:
> I'm not sure that we don't want to allow replacing by integer constants.

@tbennun  That still works. The integer constant just needs to be passed explicitly as a string now, so `replace_dict({"nclv": 5})` becomes `replace_dict({"nclv": "5"})`. This matches the existing `Dict[str, str]` contract and avoids the implicit conversion inside `replace_dict()`.

alibro005:
I have addressed the review feedback and pushed the updates in `8438868`. Could you please take another look when you get a chance?

PR #2602, opened 2026-09-17
Config Environment as Default

philip-paul-mueller:
Before `dace.Config.get()` would always consult the environment first.
The consequence of this is that the setting of explicit values in code has no effect.
Thus, the following code:
```python
with dace.Config.temporary_config():
    dace.Config.set("my.awesome.key", value=1)
    doThingsThatWouldFailIfAwesomeKeyIsNotOne()
```
has no effect and will result in a failure, if the environment variable corresponding to `my.awesome.key` is exported with a value different from `1`.

This PR changes this behaviour and now `Config.get()` resolves keys in the following order (decreasing priority):
- Values explicitly set through `set()`.
- Values exported to the environment.
- Values in the configuration file, `DACE_CONFIG`.
- Default values specified in the schema.

The main effect of this is that the result of `Config.save()` now saves the actual configuration that is in effect.
This also fixes a bug in `generate_program_folder()`, because until now the generated `dace.conf` file, ignored the environments. 

Furthermore, the `autosave` argument of  the configuration modifier functions was removed.


tbennun (approved):
Very nice!

PR #2603, opened 2026-09-17
perf: resolve the state symbols once per state during memlet propagation

edopao:
## Summary
- `propagate_memlet()` asks `SDFGState.symbols_defined_at()` for the symbols of the scope it propagates through, once per Memlet. Most of that answer does not depend on the node, and its expensive part, the walk over every data descriptor, does not even depend on the state.
- Splits it in two: `SDFGState.sdfg_symbols()` for the symbols of the SDFG and its descriptors, and `symbols_defined_at_state()` for what a state sees, which adds the inter-state edge symbols and the ones of the enclosing control flow regions. `symbols_defined_at()` takes the latter as a keyword argument.
- A `SymbolResolver` keeps one result per state and one per SDFG. The propagation entry points take one and pass it on, so the reuse belongs to the caller instead of to module state: nothing to tear down, and nothing shared between threads that compile at the same time.
- The reuse is per state, not per SDFG, because that is what the value depends on: the enclosing regions define symbols today (e.g. a `LoopRegion`'s iterator), and the inter-state edges that reach a state may come to define them too. Caching per SDFG instead would silently drop region-local symbols for every state but the one visited first — confirmed with a `LoopRegion` reproduction (`0:N`/dynamic instead of `0:it`).
- Follow-up commit documents the `symbols` parameter on every propagation entry point (only `propagate_memlets_scope()` had it untyped/undocumented before) and hoists the resolver-vs-state choice in `propagate_memlet()` out of the `defined_vars` comprehension into a ternary, since that function is the one exception that falls back to asking the state directly rather than constructing a resolver (it also runs on graph views that only offer `symbols_defined_at()`).

Measured on ICON's velocity-advection predictor (23k propagated Memlets, 1k arrays): the optimization pipeline of that program takes 132 s instead of 194 s with the other three commits of this branch in place, and produces the same SDFG.

## Test plan
- [x] `tests/sdfg/symbols_defined_at_test.py` (new/updated cases: same-result check, once-per-state resolution, propagation-unchanged, `LoopRegion` symbol scoping and propagation, resolver leaves no state behind on error)
- [x] `tests/memlet_propagation_test.py`, `tests/memlet_propagation_squeezing_test.py`

Co-authored-by: @havogt


dace/sdfg/propagation.py
```diff
@@ -1256,13 +1256,43 @@ def _propagate_state_border_memlets(state: 'SDFGState', border_memlets, arrays)
...
+    The reuse is per state on purpose: what a state sees depends on the control flow regions
+    around it, and may yet come to depend on the inter-state edges that lead to it. Its expensive
+    part, the walk over the data descriptors, is genuinely per SDFG and is reused as such.
+    """
+
+    def __init__(self) -> None:
+        self._per_sdfg: Dict['SDFG', Dict[str, dtypes.typeclass]] = {}
+        self._per_state: Dict['SDFGState', Dict[str, dtypes.typeclass]] = {}
+
+    def defined_at(self, state: 'SDFGState', node: nodes.Node) -> Dict[str, dtypes.typeclass]:
```
ThrudPrimrose, line 1275: Shouldn't this function collect symbols from parent loop regions, as loop iterators are scoped and might be defined only within the scope of a state that has parent loops cfg?

edopao: Shouldn't the symbols defined at the parent control flow region be already resolved by `symbols_defined_at_state()`, as in baseline implementation? or are you suggesting to introduce another cache level, between`_per_sdfg` and`_per_state`?

ThrudPrimrose: True, I have misunderstood the impl.

dace/sdfg/propagation.py
```diff
@@ -1256,13 +1256,43 @@ def _propagate_state_border_memlets(state: 'SDFGState', border_memlets, arrays)
                                                                         propagated, arrays[array_name])
 
 
-def propagate_memlets_nested_sdfg(parent_sdfg: 'SDFG', parent_state: 'SDFGState', nsdfg_node: nodes.NestedSDFG):
+class SymbolResolver:
+    """Resolves the symbols visible at a node, reusing the part that only depends on the state.
+
+    `SDFGState.symbols_defined_at_state()` is the same answer for every node of one state, and
```
tbennun:
```suggestion
    ``SDFGState.symbols_defined_at_state()`` is the same answer for every node of one state, and
```

dace/sdfg/propagation.py
```diff
@@ -1256,13 +1256,43 @@ def _propagate_state_border_memlets(state: 'SDFGState', border_memlets, arrays)
...
+                                  parent_state: 'SDFGState',
+                                  nsdfg_node: nodes.NestedSDFG,
+                                  symbols: Optional[SymbolResolver] = None):
     """
     Propagate memlets out of a nested sdfg.
 
     :param parent_sdfg: The parent SDFG this nested SDFG is in.
     :param parent_state: The state containing this nested SDFG.
     :param nsdfg_node: The NSDFG node containing this nested SDFG.
+    :param symbols: The `SymbolResolver` of the ongoing propagation, one is made if not given.
```
tbennun:
```suggestion
    :param symbols: The ``SymbolResolver`` of the ongoing propagation, one is made if not given.
```

dace/sdfg/propagation.py
```diff
@@ -1380,25 +1411,28 @@ def reset_state_annotations(sdfg: 'SDFG'):
         state.ranges = {}
 
 
-def propagate_memlets_sdfg(sdfg: 'SDFG'):
+def propagate_memlets_sdfg(sdfg: 'SDFG', symbols: Optional[SymbolResolver] = None):
     """ Propagates memlets throughout an entire given SDFG.
 
+        :param symbols: The `SymbolResolver` of the ongoing propagation, one is made if not given.
```
tbennun:
```suggestion
        :param symbols: The ``SymbolResolver`` of the ongoing propagation, one is made if not given.
```

dace/sdfg/propagation.py
```diff
@@ -1380,25 +1411,28 @@ def reset_state_annotations(sdfg: 'SDFG'):
...
     propagate_states(sdfg)
 
 
-def propagate_memlets_state(sdfg: 'SDFG', state: 'SDFGState'):
+def propagate_memlets_state(sdfg: 'SDFG', state: 'SDFGState', symbols: Optional[SymbolResolver] = None):
     """ Propagates memlets throughout one SDFG state.
 
         :param sdfg: The SDFG in which the state is situated.
         :param state: The state to propagate in.
+        :param symbols: The `SymbolResolver` of the ongoing propagation, one is made if not given.
```
tbennun:
```suggestion
        :param symbols: The ``SymbolResolver`` of the ongoing propagation, one is made if not given.
```

dace/sdfg/propagation.py
```diff
@@ -1447,10 +1488,13 @@ def propagate_memlets_scope(sdfg, state, scopes, propagate_entry=True, propagate
     :param scopes: The ScopeTree object or a list thereof to start from.
     :param propagate_entry: If False, skips propagating out of the scope entry node.
     :param propagate_exit: If False, skips propagating out of the scope exit node.
+    :param symbols: The `SymbolResolver` of the ongoing propagation, one is made if not given.
```
tbennun:
```suggestion
    :param symbols: The ``SymbolResolver`` of the ongoing propagation, one is made if not given.
```

dace/sdfg/propagation.py
```diff
@@ -1467,19 +1511,22 @@ def propagate_memlets_scope(sdfg, state, scopes, propagate_entry=True, propagate
...
 
 
-def propagate_memlets_map_scope(sdfg: 'SDFG', state: 'SDFGState', map_entry: nodes.MapEntry) -> None:
+def propagate_memlets_map_scope(sdfg: 'SDFG',
+                                state: 'SDFGState',
+                                map_entry: nodes.MapEntry,
+                                symbols: Optional[SymbolResolver] = None) -> None:
     """Propagate Memlets from the given Map outside.
 
     The main difference to `propagate_memlets_scope()` is that this function operates on Maps
```
tbennun:
```suggestion
    The main difference to ``propagate_memlets_scope()`` is that this function operates on Maps
```

dace/sdfg/propagation.py
```diff
@@ -1490,6 +1537,7 @@ def propagate_memlets_map_scope(sdfg: 'SDFG', state: 'SDFGState', map_entry: nod
     :param sdfg: The SDFG in which the scopes reside.
     :param state: The SDFG state in which the scopes reside.
     :param map_entry: Defining the Map scope to which propagation should be restricted.
+    :param symbols: The `SymbolResolver` of the ongoing propagation, one is made if not given.
```
tbennun:
```suggestion
    :param symbols: The ``SymbolResolver`` of the ongoing propagation, one is made if not given.
```

dace/sdfg/propagation.py
```diff
@@ -1585,6 +1641,8 @@ def propagate_memlet(dfg_state,
         :param union_inner_edges: True if the propagation should take other
                                   neighboring internal memlets within the same
                                   scope into account.
+        :param symbols: The `SymbolResolver` of the ongoing propagation, if there is one; without
```
tbennun:
```suggestion
        :param symbols: The ``SymbolResolver`` of the ongoing propagation, if there is one; without
```

tbennun (changes requested):
Good change, I would move SymbolResolver somewhere higher so that it can be reused at codegen too (there is a slower-than-optimal performance there as well). My minor comments are related to reStructuredText syntax.

